PR #28, opened 2022-03-07
Convert vision xlsx to power grid model input data

bramstoeller:


bramstoeller:
I moved this conversion script to https://github.com/alliander-opensource/power-grid-model-io/tree/feature/excel

PR #31, opened 2022-03-23
Zigzag transformer

TonyXiang8787:


include/power_grid_model/component/transformer.hpp
```diff
@@ -53,11 +53,14 @@ class Transformer : public Branch {
           z_grounding_to_{
               get_z_grounding(transformer_input.r_grounding_to, transformer_input.x_grounding_to, u2_rated)} {
         // check on clock
-        if (clock_ < 0 || clock_ > 12 ||
-            // even number should not be yd configuration
```
petersalemink95, line 57: Why did the checks for yd disappear?

include/power_grid_model/component/transformer.hpp
```diff
@@ -53,11 +53,14 @@ class Transformer : public Branch {
...
-            ((clock_ % 2) == 0 && ((winding_from_ == WindingType::delta) != (winding_to_ == WindingType::delta))) ||
-            // odd number should not be yy or dd configuration
```
petersalemink95, line 59: Same for yy or dd for odd

TonyXiang8787: It checks in another way. See below

PR #33, opened 2022-03-24
Versioning and Dependabot

TonyXiang8787:


setup.py
```diff
@@ -75,7 +76,7 @@ def generate_build_ext(pkg_dir: str, pkg_name: str):
     # include-folders
     include_dirs = [
```
bramstoeller, line 77: Must these paths be strings, or can we use Path objects?

setup.py
```diff
@@ -89,29 +90,29 @@ def generate_build_ext(pkg_dir: str, pkg_name: str):
         ("EIGEN_MPL2_ONLY", "1"),  # only MPL-2 part of eigen3
         ("POWER_GRID_MODEL_USE_MKL_AT_RUNTIME", 1),  # use mkl runtime loading
     ]
+    pkg_bin_dir = pkg_dir / "src" / pkg_name
 
     # remove old extension build
-    shutil.rmtree(os.path.join(pkg_dir, "build"), ignore_errors=True)
+    shutil.rmtree(pkg_dir / "build", ignore_errors=True)
```
bramstoeller, line 96: Do we want to ignore errors? If we expect errors in some cases, it might be good to write some comments about the expected errors.

bramstoeller: Suggestions:
```python
build_dir = pkg_dir / "build"
if build_dir.exists():
    shutil.rmtree(build_dir)
```

setup.py
```diff
@@ -187,9 +178,14 @@ def convert_long_description(raw_readme: str):
         return re.sub(r"(\[[^\(\)\[\]]+\]\()((?!http)[^\(\)\[\]]+\))", f"\\1{url}\\2", raw_readme)
 
 
-def get_version(pkg_dir):
-    with open(os.path.join(pkg_dir, "VERSION")) as f:
+def get_version(pkg_dir: Path):
```
bramstoeller, line 181: According to mypy it is recommended to add the return type as well when you use typehinting (even if the return type is None). So either use no type hinting, or full type hinting.

setup.py
```diff
@@ -226,14 +246,15 @@ def build_pkg(setup_file, author, author_email, description, url):
...
     pkg_pip_name = "power-grid-model"
     pkg_name = pkg_pip_name.replace("-", "_")
-    with open(os.path.join(pkg_dir, "README.md")) as f:
+    with open(pkg_dir / "README.md") as f:
         long_description = f.read()
         long_description = convert_long_description(long_description)
-    with open(os.path.join(pkg_dir, "requirements.txt")) as f:
+    with open(pkg_dir / "requirements.txt") as f:
         required = f.read().splitlines()
         required = [x for x in required if "#" not in x]
```
bramstoeller, line 259: It is possible to add comments to your pip package, e.g.
```pip
fancypackage==1.2.3  # Use version pin until bug #42 is solved
```

So I would recommend to use `if lstrip(x) and lstrip(x)[0] != "#"`

bramstoeller (approved):
Approved, but please review my comments before merging and decide if you want to alter some some lines or leave it as is.

setup.py
TonyXiang8787: It has to be string.

setup.py
TonyXiang8787: What should you import for `strip`?

bramstoeller: Sorry, I meant `x.strip()`

bramstoeller (approved):

PR #52, opened 2022-04-27
Fix transformer case status

nitbharambe:
Signed-off-by: Nitish Bharambe <nitish.bharambe@alliander.com>


TonyXiang8787:
@nitbharambe  is this now ready for review?

tests/data/power_flow/pandapower/components/asymmetric/transformer/input.json
```diff
@@ -21,7 +21,7 @@
       "sn": 1000000.0,
       "uk": 0.1,
       "pk": 10000.0,
-      "i0": 0.001,
```
TonyXiang8787, line 24: @nitbharambe 

Can we just make `i0` and `p0` to zero? And then we can set low tolerance for comparison.

TonyXiang8787 (approved):

PR #77, opened 2022-05-31
Use internal sparse solver to replace EIGEN and MKL

### LINKED ISSUE #75: [FEATURE] Use in-house implementation of sparse solver
# Motivation

- The SparseLU solver from `eigen3` is slow in many scenarios.
- The optional dependency of MKL PARDISO solver is heavy and not available in `arm64` (macOS, Linux).

# Proposal

- Implement in-house sparse solver using LU factorization.
- Exploit the prior knowledge to optimize the solver.
- Can deliver performance boost to MKL, let alone SparseLU.
- Can remove MKL dependency.

TonyXiang8787:
# Introduction

This PR implements an internal sparse solver which replaces EIGEN and MKL solver.

Closes #75 

# Design choices

## Adjustment outside sparse solver

- In bus reordering of topology search, the fill-ins are pre-analyzed and recorded. So all the off-diagonal fill-ins are known in advance.
- In the construction of Node Admittance Matrix, two sparse matrix structures are made
    -  The node admittance matrix
    -  The LU decomposition matrix with the fill-ins. This will have more entries. There is also a mapping between the linear index of node admittance matrix and LU matrix.
- The structure of node admittance matrix is now shared between symmetric and asymmetric math solver.

## Implementation of LU sparse solver

- It uses standard right-looking Gaussian elimination for LU decomposition.
- The buses are already pre-ordered, so no permutation or fill-in analysis is needed. At the beginning of the factorization, the LU matrix is pre-allocated including all the fill-ins.
- Since power system matrices are always diagonal major, numeric pivoting across the sparse structure is not needed.
- The entry in the matrix is either a scalar (real/complex number) or a block (2x2, 6x6)
    - For a scalar matrix it just uses the normal sparse LU decomposition.
    - For block matrix it factorizes the matrix per dense block. Inside the diagonal block a dense LU decomposition is used with full pivoting. The factorized diagonal LU block will be used to further factorize the off-diagonal elements. Because it is a full pivoting within the block, there will be row/column permutation within the block. This applies to the right hand side and unknown vectors.
- The matrix is always structurally symmetric. The code uses this feature to quickly find some non-zero entries in the matrix: if position `(i, j)` is non-zero, position `(j, i)` is also non-zero.
-The solver is a class template with types of block matrices. The compiler can then optimize (e.g. loop unrolling) for difference block sizes.


build.sh
```diff
@@ -16,13 +16,13 @@ if [ ! "$1" = "Debug" ] && [ ! "$1" = "Release" ]; then
...
 fi
 
-if [ ! "$2" = "EIGEN" ] && [ ! "$2" = "MKL" ] && [ ! "$2" = "MKL_RUNTIME" ]; then
-  echo "Missing second argument"
-  usage
-  exit 1;
+if [[ $2 == "Coverage" ]]; then
+  BUILD_COVERAGE=-DPOWER_GRID_MODEL_COVERAGE=1
+else
+  BUILD_COVERAGE=
```
petersalemink95, line 22: BUILD_COVERAGE=...

Isn't there something missing?

include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -0,0 +1,349 @@
...
+    using Scalar = typename entry_trait::Scalar;
+    using LUFactor = typename entry_trait::LUFactor;
+    using BlockPerm = typename entry_trait::BlockPerm;
+    using BlockPermArray = typename entry_trait::BlockPermArray;
+
+    SparseLUSolver(std::shared_ptr<IdxVector const> const& row_indptr,   // indptr including fill-ins
+                   std::shared_ptr<IdxVector const> const& col_indices,  // indices including fill-ins
+                   std::shared_ptr<IdxVector const> const& diag_lu)
+        : size_{(Idx)row_indptr->size() - 1},
+          nnz_{row_indptr->back()},
```
petersalemink95, line 86: What is nnz? Number of non zeros? Matbe add a comment here

tests/cpp_unit_tests/test_sparse_lu_solver.cpp
```diff
@@ -0,0 +1,124 @@
...
+        SECTION("Test (pseudo) singular") {
+            data[0] = 0.0;
+            CHECK_THROWS_AS(solver.solve(data, block_perm, rhs, x), SparseMatrixError);
+        }
+
+        SECTION("Test prefactorize") {
+            solver.prefactorize(data, block_perm);
+            solver.solve((std::vector<double> const&)data, block_perm, rhs, x);
+            check_result(x, x_ref);
+        }
```
bramstoeller, line 78: Suggestion:
```c++
SUBCASE("Data is prefactorized by solve") {
    auto prefactorized_data = data;
    auto prefactorized_block_perm = block_perm;
    solver.prefactorize(prefactorized_data, prefactorized_block_perm);
    solver.solve(data, block_perm, rhs, x);
    CHECK(prefactorized_data == data);
    CHECK(prefactorized_block_perm == block_perm);
}
```

tests/cpp_unit_tests/test_sparse_lu_solver.cpp
```diff
@@ -0,0 +1,124 @@
...
+            CHECK((cabs(x[i] - x_solver[i]) < numerical_tolerance).all());
+        }
+    }
+}
+
+// test block calculation with 2*2
+using Tensor = Eigen::Array<double, 2, 2, Eigen::ColMajor>;
+using Array = Eigen::Array<double, 2, 1, Eigen::ColMajor>;
+
+TEST_CASE("Test Sparse LU solver") {
```
bramstoeller, line 39: I would remove "Test" from the test names (everywhere).

bramstoeller (commented):
I'm not done with my review yet, but here's some comments for the. sparse_lu_solver

tests/cpp_unit_tests/test_three_phase_tensor.cpp
```diff
@@ -87,8 +87,8 @@ TEST_CASE("Three phase tensor") {
         // test layout
         double* arr = (double*)&mat1;
         CHECK(arr[0] == 1);
-        CHECK(arr[2] == 3);
-        CHECK(arr[6] == 7);
+        CHECK(arr[2] == 7);
+        CHECK(arr[6] == 3);
```
petersalemink95, line 91: Why did the order change?

petersalemink95 (dismissed):

petersalemink95:
In general, please add more documentation to classes/functions

tests/cpp_unit_tests/test_y_bus.cpp
```diff
@@ -197,12 +203,59 @@ TEST_CASE("Test one bus system") {
...
     CHECK(bus_entry == ybus.bus_entry());
-    CHECK(transpose_entry == ybus.transpose_entry());
+    CHECK(lu_transpose_entry == ybus.lu_transpose_entry());
     CHECK(y_bus_entry_indptr == ybus.y_bus_entry_indptr());
 }
 
+TEST_CASE("Test fill-in y bus") {
+    /*
+     * struct
+     * [1] --0--> [0] --[1]--> [2]
```
bramstoeller, line 214: Minor: `[1] --0--> [0] --1--> [2]`

bramstoeller (approved):
Good work, very impressive!

In general, please try to avoid refactoring code in such a large pull request, because that makes it much harder to review. And in some cases it would be more readable to use a little less template magic (even at the cost of some code duplication, imho). For example, the `is_block` constexpr in `SparseLUSolver` basically splits most functions in two; one part for scalars and one for blocks. I would prefere to have two separate functions.

tests/cpp_unit_tests/test_three_phase_tensor.cpp
TonyXiang8787: It uses now a column major order.

include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -0,0 +1,349 @@
...
+    void solve(std::vector<Tensor>& data,         // matrix data, factorize in-place
+               BlockPermArray& block_perm_array,  // pre-allocated permutation array, will be overwritten
+               std::vector<RHSVector> const& rhs, std::vector<XVector>& x) {
+        prefactorize(data, block_perm_array);
+        // call solve with const method
+        solve((std::vector<Tensor> const&)data, block_perm_array, rhs, x);
+    }
+
+    // solve with existing pre-factorization
+    void solve(std::vector<Tensor> const& data,         // pre-factoirzed data, const ref
```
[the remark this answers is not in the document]
TonyXiang8787: I have thought about this and could not now find an easy elegant way to separate the logic of block/non-block matrices.

Maybe we can have some idea later then we can refactor here a bit.

include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -90,18 +90,19 @@ class SparseLUSolver {
...
         // call solve with const method
-        solve((std::vector<Tensor> const&)data, block_perm_array, rhs, x);
+        solve_with_prefactorization((std::vector<Tensor> const&)data, block_perm_array, rhs, x);
     }
 
     // solve with existing pre-factorization
-    void solve(std::vector<Tensor> const& data,         // pre-factoirzed data, const ref
-               BlockPermArray const& block_perm_array,  // pre-calculated permutation, const ref
-               std::vector<RHSVector> const& rhs, std::vector<XVector>& x) {
+    void solve_with_prefactorization(std::vector<Tensor> const& data,         // pre-factoirzed data, const ref
```
petersalemink95, line 103: Maybe call this one `solve_prefactorised`? solve with prefactorization sounds like the prefactorization also happens in this function

petersalemink95 (approved):

PR #97, opened 2022-07-14
Bugfix: batch validation asym loads

bramstoeller:


bramstoeller (commented):
I'm not entirely sure about this solution... The question is, what would be the correct result for 
```python
[[1.1, 1.2, 1.3], [2.1, np.nan, np.nan], [np.nan, np.nan, np.nan]]
```

being updated with
```python
[[np.nan, np.nan, np.nan], [np.nan, np.nan, 5.3], [6.1, 6.2, 6.3]]
```

I would say:
```python
[[1.1, 1.2, 1.3], [2.1, np.nan, 5.3], [6.1, 6.2, 6.3]]
```

But the current result would be:
```python
[[1.1, 1.2, 1.3], [2.1, nan, nan], [6.1, 6.2, 6.3]]
```

I'll add a unit test to demonstrate this behaviour.

TonyXiang8787 (approved):

PR #106, opened 2022-07-28
Add mypy to pre-commit (and introduce data_types.py)

bramstoeller:
When applying mypy, I noticed that we had some errors in our type hinting. I got confused in the forrest of Lists, Dicts, etcetera, so in an attempt to make the data structures a bit more understandable, I've introduced type aliases in data_types.py. I'm not sure this is better or even more readable, so please be critical in your review!


src/power_grid_model/manual_testing.py
```diff
@@ -8,11 +8,21 @@
 
 import json
 from pathlib import Path
-from typing import IO, Any, Dict, List, Optional, Union
+from typing import IO, Any, List, Optional, Union, cast
 
 import numpy as np
 
 from . import initialize_array
+from .data_types import (
```
petersalemink95, line 16: Please use absolute (explicit) imports.  I see we don't do it for other imports as well; there are a lot of relative inputs in the code. Maybe good to update all from implicit to explicit imports

src/power_grid_model/manual_testing.py
```diff
@@ -103,18 +109,33 @@ def convert_python_to_numpy(
...
+    # Otherwise this should be a normal (non-batch) structure, with a list of objects (dictionaries) per component.
     if not isinstance(data, dict):
-        raise TypeError("Only list or dict is allowed in JSON data!")
+        raise TypeError("Data should be either a list or a dictionary!")
 
-    dataset: Dict[str, np.ndarray] = {}
+    return _convert_python_to_numpy_single(data=data, data_type=data_type)
+
+
+def _convert_python_to_numpy_single(data: SinglePythonDataset, data_type: str) -> SingleDataset:
```
petersalemink95, line 126: Good that you made two separate functions, makes the code more readable

src/power_grid_model/validation/utils.py
```diff
@@ -12,12 +12,15 @@
 import numpy as np
 
 from .. import power_grid_meta_data
+from ..data_types import (
```
petersalemink95, line 15: I think you forgot to commit data_types.py, since the file does not exist in this branch

petersalemink95 (changes requested):

src/power_grid_model/manual_testing.py
bramstoeller: It was added in commit [336809f](https://github.com/alliander-opensource/power-grid-model/pull/106/commits/336809facc421c13150413c2a914b5c51ec35f5e)

src/power_grid_model/validation/utils.py
bramstoeller: It was added in commit [336809f](https://github.com/alliander-opensource/power-grid-model/pull/106/commits/336809facc421c13150413c2a914b5c51ec35f5e)

src/power_grid_model/manual_testing.py
bramstoeller: Why are absolute imports better?

petersalemink95: Relative imports are allowed, but PEP 8 recommends absolute imports: https://peps.python.org/pep-0008/

petersalemink95 (approved):

src/power_grid_model/data_types.py
```diff
@@ -0,0 +1,200 @@
...
+Many data types are used throughout the power grid model project. In an attempt to clarify type hints, some types
+have been defined and explained in this file
+"""
+
+from typing import Any, Dict, List, Tuple, Union
+
+import numpy as np
+
+
+class SingleArray(np.ndarray):
```
TonyXiang8787, line 14: Subclassing `np.ndarray` is not that straightforward. See
https://numpy.org/doc/stable/user/basics.subclassing.html

I have doubts if we should do this. Or we wait until October this year. Then 3.11 is released, and 3.8 support will be dropped.

TonyXiang8787:
@bramstoeller  also put all the checks in the pipeline.

https://github.com/alliander-opensource/power-grid-model/blob/feature/mypy/.github/workflows/black-and-clang-format.yml

modify the name to a more meaningful name of file, name of workflow and jobs

src/power_grid_model/data_types.py
bramstoeller: I agree, but we're only using it for type hinting, so I guess we're fine then. I'll add some comments and I'll raise a `NotImplemented` exception in the `__new__()` method.

bramstoeller: It doesn't work :/ I've replaced SingleArray and DenseBatchArray simply with np.ndarray for now...

.github/workflows/code-quality.yml
```diff
@@ -33,23 +33,40 @@ jobs:
...
-        
-      - name: Install Black and clang-format
+
+      - name: Upgrade pip
+        run: pip install --upgrade pip
+
+      - name: Install and run isort
+        run: |
+          pip install isort
+          isort src tests scripts
```
TonyXiang8787, line 43: @bramstoeller 

why not just `isort .`?

Same for `black` and `mypy`.

TonyXiang8787 (dismissed):

.github/workflows/code-quality.yml
bramstoeller: black also tries to format jupyter notebooks, which is not desirable imho. And then I decided to explicitly list the directories that we'd like to check.

TonyXiang8787: black only check jupyter notebooks if you install black with `pip install black[jupyter]`. If you install the default black, it will ignore jupyter.

src/power_grid_model/validation/__init__.py
```diff
@@ -4,11 +4,14 @@
...
-from .assertions import (
+from power_grid_model.validation.assertions import (
     ValidationException,
     assert_valid_batch_data,
     assert_valid_input_data,
 )
-from .errors import ValidationError
-from .utils import InputData, UpdateData, errors_to_string
-from .validation import validate_batch_data, validate_input_data
+from power_grid_model.validation.errors import ValidationError
```
TonyXiang8787, line 12: So absolute import does work for `__init__.py`?

Shall we then also change the base `__init__.py` to absolute import?

bramstoeller: Yes, I'll try. I remember that there was some issue with the base __init__.py, but I don't remember what it was :)

TonyXiang8787 (approved):

TonyXiang8787 (approved):

PR #119, opened 2022-08-04
Extract extra_info from python structure in convert_python_to_numpy

bramstoeller:


bramstoeller:
I've changed my mind. I think we should remove the extra info from power-grid-model entirely and move it to power-grid-model-io

PR #121, opened 2022-08-09
In-depth performance profiler

### LINKED ISSUE #120: Build detailed performance profiler, investigate performance issue in multi threading

TonyXiang8787:
Closes #120


mgovers:
@TonyXiang8787 i personally don't like keeping PRs open for too long and it also appears to be quite out of date. What do you want to do with this? I'd recommend either one of the following:

* Close PR and delete branch
  * the commit SHA will remain valid because this is a pull request, so you can always look it up
* Fix the diff and merge
  * I did not investigate how hard this is and if there's a performance overhead
* Close PR but keep the branch
  * the downside is that there will be an open branch remaining

TonyXiang8787:
@mgovers  if you delete the branch. All unmerged commits will be gone right?

mgovers:
> @mgovers if you delete the branch. All unmerged commits will be gone right?

there is always the `restore branch` button, and if you're scared of that, you can also `tag` the latest commit with a meaningful name and list that `tag` here. you can also include the `branch` name to preserve that reference, e.g.: `feature/performance-profiler-<latest_commit_sha>`

TonyXiang8787:
@mgovers  I am in favor of fixing the diff and merge. We have four merge conflicts, here is what I think:

1. `dataset.hpp`: discard the changes in this file, use the `main` version.
2. `main_model.hpp`: you need to resolve the conflict.
3. `power_grid_model.hpp`: you need to resolve the conflict.
4. `benchmark.cpp`: This is the main file changed. Use this branch's version as leading version, adjust some code let it compile and run again.

mgovers:
> or of fixing the diff and merge. We have four merge conflicts, here i

i will have a look. i'll also make sure to benchmark before merging

mgovers:
the merge is complete but i would like to run some benchmark tests still before ready

mgovers:
Benchmark results are not comparable due to different parameters. @TonyXiang8787 since you wrote the code originally, can we maybe go over the parameters to see if there is a configuration that is comparable?

In addition, i'm getting this error message in CI and maybe you know what it is about

```cpp
      + sh -c 'auditwheel repair --plat ${AUDITWHEEL_PLAT//2_28/2_24} -w /tmp/cibuildwheel/repaired_wheel /tmp/cibuildwheel/built_wheel/power_grid_model-1.5.15a1251412073158-py3-none-linux_x86_64.whl'
  INFO:auditwheel.main_repair:Repairing power_grid_model-1.5.15a1251412073158-py3-none-linux_x86_64.whl
  usage: auditwheel [-h] [-V] [-v] command ...
  auditwheel: error: cannot repair "/tmp/cibuildwheel/built_wheel/power_grid_model-1.5.15a1251412073158-py3-none-linux_x86_64.whl" to "manylinux_2_24_x86_64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.
                                                              ✕ 1.60s
```

power_grid_model_c/power_grid_model/include/power_grid_model/power_grid_model.hpp
```diff
@@ -85,7 +86,13 @@ constexpr double default_source_rx_ratio = 0.1;
 constexpr double default_source_z01_ratio = 1.0;
 
 // calculation info
-using CalculationInfo = std::map<std::string, double>;
+#if defined(__cpp_lib_hardware_interference_size)
+constexpr size_t cache_line_size = std::hardware_destructive_interference_size;
```
TonyXiang8787, line 90: I think this one requires glibc 2.28 to work. We must stay on glibc 2.24 to support amazon linux 2.

So maybe remove this one and just set it to 64 as below.

power_grid_model_c/power_grid_model/include/power_grid_model/power_grid_model.hpp
```diff
@@ -21,6 +21,7 @@
 #include <limits>
 #include <map>
 #include <memory>
+#include <new>
```
TonyXiang8787, line 24: why do we need this header?

TonyXiang8787:
@mgovers  as a general comment, I prefer not to build benchmark project in Release mode with address sanitizer. This will not produce reasonable benchmark.

power_grid_model_c/power_grid_model/include/power_grid_model/power_grid_model.hpp
mgovers: i mean, the version test `#if defined(__cpp_lib_hardware_interference_size)` already should already take care of that dependency but i will remove it to keep everything the same between different libraries for now

TonyXiang8787: That's not the case. We compile the Python package in 2.28 environment. It just happens that all our codes do not touch 2.28 specific functionality so we can audit the wheel in 2.24 version. 

Adding this `std::hardware_destructive_interference_size` changes that fact.

mgovers:
> @mgovers as a general comment, I prefer not to build benchmark project in Release mode with address sanitizer. This will not produce reasonable benchmark.

isn't that the responsibility of the user? in addition, it could cause compile errors if we don't build it in CI.

What we can do is add a `target_compile_option` that generates a warning before running. what are your thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/power_grid_model.hpp
mgovers: OK, will remove

TonyXiang8787:
> > @mgovers as a general comment, I prefer not to build benchmark project in Release mode with address sanitizer. This will not produce reasonable benchmark.
> 
> isn't that the responsibility of the user? in addition, it could cause compile errors if we don't build it in CI.
> 
> What we can do is add a `target_compile_option` that generates a warning before running. what are your thoughts?

The benchmark script will only be run by the developer. Now we have some Cmake presets for developer which always include address sanitizer for all CMake sub-projects.

mgovers:
> > > @mgovers as a general comment, I prefer not to build benchmark project in Release mode with address sanitizer. This will not produce reasonable benchmark.
> > 
> > 
> > isn't that the responsibility of the user? in addition, it could cause compile errors if we don't build it in CI.
> > What we can do is add a `target_compile_option` that generates a warning before running. what are your thoughts?
> 
> The benchmark script will only be run by the developer. Now we have some Cmake presets for developer which always include address sanitizer for all CMake sub-projects.

this points at a different issue: the fact that `gcc-base` and `clang-base` both always include the sanitizer.

i would propose to create a separate PR to change that and after that decide on how to define the build targets

TonyXiang8787:
@mgovers can you retrigger the CI?

tests/benchmark_cpp/fictional_grid_generator.hpp
```diff
@@ -0,0 +1,428 @@
...
+        dataset.try_emplace("sym_load", sym_load.data(), batch_size, static_cast<Idx>(sym_load.size()) / batch_size);
+        dataset.try_emplace("asym_load", asym_load.data(), batch_size, static_cast<Idx>(asym_load.size()) / batch_size);
+        return dataset;
+    }
+};
+
+class FictionalGridGenerator {
+  public:
+    FictionalGridGenerator() {}
+
```
mgovers, line 92:
```suggestion
```

tests/benchmark_cpp/fictional_grid_generator.hpp
```diff
@@ -0,0 +1,428 @@
...
+        } else {
+            option_.n_lv_grid = (option_.n_node_total_specified - total_mv_connection) / node_per_lv_grid;
+        }
+        if (option_.n_lv_grid > total_mv_connection) {
+            option_.n_mv_feeder = option_.n_lv_grid / option_.n_node_per_mv_feeder + 1;
+        }
+        total_mv_connection = option_.n_mv_feeder * option_.n_node_per_mv_feeder;
+        option_.ratio_lv_grid = (double)option_.n_lv_grid / (double)total_mv_connection;
+        // each mv feeder 10 MVA, each transformer 60 MVA, scaled up by 10%
+        option_.n_parallel_hv_mv_transformer = (int)((double)option.n_mv_feeder * 10.0 * 1.1 / 60.0) + 1;
```
mgovers, line 119:
```suggestion
        option_.ratio_lv_grid = static_cast<double>(option_.n_lv_grid) / static_cast<double>(total_mv_connection);
        // each mv feeder 10 MVA, each transformer 60 MVA, scaled up by 10%
        option_.n_parallel_hv_mv_transformer = static_cast<Idx>(static_cast<double>(option.n_mv_feeder) * 10.0 * 1.1 / 60.0) + 1;
```

tests/benchmark_cpp/fictional_grid_generator.hpp
```diff
@@ -0,0 +1,428 @@
...
+                input_.line.push_back(line);
+                // generate lv grid
+                if (lv_gen(gen_)) {
+                    generate_lv_grid(current_node_id, 10.0 / option_.n_node_per_mv_feeder);
+                } else {
+                    // generate mv sym load
+                    SymLoadGenInput sym_load = mv_sym_load;
+                    sym_load.id = id_gen_++;
+                    sym_load.node = current_node_id;
+                    sym_load.type = static_cast<LoadGenType>((IntS)load_type_gen(gen_));
```
mgovers, line 245:
```suggestion
                    sym_load.type = static_cast<LoadGenType>(load_type_gen(gen_));
```

mgovers (commented):
I will apply these suggestions myself.

Can you also review once more?

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -259,40 +85,53 @@ struct PowerGridBenchmark {
...
-    benchmarker.run_benchmark(n_node, false, CalculationMethod::iterative_current, false);
+    option.has_mv_ring = false;
+    option.has_lv_ring = false;
+    benchmarker.run_benchmark<true>(option, newton_raphson, batch_size);
+    benchmarker.run_benchmark<true>(option, newton_raphson, batch_size, 6);
+    benchmarker.run_benchmark<true>(option, linear);
+    benchmarker.run_benchmark<true>(option, iterative_current);
+    benchmarker.run_benchmark<false>(option, newton_raphson);
+    benchmarker.run_benchmark<false>(option, linear);
+    // benchmarker.run_benchmark<false>(option, iterative_current);
```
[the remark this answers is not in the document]
TonyXiang8787: Iterative current method is less stable than Newton-Raphson. So this is expected that sometimes iterative current does not converge.

TonyXiang8787:
@mgovers  I agree with all your suggested changes, just push commits.

tests/benchmark_cpp/benchmark.cpp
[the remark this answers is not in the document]
mgovers: should we remove them entirely from the benchmark then?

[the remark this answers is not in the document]
TonyXiang8787: I would say keep it there to remind sometimes things do not converge.

mgovers:
@TonyXiang8787 all comments are resolved. i can approve after you have also done one more review round but AFAICT everything looks OK

TonyXiang8787:
@mgovers I have reviewed your changes. Approved

mgovers (approved):

PR #131, opened 2022-09-14
Change update for three phase values with NaN's 

### LINKED ISSUE #108: [FEATURE] Elements of a three-phase value can be set to NaN by applying update data
**Describe the problem**
When a three-phase (a.k.a. asymmetrical) value is updated with one or two NaN values, the original value is lost.

**To Reproduce**
In general, the logic for updating values is:
```C++
if (!is_nan(update_value)) {
    value = update_value;
}
```

Three phase values are tuples of three values like `(230.1, 230.2, 230.3)`. In the C++ code, a three phase value is considered a single value which can be NaN or non-NaN. Currently it is considered NaN iff **all** three values are NaN:

| value | is_nan |
|:-:|:-:|
| (NaN, NaN, NaN) | &#10004; |
| (NaN, 0.0, NaN)  | &#10060; |
| (0.0, NaN, 0.0)  | &#10060; |
| (0.0, 0.0, 0.0)  | &#10060; |

It follows that updating `(230.1, 230.2, 230.3)` with `(NaN, NaN, 0.0)` (which is not `NaN`) results in `(NaN, NaN, 0.0)`.

**Improved behavior**
After some discussion, we decided that updating `(230.1, 230.2, 230.3)` with `(NaN, NaN, 0.0)` should result in **`(230.1, 230.2, 0.0)`** as that gives the most flexibility and it follows the logic of updating scalar values.

**Additional remarks**
The python validation code follows the same behaviour as the C++ (when issue #91 is fixed), so make sure that the validation code is being updated as well.

petersalemink95:
Signed-off-by: Peter Salemink <peter.salemink@alliander.com>

### DIFF
--- a/include/power_grid_model/component/load_gen.hpp
+++ b/include/power_grid_model/component/load_gen.hpp
@@ -79,20 +79,12 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
 
     void set_power(RealValue<sym> const& new_p_specified, RealValue<sym> const& new_q_specified) {
         double scalar = direction_ / base_power<sym>;
-        //RealValue<sym> p = real(s_specified_);
-        //RealValue<sym> q = imag(s_specified_);
-        update_real_value<sym>(new_p_specified, s_specified_, scalar);
-        update_real_value<sym>(new_q_specified, s_specified_, scalar);
+        RealValue<sym> ps = real(s_specified_);
+        RealValue<sym> qs = imag(s_specified_);
+        update_real_value<sym>(new_p_specified, ps, scalar);
+        update_real_value<sym>(new_q_specified, qs, scalar);
 
-        //if (is_nan(new_p_specified))
-        //    ps = real(s_specified_);
-        //else
-        //    ps = direction_ / base_power<sym> * new_p_specified;
-        //if (is_nan(new_q_specified))
-        //    qs = imag(s_specified_);
-        //else
-        //    qs = direction_ / base_power<sym> * new_q_specified;
-        //s_specified_ = ps + 1.0i * qs;
+        s_specified_ = ps + 1.0i * qs;
     }
 
     // update for load_gen
--- a/include/power_grid_model/component/load_gen.hpp
+++ b/include/power_grid_model/component/load_gen.hpp
@@ -78,7 +78,7 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
     }
 
     void set_power(RealValue<sym> const& new_p_specified, RealValue<sym> const& new_q_specified) {
-        double scalar = direction_ / base_power<sym>;
+        double const scalar = direction_ / base_power<sym>;
         RealValue<sym> ps = real(s_specified_);
         RealValue<sym> qs = imag(s_specified_);
         update_real_value<sym>(new_p_specified, ps, scalar);
--- a/include/power_grid_model/component/load_gen.hpp
+++ b/include/power_grid_model/component/load_gen.hpp
@@ -78,16 +78,21 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
     }
 
     void set_power(RealValue<sym> const& new_p_specified, RealValue<sym> const& new_q_specified) {
-        RealValue<sym> ps, qs;
-        if (is_nan(new_p_specified))
-            ps = real(s_specified_);
-        else
-            ps = direction_ / base_power<sym> * new_p_specified;
-        if (is_nan(new_q_specified))
-            qs = imag(s_specified_);
-        else
-            qs = direction_ / base_power<sym> * new_q_specified;
-        s_specified_ = ps + 1.0i * qs;
+        double scalar = direction_ / base_power<sym>;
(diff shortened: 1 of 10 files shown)


include/power_grid_model/component/load_gen.hpp
```diff
@@ -78,15 +78,12 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
...
-        RealValue<sym> ps, qs;
-        if (is_nan(new_p_specified))
-            ps = real(s_specified_);
-        else
-            ps = direction_ / base_power<sym> * new_p_specified;
-        if (is_nan(new_q_specified))
-            qs = imag(s_specified_);
-        else
-            qs = direction_ / base_power<sym> * new_q_specified;
+        double scalar = direction_ / base_power<sym>;
```
TonyXiang8787, line 81: `double const`

include/power_grid_model/component/power_sensor.hpp
```diff
@@ -77,10 +77,13 @@ class PowerSensor : public GenericPowerSensor {
           {};
 
     UpdateChange update(PowerSensorUpdate<sym> const& power_sensor_update) {
-        if (!is_nan(power_sensor_update.p_measured) && !is_nan(power_sensor_update.q_measured)) {
-            s_measured_ = ((power_sensor_update.p_measured + 1i * power_sensor_update.q_measured) /
-                           base_power<sym>)*convert_direction();
-        }
+        double scalar = convert_direction() / base_power<sym>;
```
TonyXiang8787, line 80: `double const`

include/power_grid_model/component/voltage_sensor.hpp
```diff
@@ -70,15 +70,18 @@ class VoltageSensor : public GenericVoltageSensor {
           u_angle_measured_{voltage_sensor_input.u_angle_measured} {};
 
     UpdateChange update(VoltageSensorUpdate<sym> const& voltage_sensor_update) {
-        if (!is_nan(voltage_sensor_update.u_measured)) {
-            u_measured_ = voltage_sensor_update.u_measured / (u_rated_ * u_scale<sym>);
-        }
+        double scalar = 1 / (u_rated_ * u_scale<sym>);
```
TonyXiang8787, line 73: `double const`

try to define variable as `const` as possible

include/power_grid_model/component/voltage_sensor.hpp
```diff
@@ -70,15 +70,18 @@ class VoltageSensor : public GenericVoltageSensor {
...
+        RealValue<sym> u = u_measured_;
```
TonyXiang8787, line 74: In this case you don't need to define separate local variables, right?

include/power_grid_model/component/voltage_sensor.hpp
```diff
@@ -70,15 +70,18 @@ class VoltageSensor : public GenericVoltageSensor {
...
+        RealValue<sym> u_angle = u_angle_measured_;
+        update_real_value<sym>(voltage_sensor_update.u_measured, u, scalar);
```
TonyXiang8787, line 76: I think you can just call like
```c++
update_real_value<sym>(u_measured_, u, scalar);
```

without using a temporary variable.

include/power_grid_model/component/voltage_sensor.hpp
```diff
@@ -70,15 +70,18 @@ class VoltageSensor : public GenericVoltageSensor {
...
+        update_real_value<sym>(voltage_sensor_update.u_angle_measured, u_angle, 1.0);
```
TonyXiang8787, line 77: same here

include/power_grid_model/three_phase_tensor.hpp
```diff
@@ -321,6 +321,33 @@ inline bool is_nan(IntS x) {
...
+*/
+template <bool sym, class Proxy>
+void update_real_value(RealValue<sym> const& new_value, Proxy&& current_value, double scalar) {
+    if constexpr (sym) {
+        if (!is_nan(new_value)) {
+            current_value = scalar * new_value;
+        }
+    }
+    else {
+        for (size_t i = 0; i != 3; i++) {
```
TonyXiang8787, line 343: use `++i`

src/power_grid_model/validation/utils.py
```diff
@@ -109,12 +109,18 @@ def update_component_data(component: str, input_data: np.ndarray, update_data: n
...
+            for i in range(len(update_data[field])):
+                for phase in range(3):
+                    if mask[i][phase]:
+                        idx = np.where(input_data["id"] == update_data["id"][i])
+                        input_data[field][idx, phase] = update_data[field][i, phase]
+        else:
+            for i in range(len(update_data[field])):
+                if mask[i]:
+                    idx = np.where(input_data["id"] == update_data["id"][i])
+                    input_data[field][idx] = update_data[field][i]
```
TonyXiang8787, line 123: This is really an inefficient way of looping. For both algorithm complexity and Python tight loops. That explains why the validator so slow is.

I think we need to do something here.

TonyXiang8787 (approved):

PR #135, opened 2022-09-19
Move CI build options into pyproject.toml

TonyXiang8787:


bramstoeller (approved):
Nice! It's cool that the build options are now part of the poject config instead of the github workflow.

PR #139, opened 2022-09-26
fix bug on angle nan per phase for voltage sensor

TonyXiang8787:
Signed-off-by: Tony Xiang <tony.xiang@alliander.com>

### DIFF
--- a/include/power_grid_model/component/voltage_sensor.hpp
+++ b/include/power_grid_model/component/voltage_sensor.hpp
@@ -105,7 +105,14 @@ class VoltageSensor : public GenericVoltageSensor {
     }
 
     SensorCalcParam<false> asym_calc_param() const final {
-        bool const has_angle = !is_nan(u_angle_measured_);
+        bool const has_angle = [&]() {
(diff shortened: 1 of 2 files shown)


include/power_grid_model/component/voltage_sensor.hpp
```diff
@@ -91,10 +91,18 @@ class VoltageSensor : public GenericVoltageSensor {
...
+        else {
+            return !u_angle_measured_.isNaN().any();
+        }
+    }
+
     SensorCalcParam<true> sym_calc_param() const final {
-        bool const has_angle = !is_nan(u_angle_measured_);
         double u_variance = u_sigma_ * u_sigma_;
-        if (has_angle) {
+        if (has_angle()) {
```
petersalemink95, line 105: Since no unit test crashed when you changed this. Can you make a unit test that covers this behaviour?

petersalemink95 (changes requested):

petersalemink95 (approved):

PR #155, opened 2022-11-21
Calculation types explained

petersalemink95:


docs/user_manual/calculations.md
```diff
@@ -8,25 +8,83 @@ SPDX-License-Identifier: MPL-2.0
...
+- Power flow through branches
+
+#### State estimation
+State estimation is a statistical calculation method that determines the most probable state of the grid, based on
+network data and measurements. Measurements meaning power flow or voltage values with some kind of uncertainty, which were 
+either measured, estimated or forecasted.
+
+Input:
+- Network data: topology + component attributes
+- Power flow / voltage measurements with uncertainty
```
bramstoeller, line 35: I think it would be good to emphasize that _measurements_ don't have to be real measurements, but can be estimates or predictions as well. And maybe we can also emphasize at _uncertainty_ is not just being uncertain, but you have to specify _how_ uncertain your measurements are.

I now that this is all very logical, if you are familiar with the literature, like Phase to Phase, but a many users are not.

docs/user_manual/calculations.md
```diff
@@ -8,25 +8,83 @@ SPDX-License-Identifier: MPL-2.0
...
+$$
+
+The number of measurements can be found by the sum of the following:
+- number of nodes with a voltage sensor with magnitude only
+- two times the number of nodes with a voltage sensor with magnitude and angle
+- two times the number of nodes without appliances connected
+- two times the number of nodes where all connected appliances are measured by a power sensor
+- two times the number of branches with a power sensor
+
+Note: enough measurements doesn't necessarily mean that the system is observable. The location of the measurements is also
```
bramstoeller, line 68: Should we include a simple drawing of a net where it is obvious that the location of the measurements is important?

TonyXiang8787 (approved):

PR #158, opened 2022-11-30
batch calculation docs

nitbharambe:


petersalemink95:
I did a small update. Can you maybe include a really small example for independent and caching topology?

PR #165, opened 2022-12-15
Add configuration for VSCode, some minor change in CMake

TonyXiang8787:


.gitignore
```diff
@@ -486,7 +486,6 @@ dmypy.json
 .pyre/
 
 # build
-.vscode/
```
bramstoeller, line 489: I would keep this dir in the `.gitignore` file, so changes in these files are ignored, unless a user explicitly adds them with `git add .vscode/ --force`.

.vscode/extensions.json
```diff
@@ -0,0 +1,26 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
```
bramstoeller, line 2: Very strange... comments are not allowed in json. Apparently .vscode uses it's own json rules, ew... 😝

bramstoeller (approved):

PR #173, opened 2022-12-23
improve docstring readability

nitbharambe:


petersalemink95:
I'm not sure what I think of the fact that some docstring look different than others, just because they appear in readthedocs. I'd like all docstrings to be similar (even though both fit the Google style). Let's have a team discussion after the holiday

mgovers (dismissed):

TonyXiang8787:
@mgovers this is parked for now, there are still things to discuss

TonyXiang8787:
@nitbharambe this one can be closed with branch delete right?

PR #190, opened 2023-01-09
Create C library for the calculation core with C API/ABI

TonyXiang8787:


build.sh
```diff
@@ -22,7 +22,13 @@ else
   BUILD_COVERAGE=
 fi
 
-BUILD_DIR=cpp_build_$1
+if [[ $3 == "Sanitizer" ]]; then
```
bramstoeller, line 25: Suggestion: Use `getopts`
```bash
./build.sh -b Release -c -s
```
```bash
usage() {
  echo "Usage: $0 -b <Debug|Release> [-c] [-s]" 1>&2
  echo "  -c option enables coverage"
  echo "  -s option enables sanitizer"
  exit 1
}

while getopts "b::cs" flag; do
  case "${flag}" in
    b)
      [ "${OPTARG}" == "Debug" -o "${OPTARG}" == "Release" ] || echo "Build type should be Debug or Release."
      BUILD_TYPE=-DCMAKE_BUILD_TYPE=${OPTARG}
    ;;
    c) BUILD_COVERAGE=-DPOWER_GRID_MODEL_COVERAGE=1;;
    s) BUILD_SANITIZER=-DPOWER_GRID_MODEL_SANITIZER=1;;
    *) usage ;;
  esac
done

if [ -z "${BUILD_TYPE}" ] ; then
  usage
fi

echo "BUILD_TYPE = ${BUILD_TYPE}"
echo "BUILD_COVERAGE = ${BUILD_COVERAGE}"
echo "BUILD_SANITIZER = ${BUILD_SANITIZER}"
```

bramstoeller: In any case, update the current `usage` function.

include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -16,60 +16,77 @@ namespace power_grid_model {
...
+//          the set data is in the range [ i * length_per_batch, (i + 1) * length_per_batch )
+
 template <bool is_const>
 class DataPointer {
     template <class T>
     using ptr_t = std::conditional_t<is_const, T const*, T*>;
 
    public:
-    DataPointer() : ptr_{nullptr}, indptr_{nullptr}, size_{} {
+    DataPointer() : ptr_{nullptr}, indptr_{nullptr}, batch_size_{}, length_per_batch_{} {
```
bramstoeller, line 33: Maybe `items_per_batch` is more explanatory than `length_per_batch`

include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -169,23 +177,25 @@ struct MetaData {
...
         throw UnknownAttributeName{attr_name};
     }
 
-    bool has_attr(std::string const& attr_name) const {
-        try {
-            find_attr(attr_name);
-        }
-        catch (const UnknownAttributeName&) {
-            return false;
+    Idx find_attr(std::string const& attr_name) const {
```
bramstoeller, line 188: Nice refactor. Do we want to use the minimum value of Idx instead of -1, or do you find that redundant, as an actual Idx should always be positive?

TonyXiang8787: I think -1 should be enough here.

bramstoeller (dismissed):

bramstoeller (commented):
I discussed this PR together with Folkert, this is our joint review.

- Using the handles is a decent (and common) design choice.
- Allocating and releasing the buffers is common in C. We could register all allocations in the handle and let the handle release all the buffers, but that is not very common, so the users won't expect that. 
- It would be better to use enums instead of strings. It's faster and less prone to errors. If you can't do enum type checking, make sure that the values of the enums don't collide and don;t use 0 as a value. So for example:
```c
enum DatasetType = {INPUT = 1 << 12, UPDATE = 2 << 12, ...};
enum Component = {NODE = 1 << 8, LINE = 2 << 8, ...};
enum Attribute = {..., P_FROM = 11, P_TO = 12, ..., Q_RESIDUAL = 144};
```
- As a side effect we could even combine the values into a single int
```c
input_line_p_from = INPUT | LINE | P_FROM;
```
- The example should be split into several smaller examples.
- The examples could use more comments, especially explanation on `indptr`
- We should check for errors more consistently; i.e. use `assert(PGM_err_code(handle) == PGM_no_error);` more often.
- In the example where we expect an error, also check that `PGM_err_code(handle) != PGM_no_error`.
- The batch data part is a bit confusing for new users. Especially the index pointer.
   - For a user it makes more sense to supply an array of n_elements_per_scenario, than an index pointer (which is just the accumulated sum of the n array).
   - It would be nice if there are functions like `PGM_create_batch_buffer(..., uint_t n_elements_per_senario, uint_t n_scenarios)` and `PGM_create_sparse_batch_buffer(..., uint_t* n_elements_per_scenario, uint_t n_scenarios)`.
   - One could set the values using `PGM_batch_buffer_set_value(..., uint_t n_elements_per_senario, uint_t n_scenarios, unit_t i_scenario)` and  `PGM_sparse_batch_buffer_set_value(..., uint_t* n_elements_per_scenario, uint_t i_scenario)`.
   - The structure could also be stored in the handle, so the user doesn't have to supply the `n_elements_per_scenario`, and the index doesn't have to be recalculated every time.
 - In general, it would be good to help the user and utilize the IDE's auto completion capabilities to show which options a user has. I.e. using enums instead of strings.
 - We are using full function names, so consider to use `PGM_error_message` instead of `PGM_err_msg`.

PR #192, opened 2023-01-16
Use ctypes as binding to the calculation core

TonyXiang8787:


src/power_grid_model/errors.py
```diff
@@ -0,0 +1,62 @@
...
+    error_messages: List[str]
+
+
+def find_error() -> Optional[ValueError]:
+    """
+
+    Returns:
+
+    """
+    error_code: int = pgc.err_code()
```
bramstoeller, line 34: I'm not really sure bout this global error handling, can't we supply some handle/instance?

src/power_grid_model/errors.py
```diff
@@ -0,0 +1,62 @@
...
+        failed_idxptr = pgc.failed_scenarios()
+        failed_msgptr = pgc.batch_errs()
+        error.failed_scenarios = np.ctypeslib.as_array(failed_idxptr, shape=(n_fails,)).copy()
+        error.error_messages = [failed_msgptr[i].decode() for i in range(n_fails)]  # type: ignore
+        return error
+    else:
+        return ValueError("Unknown error!")
+
+
+def assert_error():
```
bramstoeller, line 54: Assert means, 'make sure'. So this function name is something like "make sure that there is an error". I think `assert_no_errors()` would be the appropriate name.

src/power_grid_model/errors.py
TonyXiang8787: The user is not supposed to call this directly. It is part of the back-end core. Since we can't do multi-threading in Python, a global handle is enough.

In the new commit I have moved all core part in a separate `core` folder.

pyproject.toml
```diff
@@ -85,6 +82,11 @@ line_length = 120
 
 [tool.pylint]
 max-line-length = 120
+disable = [
```
bramstoeller, line 85: You can also write `too-few-public-methods` and `too-many-arguments`, instead of the error codes.

Nevertheless, I'm not sure we should do this globally... can you explain why you'd like to do that? And maybe add a short command behind the exclusions.

setup.py
```diff
@@ -32,6 +27,21 @@
...
+    pass
+
+
+class bdist_wheel_abi_none(bdist_wheel):
+    def finalize_options(self):
+        bdist_wheel.finalize_options(self)
+        self.root_is_pure = False
+
+    def get_tag(self):
+        python, abi, plat = bdist_wheel.get_tag(self)
```
bramstoeller, line 41: What is `plat`? And what is `abi`?
If we're not using `python` and `abi`, please prefix (or replace) with underscores.

setup.py
```diff
@@ -11,17 +11,12 @@
...
-import numpy as np
-
 # noinspection PyPackageRequirements
 import requests
 from pybuild_header_dependency import HeaderResolver
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
+from wheel.bdist_wheel import bdist_wheel
 
 # determine platform, only windows or linux
```
bramstoeller, line 21: Global variable ìf_win` should be upper case.

setup.py
```diff
@@ -32,6 +27,21 @@
     raise SystemError("Only Windows, Linux, or MacOS is supported!")
 
 
+# custom class for ctypes
+class CTypesExtension(Extension):
+    pass
+
+
+class bdist_wheel_abi_none(bdist_wheel):
```
bramstoeller, line 35: class names should be PascalCase

src/power_grid_model/core/__init__.py
```diff
@@ -0,0 +1,3 @@
+# SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model project <dynamic.grid.calculation@alliander.com>
```
bramstoeller, line 1: 2023?

src/power_grid_model/errors.py
bramstoeller: What do you mean by we "can't do multi-threading in Python"?

src/power_grid_model/core/index_integer.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+Definition of integers used by the calculation core
+"""
+
+# define internal index integer
+from ctypes import c_int32, c_int64
+
+import numpy as np
+
+# pylint: disable=C0103
```
bramstoeller, line 14: Please use `invalid-name`.

src/power_grid_model/core/options.py
```diff
@@ -0,0 +1,68 @@
...
+    @property
+    def opt(self) -> OptionsPtr:
+        """
+
+        Returns: Pointer to the option object
+
+        """
+        return self._opt
+
+    def __new__(cls, *args, **kwargs):
```
bramstoeller, line 55: What is the intention of this function? Why call super(), options doesn't have a super class. Why not just implement __init__()

src/power_grid_model/core/index_integer.py
bramstoeller: On the other hand... The name is invalid... It's a global definition, so it should actually be `IDX_C`.

src/power_grid_model/core/power_grid_core.py
```diff
@@ -0,0 +1,220 @@
...
+    # not copyable
+    def __copy__(self):
+        raise NotImplementedError("Class not copyable")
+
+    def __deepcopy__(self, memodict):
+        raise NotImplementedError("class not copyable")
+
+
+# make one instance
+power_grid_core = PowerGridCore()
```
bramstoeller, line 220: `PowerGridCore` should be private, can we call it `_PowerGridCore`? Is that a valid python name? or we could rename the file to _power_grid_code.py and create the instance in the `__init__.py`. But, to be honest, I'm not a big fan of these one-instance setups, is it get's instantiated on import. I'd prefer a singleton construction, e.g. something like this:

```python
class PowerGridCore:
    _instance: Optional["PowerGridCore"] = None

def __new__():
    if PowerGridCore._instance is None:
        PowerGridCore._instance = super().__new__()
    return PowerGridCore._instance
```

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -0,0 +1,261 @@
...
+from typing import Dict, Mapping, Optional, Union
+
+import numpy as np
+
+from power_grid_model.core.error_handling import VALIDATOR_MSG
+from power_grid_model.core.index_integer import Idx_c, Idx_np
+from power_grid_model.core.power_grid_core import IdxPtr
+from power_grid_model.core.power_grid_core import power_grid_core as pgc
+
+_ctype_numpy_map = {"double": "f8", "int32_t": "i4", "int8_t": "i1", "double[3]": "(3,)f8"}
```
bramstoeller, line 20: `_CTYPE_NUMPY_MAP`

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -0,0 +1,261 @@
...
+_ENDIANNESS = "<" if pgc.is_little_endian() == 1 else ">"
+_nan_value_map = {
```
bramstoeller, line 22: `_NAN_VALUE_MAP`

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def _generate_meta_data() -> dict:
+    """
+
+    Returns: meta data for all dataset
+
+    """
+    py_meta_data = {}
+    n_dataset = pgc.meta_n_datasets()
```
bramstoeller, line 37: `n_datasets`

bramstoeller (commented):
In general, I think there are too many `type: ignore` and `pylint: disable` statements and I think we can improve type hinting. I also feel that we should explicitly write out the member/static methods in PowerGridCore.

Currently the methods are defines as member functions, while they are implemented as static methods. That's where all the `Invalid self argument` warnings come from in mypy.

I don't fully understand why PowerGridCore should be a singleton instead of a normal class. And even if it should be a singleton, it would make things much clearer if it was implemented as a normal class, with member function.

src/power_grid_model/core/power_grid_core.py
```diff
@@ -0,0 +1,220 @@
...
+    destroy_model: Callable[[ModelPtr], None]
+
+    def __new__(cls, *args, **kwargs):
+        instance = super().__new__(cls, *args, **kwargs)
+        instance._cdll = _load_core()
+        instance._handle = instance._cdll.PGM_create_handle()
+        return instance
+
+    def __init__(self):
+        for name, function in PowerGridCore.__annotations__.items():  # pylint: disable=E1101
```
bramstoeller, line 178: Can we handle this with a decorator?

For example:
```python
@cbinding
def meta_dataset_name(self, dataset_idx: int) -> str:
    pass
```
or (my preference) something like:
```python
@cbinding
def meta_dataset_name(self, dataset_idx: int) -> str:
    return self._cdll.PGM_meta_dataset_name(self._handle, dataset_idx).decode()
```

bramstoeller: I.e. I would move the `argtypes` part from the for loop in `__init__` to the decorator function. And I would like to handle the type conversions (currently defined in WrapperFunc) explicitly in the function body.

src/power_grid_model/validation/utils.py
```diff
@@ -164,7 +164,8 @@ def nan_type(component: str, field: str, data_type="input"):
     Helper function to retrieve the nan value for a certain field as defined in the power_grid_meta_data.
     It silently returns float('nan') if data_type/component/field can't be found.
     """
-    return power_grid_meta_data.get(data_type, {}).get(component, {}).get("nans", {}).get(field, float("nan"))
+    # return power_grid_meta_data.get(data_type, {}).get(component, {}).nans.get(field, float("nan"))
```
bramstoeller, line 167: Do we still want to support a default nan value for missing fields?

bramstoeller, line 167: minor: remove comment

tests/unit/test_power_grid_model.py
```diff
@@ -0,0 +1,136 @@
...
+update_1:
+    q_specified = 300 var (-j3.0A)
+u0 = 100.0 V - (j10.0 ohm * -j3.0 A) = 70.0 V
+"""
+
+
+@pytest.fixture
+def case_data():
+    case_dir = Path(__file__).parent.parent / "data" / "unit_test"
+    return import_case_data(case_dir, sym=True)
```
bramstoeller, line 38: I think for unit tests it would be better to include the data in the test file instead of separate .json files.

src/power_grid_model/core/power_grid_core.py
```diff
@@ -0,0 +1,335 @@
...
+    """
+    DLL caller
+    """
+
+    _handle: HandlePtr
+    _instance: Optional["PowerGridCore"] = None
+
+    # singleton of power grid core
+    def __new__(cls, *args, **kwargs):
+        if PowerGridCore._instance is None:
```
bramstoeller, line 144: You could use `cls` instead of `PowerGridCode`. I don't have a strong preference.

TonyXiang8787: @bramstoeller  done.

bramstoeller (approved):

bramstoeller (approved):

PR #202, opened 2023-02-08
Feature: node injection

bramstoeller:


tests/cpp_unit_tests/CMakeLists.txt
```diff
@@ -8,29 +8,29 @@ find_package(doctest REQUIRED)
...
+	test_math_solver.cpp
+	test_topology.cpp
+	test_container.cpp
+	test_sparse_mapping.cpp
+	test_meta_data_generation.cpp
+	test_voltage_sensor.cpp
+	test_power_sensor.cpp
 	test_validation.cpp
-    test_three_winding_transformer.cpp
+	test_three_winding_transformer.cpp
```
TonyXiang8787, line 33: maybe you can use a `glob` to just get all `cpp` files in this folder.

bramstoeller: Yes, we could do that, but I like to disable test source files while developing, which makes the building process much faster.

TonyXiang8787:
@bramstoeller You need to add the new enum of measured terminal in the documentation.

include/power_grid_model/math_solver/iterative_pf_solver.hpp
```diff
@@ -121,6 +123,7 @@ class IterativePFSolver {
                 ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
                 output.source[source].i = dot(y_ref, u_ref - output.u[bus]);
                 output.source[source].s = output.u[bus] * conj(output.source[source].i);
+                output.bus_injection[bus] += output.source[source].s;
```
TonyXiang8787, line 126: Please calculate bus injection with the function `calculate_injection` from `YBus`

tests/cpp_unit_tests/test_math_solver.cpp
```diff
@@ -8,6 +8,8 @@
 #include "power_grid_model/math_solver/newton_raphson_pf_solver.hpp"
 #include "power_grid_model/three_phase_tensor.hpp"
 
+using namespace std::placeholders;
```
TonyXiang8787, line 11: Why do you need this?

Maybe put the using namespace inside the `power_grid_model` namespace?

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -22,7 +22,25 @@ namespace power_grid_model {
...
-// block class for the unknown vector and/or right hand side in state estimation equation
+constexpr Idx DISCONNECTED = -1;
+constexpr Idx UNMEASURED = -2;
+constexpr Idx UNDEFINED = -3;
+
+// struct to store bus injection information
+struct BusInjection {
+    // The index in main_value_ where the total measured bus injection is stored.
+    // This includes node injection measurements, source power measurements and load/gen power measurements.
+    Idx idx_bus_injection = UNDEFINED;
```
TonyXiang8787, line 33: use curly brace to set default value of struct
```
    Idx idx_bus_injection{UNDEFINED};

```

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -22,7 +22,25 @@ namespace power_grid_model {
...
+
+    // The index in main_value_ where the measured appliance injection is stored.
+    // This includes source power measurements and load/gen power measurements.
+    Idx idx_appliance_injection = UNDEFINED;
```
TonyXiang8787, line 37: same here

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -22,7 +22,25 @@ namespace power_grid_model {
...
+
+    // The number of unmeasured appliances
+    Idx n_unmeasured_appliances = 0;
```
TonyXiang8787, line 40: same here

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -22,7 +22,25 @@ namespace power_grid_model {
 // hide implementation in inside namespace
 namespace math_model_impl {
 
-// block class for the unknown vector and/or right hand side in state estimation equation
+constexpr Idx DISCONNECTED = -1;
```
TonyXiang8787, line 25: maybe define these values inside the `struct BusInjection` as `static constexpr` and make them small letters?

TonyXiang8787: Okay, you also use it for other places besides bus injection. Then just put here. I still don't think we need all capital letters. It is scoped inside namespace.

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -22,7 +22,25 @@ namespace power_grid_model {
...
+constexpr Idx UNMEASURED = -2;
+constexpr Idx UNDEFINED = -3;
+
+// struct to store bus injection information
+struct BusInjection {
+    // The index in main_value_ where the total measured bus injection is stored.
+    // This includes node injection measurements, source power measurements and load/gen power measurements.
+    Idx idx_bus_injection = UNDEFINED;
+
+    // The index in main_value_ where the measured appliance injection is stored.
```
TonyXiang8787, line 35: You mean `extra_value_` here right?

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -249,30 +269,27 @@ class MeasuredValues {
     // not relevant for the main calculation, as extra data for load_gen/source calculation
     std::vector<SensorCalcParam<sym>> extra_value_;
     // array of partial injection measurement, the bus with not all connected appliances measured
-    std::vector<SensorCalcParam<sym>> partial_injection_;
+    std::vector<SensorCalcParam<sym>> partial_load_injection_;
```
TonyXiang8787, line 272: why this change?

Partial injection includes load, gen and source injection together, if at least one connected source or load or gen is not measured.

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -249,30 +269,27 @@ class MeasuredValues {
...
 
     // indexing array of the entries
-    // for -1 (non bus injection): connected, but no measurement
-    // for -2 (non bus injection): not connected
+    // for UNMEASURED (non bus injection): connected, but no measurement
+    // for DISCONNECTED (non bus injection): not connected
     // for <0 (bus injection): the number of unmeasured appliances in negative
-    // 	   -2 means 2 appliances are unmeasured, bus-appliance is under-determined
     //     -1 means 1 appliance is unmeasured, bus-appliance is exactly determined
+    // 	   -2 means 2 appliances are unmeasured, bus-appliance is under-determined
```
TonyXiang8787, line 279: This comment block needs to be changed since you use the new values. And you don't use negative number for unmeasured appliance anymore.

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -171,8 +188,11 @@ class MeasuredValues {
     }
 
     // power measurement
-    SensorCalcParam<sym> const& bus_injection_power(Idx bus) const {
-        return main_value_[idx_bus_injection_power_[bus]];
+    SensorCalcParam<sym> const& bus_injection(Idx bus) const {
+        return main_value_[bus_injection_[bus].idx_bus_injection];
+    }
+    SensorCalcParam<sym> const& appliance_injection(Idx bus) const {
```
TonyXiang8787, line 194: I have difficulty to understand why you introduce the concept of `appliance_injection`. It seems to be vague.

If a bus injection if fully measured, we can refer to the `main_value_`, if it is not fully measured, we refer to `partial_injection_`.

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -674,7 +704,7 @@ class IterativeLinearSESolver {
                 // get data idx of y bus,
                 // skip for a fill-in
                 Idx const data_idx = y_bus.map_lu_y_bus()[data_idx_lu];
-                if (data_idx == -1) {
+                if (data_idx == DISCONNECTED) {
```
TonyXiang8787, line 707: This does not make sense. `-1` here means different thing.

TonyXiang8787: It's then a good reason to put the contant definition inside `MeasuredValues`. So it is scoped and will not be misused.

include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -741,7 +771,7 @@ class IterativeLinearSESolver {
         // assign the hermitian transpose of the transpose entry of Q
         for (Idx data_idx_lu = 0; data_idx_lu != y_bus.nnz_lu(); ++data_idx_lu) {
             // skip for fill-in
-            if (y_bus.map_lu_y_bus()[data_idx_lu] == -1) {
+            if (y_bus.map_lu_y_bus()[data_idx_lu] == DISCONNECTED) {
```
TonyXiang8787, line 774: same here

tests/cpp_unit_tests/test_math_solver.cpp
bramstoeller: I'm using the `_1` and `_2` placeholder in the unit test to sum the `s` attributes of all (one or two) `load_gens`.
```c++
const ComplexValue<true> load_gen_s = std::accumulate(
    output.load_gen.begin(),
    output.load_gen.end(),
    ComplexValue<true>{},
    bind(std::plus<ComplexValue<true>>(), _1, bind(&ApplianceMathOutput<true>::s, _2)));
```
And since it is inside a unit test I didn't really worry about namespaces. But I moved it inside the `power_grid_model ` namespace now.

bramstoeller:
> @bramstoeller You need to add the new enum of measured terminal in the documentation.

I think that should be generated automatically, as the only reference to `MeasuredTerminalType` I can find in the docs folder is:
```
{py:class}`MeasuredTerminalType <power_grid_model.enum.MeasuredTerminalType>`
```

TonyXiang8787:
@bramstoeller  In the pipeline you have a buffer overflow in Linux and macOS. Please fix that.

Another TODO is adding new node output attributes into component documentation.

TonyXiang8787 (approved):

PR #203, opened 2023-02-08
Use constexpr instead of magic numbers

bramstoeller:


include/power_grid_model/main_model.hpp
```diff
@@ -700,7 +702,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     template <bool sym>
     BatchParameter calculate_power_flow(double err_tol, Idx max_iter, CalculationMethod calculation_method,
                                         Dataset const& result_data, ConstDataset const& update_data,
-                                        Idx threading = -1) {
+                                        Idx threading = State::undefined) {
```
TonyXiang8787, line 705: This will be breaking.... The Python side user expect to have -1 as threading sequential. Maybe just define State::sequential = -1

bramstoeller: Doesn't -1 mean "as many threads as cpu's"?

bramstoeller: O, no, that's 0...

bramstoeller:
Too much hassle to fix this right now... Maybe later (:

PR #214, opened 2023-03-07
add short circuit component

petersalemink95:


code_generation/requirements.txt
```diff
@@ -0,0 +1,2 @@
+jinja2
+dataclasses_json
```
TonyXiang8787, line 2: Nitish already add these two in the dev dependency. We don't need this file here.

code_generation/requirements.txt.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model project <dynamic.grid.calculation@alliander.com>
+
```
TonyXiang8787, line 2: Nitish already add these two in the dev dependency. We don't need this file here.

include/power_grid_model/calculation_parameters.hpp
```diff
@@ -59,6 +59,17 @@ struct BranchMathOutput {
     ComplexValue<sym> i_t;
 };
 
+// fault math calculation parameters and math output
+struct FaultCalcParam {
+    Idx math_fault_object{-1};
+    DoubleComplex y_fault;
+};
+
```
TonyXiang8787, line 67: we need to define a `ShortCircuitMathInput`, which is actually a collection of faults. This will be the math input.

```c++
struct ShortCircuitMathInput {
    std::vector<FaultCalcParam> faults;
};
```

include/power_grid_model/calculation_parameters.hpp
```diff
@@ -59,6 +59,17 @@ struct BranchMathOutput {
...
+template <bool sym>
+struct FaultMathOutput {
+    ComplexValue<sym> i_f;
```
TonyXiang8787, line 70: We need also more stuff here for 1) faults, 2) buses, 3) branches, 4) sources.

```c++
template <bool sym>
struct FaultMathOutput {
    std::vector<ComplexValue<sym>> i_fault;
    std::vector<ComplexValue<sym>> u_bus;
    std::vector<ComplexValue<sym>> i_branch_from;
    std::vector<ComplexValue<sym>> i_branch_to;
    std::vector<ComplexValue<sym>> i_source;
};
```

include/power_grid_model/component/fault.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include "base.hpp"
+
+namespace power_grid_model {
+
+class Fault final : public Base {
+   public:
+    using InputType = FaultInput;
+    using UpdateType = FaultUpdate;
+    template <bool sym>
+    using OutputType = FaultOutput;
```
TonyXiang8787, line 22: also define a `ShortCircuitOutputType`
```c++
    template <bool sym>
    using ShortCircuitOutputType= FaultShortCircuitOutput<sym>;
```

code_generation/requirements.txt
TonyXiang8787: then remove the dependency in the toml

include/power_grid_model/component/fault.hpp
```diff
@@ -0,0 +1,100 @@
...
+    static constexpr char const* name = "fault";
+    ComponentType math_model_type() const final {
+        return ComponentType::fault;
+    }
+
+    Fault(FaultInput const& fault_input)
+        : Base{fault_input},
+          fault_object_{fault_input.fault_object},
+          r_f_{is_nan(fault_input.r_f) ? (bool)0.0 : fault_input.r_f},
+          x_f_{is_nan(fault_input.x_f) ? (bool)0.0 : fault_input.x_f} {
```
TonyXiang8787, line 32: same here

include/power_grid_model/component/fault.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    FaultCalcParam calc_param(double const& u_rated, bool const& is_connected_to_source = true) const {
+        if (!energized(is_connected_to_source)) {
+            return FaultCalcParam{};
+        }
+        // param object
+        FaultCalcParam param{};
+        // calculate the fault admittance in p.u.
+        double const base_y = base_power_3p / u_rated / u_rated;
+        DoubleComplex y_f = 1.0 / (r_f_ + 1.0i * x_f_) / base_y;
```
TonyXiang8787, line 43: Can you do a research and/or experiment to be sure that if you do `1.0 / z`  where `z` is a zero complex number, the result is indeed `inf` for both real and imaginary part?

Also check if a `inf / base_y` also results in `inf`.

TonyXiang8787 (changes requested):
Apart from the comments, also add fault object into the component documentation.

include/power_grid_model/component/fault.hpp
TonyXiang8787: I have Googled a bit. The behaviour is not actually consistant. Sometimes you get `NaN`, sometimes you get `inf`. Maybe it is good to define `z_f` in the math input? And we can check if `z_f == 0.0` in the core to decide our operation.

petersalemink95: implemented it in this function instead

include/power_grid_model/component/fault.hpp
```diff
@@ -0,0 +1,109 @@
...
+        // calculate current magnitude and angle
+        output.i_f = cabs(i_f);
+        output.i_f_angle = arg(i_f);
+        return output;
+    }
+
+    // update faulted object
+    UpdateChange update(FaultUpdate const& update) {
+        assert(update.id == id());
+        if (update.fault_object != na_IntS) {
```
TonyXiang8787, line 85: you should compare against `na_ID`.

include/power_grid_model/component/fault.hpp
```diff
@@ -0,0 +1,109 @@
...
+          r_f_{is_nan(fault_input.r_f) ? (double)0.0 : fault_input.r_f},
+          x_f_{is_nan(fault_input.x_f) ? (double)0.0 : fault_input.x_f} {
+    }
+
+    FaultCalcParam calc_param(double const& u_rated, bool const& is_connected_to_source = true) const {
+        if (!energized(is_connected_to_source)) {
+            return FaultCalcParam{};
+        }
+        // param object
+        FaultCalcParam param{};
```
TonyXiang8787, line 42: If the fault impedance is zero, which is quite common, you still calculate `base_y`. We can improve the performance here by doing two separate returns. 

Also, try not to define separate variables for each return path. This will prevent the Return Value Optimization.

Suggestion:


```c++
    FaultCalcParam calc_param(double const& u_rated, bool const& is_connected_to_source = true) const {
        // param object
        FaultCalcParam param{};  // only declare one object to be returned, this enables RVO.
        if (!energized(is_connected_to_source)) {
            return param;
        }
        if (r_f_ == 0.0 && x_f_ == 0.0) {
           param.y_fault.real(std::numeric_limits<double>::infinity());
           param.y_fault.imag(std::numeric_limits<double>::infinity());
           return param;
        }
        double const base_z = u_rated * u_rated / base_power_3p;
        param.y_fault = base_z  / (r_f_ + 1.0i * x_f_);
        return param;
    }
```

include/power_grid_model/component/fault.hpp
```diff
@@ -0,0 +1,109 @@
...
+        return output;
+    }
+    FaultOutput get_output() const {
+        // During power flow and state estimation the fault object will have an empty output
+        return get_null_output();
+    }
+
+    // energized
+    template <bool sym>
+    FaultShortCircuitOutput<sym> get_short_circuit_output(ComplexValue<sym> i_f, double const u_rated) {
```
TonyXiang8787, line 69: This is also a `const` function

TonyXiang8787 (approved):

PR #219, opened 2023-03-14
Output short circuit attributes

nitbharambe:


petersalemink95:
Seems logical to me, I've requested @TonyXiang8787 's review to double check

nitbharambe:
Right now it is:

base  -- |--> branch 
              |--> shortcircuitbranch
        
 
Is it a good choice to do this way instead?:
base --> branch  --|--> PowerFlowbranch
                               |--> shortcircuitbranch
                           
Then they would share output variables, one issue of having a separate but same `get_null_output()` function will be resolved. But the meaning behind these current and voltages are different

code_generation/data/attribute_classes/output.json
```diff
@@ -125,6 +125,73 @@
...
+            ]
+        },
+        {
+            "name": "BranchShortCircuitOutput",
+            "base": "BaseOutput",
+            "is_template": true,
+            "attributes": [
+                {
+                    "data_type": "RealValue<sym>",
+                    "names": ["i_from"],
```
petersalemink95, line 152: I'd like to have single values not written as a list, so:
`"names": "i_from",`

Happen on more places ;)

petersalemink95:
@nitbharambe answer to your suggestion to have:
"ShortCircuitBranchOutput and PowerFlowOutput derive from BranchOutput"

I don't think that's a good idea, since `BranchOutput` has attributes that are not needed in `ShortCircuitBranchOutput `

TonyXiang8787 (changes requested):
@nitbharambe two TODO's:

1. add angles to all current results (branch, branch3, appliance)
2. update the documentation in the component table. For each component, use three separate tables: input/update, steady-state output, short circuit output

TonyXiang8787 (changes requested):
@nitbharambe some open points:

1. In all input attribute tables, remove the column `input`. This does not make sense as it will always have a check sign.
2. Add a separate section between `Component Type Hierarchy and Graph Data Model` and `Components` to explain the structure of the datasets, including the concepts of dataset, component, attribute and value. As we have described in https://alliander.atlassian.net/wiki/spaces/SASOG/pages/3385131970/Glossary
3. In that section explain the current supported dataset types, namely `input`, `update`, `sym_output`, `asym_output`, `sym_sc_output`, `asym_sc_output`. Explain their intended usage.

TonyXiang8787 (approved):

PR #220, opened 2023-03-14
add dataclasses_json to pyproject

nitbharambe:
Does it make sense to add this? Close if this is irrelevant.
I get import module error when I use `code_gen.py`


TonyXiang8787:
@nitbharambe it is okay to include this. Maybe also include `jinja2`? 
https://pypi.org/project/Jinja2/

TonyXiang8787 (approved):

PR #224, opened 2023-03-23
Short circuit solver

petersalemink95:


include/power_grid_model/math_solver/short_circuit_solver.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+        // getter
+        ComplexTensorVector<sym> const& ydata = y_bus.admittance();
+        IdxVector const& bus_entry = y_bus.lu_diag();
+        // output
+        ShortCircuitMathOutput<sym> output;  // TODO: resize output values that are updated
+
+        // copy y_bus data
+        std::transform(y_bus.map_lu_y_bus().cbegin(), y_bus.map_lu_y_bus.cend(), mat_data_.begin(), [&](Idx k) {
+            if (k == -1) {
```
mgovers, line 65: does it make sense to use this? or are there use cases where a different `k` would be used?

```suggestion
            if (k >= 0) {
```

otherwise, maybe make this a `static constexpr` named value, e.g. `static constexpr newIdx = -1`

mgovers:
code smells are `std::format` which is not possible to (cleanly) solve as of now due to clang lagging behind on its implementation.

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
```diff
@@ -34,6 +35,15 @@ class Fault final : public Base {
...
           x_f_{is_nan(fault_input.x_f) ? double{} : fault_input.x_f} {
+        if (fault_type_ == FaultType::three_phase) {
+            if (fault_phase_ != FaultPhase::abc)
+                throw InvalidShortCircuitPhases(fault_type_, fault_phase_);
+        }
+        else {
+            if (fault_phase_ == FaultPhase::abc) {
+                throw InvalidShortCircuitPhases(fault_type_, fault_phase_);
+            }
+        }
```
TonyXiang8787, line 46: @mgovers  please move this check in a separate function. And call the function both in the constructor and the `update` function.

power_grid_model_c/power_grid_model/include/power_grid_model/exception.hpp
```diff
@@ -160,6 +161,23 @@ class UnknownAttributeName : public PowerGridError {
     }
 };
 
+class InvalidShortCircuitType : public PowerGridError {
```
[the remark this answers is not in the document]
TonyXiang8787: It will be used later

.github/workflows/main.yml
```diff
@@ -112,6 +112,7 @@ jobs:
 
       - name: Build and test
         run: |
+          $ErrorActionPreference = 'Stop'
```
mgovers, line 115: test failed but CI passed. see also https://github.com/actions/runner-images/issues/6668#issuecomment-1364540817

mgovers: doesn't work. can repro locally. i will create a PR on `main` after i find a fix

mgovers:
```suggestion
```

mgovers: see https://github.com/PowerGridModel/power-grid-model/pull/287

TonyXiang8787 (approved):

PR #230, opened 2023-04-07
Feature/cmake presets CI

mgovers:
* Add configuration + build presets for CI and Linux / MacOS development
* Update the `build.sh`, CI workflows and documentation

NOTE: this PR is still in draft because alignment is required on whether we want to go down this route for CI. TBD

* Pros:
  * all developers have the exact same compiler settings
  * if desired, developers can configure their settings however they want
  * the `CMakeLists.txt` (enforced code path) is clean
* Cons:
  * the `CMakePresets.json` file is quite long and therefore not always readable


mgovers:
approved verbally in quick call. merging

PR #232, opened 2023-04-07
Fix read the docs equations

petersalemink95:
In the read the docs documentation, the equations exceed the line length. This should fix that


petersalemink95:
Before merging, we need to check the alignment of the documentation in read the docs

petersalemink95:
Can be reviewed. The extra benefit is that the read the docs page now does not have a fixed line width, but scales with the page size :)

TonyXiang8787 (approved):

PR #233, opened 2023-04-10
Short circuit validation case

nitbharambe:
Added 3 phase now. Will add others later


TonyXiang8787 (changes requested):
Two comments:

1. You need to add more values in the `params.json`. In short circuit function arguments you need to specify faulted phase and fault type. These two arguments have to be present in `params.json`. So that the test script will run the short circuit calculation with correct arguments.
2. The output should not be called `asym_output` or `sym_output`. That's for steady state calculation. The dataset name for short circuit results are `asym_sc_output` and `sym_sc_output`.

TonyXiang8787 (changes requested):
1. Why does three phase test case have a `asym_sc_output`?
2. The fault type and fault phase in the `params.json`, the name should match exactly as in the enumeration in https://github.com/PowerGridModel/power-grid-model/blob/feature/short-circuit-calculator/power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp

TonyXiang8787:
@nitbharambe I mean. The fault type and fault phase can be a string (does not have to be integer). But the name of the string should be exactly the same as the enumeration name.

TonyXiang8787:
@nitbharambe  now only one comment
Why does three phase test case have a asym_sc_output?

TonyXiang8787 (approved):

mgovers:
> 1. Why does three phase test case have a `asym_sc_output`?

why shouldn't it? it's a valid test case, right? I know the concept of `asym` doesn't make sense for a three phase fault, but it feels weirder to not have this test case.

> 2. The fault type and fault phase in the `params.json`, the name should match exactly as in the enumeration in https://github.com/PowerGridModel/power-grid-model/blob/feature/short-circuit-calculator/power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp

cfr. discussion in https://github.com/PowerGridModel/power-grid-model/pull/273#discussion_r1230907750 , I am unsure whether `params.json` is the right place to put these `fault_type` and `fault_phase` parameters

PR #234, opened 2023-04-11
split core library from C API

mgovers:
* Move all core PGM functionality to `power_grid_model_c/power_grid_model/`
  * headers can be included using `#include <power_grid_model/*.hpp>`
* Move C-API to `power_grid_model_c/power_grid_model_c`
  * header can be included using `#include "power_grid_model_c.h"`
* Split C-API tests (actual integration tests) and C++ unit tests
* Move library code to respective library-specific `CMakeLists.txt`
* Link to `Boost::boost` instead of using `include_directories(${Boost_INCLUDE_DIRS})`
* Link `Threads::Threads` if found

PR #238, opened 2023-04-13
Feature/install targets

mgovers:
* Make C API installable
* Make C API findable using `find_package`
* Install license
* Add a separate project that uses `find_package` on an installed PGM package and communicates with the C API to check whether it is available

NOTE: no github actions or install script were updated yet because that would require dependency on https://github.com/PowerGridModel/power-grid-model/pull/230 while this PR already depends on https://github.com/PowerGridModel/power-grid-model/pull/234


CMakeLists.txt
```diff
@@ -25,6 +28,8 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 include(CTest)
 enable_testing()
 
+include(GNUInstallDirs)
```
TonyXiang8787, line 31: What is this? Does it work also in Windows?

TonyXiang8787 (approved):

CMakeLists.txt
mgovers: yes and no 😄 it's just a generic standard CMake configuration file that just so happens to be developed by GNU. What it does is set a lot of CMake definitions and variables for installation, mainly for Linux (but with platform compatibility). E.g. shared libraries should have their RPATH set if that property exists (which means: Unix). See also the documentation in https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

For the Windows shared libraries, the user dynamically linking to this DLL, the `install(IMPORTED_RUNTIME_ARTIFACTS library)` already does the job

mgovers: NOTE: for Unix, the user also still needs to set `set_target_properties(<target> PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE)` to ensure that the properties propagate accordingly

PR #243, opened 2023-04-18
Feature/separate validation test

mgovers:
* separate the validation tests from the cpp unit tests
* combine all test projects into single test project with multiple targets
* remove `add_project`
  * `add_subdirectory` already does that for you
  * the new CMake Targets view in Visual Studio and Visual Studio Code now has the same structure as the folder view


tests/cpp_unit_tests/CMakeLists.txt
TonyXiang8787, line 1: We can remove the `power_grid_model_c` link target here I guess?
```cmake
target_link_libraries(power_grid_model_unit_tests
	PRIVATE power_grid_model
	PRIVATE power_grid_model_c
	PRIVATE doctest::doctest
	PRIVATE nlohmann_json nlohmann_json::nlohmann_json
)
```

mgovers: yes, you're absolutely right. i did that but then those changes were lost when i reverted them 😬 wilco

TonyXiang8787 (approved):

Issue #193, opened 2023-01-18
[BUG] Program will crash if you have NaN in the input, but valid data in batch set


TonyXiang8787:
Approach:

Add default value in the `preapre_power_flow_input` and `prepare_state_estimation_input` function for all inputs vector.

https://github.com/PowerGridModel/power-grid-model/blob/9d9fa4e6348b2ea3d42ab6ea93eb220d89eb1b07/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp#L1182

https://github.com/PowerGridModel/power-grid-model/blob/9d9fa4e6348b2ea3d42ab6ea93eb220d89eb1b07/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp#L1199

# Default values for Power Flow

`s_injection`: `ComplexValue<sym>{1.0}`
`source`: `ComplexValue<sym>{1.0}`

# Default values for State Estimation

`shunt_status`: `0`
`load_gen_status`: `0`
`measured_voltage`: `{ ComplexValue<sym>{1.0}, 1.0 }`
`measured_*_power`: `{ ComplexValue<sym>{0.0}, 1.0 }`
`measured_bus_injection`: `{ ComplexValue<sym>{0.0}, 1.0 }`

## source status

```
se_input[i].source_status.resize(math_topology_[i]->n_source());
se_input[i].source_status[math_topology_[i]->slack_bus_] = ComplexValue<sym>{1.0};

```

# Batch calculation

When you (indirectly) call `prepare_*_input` in the topology caching step in batch calculation, skip the retrieval of the real values, only use the default values.

mgovers:
@TonyXiang8787 isn't this enough to just fix everything?

```cpp
        // calculate once to cache topology, ignore results, all math solvers are initialized
        (this->*calculation_fn)(std::numeric_limits<double>::infinity(), 0, calculation_method);
```

mgovers:
> @TonyXiang8787 isn't this enough to just fix everything?
> 
> ```c++
>         // calculate once to cache topology, ignore results, all math solvers are initialized
>         (this->*calculation_fn)(std::numeric_limits<double>::infinity(), 0, calculation_method);
> ```

After discussion: this does not suffice. There are direct calculation methods that do not use tolerance and iterations

PR #244, opened 2023-04-21
Feature/restore precomputation state

### LINKED ISSUE #107: [BUG] False positives in is_component_update_independent
**False positives in is_component_update_independent**

Currently the function checks if all batches contain exactly the same set of object IDs. If so, the batches are considered independent. However, the actual data in the batch can contain NaN values, in which case the update is not independent.

A (partial) example of input data for two sources and two seemingly independent batch updates:
```
Input
id  node  status   u_ref  u_ref_angle  ...
 1  1001       1 10000.0          0.0  ...
 2  1002       1 10000.0          0.0  ...
```
```
Update batch #1
id  status    u_ref  u_ref_angle
 1     NaN  10100.0          NaN
 2     NaN      NaN         0.01
```
```
Update batch #2
id  status    u_ref  u_ref_angle
 1     NaN      NaN         0.02
 2     NaN  10200.0          NaN
```

In general, if the batches are independent, there is no need to copy the input data before applying each batch. However, when a batch contains a NaN value (at a position where another batch contains a non-NaN values) some values may not be overwritten, so you might up with values from batch 1 in batch 2 etc.
```
Batch #1
id  node  status   u_ref  u_ref_angle  ...
 1  1001       1 10100.0         0.0   ...
 2  1002       1 10000.0         0.01  ...
```
```
Batch #2
id  node  status   u_ref  u_ref_angle  ...
 1  1001       1 10100.0         0.02  ...
 2  1002       1 10200.0         0.01  ...
```
Notice how the batch update data from batch 1 is still present in batch 2.

**Expected behavior**

We'd expect `is_component_update_independent` to return `false` in this case.

**Guiding directions**

We should add a check for each attribute of each object to see if there is a NaN value following a non-NaN value when we loop through the batches from batch 0 to N. A simpler easier implementation would be to check if all values are NaN or non of the values are NaN.

#177 might fix the problem automatically.

### LINKED ISSUE #177: [FEATURE] Allow shadow copy/update for components in batch calculation
# Effect

This should remove the concept of independent update dataset.

mgovers:
* Add `cache_item` and `restore_values` functionality to to `Container`
* In main model: after every scenario, `restore` original state
* Fixes issue where separate scenarios influenced eachother when containing updates to different fields within the same component (#107)

Closes #107 
Closes #177 

To be merged together with https://github.com/PowerGridModel/power-grid-model-workshop/pull/10


power_grid_model_c/power_grid_model_c/power_grid_model_c.cpp
```diff
@@ -300,7 +300,7 @@ void PGM_update_model(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Idx n_c
         dataset[components[i]] = ConstDataPointer{update_data[i], component_sizes[i]};
     }
     try {
-        model->update_component(dataset);
+        model->update_component<MainModel::permanent_update_t>(dataset);
```
mgovers, line 303: TODO: make this `cached_update_t`

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -281,6 +296,14 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         }
     }
 
+    // restore the initial values of all components
+    void restore_components() {
+        components_.restore_values();
+        is_topology_up_to_date_ = false;
+        is_sym_parameter_up_to_date_ = false;
+        is_asym_parameter_up_to_date_ = false;
```
mgovers, line 304: Needs help: this kills performance.

Possible solutions:

* [ ] Also cache and restore the topology
* [ ] Remember whether the `UpdateChange` mutated topology
  * This may cause repeatedly recalculate the same topology
* [ ] Completely different approach: Instead of caching components, cache the inverse patches/mutations and define operation `patch_y o patch_x_inverse`
  * E.g.
    * Input: component A `{.p = a, .q = b}`
    * Scenario 1: update component A `{.p = c}`
      * The inverse patch is `{.p = a}`
    * Scenario 2: update component A `{.q = d}`
      * The inverse patch is `{.q = b}`
    * Parallel: OK
    * Sequential (1 -> 2): `{.p = a, .q = b}`
      * Scenario 1:
        * apply patch `(1) o (inverse noOp) = {.p = c}` -> `{.p = c}`
        * check topology changed etc.
        * run calculation
      * Scenario 2:
        * apply combined patch `(2) o (inverse 1) = {.p = a, .q = d}` -> `{.p = a, .q = d}`
        * check topology changed etc.
        * run calculation
  * This way, topology is diffed against the latest run, not against the main topology. This could be useful if e.g. an entire batch of scenarios is run that has the same topology within the batch, but that topology is different from the input state

TonyXiang8787:
@mgovers I have cherry-picked your commits about only renaming the template arguments of container class into a separate PR. So this PR becomes more easy to review for the content change.

power_grid_model_c/power_grid_model/include/power_grid_model/container.hpp
```diff
@@ -199,14 +199,34 @@ class Container<RetrievableTypes<GettableTypes...>, StorageableTypes...> {
...
+
+        auto it = std::find_if(cached_vec.begin(), cached_vec.end(), [pos](const auto& e) {
+            return e.first == pos;
+        });
+        if (it == cached_vec.end()) {
+            cached_vec.emplace_back(pos, value);
+        }
+        else {
+            it->second = value;
+        }
```
TonyXiang8787, line 216:
```suggestion
        cached_vec.emplace_back(pos, value);
```

power_grid_model_c/power_grid_model/include/power_grid_model/container.hpp
```diff
@@ -272,6 +292,15 @@ class Container<RetrievableTypes<GettableTypes...>, StorageableTypes...> {
         return res;
     }
 
+    template <class Storageable>
+    void restore_values_impl() {
+        auto& cached_vec = std::get<std::vector<std::pair<Idx, Storageable>>>(cached_reset_values_);
+        for (auto const& cache : cached_vec) {
```
TonyXiang8787, line 298: As commented above, reverse restore the object garantees that the original object is restored, if (in rare case) one object is updated multiple times.

```suggestion
        for (auto it = cached_vec.rcbegin(); it != cached_vec.rcend(); ++it) {
            auto const& cache = *it;
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -81,6 +81,14 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     using GetIndexerFunc = void (*)(MainModelImpl const& x, ID const* id_begin, Idx size, Idx* indexer_begin);
 
    public:
+    struct cached_update_t {
+        static constexpr auto cached_update = true;
+    };
```
TonyXiang8787, line 86: Nice to use a tag type so it is more clear. As a convention for tag type with boolean value, we just inherit from `std::true_type` or `std::false_type`.

```suggestion
    struct cached_update_t : std::true_type {};
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -81,6 +81,14 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+
+    struct permanent_update_t {
+        static constexpr auto cached_update = false;
+    };
```
TonyXiang8787, line 90: Same here.

```suggestion
    struct cached_update_t : std::false_type {};
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -241,8 +243,14 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
             // get component
             // either using ID via hash map
             // either directly using sequence id
-            CompType& comp = has_sequence_id ? components_.template get_item<CompType>(sequence_idx[seq])
-                                             : components_.template get_item<CompType>(it->id);
+            Idx2D const sequence_single =
+                has_sequence_id ? sequence_idx[seq] : components_.template get_idx_by_id<CompType>(it->id);
+
+            if constexpr (CacheType::cached_update) {
```
TonyXiang8787, line 249: If you use `std::true_type` and `std::false_type`.

```suggestion
            if constexpr (CacheType::value) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -545,9 +568,10 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
                 }
```
TonyXiang8787, line 568: we can remove this part as it is the whole intention here to avoid copy the whole model.

```
                if (!independent) {
                    model = base_model;
                }
```

power_grid_model_c/power_grid_model_c/power_grid_model_c.cpp
TonyXiang8787: This call is intended to do permanent update.

power_grid_model_c/power_grid_model/include/power_grid_model/container.hpp
mgovers: makes sense

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: right, i did this and then i didn't due to the investigation in my comment above 😬 will fix

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: @TonyXiang8787 I updated the code with your comments. This question, however, still stands and requires your input.

mgovers: cfr. call: go with option 2.

The user should be smart enough to think about which types of calculations to do with a single permanent update (e.g.: a set of simulations with a the same topology that is different from that of the `0`-case).

In the future, we may want to implement a "smart dispatch" which batches for these kind of expensive changes like updating topology, but out of scope of this PR

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -260,9 +260,14 @@ struct ComponentToMathCoupling {
 
 // change of update cause topology and param change, or just param change
 struct UpdateChange {
-    bool topo, param;
+    bool topo{};
+    bool param{};
 };
 
+constexpr UpdateChange operator||(UpdateChange const& a, UpdateChange const& b) {
```
TonyXiang8787, line 267: It is preferred to define `operator||` inside the class as a friend function. This will result a so-called hidden friend. It can be only access via ADL and will not participate the overload resolution in a random `||` operator. This increase compilation performance.

see:
https://youtu.be/iDX2d7poJnI?t=1553

TonyXiang8787:
@mgovers nice work! Only one comment left over for the code part. Now just documentation.

TonyXiang8787:
@bramstoeller you might interested in this PR. Feel free to have a look and give comments if you have time.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: right, i'm so used to the `concepts` way of implementing this that i completely forgot about the advantages of using `friend` 😬 will fix

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -260,7 +260,12 @@ struct ComponentToMathCoupling {
 
 // change of update cause topology and param change, or just param change
 struct UpdateChange {
-    bool topo, param;
+    bool topo{};
+    bool param{};
+
+    static constexpr UpdateChange operator||(UpdateChange const& a, UpdateChange const& b) {
```
TonyXiang8787, line 266: Should it be a `friend` keyword instead of `static`?

TonyXiang8787: Ah, I see your comment that `static` is also good.

TonyXiang8787: Hmmm, `clang` says `operator||` cannot `static`.

mgovers: clang does not like this

```
     /home/runner/work/power-grid-model/power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp: In member function ‘void power_grid_model::MainModelImpl<power_grid_model::container_impl::ExtraRetrievableTypes<TR ...>, power_grid_model::ComponentList<ComponentType ...> >::update_component(ForwardIterator, ForwardIterator, const std::vector<power_grid_model::Idx2D>&)’:
      /home/runner/work/power-grid-model/power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp:254:63: error: no match for ‘operator||’ (operand types are ‘power_grid_model::UpdateChange’ and ‘power_grid_model::UpdateChange’)
        254 |                 cached_state_changes_ = cached_state_changes_ || changed;
            |                                         ~~~~~~~~~~~~~~~~~~~~~ ^~ ~~~~~~~
            |                                         |                        |
            |                                         |                        power_grid_model::UpdateChange
            |                                         power_grid_model::UpdateChange
      /home/runner/work/power-grid-model/power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp:254:63: note: candidate: ‘operator||(bool, bool)’ (built-in)
        254 |                 cached_state_changes_ = cached_state_changes_ || changed;
            |                                         ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
```

will repro locally using WSL to fix

mgovers: oh right, `operatorX` **must** be friend because it's a special function 😅

TonyXiang8787:
@mgovers I think we can remove `independent` and `cache_topology` completely from the C/Python API. This is theoretically a breaking change. But these two attributes are used in tests only to verify if the (old) topology and independent check are correct. I can be almost 100% sure no users use these two attributes.

Steps to remove them include:

1. Define `BatchParameter` as an empty struct. So it does not have the two attributes anymore. To keep it empty is to reserve the possibility to return new batch attributes we might add in the future.
1. Remove the static function `is_topology_cacheable` and `is_topology_cacheable_component`. We don't need them anymore.
1. Adjust C++ Validation test to not to check `independent` and `cache_topology` anymore in the batch validation.
1. Adjust C API, remove `PGM_is_batch_independent` and `PGM_is_batch_cache_topology`.
1. Adjust Python API, remove the `ctypes` binding call of those two C API functions. Remove the two properties `independent` and in the `cache_topology` in [Python Class Definition](https://github.com/PowerGridModel/power-grid-model/blob/feature/container-view/src/power_grid_model/core/power_grid_model.py)
1. Adjust Python validation test to not to check `independent` and `cache_topology` anymore in the batch validation.
1. Remove the entry `independent` and `cache_topology` in the `params.json` in all the validation [test cases](https://github.com/PowerGridModel/power-grid-model/tree/feature/container-view/tests/data).

TonyXiang8787:
@mgovers  Another interesting stuff, when we now run `ctest` and it fails. It does not show the failed message by default. See this pipeline message:

https://github.com/PowerGridModel/power-grid-model/actions/runs/4787616547/jobs/8513085223?pr=244

Maybe we need to adjust `build.sh` and/or CI scripts to add `--output-on-failure` in the call? So we can actually see the failed test message.

mgovers:
> 

the danger you can run into is that the tests become extremely verbose and it's hard to find which test is the main culprit, e.g. when one issue in location `X` tested by test `x` cascades down and affects location `Y` covered by test `y`.

Since we're not using thousands of tests yet, i do believe we can add that for now, but keep in mind that we might want to revert that change

mgovers:
> @mgovers I think we can remove `independent` and `cache_topology` completely from the C/Python API. This is theoretically a breaking change. But these two attributes are used in tests only to verify if the (old) topology and independent check are correct. I can be almost 100% sure no users use these two attributes.
> 
> Steps to remove them include:
> 
>     1. [ ]  Define `BatchParameter` as an empty struct. So it does not have the two attributes anymore. To keep it empty is to reserve the possibility to return new batch attributes we might add in the future.
> 
>     2. [ ]  Remove the static function `is_topology_cacheable` and `is_topology_cacheable_component` in `MainModelImpl`. We don't need them anymore.
> 
>     3. [ ]  Adjust C++ Validation test to not to check `independent` and `cache_topology` anymore in the batch validation.
> 
>     4. [ ]  Adjust C API, remove `PGM_is_batch_independent` and `PGM_is_batch_cache_topology`.
> 
>     5. [ ]  Adjust Python API, remove the `ctypes` [binding](https://github.com/PowerGridModel/power-grid-model/blob/feature/container-view/src/power_grid_model/core/power_grid_core.py) call of those two C API functions. Remove the two properties `independent` and in the `cache_topology` in [Python Class Definition](https://github.com/PowerGridModel/power-grid-model/blob/feature/container-view/src/power_grid_model/core/power_grid_model.py)
> 
>     6. [ ]  Adjust Python validation test to not to check `independent` and `cache_topology` anymore in the batch validation.
> 
>     7. [ ]  Remove the entry `independent` and `cache_topology` in the `params.json` in all the validation [test cases](https://github.com/PowerGridModel/power-grid-model/tree/feature/container-view/tests/data).

we may also mark it as `[[deprecated]]`, right? that way, we can give the users the opportunity to switch if they need to

mgovers:
The failing test case was a result of invalid input data caused by #193 . Fix in https://github.com/PowerGridModel/power-grid-model/pull/244/files#diff-fbee3fb5854da47e323994f42cba983d066a4c45b4596453f9d1516bf1e98a10

Interestingly, there wasn't a `NaN` value, but the test case still failed

mgovers:
paired with https://github.com/PowerGridModel/power-grid-model-workshop/pull/10

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -260,7 +260,12 @@ struct ComponentToMathCoupling {
 
 // change of update cause topology and param change, or just param change
 struct UpdateChange {
-    bool topo, param;
+    bool topo{};
+    bool param{};
+
+    friend constexpr UpdateChange operator||(UpdateChange const& a, UpdateChange const& b) {
+        return UpdateChange{a.topo || b.topo, a.param || b.param};
```
mgovers, line 267: sonarcube suggestion due to variable shadowing

```suggestion
    friend constexpr UpdateChange operator||(UpdateChange const& x, UpdateChange const& y) {
        return UpdateChange{x.topo || y.topo, x.param || y.param};
```

mgovers:
the remaining suggested code smell is not introduced in this PR and therefore out of scope

TonyXiang8787:
> 

@mgovers you can remove that comment

docs/user_manual/calculations.md
```diff
@@ -408,71 +402,68 @@ Algorithm call: `CalculationMethod.iterative_linear`
...
+The parameters of the individual scenarios within a batch can be done by providing deltas compared to the existing state of the model.
+The values of unchanged attributes and components parameters within a scenario may be implicit (like a delta update) or explicit (similarly to how one would provide a full state).
+In the context of the power-grid-model, these are called **dependent** (implicit) and **independent** (explicit) batch updates, respectively.
+In both cases, all scenario updates are relative to the state of the model before the call of the calculation.
+See the examples below for usage.
 
-- Only specify the objects and attributes that are changed in this batch.
-Here original model is copied everytime for each batch.
+- Dependent batches are useful for a sparse sampling for many different components, e.g. for N-1 checks.
+- Independent batches are useful for a dense sampling of a small subset of components, e.g. time seris power flow calculation.
```
TonyXiang8787, line 427:
```suggestion
- Independent batches are useful for a dense sampling of a small subset of components, e.g. time series power flow calculation.
```

TonyXiang8787 (approved):

PR #248, opened 2023-04-26
Performance improvements: Enable C++20 standard

### LINKED ISSUE #247: Enable `constexpr` STL functionality
Many constructors and element accessors were made `constexpr` in the C++20 STL. This often results in a "free" performance gain, especially since we're already templating much of the power-grid-model C++ core library. We're not optimally making use of this.

Setting the C++ standard to `C++20` will enable many of these `constexpr` versions of STL even if we do not use any other C++20 core features.

**NOTE:** this is completely independent of using C++20 core features as mentioned in https://en.cppreference.com/w/cpp/compiler_support and instead merely enables things that have been thoroughly tested by the STL headers' suppliers. STL headers often are filled with opt-in checks for compiler support and generally do not impact computation results (although some strict warning levels may be triggered when using the STL with one supplier for another's compiler).

mgovers:
Closes #247 

Enables C++20 STL features - especially the `constexpr` ones

### DIFF
--- a/docs/advanced_documentation/build-guide.md
+++ b/docs/advanced_documentation/build-guide.md
@@ -48,7 +48,7 @@ You can define the environment variable `CXX` to for example `clang++` to specif
 
 **Windows**
 
-* MSVC >= 16.2 (Visual Studio 2019, IDE or build tools)
+* MSVC >= 16.0 (Visual Studio 2019, IDE or build tools)
 
 **macOS**
 
--- a/docs/advanced_documentation/build-guide.md
+++ b/docs/advanced_documentation/build-guide.md
@@ -37,7 +37,7 @@ minimum.
 
 ### Compiler Support
 
-You need a C++ compiler with C++17 support. Below is a list of tested compilers:
+You need a C++ compiler with C++20 support. Below is a list of tested compilers:
 
 **Linux**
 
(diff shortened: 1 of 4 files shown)


TonyXiang8787:
@mgovers you also need to change `setup.py`. There are compiler flag there for version.

docs/advanced_documentation/build-guide.md
```diff
@@ -48,7 +48,7 @@ You can define the environment variable `CXX` to for example `clang++` to specif
 
 **Windows**
 
-* MSVC >= 14.2 (Visual Studio 2019, IDE or build tools)
+* MSVC >= 16.2 (Visual Studio 2019, IDE or build tools)
```
mgovers, line 51: The documented MSVC version `14.2` cannot have been correct, because MSVC only introduced the first C++17 features in version `15.0`.

C++20 support has been enabled in MSVC version `16.0`. Since we're only using the C++20 STL and no other language features, I believe `16.0` should also do. Do you agree @TonyXiang8787 ?

mgovers:
I did a full search for `17` and these seem to be the only places where it was mentioned (other occurrences were values in the validation parameters, svgs, component ids, or visual studio versions)

TonyXiang8787 (dismissed):

docs/advanced_documentation/build-guide.md
mgovers:
```suggestion
* MSVC >= 16.0 (Visual Studio 2019, IDE or build tools)
```

TonyXiang8787 (dismissed):

mgovers:
@TonyXiang8787 i don't know why but there was a DCO issue after the merge with `main`. this fixed that

TonyXiang8787 (approved):

PR #250, opened 2023-04-28
Bugfix/incomplete input valid update

### LINKED ISSUE #193: [BUG] Program will crash if you have NaN in the input, but valid data in batch set

mgovers:
* Allow `infinite` error tolerance to be passed to the math solver.
  * Accept any valid output
  * Prevent fatal crashes when solution is ambiguous (e.g. not invertible)
* Small refactor of math solver for readability
  * Use `switch`/`case` statement to select solver
  * Remove hardcoded values for `linear_current` solver
* Tests that show this solves #193

Closes #193


tests/cpp_unit_tests/test_main_model.cpp
```diff
@@ -941,6 +941,129 @@ TEST_CASE_TEMPLATE("Test main model", settings, regular_update, cached_update) {
             CHECK(sym_node_2[8].u_pu == doctest::Approx(0.67).epsilon(0.005));
         }
     }
+
+    SUBCASE("Test incomplete input but complete update dataset") {
```
mgovers, line 945: @TonyXiang8787 what's the best way to test this for state estimation?

tests/data/state_estimation/distribution-case/input.json
mgovers, line 1: this only reverts the validation test change from #245

tests/cpp_unit_tests/test_main_model.cpp
mgovers: made an attempt below. and yes, i know it doesn't make sense to use both a symmetric voltage sensor and an asymmetric one in the same place, but it's just to check that the code in principle works.

EDIT: actually, it does make sense cause they represent measurements, not exact values

mgovers:
> Kudos, SonarCloud Quality Gate passed!    [Quality Gate passed](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=250)
> 
> [Bug](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=BUG) [A](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=BUG) [0 Bugs](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=BUG) [Vulnerability](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=VULNERABILITY) [A](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=VULNERABILITY) [0 Vulnerabilities](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=VULNERABILITY) [Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=SECURITY_HOTSPOT) [A](https://sonarcloud.io/project/security_hotspots?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=SECURITY_HOTSPOT) [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=SECURITY_HOTSPOT) [Code Smell](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=CODE_SMELL) [A](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=CODE_SMELL) [2 Code Smells](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=250&resolved=false&types=CODE_SMELL)
> 
> [](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=250&metric=new_coverage&view=list) [88.9% Coverage](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=250&metric=new_coverage&view=list) [](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=250&metric=new_duplicated_lines_density&view=list) [0.0% Duplication](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=250&metric=new_duplicated_lines_density&view=list)

@TonyXiang8787 the introduced code smells require `MSVC >= 16.4`. I think it's a good idea to have a discussion on the minimal version we want to support so we don't have to bump that every single time

wrt the missing coverage, the `test_main_model.cpp` tests do test that. i don't know which metric is used to check coverage, but if it's `test_<header_file_name>.cpp` then that makes sense. i can have a stab at it on monday if you want me to.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -638,7 +638,7 @@ class IterativeLinearSESolver {
 
         // loop to iterate
         Idx num_iter = 0;
-        while (max_dev > err_tol) {
+        while (max_dev > err_tol || (err_tol == std::numeric_limits<double>::infinity() && num_iter == 0)) {
```
TonyXiang8787, line 641: Use `std::isinf` intead of `==`

mgovers: i knew there was one but i couldn't find it 😄 TY

TonyXiang8787: Why do you use `do` - `while` loop in another refactor, but not here?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
```diff
@@ -38,39 +38,28 @@ class MathSolver {
...
+                default:
+                    throw InvalidCalculationMethod{};
             }
-            return iterative_current_pf_solver_.value().run_power_flow(y_bus_, input, err_tol, max_iter,
-                                                                       calculation_info);
         }
-
-        else {
-            throw InvalidCalculationMethod{};
+        catch (const SparseMatrixError&) {
```
TonyXiang8787, line 58: Catch the error here is a bit wired. You want the error to propagate above even when `err_tol = inf`. Users can theoretically run a  single (not-batch) calculation with set specifically `err_tol = inf`. We only need to catch and silence the error in the cache topology step in the batch calculation. So we can catch and silent there.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -535,7 +535,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         }
 
         // calculate once to cache topology, ignore results, all math solvers are initialized
-        (this->*calculation_fn)(err_tol, max_iter, calculation_method);
+        (this->*calculation_fn)(std::numeric_limits<double>::infinity(), 1, calculation_method);
```
TonyXiang8787, line 538: Try and catch `SparseMatrixError` here, and silence it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
```diff
@@ -38,39 +38,28 @@ class MathSolver {
...
-                 calculation_method == CalculationMethod::linear_current) {
-            if (!iterative_current_pf_solver_.has_value()) {
-                Timer timer(calculation_info, 2210, "Create math solver");
-                iterative_current_pf_solver_.emplace(y_bus_, topo_ptr_);
-            }
-            if (calculation_method == CalculationMethod::linear_current) {
-                err_tol = 1000;
-                max_iter = 2;
+        try {
+            switch (calculation_method) {
```
TonyXiang8787, line 45: nice refactor to make them separate

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
mgovers: well, i was thinking about the case where a user actually provides `err_tol = inf` where they really don't care about the outcome, but i'll change it. it does clean up the code a bit

TonyXiang8787:
Looks pretty good. Another thing to modify, you can remove the attribute `q_specified` of all `sym_load` in one validaiton dataset:

https://github.com/PowerGridModel/power-grid-model/blob/feature/incomplete-input-valid-update/tests/data/power_flow/dummy-test-batch-newton/input.json

The test should still be good.

TonyXiang8787:
Also this one

https://github.com/PowerGridModel/power-grid-model/blob/feature/incomplete-input-valid-update/tests/data/power_flow/dummy-test-batch/input.json

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
mgovers: > Why do you use `do` - `while` loop in another refactor, but not here?

whoops i forgot about this one

mgovers:
> Looks pretty good. Another thing to modify, you can remove the attribute `q_specified` of all `sym_load` in one validaiton dataset:
> 
> https://github.com/PowerGridModel/power-grid-model/blob/feature/incomplete-input-valid-update/tests/data/power_flow/dummy-test-batch-newton/input.json
> 
> The test should still be good.

for this one, it is not allowed, because the `update` data contains three batches - one which updates only id `7`, one which updates id `8`, and one that updates both.

Same for the `dummy-test-batch`, actually

TonyXiang8787:
> > Looks pretty good. Another thing to modify, you can remove the attribute `q_specified` of all `sym_load` in one validaiton dataset:
> > https://github.com/PowerGridModel/power-grid-model/blob/feature/incomplete-input-valid-update/tests/data/power_flow/dummy-test-batch-newton/input.json
> > The test should still be good.
> 
> for this one, it is not allowed, because the `update` data contains three batches - one which updates only id `7`, one which updates id `8`, and one that updates both.
> 
> Same for the `dummy-test-batch`, actually

You are correct. I overlooked too soon.

mgovers:
> > > Looks pretty good. Another thing to modify, you can remove the attribute `q_specified` of all `sym_load` in one validaiton dataset:
> > > https://github.com/PowerGridModel/power-grid-model/blob/feature/incomplete-input-valid-update/tests/data/power_flow/dummy-test-batch-newton/input.json
> > > The test should still be good.
> > 
> > 
> > for this one, it is not allowed, because the `update` data contains three batches - one which updates only id `7`, one which updates id `8`, and one that updates both.
> > Same for the `dummy-test-batch`, actually
> 
> You are correct. I overlooked too soon.

i can make a new validation set with missing input data if you like

TonyXiang8787:
> > > > Looks pretty good. Another thing to modify, you can remove the attribute `q_specified` of all `sym_load` in one validaiton dataset:
> > > > https://github.com/PowerGridModel/power-grid-model/blob/feature/incomplete-input-valid-update/tests/data/power_flow/dummy-test-batch-newton/input.json
> > > > The test should still be good.
> > > 
> > > 
> > > for this one, it is not allowed, because the `update` data contains three batches - one which updates only id `7`, one which updates id `8`, and one that updates both.
> > > Same for the `dummy-test-batch`, actually
> > 
> > 
> > You are correct. I overlooked too soon.
> 
> i can make a new validation set with missing input data if you like

I think it is a good idea.

mgovers:
i ran into an unexpected roadblock: when `p_specified` is given, but `q_specified` not, then `s_specified` is `nan`. when later `q_specified` is also provided, then `s_specified` is still `nan`

i will investigate if and how we could set the real resp imaginary components for `s_specified` when only one of the two values is given

mgovers:
> i ran into an unexpected roadblock: when `p_specified` is given, but `q_specified` not, then `s_specified` is `nan`. when later `q_specified` is also provided, then `s_specified` is still `nan`
> 
> i will investigate if and how we could set the real resp imaginary components for `s_specified` when only one of the two values is given

As discussed, this was a separate bug that has the same impact on a customer.

I fixed it in [292f7de](https://github.com/PowerGridModel/power-grid-model/pull/250/commits/292f7ded1be82f6e79f49267b02feede8c5b7111)

TonyXiang8787 (approved):

PR #251, opened 2023-05-02
IDE presets

mgovers:
* Disable coverage and sanitizer for IDE on Linux
  * this prevents the annoying false positive `profiling: /path/to/xyz.cpp.gcda: cannot merge previous run count: corrupt object tag` and `profiling: /path/to/xyz.cpp.gcda: cannot merge previous GCDA file: mismatched number of counters` warning messages from happening (see https://stackoverflow.com/questions/59259069/what-does-the-gcc-warning-coverage-mismatch-mean )
  * Introduces `clang-release`, `clang-debug`, `gcc-release` and `gcc-debug` presets (as opposed to `ci-*`)
* Update documentation on supported vs tested compiler versions

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -151,6 +151,42 @@
       "hidden": true
     },
     {
+      "name": "clang-release",
+      "inherits": [
+        "release",
+        "unix-base"
+      ]
+    },
+    {
+      "name": "clang-debug",
+      "inherits": [
+        "release",
+        "unix-base"
+      ]
+    },
+    {
+      "name": "gcc-base",
+      "environment": {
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -150,6 +150,42 @@
...
+      "name": "clang-release",
+      "inherits": [
+        "release",
+        "unix-base"
+      ]
+    },
+    {
+      "name": "clang-debug",
+      "inherits": [
+        "release",
```
mgovers, line 163:
```suggestion
        "debug",
```

TonyXiang8787 (changes requested):
Also add `clang-release`, `clang-debug`, `gcc-release` and `gcc-debug` to build preset and test preset

TonyXiang8787 (approved):

PR #256, opened 2023-05-12
Configure and test conda build

TonyXiang8787:
# Purpose

This PR configures the build process for conda environment and adds relevant tests in the CI.

This is needed if we want to publish PGM into conda.

# Approach

`conda-build` does not use `pip` build isolations. `conda-forge` requires that all the build dependencies have to be installed from `conda-forge`. To satisfy this, `setup.py` is adjusted to either add include folder from `CONDA_PREFIX`, or use the `pip` package to fetch the header files.

### DIFF
--- a/.github/conda_pgm_env.yml
+++ b/.github/conda_pgm_env.yml
@@ -1,0 +1,11 @@
+name: conda-pgm-env
+channels:
+  - conda-forge
+dependencies:
+  - python=3.11
+  - pip
+  - wheel
+  - setuptools
+  - boost-cpp
+  - eigen
+  - numpy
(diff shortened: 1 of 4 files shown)


mgovers:
looks good. Is there any particular reason why the python builds and tests now do not depend on the C++ ones anymore? I'm happy with the major speedup, but I want to know if there's any particular reason

image

mgovers (approved):

TonyXiang8787:
> looks good. Is there any particular reason why the python builds and tests now do not depend on the C++ ones anymore? I'm happy with the major speedup, but I want to know if there's any particular reason
> 
> image

This is a legacy problem. Before PGM became open-source, we had only limit number of GitHub Actions runners for private repository. So we needed to limit parallel jobs. Now it is not a problem anymore.

PR #257, opened 2023-05-16
long path fail doc add

nitbharambe:
Added a note for build failing because of long paths and which might give vague errors

### DIFF
--- a/docs/advanced_documentation/build-guide.md
+++ b/docs/advanced_documentation/build-guide.md
@@ -293,6 +293,11 @@ Install from source in develop mode, and run `pytest`.
 pip install -e .[dev]
 pytest
 ```
+```{note}
+Using long path for miniconda environment location for the installation might lead to crossing the `maximum path length limitation` set by windows.
+This might cross the installtion to fail.
+It is then possible to enable long paths in windows or alternatively place the environment in a small path location.  
+```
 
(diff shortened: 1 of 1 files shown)


docs/advanced_documentation/build-guide.md
```diff
@@ -293,6 +293,11 @@ Install from source in develop mode, and run `pytest`.
 pip install -e .[dev]
 pytest
 ```
+```{note}
+Using long path for miniconda environment location for the installation might lead to crossing the `maximum path length limitation` set by windows.
+This might cross the installtion to fail.
+It is then possible to enable long paths in windows or alternatively place the environment in a small path location.
```
mgovers, line 299:
```suggestion
Long paths for (dependencies in) the miniconda installation environment might exceed the `maximum path length limitation` set by Windows, causing the installation to fail.
It is possible to enable long paths in Windows by following the steps in the [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry)
```

it is probably a bad idea to suggest changing the miniconda path, because any dependency (especially those developed on Unix) could exceed the default maximum path length by itself, even without the miniconda environment

mgovers (approved):

PR #265, opened 2023-06-07
Feature/clang tidy

### LINKED ISSUE #262: Add clang-tidy checks to CI
Complementary to Sonar Cube (there is overlap but not completely).

## Desired solution

* Add CI job that runs clang-tidy
* Any warnings should be disabled
* Make follow-up issues regarding warning disabling

mgovers:
Closes #262 

This PR only adds support for running `clang-tidy`. Currently, all settings are optimized for `clang-tidy-15`, but CI only checks for `clang-tidy-14`.

A new workflow is created because it is more similar to the sonar cube workflow than it is to the regular release workflow and it is complementary to the sonar cube pipeline.

No attempt is made to fix the code. Those changes should be postponed to follow-up PRs.


.github/workflows/clang-tidy.yml
TonyXiang8787, line 1: I think we don't need a separate CI workflow. We already have a lot.

I would say add clang-tidy always to the `ci-clang` presents. And we run clang-tidy in the Linux CI build. We can even add it to too root clang presets, so it is also checked locally.

docs/advanced_documentation/build-guide.md
```diff
@@ -160,10 +160,11 @@ WSL), or in a physical/virtual machine.
 Append the following lines into the file `${HOME}/.bashrc`.
 
 ```shell
-export CXX=clang++-14  # or g++-11
-export CC=clang-14  # gcc-11
+export CXX=clang++-14            # or g++-11
+export CC=clang-14               # gcc-11
 export CMAKE_PREFIX_PATH=/home/linuxbrew/.linuxbrew
 export LLVM_COV=llvm-cov-14
+export CLANG_TIDY=clang-tidy-14  # only if you want to use one of the clang-tidy presets
```
TonyXiang8787, line 167: Just make it mandatory in the build guide. We just use it always in the presets.

TonyXiang8787 (changes requested):
Propose to always run clang-tidy in the Linux Clang build. And also locally in clang presets.

.clang-tidy
```diff
@@ -0,0 +1,92 @@
+# SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model project <dynamic.grid.calculation@alliander.com>
+#
+# SPDX-License-Identifier: MPL-2.0
+---
+Checks: '
```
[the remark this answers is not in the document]
mgovers: the checks preceded by `-` are disabled/ignored checks, the others are enabled. for instance:

* `-*` disables all defaults (this is good practice because it allows choosing your own style)
* `bugprone-*` enables all checks in the `bugprone` namespace (overwriting `-*`)
* `-bugprone-blabla` overwrites the previous statement for that specific check and disables it again

.github/workflows/clang-tidy.yml
mgovers: the problem is that `clang-tidy` cannot run in parallel, so those builds would take a long time. i did consider this option, but when i realised a single `clang-tidy` build can take 30min, i realised this probably not a good idea.

i discussed adding the `clang-tidy` job to the `main` workflow, but it's more like the `sonar` workflow, but the `sonar` workflow is sonarcube-specific, so either that name is wrong or it's a bad idea to add it to that workflow. Any thoughts?

TonyXiang8787: If that's the case then keep it as separate workflow.

TonyXiang8787: Also do we need to run both debug and release build on this?

mgovers: imo no, but i did notice that there's a difference in tests due to `assert` being optimized away in release mode, and therefore causing the 'unread-variable' check to fail.

i'm also fine with either

* just disabling that check in the long term as well; or
* adding separate debug and release builds later whenever needed

your  call

TonyXiang8787: Let's keep it as is.

TonyXiang8787 (approved):

PR #266, opened 2023-06-07
Upgrade to Python 3.8 and manylinux_2_24 (2_28 in build environment)

TonyXiang8787:
# Python version
Upgrade the supported python to >= 3.8.

# Linux version
Upgrade the supported linux to `manylinux_2_24 `. The wheels will be built in `manylinux_2_28`, but it checks compatibility in `manylinux_2_24`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
     "Operating System :: MacOS",
     "Topic :: Scientific/Engineering :: Physics",
 ]
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 dependencies = [
     "numpy>=1.21.0",
 ]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -101,21 +101,21 @@ build-frontend = "build"
 build-verbosity = 1
 test-extras = ["dev"]
 test-command = "pytest {package}/tests"
+# we do not support PyPy, musllinux
+skip = ["pp*", "*musllinux*"]
```
mgovers, line 105: why not?

TonyXiang8787: PyPy is very tricky to get it working. I tried once but there are so many errors so I decided not to prioritize it.

`musl` is not supported by `numpy` yet. It does not make sense that we support it. Numpy 1.25 will support `musl`, we can then also add the support.
https://github.com/numpy/numpy/releases/tag/v1.25.0rc1

mgovers (approved):

PR #272, opened 2023-06-15
implement C API

mgovers:
Prepare C API for short circuit:

* Add a short circuit calculation type
* Add a sequential short circuit calculation method
* For now: not implemented (full impl requires model update)

Cfr. discussion in https://github.com/PowerGridModel/power-grid-model/pull/273#discussion_r1230907750 , I do not prefer the solution where the short circuit fault type and phase are added to the C API , but instead would prefer to add those to the `Fault` component attributes


power_grid_model_c/power_grid_model_c/include/power_grid_model_c.h
```diff
@@ -79,20 +79,24 @@ typedef struct PGM_Options PGM_Options;
...
-    PGM_newton_raphson = 1,    /**< Newton-Raphson method for power flow */
-    PGM_iterative_linear = 2,  /**< iterative linear method for state estimation */
-    PGM_iterative_current = 3, /**< linear current method for power flow */
-    PGM_linear_current = 4     /**< iterative constant impedance method for power flow */
+    PGM_linear = 0,                  /**< linear constant impedance method for power flow */
+    PGM_newton_raphson = 1,          /**< Newton-Raphson method for power flow */
+    PGM_iterative_linear = 2,        /**< iterative linear method for state estimation */
+    PGM_iterative_current = 3,       /**< linear current method for power flow */
+    PGM_linear_current = 4,          /**< iterative constant impedance method for power flow */
+    PGM_sequential_short_circuit = 5 /**< sequential fault analysis for short circuits */
```
TonyXiang8787, line 97: Use the name IEC60909 as the short circuit method, so `PGM_iec60909`

power_grid_model_c/power_grid_model_c/power_grid_model_c.cpp
```diff
@@ -385,6 +394,8 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
                         update_dataset, opt->threading);
                 }
                 break;
+            case PGM_short_circuit:
+                [[fallthrough]];
```
TonyXiang8787, line 398: If you fall through, it will still throw error? Maybe it should be `break` here. And add a TODO if we implement the actual C++ method.

power_grid_model_c/power_grid_model_c/power_grid_model_c.cpp
```diff
@@ -15,7 +15,9 @@
 
 using namespace power_grid_model;
 
-static meta_data::AllPowerGridMetaData const& pgm_meta = meta_data::meta_data();
+namespace {
```
TonyXiang8787, line 18: nice to wrap it in anonymous namespace instead of static.

TonyXiang8787 (changes requested):
Some changes needed, see comments

power_grid_model_c/power_grid_model_c/include/power_grid_model_c.h
mgovers:
```suggestion
    PGM_iec60909 = 5 /**< sequential fault analysis for short circuits */
```

power_grid_model_c/power_grid_model_c/power_grid_model_c.cpp
mgovers: `[[fallthrough]]` indeed implies executing the next statement (by design of the attribute)

since the next statement already is a `not implemented` anyway, i do think adding it like this is better. if you think it's a good idea to also already _create_ (not implement) the C++ method, then indeed we can remove this `[[fallthrough]]` statement, but otherwise, I would keep it for the time being.

mgovers: @TonyXiang8787 if you agree with this change, this PR is ready for merge, regardless of the implementation of the validation (discussion tomorrow) itself

power_grid_model_c/power_grid_model_c/include/power_grid_model_c.h
```diff
@@ -91,8 +93,10 @@ enum PGM_CalculationMethod {
     PGM_newton_raphson = 1,    /**< Newton-Raphson method for power flow */
     PGM_iterative_linear = 2,  /**< iterative linear method for state estimation */
     PGM_iterative_current = 3, /**< linear current method for power flow */
-    PGM_linear_current = 4     /**< iterative constant impedance method for power flow */
+    PGM_linear_current = 4,    /**< iterative constant impedance method for power flow */
+    PGM_iec60909 = 5           /**< sequential fault analysis for short circuits */
```
mgovers, line 97: we can also add bit flags here if you prefer

power_grid_model_c/power_grid_model_c/include/power_grid_model_c.h
nitbharambe: @TonyXiang8787 Is it fine to name it as IEC 60909? The results would be a bit different (more accurate) compared to the standard as we make lesser assumptions right?

mgovers:
@TonyXiang8787 can you please answer the remaining open threads and review? this should be ready and independent of the other work in progress

power_grid_model_c/power_grid_model_c/include/power_grid_model_c.h
TonyXiang8787: I think that's appropriate. We don't want to invent a new name.

power_grid_model_c/power_grid_model_c/power_grid_model_c.cpp
TonyXiang8787: agreed, I will go through this PR

power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp
```diff
@@ -27,7 +27,8 @@ enum class CalculationMethod : IntS {
     newton_raphson = 1,
     iterative_linear = 2,
     iterative_current = 3,
-    linear_current = 4
+    linear_current = 4,
+    sequential_short_circuit = 5,
```
TonyXiang8787, line 31: this should be changed also to `iec60909`

power_grid_model_c/power_grid_model_c/include/power_grid_model_c.h
```diff
@@ -91,8 +93,10 @@ enum PGM_CalculationMethod {
...
 };
```
TonyXiang8787, line 98: I think it would be appropriate to add a default calculation method enum: `PGM_default_method`. In the CPP file of `PGM_Options` definition, `PGM_default_method` should be the default value for `calculation_method`.

Reasoning:
1. Different calculation types require different method. We now default the method to `newton_raphson` which is only valid for power flow. If user runs short circuit or state estimation via C-API, he always has to specify a valid method, otherwise it will crash.
2. By add a new enumeration `PGM_default_method`, user do not have to specify a method when he changes the calculation type. In the `calculate` function of C-API, we can detect this and based on calculation type to change to a real default method as below:

* power flow: newton raphson
* state estimation: iterative linear
* short circuit: iec60909

power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp
TonyXiang8787: Add a new enum `default_method` as below. Make it -128.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c.h
mgovers: i agree. i will apply in this PR

TonyXiang8787: This only applies to C-API. In Python side, there is already a default keyword argument for calculaiton method per calculation type.

power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp
```diff
@@ -23,11 +23,13 @@ enum class BranchSide : IntS { from = 0, to = 1 };
...
 
 enum class CalculationMethod : IntS {
+    default_method = -128,
     linear = 0,
     newton_raphson = 1,
     iterative_linear = 2,
     iterative_current = 3,
-    linear_current = 4
+    linear_current = 4,
+    iec60909 = 5,
```
TonyXiang8787, line 32: You need to also add `default_method` here, otherwise you are casting -128 in the C-API to an undefined value in C++ enum.

TonyXiang8787: Never mind, you added at the beginning.

TonyXiang8787 (approved):

PR #273, opened 2023-06-15
Feature/short circuit validation impl

mgovers:
Implement short circuit validation tests for C++

* Add `Fault` component to list of all components
* Use a test execution wrapper to allow (temporarily skipped) tests
* For now: `short circuit` calculation type is not part of the C++ API yet


tests/data/short_circuit/single_phase_to_ground_a/input.json
```diff
@@ -26,6 +26,6 @@
     ],
   "fault":
     [
-      {"id": 10, "r": 0.1, "x": 0.1, "node": 3, "status": 1}
+      {"id": 10, "r": 0.1, "x": 0.1, "fault_type": "single_phase_to_ground", "phase": "a", "fault_object": 3, "status": 1}
```
mgovers, line 29: I don't know how to solve these fault types in the input/update type.

My proposal is either have a separate field for the `fault_type` and the phases like so:

```json
        {
            "name": "FaultInput",
            "base": "BaseInput",
            "attributes": [
                {
                    "data_type": "ID",
                    "names": "fault_object",
                    "description": "ID of the faulty object"
                },
                {
                    "data_type": "IntS",
                    "names": "status",
                    "description": "Whether the fault is active (object is faulty)"
                },
                {
                    "data_type": "FaultType",
                    "names": "fault_type",
                    "description": "type of fault"
                },
                {
                    "data_type": "FaultPhases",
                    "names": "fault_phases",
                    "description": "phases of fault"
                },
                {
                    "data_type": "double",
                    "names": [
                        "r_f",
                        "x_f"
                    ],
                    "description": "short circuit impedance"
                }
            ]
        }
```

or to have separate fault types for each phase set:

```json
        {
            "name": "FaultInput",
            "base": "BaseInput",
            "attributes": [
                {
                    "data_type": "ID",
                    "names": "fault_object",
                    "description": "ID of the faulty object"
                },
                {
                    "data_type": "IntS",
                    "names": "status",
                    "description": "Whether the fault is active (object is faulty)"
                },
                {
                    "data_type": "FaultType",
                    "names": "fault_type",
                    "description": "type of fault"
                },
                {
                    "data_type": "double",
                    "names": [
                        "r_f",
                        "x_f"
                    ],
                    "description": "short circuit impedance"
                }
            ]
        }
```

mgovers: In https://github.com/PowerGridModel/power-grid-model/pull/233 it was decided to remove this from the component and into the calculation options interface. To be honest, I don't know whether that is the correct option. I'm thinking about compatibility here. I'm considering the following use cases:

* batch calculation for different fault types on the same node
  * requires to have it available in the update data
* if the case of having multiple faults is considered (cfr. the reason as to why the sequential short circuit exists mentioned in the chapter introduction in https://www.phasetophase.nl/boek/boek_2_10.html#_10 ), then i want to add multiple faults, possibly with different fault types.
  * In those cases, i guess one wants to treat multiple faults sequentially (e.g. by first solving for the first, then feeding that result as input to the second fault, etc., rather than by constructing a single optimization function and then solving that).
  * To avoid exposing this complexity to the user, we should just allow a list of faults to be provided. How we handle that is a `calculation_method` choice and should not require multiple API calls to `calculate` by the user

mgovers:
Code smells are not related to this PR. this was intentional copy-paste from another function

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -939,6 +939,21 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+    std::enable_if_t<
+        std::is_base_of_v<std::forward_iterator_tag, typename std::iterator_traits<ResIt>::iterator_category> &&
+            std::is_same_v<Fault, Component>,
+        ResIt>
+    output_result(std::vector<MathOutput<sym>> const& /* math_output */, ResIt res_it) {
+        assert(construction_complete_);
+        return std::transform(components_.template citer<Component>().begin(),
+                              components_.template citer<Component>().end(), comp_coup_->fault.cbegin(), res_it,
+                              [](Fault const& fault, Idx2D /* math_id */) {
+                                  return fault.get_output();
```
nitbharambe, line 953: Fault would have a short circuit output right? And that output will only be available for short circuit calculation.

mgovers: yes but this is just to make it compile. Modifying the main model will be addressed in [DNR-1742](https://alliander.atlassian.net/browse/DNR-1742)

mgovers: oh sorry i see what you mean. This is just for `MathOutput`, not `ShortCircuitMathOutput`

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -321,11 +343,10 @@ inline void add_cases(std::filesystem::path const& case_dir, std::string const&
...
     // loop sym and batch
     for (bool const sym : {true, false}) {
-        std::string const output_prefix = sym ? "sym_output" : "asym_output";
-        for (std::string const& calculation_method : calculation_methods) {
-            std::string const batch_suffix = is_batch ? "_batch" : "";
+        for (auto const& calculation_method : calculation_methods) {
             // add a case if output file exists
-            std::filesystem::path const output_file = case_dir / (output_prefix + batch_suffix + ".json");
+            std::filesystem::path const output_file =
+                case_dir / (get_output_type(calculation_method, sym) + batch_suffix + ".json"s);
```
nitbharambe, line 349: It would create a duplicate case for sc_output right? For both sym=true and false?

mgovers: ah you're right, let me try to fix this

tests/data/short_circuit/single_phase_to_ground_a/input.json
mgovers: cfr. discussions: this was accepted and resolved

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: `ShortCircuitMathOutput` is not supported yet by the main model (cfr. [DNR-1742](https://alliander.atlassian.net/browse/DNR-1742)). therefore, this is out of scope. resolving this thread

tests/data/short_circuit/three_phase/input.json
```diff
@@ -26,6 +26,6 @@
     ],
   "fault":
     [
-      {"id": 10, "r": 0.1, "x": 0.1, "node": 3, "status": 1}
+      {"id": 10, "status": 1, "fault_type": "three_phase", "phase": null, "fault_object": 3, "r": 0.1, "x": 0.1}
```
nitbharambe, line 29:
```suggestion
      {"id": 10, "status": 1, "fault_type": "three_phase", "phase": "abc", "fault_object": 3, "r": 0.1, "x": 0.1}
```

nitbharambe (approved):

nitbharambe (approved):

PR #274, opened 2023-06-18
Add Musllinux support

TonyXiang8787:
This PR adds `musllinux` support for `x86_64`. This is now possible because `numpy` supports `musllinux` from version `1.25`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,8 +101,11 @@ build-frontend = "build"
 build-verbosity = 1
 test-extras = ["dev"]
 test-command = "pytest {package}/tests"
-# we do not support PyPy, musllinux
-skip = ["pp*", "*musllinux*"]
+# we do not support 
+#    PyPy
+#    musllinux when python < 3.9
+#    musllinux in aarch64
+skip = ["pp*", "cp38-musllinux*", "*-musllinux_aarch64"]
 
 [tool.cibuildwheel.linux]
 archs = ["x86_64", "aarch64"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,8 +106,8 @@ skip = ["pp*", "*musllinux*"]
 
 [tool.cibuildwheel.linux]
 archs = ["x86_64", "aarch64"]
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -172,6 +170,12 @@ def generate_build_ext(pkg_dir: Path, pkg_name: str):
             "-O3",
             "-fvisibility=hidden",
         ]
+        # add c++20 or c++2a flag depending on gcc version
+        # TODO remove this if musllinux upgrade to newer gcc version
+        if "GCC 9." in sys.version:
+            cflags += ["-std=c++2a"]
```
mgovers, line 176: how important is it that we support musllinux? this strongly restricts us in the specific C++20 features and library features we can use. e.g. the spaceship `<=>` operator is not supported until GCC 10

TonyXiang8787: @SteffanWullems wants this in his team because he would like to use alpine based image. I can understand that. Since numpy 1.25 now supports musl, I would also go for it for PGM. I suppose `musllinux_1_2` should have more receng GCC version. You can try to contribute a `musllinux_1_2` image. See https://github.com/pypa/manylinux/issues/1488.

mgovers: ok if there are known users requesting it then that is a good reason :+1:

mgovers (approved):

SteffanWullems (approved):

setup.py
```diff
@@ -166,7 +166,6 @@ def generate_build_ext(pkg_dir: Path, pkg_name: str):
     else:
         # flags for Linux and Mac
         cflags += [
-            "-std=c++20",
```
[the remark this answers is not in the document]
mgovers: well it is added below. the temp fix is that because the musllinux image uses an old version of GCC (9.x) that does not yet support `-std=c++20` but only `-std=c++2a` (`-std=c++20` is supported from GCC 10 onwards). as soon as there's a new musllinux image out with a more recent GCC version, we can switch back to `-std=c++20`. until then, only a subset of C++20 features are supported, unfortunately

PR #276, opened 2023-06-20
Bugfix/fix clang cl

### LINKED ISSUE #241: FYI: @nitbharambe found multiple bugs for the Clang CL presets (SEGFAULTs, type mismatch, ...). The compiler compiles the code, but the tests fail.
FYI: @nitbharambe found multiple bugs for the Clang CL presets (SEGFAULTs, type mismatch, ...). The compiler compiles the code, but the tests fail.

After investigating:
* I am able to repro on `main`, which means that it's not a problem on my branch
* we do need to disable the Clang CL presets for the time being.

Apparently, I did not find the issue before because I just so happened to not run the tests for these specific presets (I did do a matrix test with `(gcc, clang, msvc) x (release, debug)`, but apparently not for `clang-cl`)

Repro:

```
cmake -DCMAKE_C_COMPILER="clang-cl.exe" -DCMAKE_CXX_COMPILER=clang-cl.exe "-DCMAKE_CXX_FLAGS=/DWIN32 /D_WINDOWS /W3 /WX /GR /EHsc -Wno-uninitialized-const-reference" -DCMAKE_BUILD_TYPE=RelWithDebInfo -S"." -B"cpp_build/clang-cl-release" -G Ninja  
cmake --build .\cpp_build\clang-cl-release\
.\cpp_build\clang-cl-debug\bin\power_grid_model_unit_tests.exe
```

Potentially related to the `-Wno-uninitialized-const-reference` which was necessary to make it compile

c.c. @TonyXiang8787 @petersalemink95

_Originally posted by @mgovers in https://github.com/PowerGridModel/power-grid-model/issues/239#issuecomment-1511010315_

mgovers:
closes #241 

It was found that in #241 the culprit was the fact that the `trait_pointer_to_member` struct pointed to the wrong member variable in the memory layout. Maybe this was caused by some different memory or template overload order when dealing with inheritance

Overloading the template with a `requires` block that only generates the template when the provided base class is **identical** to the identified base class fixes this template choice mismatch

Please try to repro this on your local machine (build with Clang CL Debug/Release + run all tests)

### Testing

I did a matrix test on my computer using `[msvc, clang-cl] x [debug, release]` with clean reconfigure and clean rebuild all targets and all tests passed after this change

(note: before this change, i verified that the bug in #241 was still there, so the change actually fixed the tests)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,9 +94,10 @@ jobs:
     strategy:
       matrix:
         build-option: [ debug, release ]
+        compiler: [msvc, clang-cl]
 
     env:
-      PRESET: msvc-${{ matrix.build-option }}
+      PRESET: ${{ matrix.compiler }}-${{ matrix.build-option }}
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -120,7 +120,7 @@ jobs:
           # build
           cmake --build --preset ${{ env.PRESET }} --verbose -j 1
           # test
-          ctest --test-dir . --output-on-failure
+          ctest --test-dir cpp_build\${{ env.PRESET }} --output-on-failure
           # install
           cmake --build --preset ${{ env.PRESET }} --verbose -j 1 --target install
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -110,11 +110,21 @@
       "hidden": true
     },
     {
-      "name": "clang-tidy-base",
+      "name": "clang-tidy-windows-base",
+      "cacheVariables": {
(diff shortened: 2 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -94,9 +94,10 @@ jobs:
     strategy:
       matrix:
         build-option: [ debug, release ]
+        compiler: [msvc, clang-cl]
```
mgovers, line 97: this works: CI has the following lines in the logs

```
-- The C compiler identification is Clang 15.0.7 with MSVC-like command-line
-- The CXX compiler identification is Clang 15.0.7 with MSVC-like command-line
```

TonyXiang8787 (approved):

PR #279, opened 2023-06-22
Feature/short circuit fault object

mgovers:
https://alliander.atlassian.net/browse/DNR-1880

Outcome of design discussion of 2023/06/21:

## New fault input attributes:

- fault_type
- fault_phase
- status

## New sc_output dataset

- Remove sym_sc_output, and asym_sc_output
- Define new sc_output
  - For nodes
    - u_pu, u, u_angle, u: all Realvalue<false>, line to ground voltage
    - u1ll_pu, u1ll, u1ll_angle: all double, positive sequence line-line voltage
  - For branches
    - i_from, i_from_angle, : all Realvalue<false>, three phase current
    - i1_from, i1from_angle, all double, positive sequence current
    - do the same for to side, and for branch3, do this this three sides
  - For appliances
    - i, i_angle: all Realvalue<false>, three phase current
    - i1, i1_angle: all double, positive sequence current
  - For faults
    - i_f, i_f_angle, all Realvalue<false>, three phase current
    - i1_f, i1_f_angle, all double, positive sequence current


code_generation/data/attribute_classes/update.json
```diff
@@ -147,6 +147,21 @@
...
+                {
+                    "data_type": "FaultType",
+                    "names": "fault_type",
+                    "description": "type of the fault"
+                },
+                {
+                    "data_type": "FaultPhase",
+                    "names": "fault_phase",
+                    "description": "phase(s) of the fault"
+                },
```
mgovers, line 164: should these both be part of the `update`?

power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp
```diff
@@ -72,6 +72,16 @@ enum class YBusElementType : IntS {
     fill_in_tf = 0b110
 };
 
+enum class FaultType : IntS {
+    three_phase = 0,
+    two_phase = 1,
+    two_phase_to_ground = 2,
+    single_phase_to_ground = 3,
+    default_value = na_IntS
```
mgovers, line 80: this is required if we want to add this to `update_data`

code_generation/data/attribute_classes/update.json
```diff
@@ -22,7 +22,7 @@
                         "from_status",
                         "to_status"
                     ],
-                    "description": "If the branch is connected at each side"
+                    "description": "Whether the branch is connected at each side"
```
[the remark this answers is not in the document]
mgovers: you're right, thanks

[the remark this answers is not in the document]
mgovers: i homogenized in one but forgot the other

code_generation/data/attribute_classes/update.json
petersalemink95: My feeling is that a fault is a certain type of fault, so you should not be able to change the fault type and phase, only the `status`. So if you want to test various faults on 1 location you should specify multiple faults and switch them on or off. 
But this is a design choice. What do you think?
Also curious what @TonyXiang8787 thinks about this

docs/user_manual/components.md
```diff
@@ -494,7 +509,7 @@ the meaning of `RealValueInput` is different, as shown in the table below.
 
 #### Input
```
mgovers, line 510: TODO: update Fault input

code_generation/data/attribute_classes/update.json
mgovers: i'm on your boat for the time being, but i wonder what that means for - e.g. - choosing the fault phase. i feel like if we make `fault_phase` updateable, then we should do the same for the `fault_type`

docs/user_manual/components.md
```diff
@@ -493,18 +508,22 @@ the meaning of `RealValueInput` is different, as shown in the table below.
...
-| `x_f`          | `double`  | ohm (Ω) | Short circuit reactance                             | &#10060; default 0.0 | &#10060; |                   |
+| name           | data type                                                | unit    | description                                         |       required       |  update  |   valid values    |
+| -------------- | -------------------------------------------------------- | ------- | --------------------------------------------------- | :------------------: | :------: | :---------------: |
+| `status`       | `int8_t`                                                 | -       | whether the fault is active                         |       &#10004;       | &#10004; |    `0` or `1`     |
+| `fault_type`   | {py:class}`FaultType <power_grid_model.enum.FaultType>   | -       | the type of the fault                               |       &#10004;       | &#10004; |                   |
+| `fault_phase`  | {py:class}`FaultPhase <power_grid_model.enum.FaultPhase> | -       | the phase(s) of the fault                           |       &#10004;       | &#10004; |                   |
```
mgovers, line 515: if decided to not add to update data for the time being: remove

mgovers:
Follow-up discussion 2023/06/23

> [Yesterday 3:15 PM] Peter Salemink
> 
> I just had a long thought about it. I agree with the design of the API from the input/update perspective.
> 
> However, I do have some doubts about the way we constructed the output now. To me is does not make sense to always give 2 outputs (single and three phase), where the user has to know when to look at which. This feels messy and does not comply with the "easy to use API" that we always promote. 
> 
> An example why I'm hesitating:
> 
> Say we have a batch with three scenario's, where in each scenario a different fault occurs:
> 
> - sc1: three-phase fault
> - sc2: two-phase fault
> - sc3: single phase fault
> 
> Now the user gets two ouput sets for u:
> 
> - u
> - u1ll
> 
> To interpret the data the user must look at u1ll for scenario 1 and to u for scenario 2 and 3. This does not feel right from a user perspective
> 
> [Yesterday 3:17 PM] Peter Salemink
> 
> I still think we should just give one output per variable.
> 
> In this case that would mean always give u as Realvalue<false>, stating that we want to allow different fault types within one batch
> 
> [Yesterday 3:21 PM] Peter Salemink
> 
> I know giving an asym output when you have a sym fault does not feel good. However, giving the user 2 sets and tell them to figure it out themselves feels even worse
> 
> [Yesterday 4:36 PM] Nitish Bharambe
> 
> Peter Salemink Another point to think is that user actually would not need only positive sequence in case of asymmetrical faults. And the point made by you is totally valid about user having to look at two different things. There are two options me and Tony have thought of:
> 
> We give u as result voltage of all abc phases. And another attribute u012 as the result of zero, positive, negative sequences. 
> We only give u as result voltage of all abc phases. User would interpret / Convert to a single value on his own in case of a three-phase fault. (All voltages would be line-ground)
> 
> We are both leaning towards the 2nd option. What do you think?
> 
> [7:53 AM] Martijn Govers
> 
> i'm also on boat 2.
> 
> if the users desire, we can also decide to add some functionality to the python library to do the conversion to a single value for them so that they only have to do a single simple function call.

Outcome TBD

mgovers:
all code smells are TODO's to be fixed in #280

power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp
nitbharambe: True. The name of default value for enum might however be a bit confusing for user. I cant think of a better name though.

tests/cpp_unit_tests/test_fault.cpp
```diff
@@ -50,21 +53,20 @@ TEST_CASE("Test fault") {
...
-        ComplexValue<true> const i_f = i_f_pu * base_i;
 
-        FaultShortCircuitOutput<true> output = fault.get_short_circuit_output<true>(i_f_pu, u_rated);
+        FaultShortCircuitOutput output = fault.get_short_circuit_output<true>(i_f_pu, u_rated);
         CHECK(output.id == 1);
         CHECK(output.energized);
-        CHECK(output.i_f == doctest::Approx(cabs(i_f)));
-        CHECK(output.i_f_angle == doctest::Approx(0.25 * pi));
+
+        // TODO(NITISH) sym sc output case
```
nitbharambe, line 61:
```suggestion
        // TODO add fault test post implementation
```

The output functions to modify are only of components other than fault. The fault output function would depend on how its implemented. Lets do this later then.

power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp
mgovers: `unknown`? `unspecified`?

nitbharambe:
There are lot of changes on output functions. I shall skip review them as I will modify them anyway.

power_grid_model_c/power_grid_model/include/power_grid_model/enum.hpp
nitbharambe: Equally vague. I guess we can leave it at default value for now.

nitbharambe (approved):

PR #280, opened 2023-06-23
Update SC output functions

nitbharambe:
Update SC output functions as per new output attributes. New output attributes contain only asymmetrical values as result


mgovers:
i recommend taking https://github.com/PowerGridModel/power-grid-model/pull/279 as the base (branch: feature/short-circuit-fault-object )

power_grid_model_c/power_grid_model/include/power_grid_model/component/appliance.hpp
```diff
@@ -94,17 +94,17 @@ class Appliance : public Base {
...
-            output.i_angle = arg(i * injection_direction());
-        }
+        output.i = base_i_ * cabs(i);
+        output.i_angle = arg(i * injection_direction());
         return output;
     }
+    ApplianceShortCircuitOutput get_sc_output(ComplexValue<true> const& i) const {
+        ComplexValue<false> iabc{i, i * a2, i * a};
+        return get_sc_output(iabc);
+    }
```
mgovers, line 107: any particular reason why you're not using `if constexpr (sym)`? i'm fine with this way, just interested 😬

tests/cpp_unit_tests/test_line.cpp
```diff
@@ -202,8 +197,9 @@ TEST_CASE("Test line") {
         CHECK(output.q_to(1) == doctest::Approx(imag(s_t) / 3.0));
     }
 
-    SUBCASE("Asymmetric short circuit results") {
-        BranchShortCircuitOutput output = branch.get_sc_output<false>(if_sc_asym, it_sc_asym);
+    SUBCASE("Short circuit results") {
+        BranchShortCircuitOutput sym_output = branch.get_sc_output(if_sc, it_sc);
+        BranchShortCircuitOutput output = branch.get_sc_output(if_sc_asym, it_sc_asym);
```
mgovers, line 202: `asym_output`? if not, please create a new `SUBCASE`

(also for the other files)

power_grid_model_c/power_grid_model/include/power_grid_model/component/branch3.hpp
```diff
@@ -113,24 +113,28 @@ class Branch3 : public Base {
         return output;
     }
 
-    template <bool sym>
-    Branch3ShortCircuitOutput get_sc_output(ComplexValue<sym> const& i_1, ComplexValue<sym> const& i_2,
-                                            ComplexValue<sym> const& i_3) const {
+    Branch3ShortCircuitOutput get_sc_output(ComplexValue<false> const& i_1, ComplexValue<false> const& i_2,
+                                                 ComplexValue<false> const& i_3) const {
```
mgovers, line 117: please run the formatter

power_grid_model_c/power_grid_model/include/power_grid_model/component/appliance.hpp
nitbharambe: No reason actually. The objective is to convert the ComplexValue<true> to ComplexValue<false> then rest of  the things remain the same.
Can be done with a condition statement too. Should I do it that way?

mgovers: it's fine like this. i prefer the `constexpr` solution because it makes it clear that it's the same function, just with two different implementations. that also makes refactoring easier in the future 

e.g. if we decide to go with `tag` structs rather than boolean `sym` values, which would look something like this

```cpp
template <typename T>
concept tag = std::is_empty<T>;

struct sym_tag {};
struct asym_tag {};

template <typename T>
concept calculation_type = tag<T> && (std::same_as<sym_tag, T> || std::same_as<asym_tag, T>);

template <calculation_type T> constexpr is_sym = false;
template <std::same_as<sym_tag> T> constexpr is_sym = true;
```

the refactoring would only require changing the `template` + `if constexpr` statements

but it doesn't really matter because the code already contains a lot of similar cases

tests/cpp_unit_tests/test_line.cpp
nitbharambe: Made it in different subcases

mgovers (approved):

mgovers (dismissed):

mgovers:
tests fail

power_grid_model_c/power_grid_model/include/power_grid_model/component/appliance.hpp
```diff
@@ -94,17 +94,17 @@ class Appliance : public Base {
...
-        if constexpr (!sym) {
-            output.i = base_i_ * cabs(i);
-            output.i_angle = arg(i * injection_direction());
-        }
+        output.i = base_i_ * cabs(i);
+        output.i_angle = arg(i * injection_direction());
         return output;
     }
+    ApplianceShortCircuitOutput get_sc_output(ComplexValue<true> const& i) const {
+        ComplexValue<false> iabc{i, i * a2, i * a};
```
TonyXiang8787, line 105: add `const`

power_grid_model_c/power_grid_model/include/power_grid_model/component/branch.hpp
```diff
@@ -112,22 +112,25 @@ class Branch : public Base {
...
+        output.i_to = base_i_to() * cabs(i_t);
+        output.i_from_angle = arg(i_f);
+        output.i_to_angle = arg(i_t);
         return output;
     }
 
+    BranchShortCircuitOutput get_sc_output(ComplexValue<true> const& i_f, ComplexValue<true> const& i_t) const {
+        // Convert the input of only positive sequence to abc
+        ComplexValue<false> iabc_f{i_f, i_f * a2, i_f * a};
+        ComplexValue<false> iabc_t{i_t, i_t * a2, i_t * a};
```
TonyXiang8787, line 130: add `const`

power_grid_model_c/power_grid_model/include/power_grid_model/component/branch3.hpp
```diff
@@ -113,24 +113,27 @@ class Branch3 : public Base {
...
+        output.i_1_angle = arg(i_1);
+        output.i_2_angle = arg(i_2);
+        output.i_3_angle = arg(i_3);
         return output;
     }
+    Branch3ShortCircuitOutput get_sc_output(ComplexValue<true> const& i_1, ComplexValue<true> const& i_2,
+                                            ComplexValue<true> const& i_3) const {
+        ComplexValue<false> iabc_1{i_1, i_1 * a2, i_1 * a};
+        ComplexValue<false> iabc_2{i_2, i_2 * a2, i_2 * a};
+        ComplexValue<false> iabc_3{i_3, i_3 * a2, i_3 * a};
```
TonyXiang8787, line 134: add `const`

power_grid_model_c/power_grid_model/include/power_grid_model/component/node.hpp
```diff
@@ -47,24 +47,28 @@ class Node final : public Base {
...
-        }
+        output.u_pu = cabs(u_pu);
+        output.u = u_scale<false> * u_rated_ * output.u_pu;
+        output.u_angle = arg(u_pu);
+
         return output;
     }
+    NodeShortCircuitOutput get_sc_output(ComplexValue<true> const& u_pu) const {
+        // Convert the input positive sequence voltage to phase voltage
+        ComplexValue<false> uabc_pu{u_pu, u_pu * a2, u_pu * a};
```
TonyXiang8787, line 62: add `const`

TonyXiang8787 (changes requested):
Add `const` for some variables

power_grid_model_c/power_grid_model/include/power_grid_model/component/node.hpp
nitbharambe: Added const at all such instances.

power_grid_model_c/power_grid_model/include/power_grid_model/component/appliance.hpp
```diff
@@ -94,17 +94,17 @@ class Appliance : public Base {
...
-        if constexpr (!sym) {
-            output.i = base_i_ * cabs(i);
-            output.i_angle = arg(i * injection_direction());
-        }
+        output.i = base_i_ * cabs(i);
+        output.i_angle = arg(i * injection_direction());
         return output;
     }
+    ApplianceShortCircuitOutput get_sc_output(ComplexValue<true> const& i) const {
+        ComplexValue<false> const iabc{i, i * a2, i * a};
```
TonyXiang8787, line 105: I think we can do it using just one value.

https://github.com/PowerGridModel/power-grid-model/blob/a7eae1a3ad013191bd0a9f1afe074c7f907e7008/power_grid_model_c/power_grid_model/include/power_grid_model/three_phase_tensor.hpp#L49

`ComplexValue<false>` has a constructor which take one complex value, and automatically rotate for three phase. So you don't need to multiply it as you did there.

Please change it for all the places

TonyXiang8787:
@nitbharambe why did you not change the fault object output function?

nitbharambe:
> @nitbharambe why did you not change the fault object output function?

Was not in scope of this ticket before. Should I do that too? 
Also the output function in ybus is pending too.

TonyXiang8787 (approved):

PR #286, opened 2023-06-27
split up python calculation core

mgovers:
Refactor python implementation of the calculation core


src/power_grid_model/core/options.py
```diff
@@ -17,16 +17,21 @@ class OptionSetter:
...
         self._setter(instance.opt, value)
+        self._value = value
 
-    def __get__(self, instance, owner):
-        raise NotImplementedError("Cannot get option value!")
+    def __get__(self, instance, owner) -> Any:
+        """Return the value that was most recently set.
+
+        No guarantees are given on whether the value was set correctly."""
+        return self._value
```
TonyXiang8787, line 34: Why do we want this? The descriptor is a global object, it will not have the correct value.

src/power_grid_model/core/power_grid_model.py
```diff
@@ -149,92 +157,95 @@ def get_indexer(self, component_type: str, ids: np.ndarray):
...
+    # pylint: disable=too-many-arguments
+    @staticmethod
+    def _options(
         calculation_type: CalculationType,
         symmetric: bool,
         error_tolerance: float,
         max_iterations: int,
         calculation_method: Union[CalculationMethod, str],
-        update_data: Optional[Dict[str, Union[np.ndarray, Dict[str, np.ndarray]]]],
         threading: int,
```
TonyXiang8787, line 195: some of the options are not relevant for short circuit. Maybe instead of list all the options, just use `**kw_options` in the argument?

src/power_grid_model/core/power_grid_model.py
```diff
@@ -149,92 +157,95 @@ def get_indexer(self, component_type: str, ids: np.ndarray):
...
+    ) -> Options:
+        if isinstance(calculation_method, str):
+            calculation_method = CalculationMethod[calculation_method]
+
+        opt = Options()
+        opt.calculation_type = calculation_type.value
```
TonyXiang8787, line 201: here you just do 
```python
for k, v in kw_options.items():
    setattr(opt, k, v)
```

src/power_grid_model/core/power_grid_model.py
```diff
@@ -248,25 +259,54 @@ def _calculate(
...
-        if not batch_calculation:
-            result_dict = {k: v.ravel() for k, v in result_dict.items()}
-
-        return result_dict
+        self._handle_errors(continue_on_batch_error=continue_on_batch_error, batch_size=batch_size)
+
+        return reduce_output_data(output_data=output_data, batch_calculation=batch_calculation)
+
+    # pylint: disable=too-many-arguments
+    def _calculate(
```
TonyXiang8787, line 267: I would remove this function, just call `_options` and `_calculate_impl` in power flow, state estimation, and short circuit.

src/power_grid_model/core/options.py
mgovers: my reasoning was that rather than keeping a dict of all desired settings, you can read it from the options, which should contain those exact setting anyway. this keeps e.g. the amount of arguments low.

if you have any better proposal, i'm more than happy to implement that

src/power_grid_model/core/power_grid_model.py
mgovers: i thought about that beforehand, yes, but i did struggle with implementation. will retry 😄

src/power_grid_model/core/options.py
TonyXiang8787: You mean the amount of argument in `calculate_power_flow`?

mgovers: yes but also in `_calculate_impl`

TonyXiang8787: We don't have any option related stuff in `_calculate_impl` right? Just a single argument `Options`. I still do not get it why do we need a getter here.

TonyXiang8787: I think `symmetric` is too fundamental and you need to pass it as a separate argument in `_calculate_impl`

mgovers: fixed

src/power_grid_model/core/power_grid_model.py
mgovers: ## option 1

```py
class Options:
    """
    Option class
    """

    _opt: OptionsPtr
    # option setter
    calculation_type = OptionSetter(pgc.set_calculation_type)
    calculation_method = OptionSetter(pgc.set_calculation_method)
    symmetric = OptionSetter(pgc.set_symmetric)
    error_tolerance = OptionSetter(pgc.set_err_tol)
    max_iterations = OptionSetter(pgc.set_max_iter)
    threading = OptionSetter(pgc.set_threading)
```

## option 2a

```py
    @staticmethod
    def _options(**kwargs) -> Options:
        calculation_method = kwargs.get("calculation_method")
        if calculation_method is not None and isinstance(calculation_method, str):
            kwargs["calculation_method"] = CalculationMethod[calculation_method]

        if "max_iterations" in kwargs:
            kwargs["max_iteration"] = kwargs["max_iterations"]

        opt = Options()
        for key, value in kwargs.items():
            setattr(opt, key, value.value if isinstance(value, IntEnum) else value)
        return opt
```


## option 2b

```py
class Options:
    """
    Option class
    """

    _opt: OptionsPtr
    # option setter
    calculation_type = OptionSetter(pgc.set_calculation_type)
    calculation_method = OptionSetter(pgc.set_calculation_method)
    symmetric = OptionSetter(pgc.set_symmetric)
    error_tolerance = OptionSetter(pgc.set_err_tol)
    max_iterations = OptionSetter(pgc.set_max_iter)
    threading = OptionSetter(pgc.set_threading)
    max_iteration = max_iterations
```

TonyXiang8787: No, that's nothing to do with C-API. It's just the name of the descriptor. You can change it. Everything inside `core` is not public Python API, unless imported at root level.

mgovers: no the name in the C-API is completely different:

```c
/**
 * @brief Specify maximum number of iterations. Only applicable if using iterative method.
 *
 * @param handle
 * @param opt pointer to option instance
 * @param max_iter Maximum number of iterations
 */
PGM_API void PGM_set_max_iter(PGM_Handle* handle, PGM_Options* opt, PGM_Idx max_iter);
```

```cpp
// options
struct PGM_Options {
    Idx calculation_type{PGM_power_flow};
    Idx calculation_method{PGM_default_method};
    Idx symmetric{1};
    double err_tol{1e-8};
    Idx max_iter{20};
    Idx threading{-1};
};
```

mgovers: > Go for Option 1

ok 😄 that has my preference as well

src/power_grid_model/core/options.py
```diff
@@ -17,16 +17,21 @@ class OptionSetter:
...
     def __init__(self, setter):
         self._setter = setter
 
     def __set__(self, instance: "Options", value: Any):
         self._setter(instance.opt, value)
+        self._value = value
 
-    def __get__(self, instance, owner):
-        raise NotImplementedError("Cannot get option value!")
+    def __get__(self, instance, owner) -> Any:
```
TonyXiang8787, line 30: You are adding this back again in this commit...

mgovers: huh, weird :\ thanks for catching this

src/power_grid_model/core/options.py
```diff
@@ -17,13 +17,15 @@ class OptionSetter:
     setter for options
     """
 
+    _value: Any
```
TonyXiang8787, line 20: also remove this

src/power_grid_model/core/options.py
```diff
@@ -17,13 +17,15 @@ class OptionSetter:
...
     _setter: Callable
 
     def __init__(self, setter):
         self._setter = setter
 
     def __set__(self, instance: "Options", value: Any):
         self._setter(instance.opt, value)
+        self._value = value
```
TonyXiang8787, line 28: also remove this

TonyXiang8787 (approved):

PR #288, opened 2023-06-29
Feature/modernize headers

### LINKED ISSUE #236: Modernize header includes
Headers are currently included using double-quotes and starting with global namespace.

## Proposed structure

While no consensus exists on the header include order and specific structure, and different parties choose different include styles, there are good arguments to start from local private headers, and cascading up. [This thread](https://stackoverflow.com/questions/2762568/c-c-include-header-file-order) (and mainly its comments to the accepted answer) explains why it is better to use this order:

1. `#include "private_header.cpp"` locally to the current sources dir (e.g.: `test_helper.hpp` living next to `test_xyz.cpp`)
2. `#include <current_library/current_file.hpp>` or `#include "current_library/current_file.hpp"`
3. `#include <current_library/other_headers.hpp>` or `#include "current_library/other_headers.hpp"`
4. `#include <other_project_libraries/headers.hpp>`
5. `#include <third_party_libraries/headers.hpp>`
6. `#include <system_headers>`
7. `#include <stl_headers>`

### Use of double quotes vs brackets

Cfr. [gcc documentation](http://gcc.gnu.org/onlinedocs/gcc-2.95.3/cpp_1.html#SEC6)

`#include "xyz.hpp"` looks into the current directories first, and then into other party's libraries. For private headers, this is fine - you know the desired path perfectly - but for public library headers, different OSes may use different conventions.

Instead, `#include <xyz.hpp>` starts looking into the system headers first. Any dependency / library linked to the current library or executable may be considered a system header

### Include order

Because of the way `#include` works (by raw text insertion), any functions, declarations and definitions defined in a later header will hide a former declaration. If a user tries to include our header, he/she expects that e.g. the `min` and `sqrt` functions - if defined - are the ones from STL. Therefore, it is a good idea do include those STL headers last, then 3pp libraries, etc.

In addition, for framework developers, this approach will highlight issues with the current approach when e.g. STL or a 3rd party dependency is updated with a function that has e.g. the same name as one of our own, because our unit tests will fail. This improves Agile WoW by implementing fail-fast behaviour.

Laurynas-Jagutis:
Closes #236

- The first header files will now be the ones which are contained in quotation marks as in "HEADER FILE", then others will follow
- Clang formatter used to format the headers in an undesirable way, placing <HEADER FILE> above "HEADER FILE", thus //clang-format off and //clang-format on were used to indicate the places where the formatter should not reformat anything


tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -2,6 +2,13 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
+#include "doctest/doctest.h"
+#include "nlohmann/json.hpp"
```
mgovers, line 6: `nlohmann` is an external library and should be included as `<nlohmann/json.hpp>`

tests/c_api_tests/test_c_api_meta_data.cpp
```diff
@@ -2,12 +2,13 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-#include <memory>
-
 #include "c_api_cpp_handle.hpp"
 #include "doctest/doctest.h"
```
mgovers, line 6: `doctest` is an external library and should be included `#include <doctest/doctest.h>`

power_grid_model_c/power_grid_model/include/power_grid_model/power_grid_model.hpp
```diff
@@ -6,7 +6,10 @@
 #ifndef POWER_GRID_MODEL_POWER_GRID_MODEL_HPP
 #define POWER_GRID_MODEL_POWER_GRID_MODEL_HPP
 
+#include "boost/iterator/counting_iterator.hpp"
```
mgovers, line 9: boost is an external library and should be included `#include <boost/...>`

power_grid_model_c/power_grid_model/include/power_grid_model/three_phase_tensor.hpp
```diff
@@ -9,11 +9,13 @@
 #include "power_grid_model.hpp"
 
 // eigen properties
+// clang-format off
 #include <Eigen/Dense>
```
mgovers, line 13: I guess `Eigen` should have it's own block (separated by whitelines) together with other external libraries `boost`, `nlohmann` and `doctest`. that may be one potential argument in favour of the `IncludeCategories` (but as I said before, I prefer not to add that, so if it autodetects, that would be great)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -6,13 +6,14 @@
...
-#include <string>
-
 #include "../enum.hpp"
 #include "../exception.hpp"
 #include "../power_grid_model.hpp"
 #include "../three_phase_tensor.hpp"
+// clang-format off
+#include <bit>
+#include <string>
+// clang-format on
```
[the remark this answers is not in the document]
mgovers: because they effectively disable some of our own checks. that's something that should only be done as a last resort

.clang-tidy
```diff
@@ -18,7 +18,7 @@ concurrency-*,
 cppcoreguidelines-*,
 -cppcoreguidelines-avoid-c-arrays,
 -cppcoreguidelines-avoid-magic-numbers,
--cppcoreguidelines-init-variables,
+-cppcoreguidelines-init-variables,F
```
mgovers, line 21:
```suggestion
-cppcoreguidelines-init-variables
```

.clang-format
```diff
@@ -17,3 +17,24 @@ BraceWrapping:
   BeforeCatch: true
   BeforeElse: true
 
+IncludeBlocks: Regroup
+IncludeCategories:
+- Regex:           '^"power_grid_model\.hpp"$'
```
mgovers, line 22: why not just use the one below? i'm not sure whether this should be a separate `Priority` (group). Instead, you could make it `Priority: 1` and `SortPriority: 0`, but personally, I believe you can just leave this entry out.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
```diff
@@ -11,16 +11,13 @@
 #include "../power_grid_model.hpp"
 #include "../three_phase_tensor.hpp"
 #include "../timer.hpp"
```
mgovers, line 13: maybe create a separate regex for this type of entries. they lie outside the current "submodule" and therefore should be treated as external relative to the current module. And yes that may cause some files to first include local things and then `../power_grid_model.hpp`, but in those cases, that `../power_grid_model.hpp` index may be removed anyway, so there is no issue

.clang-tidy
```diff
@@ -18,7 +18,7 @@ concurrency-*,
 cppcoreguidelines-*,
 -cppcoreguidelines-avoid-c-arrays,
 -cppcoreguidelines-avoid-magic-numbers,
--cppcoreguidelines-init-variables,
+-cppcoreguidelines-init-variables
```
mgovers, line 21:
```suggestion
-cppcoreguidelines-init-variables,
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
```diff
@@ -6,20 +6,19 @@
 #ifndef POWER_GRID_MODEL_MATH_SOLVER_MATH_SOLVER_HPP
 #define POWER_GRID_MODEL_MATH_SOLVER_MATH_SOLVER_HPP
 
-#include <optional>
+#include "iterative_current_pf_solver.hpp"
+#include "iterative_linear_se_solver.hpp"
+#include "linear_pf_solver.hpp"
+#include "newton_raphson_pf_solver.hpp"
+#include "y_bus.hpp"
```
mgovers, line 13: @TonyXiang8787 is there any particular reason why this was marked `// clang-format off` in `main`? I prefer the new implementation but want to double-check with you

mgovers (approved):
LGTM, but one small question remaining

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
TonyXiang8787: For some reason the compilation failed if we allowed clang-format to reorder those two headers. @bramstoeller and I have investigated a bit but could not find the course. It is now compiling fine. So maybe that issue is gone. Let merge it in the new way.

mgovers: great 😄

PR #292, opened 2023-07-05
Feature/short circuit input validation

mgovers:
- [x] Validate the input of fault objects + short circuits
  - [x] fault_phase is not required
    - [x] Verify that this is processed correctly in the main model  <= THIS IS NOT HANDLED CORRECTLY YET. OUT OF SCOPE OF THIS PR
  - [x] Within one scenario, multiple faults may exist but they have to have the same fault_type and fault_phase
  - [x] fault object: only check short_circuit for all objects
    - [x] cross referencing + ids should always be valid (power_flow)
  - [x] x can be anything (default 0.0), r should be >= 0 (default 0.0)
  - [x] if all scenarios in batch are three-phase fault, you don’t need zero-sequence parameters
- [x] Update the documentation
  - [x] Make `fault_phase` optional
- [x] Fix output of `all_unique` validation rule (also update its tests)


docs/user_manual/components.md
```diff
@@ -127,6 +127,12 @@ If `i_n` is not provided, `loading` of line will be a `nan` value.
 | `tan0` | `double`  | -          | zero-sequence shunt loss factor (tan𝛿)     | &#10024; only for asymmetric calculations | &#10060; |                                   |
 | `i_n`  | `double`  | ampere (A) | rated current                              |                 &#10060;                  | &#10060; |               `> 0`               |
 
+```{note}
```
nitbharambe, line 130: Lets probably mention this in the calculation page of short circuit. Does not fit under the input of node title.

nitbharambe: Or maybe we can modify the required column instead

src/power_grid_model/validation/errors.py
```diff
@@ -428,11 +446,24 @@ class InfinityError(SingleFieldValidationError):
...
+class FaultPhaseError(MultiFieldValidationError):
+    """
+    The fault phase does not match the fault type.
+    """
+
+    _message = (
+        "The fault phase does not match the fault type for {n} {objects}. "
+        "If the fault type is three_phase, the fault phase should be abc. "
+        "If the fault type is single_phase_to_ground, the fault phase should be either a, b or c. "
+        "If the fault type is either two_phase or two_phase_to_ground, the fault phase should be either ab, ac or bc."
```
nitbharambe, line 468: The message seems a bit long and i think we can leave this to user  to find in the documentation.

A better message would be :

```suggestion
        "The fault phase is not applicable to the corresponding fault type for {n} {objects}. "
```

docs/user_manual/components.md
mgovers: what do you propose?

mgovers: I do believe that this should be mentioned here as well. Otherwise, not all info will be present where expected. I also added an entry for the short circuit calculations section with this node.

TonyXiang8787:
@mgovers  is this ready for review?

mgovers:
> @mgovers is this ready for review?

yes :+1:

docs/user_manual/components.md
nitbharambe: You are right. It would kind of seem odd in the short circuit calculation page. My concern was that there too many notes below the tables around the page which makes it less readable, plus its mentioned twice.

 Technically, the required column mentions `&#10024; only for asymmetric calculations` which should be self sufficient. Just `&#10024;` in table and `&#10024; : Only for asymmetric power-flow and short-circuit calculations` below the table is also a possibility. What do you  think?

TonyXiang8787:
@nitbharambe can you review this PR?

nitbharambe:
Yes. I shall do that @TonyXiang8787

tests/unit/validation/test_rules.py
```diff
@@ -212,6 +218,34 @@ def test_none_match_comparison():
...
+    data = {
+        "bar": np.array([(0, 10), (1, 10), (2, 10)], dtype=dtype),
+        "baz": np.array([(3, 11), (4, 12), (5, 12)], dtype=dtype),
+    }
+    errors = all_identical(data, "bar", "foo")
+    assert not errors
+
+    errors = all_identical(data, "baz", "foo")
+    assert len(errors) == 1
+    assert NotIdenticalError("baz", "foo", ids=[3, 4, 5], values=[11, 12, 12])
```
nitbharambe, line 232: Missed `in errors` here and below.

docs/user_manual/calculations.md
```diff
@@ -409,6 +410,16 @@ crash due to an unobservable system, but succeeding with the calculations and gi
...
 
+### Short circuit algorithms
+
+**TODO**
+
+```{note}
+In case of short circuit calculations, the zero-sequence parameters are required only
+if any of the faults in any of the scenarios within a batch are not three-phase faults
+(i.e. `fault_type` is not `FaultType.three_phase`).
+```
```
nitbharambe, line 421:
```suggestion
```{note}
In case of short circuit calculations, the zero-sequence parameters are required only
if any of the faults in any of the scenarios within a batch are not three-phase faults
(i.e. `fault_type` is not `FaultType.three_phase`).
```
```
Agreed with the part that this is not where the user would try to find the info.

docs/user_manual/components.md
nitbharambe: Or we leave it as you originally did and deal with notes later if they become a big deal, apologies for this nitpick :D

nitbharambe (approved):

nitbharambe (approved):

PR #293, opened 2023-07-05
short circuit solver unit test

nitbharambe:
Solver unit test with a source line and fault. Test solid and non solid fault of all types


mgovers:
since you're rebased off of my branch, you'll have to change base to `feature/short-circuit-solver`

tests/cpp_unit_tests/test_math_solver.cpp
```diff
@@ -536,6 +556,264 @@ TEST_CASE("Test math solver") {
...
+    return sc_output_ref;
+}
+
+// TODO Un-skip unit test later
+TEST_CASE("Short circuit solver" * doctest::skip(true)) {
+    /*
+
+    source -- bus --- line -- bus -- fault(type varying as per subcase)
+
+    */
```
mgovers, line 648: please use `//` for this type of comments

mgovers (commented):
a couple small remarks but overall looks good

tests/cpp_unit_tests/test_math_solver.cpp
```diff
@@ -83,6 +87,26 @@ void assert_output(MathOutput<sym> const& output, MathOutput<sym> const& output_
     }
 }
 
+template <bool sym>
```
nitbharambe, line 90: Should this and the existing function for powerflow be in a unnamed namespace too? @mgovers

mgovers: technically yes, but keep it out of scope for now.

mgovers:
The failing function on `macos` is being fixed by @petersalemink95 on his branch. we do not need to care about that in this PR.

mgovers (approved):
the code looks really clean and readable now. i can understand exactly what's happening, what's being tested and why something fails or passes.

PR #294, opened 2023-07-05
Feature/separate nan default enum values

mgovers:
* separate `default_value` and from `nan` for `FaultPhase` and `FaultType`
  * document this best practice for all future updateable optional enums
* remove `FaultType::default_value`
* implement the `fault_phase` value depending on `fault_type`

This PR deliberately depends on the input validation PR #292 , because of the potentially large conflicts.

This is technically a breaking change, but because it is only in the `release/` branch and not in `main`, this is OK. Rather now than in the future.


power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
```diff
@@ -125,6 +125,22 @@ class Fault final : public Base {
...
+            switch (fault_type_) {
+                case FaultType::three_phase:
+                    return abc;
+                case FaultType::single_phase_to_ground:
+                    return a;
+                case FaultType::two_phase:
+                    [[fallthrough]];
+                case FaultType::two_phase_to_ground:
+                    return bc;
+                case FaultType::nan:
```
TonyXiang8787, line 140: `fault_type` is a mandatory attribute, so it will never has the value `nan` in the class object. So here you need to throw `InvalidEnum`, because it should not happen.

mgovers: it could happen that the user sets it to `nan` in the input data (constructor) and then does a later batch update where he sets the value, right?

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
```diff
@@ -141,15 +157,25 @@ class Fault final : public Base {
...
     void check_sanity() const {
-        if (fault_type_ == FaultType::three_phase) {
-            if (fault_phase_ != FaultPhase::abc) {
-                throw InvalidShortCircuitPhases(fault_type_, fault_phase_);
-            }
-        }
-        else {
-            if (fault_phase_ == FaultPhase::abc) {
-                throw InvalidShortCircuitPhases(fault_type_, fault_phase_);
+        auto supported = [](FaultType fault_type) -> std::vector<FaultPhase> {
```
TonyXiang8787, line 160: Is using a dynamic vector here too much? Can you find a way of only using a stack variable? Maybe an `std::array` with maximum length?

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
```diff
@@ -141,15 +157,25 @@ class Fault final : public Base {
...
+            switch (fault_type) {
+                case FaultType::three_phase:
+                    return {FaultPhase::nan, default_value, abc};
+                case FaultType::single_phase_to_ground:
+                    return {FaultPhase::nan, default_value, a, b, c};
+                case FaultType::two_phase:
+                    [[fallthrough]];
+                case FaultType::two_phase_to_ground:
+                    return {FaultPhase::nan, default_value, ab, ac, bc};
+                case FaultType::nan:
```
TonyXiang8787, line 171: same here, a `InvalidEnum` should be thrown.

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
mgovers: i can try but that feels like overengineering to me. i'll have a look

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
TonyXiang8787: This function `get_fault_phase` is not called in the constructor right? Only at the moment you call `calc_param`. There you should have a valid `fault_type`.

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
TonyXiang8787: At this moment all the component classes do not have any dynamic allocation. It would be good to keep in this way.

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
mgovers: this one **is** checked during the update, so should i accept the `nan` value here or not?

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
```diff
@@ -125,6 +125,22 @@ class Fault final : public Base {
         return fault_type_;
     }
     FaultPhase get_fault_phase() const {
```
TonyXiang8787, line 127: This function should be called in `calc_param`, there you assign `fault_type` and `fault_phase` to the math input.

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
TonyXiang8787: You are correct. The function is called in the constructor and the `fault_type` can be `nan` there. So we can keep this as is for this one.

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
TonyXiang8787: I feel like using the whole lambda and a `std::find` itself is a bit too much? Maybe just a simple switch on `fault_type` and each case just a simple if throw statement is enough?

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
TonyXiang8787, line 1: In the constructor of `Fault`, if the `fault_phase` is `nan`, you change it to `default_value` and assign to the class member.

power_grid_model_c/power_grid_model/include/power_grid_model/component/fault.hpp
mgovers: i tried that but it got incredibly verbose. can you have a quick look at my new solution? i think this one looks quite clean and can be resolved compile-time (aside from the `throw` statements)

TonyXiang8787: Your current implementation should be fine.

PR #301, opened 2023-07-11
Reorganize validation cases

nitbharambe:


tests/data/short_circuit/three_phase/update_batch.json
```diff
@@ -2,7 +2,7 @@
   {
     "fault":
       [
-        {"id": 10, "fault_object": 1, "r": 0, "x": 0}
```
mgovers, line 5: don't we need cases with `nan` and/or `default_value` for `fault_phase`?

tests/data/short_circuit/single_phase_to_ground/params.json
```diff
@@ -0,0 +1,5 @@
+{
+  "calculation_method": "short_circuit",
```
mgovers, line 2: here and below: this is incorrect and outdated. this should be `ie60909`.

did you make sure to run `python -m pip install .` on the `pgm-test-generator` and/or merge with main before running `python main.py` on the `pgm-test-generator`?

maybe it's a good idea to just clean your build environment and/or `venv` on the `pgm-test-generator` and start with a clean workspace

mgovers (changes requested):
Please first use the correct `calculation_method` etc

tests/data/short_circuit/three_phase/update_batch.json
nitbharambe: Let us not include those in the validation cases since their results will be just a duplicate and crowd the already big cases. We can keep the validation cases only to theoretical validation of the algorithm. 

I can add default value as a case in unit tests instead.

tests/data/short_circuit/single_phase_to_ground/params.json
nitbharambe: Yeah, there was a merge conflict ;(

mgovers (approved):

Issue #267, opened 2023-06-08
Resolve code smells from `clang-tidy` on C++ side


mgovers:
The [`clang-tidy` action](https://github.com/PowerGridModel/power-grid-model/actions/workflows/clang-tidy.yml) checks for static analyzer warnings (similarly to SonarCube). Currently, there are quite some checks that are disabled in the [.clang-tidy](https://github.com/PowerGridModel/power-grid-model/blob/main/.clang-tidy) file. Disabled warnings are prefixed `-` in the `.clang-tidy` file, e.g. `-llvm-else-after-return`.

Note that some checks are more easily fixable than others! If you get stuck on one check, just move on to the next one! Check the list of easily fixable checks below.

## Desired solution

* Create a branch
* Remove one disabled check from the list easily fixable checks below
  * Reasonably good explanations of the error exist in the LLVM docs (just do a Google search on the name of the error).
  * Sometimes, the `clang-tidy` error message and following notes provide suggestions on how to fix the issue. Most of the times, these proposed changes are good, but beware about applying them blindly as sometimes, they fail to consider the greater picture.
* Compile using one of the CMake Presets with `clang-tidy` in the name, as explained in the [Documentation](https://power-grid-model.readthedocs.io/en/stable/advanced_documentation/build-guide.html#environment-variables), or wait for CI to fail on the given check.
* Fix the `clang-tidy` errors
* Create a pull request

It is good practice to fix only one check per pull request to keep the amount of changes manageable.

## Compilation of checks

As mentioned above, some clang-tidy checks may be hard to fix, especially for people that are relatively new to C++. Here is a list of good starting points. Again, if you get stuck, don't focus too long on one check but just move on to the next one. There are plenty low hanging fruits left.

### Suggestions for easily fixable checks

* `cppcoreguidelines-*` checks are important but sometimes hard to resolve. Here are some of the few exceptions that may be easy regardless:`
  * [ ] `-cppcoreguidelines-narrowing-conversions`
  * [ ] `-cppcoreguidelines-prefer-member-initializer`
  * [ ] `-cppcoreguidelines-special-member-functions`
* `performance-*` checks are almost always useful and easy to fix!
  * [ ] `-performance-no-automatic-move`
  * [ ] `-performance-unnecessary-copy-initialization`
* `bugprone-*` checks help write code better, faster
  * [ ] `-bugprone-macro-parentheses`
  * [ ] `-bugprone-narrowing-conversions`
[...]

PR #302, opened 2023-07-11
fix cppcoreguidelines-prefer-member-initializer

Laurynas-Jagutis:
**Fixes issue**: fixes cppcoreguidelines-prefer-member-initializer from Resolve code smells from clang-tidy on C++ side #267


power_grid_model_c/power_grid_model/include/power_grid_model/component/source.hpp
```diff
@@ -28,8 +28,11 @@ class Source : public Appliance {
     }
 
     explicit Source(SourceInput const& source_input, double u)
-        : Appliance{source_input, u}, u_ref_{source_input.u_ref}, u_ref_angle_{}, y1_ref_{}, y0_ref_{} {
-        u_ref_angle_ = is_nan(source_input.u_ref_angle) ? 0.0 : source_input.u_ref_angle;
+        : Appliance{source_input, u},
+          u_ref_{source_input.u_ref},
+          u_ref_angle_{is_nan(source_input.u_ref_angle) ? 0.0 : source_input.u_ref_angle},
+          y1_ref_{},
+          y0_ref_{} {
```
mgovers, line 35: to get rid of the sonar cloud issue, you can probably modify the statement where this variable is created and add `{}` before the `;`, like I did in this commit: https://github.com/PowerGridModel/power-grid-model/pull/296/commits/8e5b2954020014399b6294f741c5f681579e3597

mgovers, line 35: since you now call the default initializer in the member definition (compile time) instead of in the constructor (runtime), you may leave leave these initializations in the constructor
```suggestion
          u_ref_angle_{is_nan(source_input.u_ref_angle) ? 0.0 : source_input.u_ref_angle} {
```

mgovers: resolved in favour of https://github.com/PowerGridModel/power-grid-model/pull/302/files#r1260654721

mgovers: cfr. the code smell:

> Remove this use of the constructor's initializer list for data member "y1_ref_". It is redundant with the in-class initializer.

mgovers: according to the explanation in the sonar cloud code smell:

> For C++11 or later, when you initialize a member variable in the initializer list of a constructor, but could have done so directly in the class:
> 
>     The variable has either no in-class initializer, or an in-class initializer with the same value as in the constructor
>     The initial value does not depend on a constructor parameter

mgovers (approved):

PR #306, opened 2023-07-13
Use enum for ctype indication in the C-API

TonyXiang8787:
# Problem

Previously we use string to indicate the ctype of an attribute in the C-API. This results in slow string look-up.

# Solution

We change it to enum.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
@@ -36,29 +36,30 @@ template <class T>
 struct get_component_nan;
 
 // ctype string
-template <class T, bool is_enum = std::is_enum_v<T>>
+template <class T>
 struct ctype_t;
 template <>
-struct ctype_t<double, false> {
-    static constexpr const char* value = "double";
+struct ctype_t<double> {
+    static constexpr CType value = CType::c_double;
 };
 template <>
-struct ctype_t<int32_t, false> {
-    static constexpr const char* value = "int32_t";
+struct ctype_t<int32_t> {
+    static constexpr CType value = CType::c_int32;
 };
 template <>
-struct ctype_t<int8_t, false> {
-    static constexpr const char* value = "int8_t";
+struct ctype_t<int8_t> {
+    static constexpr CType value = CType::c_int8;
 };
 
 template <>
-struct ctype_t<RealValue<false>, false> {
-    static constexpr const char* value = "double[3]";
+struct ctype_t<RealValue<false>> {
+    static constexpr CType value = CType::c_double3;
 };
 template <class T>
-struct ctype_t<T, true> : ctype_t<std::underlying_type_t<T>> {};
+    requires(std::is_enum_v<T>)
+struct ctype_t<T> : ctype_t<std::underlying_type_t<T>> {};
 template <class T>
-constexpr const char* ctype_v = ctype_t<T>::value;
+constexpr CType ctype_v = ctype_t<T>::value;
 
 // set nan
 inline void set_nan(double& x) {
@@ -74,7 +75,7 @@ inline void set_nan(RealValue<false>& x) {
     x = RealValue<false>{nan};
 }
 template <class Enum>
-requires std::same_as<std::underlying_type_t<Enum>, IntS>
+    requires std::same_as<std::underlying_type_t<Enum>, IntS>
 inline void set_nan(Enum& x) {
     x = static_cast<Enum>(na_IntS);
 }
@@ -115,6 +116,7 @@ struct MetaAttributeImpl {
 // attribute in global namespace
 struct PGM_MetaAttribute {
     using Idx = power_grid_model::Idx;
+    using CType = power_grid_model::CType;
     template <class T>
     using trait_pointer_to_member = power_grid_model::meta_data::trait_pointer_to_member<T>;
     template <class StructType, auto member_ptr>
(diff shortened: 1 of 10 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -36,29 +36,30 @@ template <class T>
...
 template <>
-struct ctype_t<RealValue<false>, false> {
-    static constexpr const char* value = "double[3]";
+struct ctype_t<RealValue<false>> {
+    static constexpr CType value = CType::c_double3;
 };
 template <class T>
-struct ctype_t<T, true> : ctype_t<std::underlying_type_t<T>> {};
+requires(std::is_enum_v<T>) struct ctype_t<T> : ctype_t<std::underlying_type_t<T>> {
+};
```
mgovers, line 60: if i'm correct, you don't need the braces after the `requires` clause

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -36,29 +36,30 @@ template <class T>
...
+    static constexpr CType value = CType::c_int8;
 };
 
 template <>
-struct ctype_t<RealValue<false>, false> {
-    static constexpr const char* value = "double[3]";
+struct ctype_t<RealValue<false>> {
+    static constexpr CType value = CType::c_double3;
 };
 template <class T>
```
mgovers, line 58: maybe for readability and to keep the various template struct specifications similar, you can use a `requires` clause for all of these (using the concept `std::same_as`, you may not even need the `requires` but you can use `template <std::same_as<int8_t> T>` e.g.)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -36,29 +36,30 @@ template <class T>
...
-struct ctype_t<T, true> : ctype_t<std::underlying_type_t<T>> {};
+requires(std::is_enum_v<T>) struct ctype_t<T> : ctype_t<std::underlying_type_t<T>> {
```
mgovers, line 59: i'm not sure but maybe it's possible to solve this without using inheritance by using `using`

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -17,7 +17,14 @@
 from power_grid_model.core.power_grid_core import AttributePtr, ComponentPtr, DatasetPtr, IdxPtr
 from power_grid_model.core.power_grid_core import power_grid_core as pgc
 
-_CTYPE_NUMPY_MAP = {"double": "f8", "int32_t": "i4", "int8_t": "i1", "double[3]": "(3,)f8"}
+# constant enum for ctype
+PGM_INT32 = 0
+PGM_INT8 = 1
+PGM_DOUBLE = 2
+PGM_DOUBLE3 = 3
```
mgovers, line 24: maybe it's a good idea to put this in an `IntEnum`: `class PGM_CType(IntEnum)`. You can then automatically cast from `int` to `PgmCType` and vice versa

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/meta_data.h
```diff
@@ -160,15 +160,10 @@ PGM_API char const* PGM_meta_attribute_name(PGM_Handle* handle, PGM_MetaAttribut
...
+ * @return  Type of the attribute as in enum PGM_CType.
  *
- * Valid types are:
- *   - int32_t
- *   - int8_t
- *   - double
- *   - double[3]
  */
-PGM_API char const* PGM_meta_attribute_ctype(PGM_Handle* handle, PGM_MetaAttribute const* attribute);
+PGM_API PGM_Idx PGM_meta_attribute_ctype(PGM_Handle* handle, PGM_MetaAttribute const* attribute);
```
mgovers, line 166: I know we don't have any C API users yet, but are you sure you want to break the API by removing the old function? I actually would argue providing a `str` representation as well has its advantages

mgovers (approved):
minor comments, overall a really nice improvement

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
TonyXiang8787: Inheritance is the standard way of template meta programming.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
TonyXiang8787: I am not in favor of this. If we can have a template full specialization for other types, we should do it.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/meta_data.h
TonyXiang8787: I think this is our good chance to make breaking changes, given the timing. When we have version 1.5 released and also concrete users on the C-API, we will need to be more careful about breaking changes.

src/power_grid_model/core/power_grid_meta.py
TonyXiang8787: done

mgovers (dismissed):
agreed with all comments

mgovers (approved):

PR #307, opened 2023-07-13
Feature/refactor main model io

mgovers:
* Move the `add_component`, `update_component` and `output_result` functionality outside the `MainModel` class and into `main_core/(input|update|output).hpp`
* Also added a `main_core/state.hpp` to remove the location where the state is stored from the math model cache
* Unblock adding short circuit output due to ambiguity between `get_output` and `get_sc_output` and the potential danger of code duplication

I decided to go with static functions because they are in their own `main_core` namespace so the compiler should not be impacted too much by this. Alternatively, it becomes excessive, we can make a `struct` with only `static` public functions but I feel like that is a suboptimal solution

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -222,6 +222,50 @@ struct ShortCircuitMathOutput {
     std::vector<ComplexValue<sym>> i_source;
 };
 
+template <typename T>
+concept symmetric_math_output_type = std::same_as<T, MathOutput<true>> || std::same_as<T, ShortCircuitMathOutput<true>>;
+
+static_assert(symmetric_math_output_type<MathOutput<true>>);
+static_assert(!symmetric_math_output_type<MathOutput<false>>);
+static_assert(symmetric_math_output_type<ShortCircuitMathOutput<true>>);
+static_assert(!symmetric_math_output_type<ShortCircuitMathOutput<false>>);
+
+template <typename T>
+concept asymmetric_math_output_type =
+    std::same_as<T, MathOutput<false>> || std::same_as<T, ShortCircuitMathOutput<false>>;
+
+static_assert(!asymmetric_math_output_type<MathOutput<true>>);
+static_assert(asymmetric_math_output_type<MathOutput<false>>);
+static_assert(!asymmetric_math_output_type<ShortCircuitMathOutput<true>>);
+static_assert(asymmetric_math_output_type<ShortCircuitMathOutput<false>>);
+
+template <typename T>
+concept steady_state_math_output_type = std::same_as<T, MathOutput<true>> || std::same_as<T, MathOutput<false>>;
+
+static_assert(steady_state_math_output_type<MathOutput<true>>);
+static_assert(steady_state_math_output_type<MathOutput<false>>);
+static_assert(!steady_state_math_output_type<ShortCircuitMathOutput<true>>);
+static_assert(!steady_state_math_output_type<ShortCircuitMathOutput<false>>);
+
+template <typename T>
+concept short_circuit_math_output_type =
+    std::same_as<T, ShortCircuitMathOutput<true>> || std::same_as<T, ShortCircuitMathOutput<false>>;
+
+static_assert(!short_circuit_math_output_type<MathOutput<true>>);
+static_assert(!short_circuit_math_output_type<MathOutput<false>>);
+static_assert(short_circuit_math_output_type<ShortCircuitMathOutput<true>>);
+static_assert(short_circuit_math_output_type<ShortCircuitMathOutput<false>>);
+
+template <typename T>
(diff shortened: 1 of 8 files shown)


TonyXiang8787 (changes requested):
Please fix the build errors first

mgovers:
@TonyXiang8787 you can either review now or wait for a bit so i can clean up a bit more (otherwise, I will just do that in the short circuit implementation PR)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -0,0 +1,287 @@
...
+          typename ResFunc>
+requires model_component_state<MainModelState, ComponentContainer, Component> &&
+    std::invocable<std::remove_cvref_t<ResFunc>, Component const&, IndexType>&&
+        std::convertible_to<std::invoke_result_t<ResFunc, Component const&, IndexType>, std::iter_value_t<ResIt>>&&
+            std::convertible_to<IndexType, decltype(*comp_base_sequence_cbegin<Component>(
+                                               MainModelState<ComponentContainer>{}))> constexpr ResIt
+            produce_output(MainModelState<ComponentContainer> const& state, ResIt res_it, ResFunc&& func) {
+    return std::transform(state.components.template citer<Component>().begin(),
+                          state.components.template citer<Component>().end(),
+                          comp_base_sequence_cbegin<Component>(state), res_it, func);
```
mgovers, line 90: doing it this way ensures the logic between actual conversion (declared later), writing output (declared here) and declaring input (declared earlier in `comp_base_sequence_cbegin`) is split up.

i know it costs quite a lot more lines than initially, but compile time is actually faster than it was before because it can pre-compile all functions separately

mgovers:
all code smells and coverage issues are directly related to equivalent issues in the existing main model

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -308,87 +218,92 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         // set construction_complete for debug assertions
         construction_complete_ = true;
 #endif  // !NDEBUG
-        components_.set_construction_complete();
+        state_.components.set_construction_complete();
         // set component topo
         ComponentTopology comp_topo;
-        comp_topo.n_node = components_.template size<Node>();
+        comp_topo.n_node = state_.components.template size<Node>();
         // fill topology data
```
mgovers, line 225: i would like to move this to a separate file as well but i'd say that is out of scope of this PR cause it's already quite large

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -732,218 +647,25 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+            [](MainModelImpl& model, std::vector<MathOutputType> const& math_output_,
                DataPointer<false> const& data_ptr, Idx position) {
                 auto const begin =
-                    data_ptr.get_iterators<typename ComponentType::template OutputType<sym>>(position).first;
-                model.output_result<sym, ComponentType>(math_output_, begin);
+                    data_ptr
+                        .get_iterators<std::conditional_t<
+                            steady_state_math_output_type<MathOutputType>,
+                            typename ComponentType::template OutputType<symmetric_math_output_type<MathOutputType>>,
+                            typename ComponentType::ShortCircuitOutputType>>(position)
```
mgovers, line 666: i know this looks really ugly but the alternative is to also change all `OutputType` and `ShortCircuitOutputType`s in the individual components to support the calculation type logic and i would like to do that in a separate PR to prevent too many things from being touched at once

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -222,6 +222,50 @@ struct ShortCircuitMathOutput {
     std::vector<ComplexValue<sym>> i_source;
 };
 
+template <typename T>
+concept symmetric_math_output_type = std::same_as<T, MathOutput<true>> || std::same_as<T, ShortCircuitMathOutput<true>>;
```
mgovers, line 226: i would like to replace the `<sym>` template with a tag relating to the calculation type but that requires an even larger overhaul (using `tag` structs instead of `bool sym`) and it's probably better to hold that off until after short circuit is done. this change at least scopes down the other overhaul

mgovers:
@TonyXiang8787 this PR is ready to be reviewed

TonyXiang8787:
@mgovers a general question. Is it conventional to name concepts as snake case instead of camel case?

TonyXiang8787 (approved):
Nice work. Just the question regarding the naming conversion of concepts. If it is indeed conventional to use snake case. Feel free to merge it.

PR #311, opened 2023-07-17
cppcoreguidelines-special-member-functions fix

Laurynas-Jagutis:
1. Fix for "cppcoreguidelines-special-member-functions" from Resolve code smells from clang-tidy on C++ side #267
2. Fix for "performance-no-automatic-move"  from Resolve code smells from clang-tidy on C++ side #267

Set copy constructor, copy assignment operator, move constructor, move assignment operator to default

related to #267


.clang-tidy
```diff
@@ -59,7 +59,6 @@ modernize-*,
 -modernize-use-transparent-functors,
 -modernize-use-using,
 performance-*,
--performance-no-automatic-move,
```
mgovers, line 63: nice :+1:

power_grid_model_c/power_grid_model/include/power_grid_model/component/base.hpp
```diff
@@ -34,6 +34,11 @@ class Base {
     }
     virtual bool energized(bool is_connected_to_source) const = 0;
 
+    Base(const Base&) = default;
+    Base& operator=(const Base&) = default;
+    Base(Base&&) = default;
+    Base& operator=(Base&&) = default;
+
```
mgovers, line 41: instead of adding all `default`, the default `virtual deleter` should be removed in line `28` (`virtual ~Base() = default;`).

Remember: the rule of three-five-zero prefers having zero specializations over five. only if the rule of zero cannot be used should the rule of five be followed

```suggestion
```

Laurynas-Jagutis: If i remove all five of them and rebuild validation tests, I get 18 more warnings treated as errors. I suppose leaving it as it is now might be a good idea?

mgovers: Oh wait my bad, since you explicitly mark it `virtual` and `public`, it **is** needed (cfr. explanation of rule of zero in https://en.cppreference.com/w/cpp/language/rule_of_three )

see also https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c67-a-polymorphic-class-should-suppress-public-copymove

let me investigate how to work around this

mgovers: ok i understand the reason now. you **do** need to have all declared, but please make the copy and copy assignment constructors `protected`. You will likely also have to explicitly declare the copy and copy assignment constructors in the individual components that are **NOT** intended for inheritance (the ones marked `final`, e.g. Line rather than in Branch) as `public`.

mgovers: please note that this is one of the many reasons why people are moving away from polymorphism models (inheritance) 😬

mgovers:
random sidenote: please also refer to https://github.com/PowerGridModel/power-grid-model/issues/267 in the description for this type of PRs. that helps with crosslinking and quick investigations if ever needed

power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
```diff
@@ -56,11 +56,11 @@ class GenericLoadGen : public Appliance {
 };
 
 // abstraction of load/generation
-class GenericLoad : public GenericLoadGen {
+class GenericLoad : public GenericLoadGen {  // generic load private if errors then prot
    public:
     using GenericLoadGen::GenericLoadGen;
 };
-class GenericGenerator : public GenericLoadGen {
+class GenericGenerator : public GenericLoadGen {  // add genload and gengen to asserts
```
mgovers, line 63: remove comment
```suggestion
class GenericGenerator : public GenericLoadGen {
```

tests/cpp_unit_tests/test_all_components.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        U {
+            t
+        }
+        } -> std::same_as<U>;       // copy
+    { u = t } -> std::same_as<U&>;  // copy assignment
+};
+
+// is_abstract_v<GenericLoadGen> && !is_copyable_to<GenericLoadGen, GenericLoadGen>
+
```
mgovers, line 40: remove comment

```suggestion
// is_abstract_v<GenericLoadGen> && !is_copyable_to<GenericLoadGen, GenericLoadGen>

```

mgovers (approved):

mgovers (dismissed):

mgovers:
@Laurynas-Jagutis please run the formatter again using `pre-commit run --all-files` https://github.com/PowerGridModel/power-grid-model/actions/runs/5585874736/jobs/10209177013?pr=311

mgovers (approved):

PR #312, opened 2023-07-18
Feature/short circuit ybus

mgovers:
The failing compilation is caused by the compilation failing on the base branch


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/short_circuit_solver.hpp
```diff
@@ -210,36 +210,35 @@ class ShortCircuitSolver {
                           int const& phase_1, int const& phase_2, double const& source_voltage_ref) {
         // loop through all buses
         for (Idx bus_number = 0; bus_number != n_bus_; ++bus_number) {
-            ComplexValue<sym> const x_tmp = output.u_bus[bus_number];  // save x to temp variable
+            ComplexValue<sym> const u_bus_initial = output.u_bus[bus_number];
```
nitbharambe, line 213: The variable naming as x instead of u here was intentional as the `x_tmp` at this point contains primarily u_bus with few i_fault at `y_fault=inf` condition.

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/short_circuit_solver.hpp
nitbharambe: And we swap i_fault in the coming lines so that it is only u_bus then

mgovers: then still `x_tmp` is not really an elaborate name. do you know anything better than `u_bus_initial`?

mgovers: maybe `sub_total_impedance`?

mgovers:
the issues are with the base branch. will be solved there

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/short_circuit_solver.hpp
mgovers: went with `x_bus_subtotal`

PR #316, opened 2023-07-25
Feature/code smells

Laurynas-Jagutis:

fixes misc-const-correctness


tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -496,9 +496,11 @@ void validate_single_case(CaseParam const& param) {
     execute_test(param, [&]() {
         ValidationCase const validation_case = create_validation_case(param);
         std::string const output_prefix = get_output_type(param.calculation_type, param.sym);
-        SingleData result = create_result_dataset(validation_case.input, output_prefix);
+        SingleData const result = create_result_dataset(validation_case.input, output_prefix);
 
         // create model and run
+        // TODO (mgovers): fix false positive of misc-const-correctness
+        // NOLINTNEXTLINE(misc-const-correctness,-warnings-as-errors)
```
mgovers, line 503: why do you need the `-warnings-as-errors` as well? shouldn't it just be

```suggestion
        // NOLINTNEXTLINE(misc-const-correctness)
```

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -511,9 +513,11 @@ void validate_batch_case(CaseParam const& param) {
     execute_test(param, [&]() {
         ValidationCase const validation_case = create_validation_case(param);
         std::string const output_prefix = get_output_type(param.calculation_type, param.sym);
-        SingleData result = create_result_dataset(validation_case.input, output_prefix);
+        SingleData const result = create_result_dataset(validation_case.input, output_prefix);
 
         // create model
+        // TODO (mgovers): fix false positive of misc-const-correctness
+        // NOLINTNEXTLINE(misc-const-correctness,-warnings-as-errors)
```
mgovers, line 520: why do you need the `-warnings-as-errors` as well? shouldn't it just be

```suggestion
        // NOLINTNEXTLINE(misc-const-correctness)
```

mgovers:
> In the unit tests, I had an issue with test_node.cpp line 63 where the const check complained that we lacked const. However, when I made base_update const I got an issue: **error: default initialization of an object of const type without a user-provided default constructor**. I found this: lballabio/QuantLib#1091 and used their solution which seemed to work just fine, the solution was to simply add {} to the statement

yes this is the correct solution (actually both marking it `const` and adding the `{}` should be the correct solution).

this is related to the fact that STL standard does not specify how the `std::function` class should be implemented. Different libraries may use different solutions, so depending on the provider, it is allowed to contain fundamental types only.

For such objects, there is no user defined constructor and also no default constructor.

* When marking it `const`, there is no issue because a user can assign a value later (at the risk of using the members before assigning them, notice that this is bugprone!!!)
* When marking it `const`, you can only use the members but only assign the value when declaring it. If you don't assing a value, you will **always** use the member before assigning it, and you may always run into bugs

I know this is quite technical but welcome to the C++ world 😬

mgovers (approved):

PR #317, opened 2023-07-25
Feature/macro parantheses

Laurynas-Jagutis:

Fixes:

- bugprone-macro-parentheses
- readability-braces-around-statements
- readability-container-size-empty


power_grid_model_c/power_grid_model/include/power_grid_model/container.hpp
```diff
@@ -35,7 +35,7 @@ inline constexpr size_t get_index_bool_array(std::array<bool, N> arr, size_t idx
 
 template <typename U, typename First, typename... Rest>
 constexpr size_t get_type_index() {
-    if constexpr (std::is_same<U, First>::value) {
+    if constexpr (std::is_same_v<U, First>) {
```
mgovers, line 38: just a matter of taste but you could also use `std::same_as<U, First>` here. it's not exactly the same but for most use cases it is. saves another `_v`. Don't worry about it too much

.clang-tidy
```diff
@@ -59,8 +58,6 @@ modernize-*,
 performance-*,
 portability-*,
 readability-*,
--readability-braces-around-statements,
```
mgovers, line 62: @TonyXiang8787 this relates to the `.clang-format` as well. do you have a strong opinion about not adding braces around statements? If not, I suggest we go with the LLVM convention of adding them. In that case, we should also update the `.clang-format` file with that (provided that the `diff` is small, which I expect to be the case after this PR)

tests/cpp_unit_tests/test_math_solver.cpp
```diff
@@ -51,12 +51,12 @@ TEST_CASE("Test block") {
...
-        if constexpr (sym)                              \
-            CHECK(cabs((x) - (y)) < tolerance);         \
-        else                                            \
-            CHECK((cabs((x) - (y)) < tolerance).all()); \
+#define CHECK_CLOSE(x, y, tolerance)                      \
+    do {                                                  \
+        if constexpr (sym)                                \
+            CHECK(cabs((x) - (y)) < (tolerance));         \
+        else                                              \
+            CHECK((cabs((x) - (y)) < (tolerance)).all()); \
```
mgovers, line 59: please also use the `break before braces` here:

```suggestion
        if constexpr (sym) {                              \
            CHECK(cabs((x) - (y)) < (tolerance));         \
        }                                                 \
        else {                                            \
            CHECK((cabs((x) - (y)) < (tolerance)).all()); \
        }
```

.clang-tidy
TonyXiang8787: What braces are you referring?

mgovers: the braces in the case of short `if` statements, e.g.:

```cpp
if (foo)
    bar;
else
    baz;

etc
```

vs.

```cpp
if (foo) {
    bar;
}
else {
    baz;
}

etc
```

TonyXiang8787: My opinion is: always with braces when you put the statement in a new line. If you put the statement at the same line of `if`, you can omit braces.

mgovers: ok, then I will update the `.clang-format` file accordingly

mgovers: investigated this and it's probably not desired to add this to the `.clang-format` file since it doesn't know semantics and therefore could lead to incorrect formatting. instead, let's keep it in the `.clang-tidy` file (because that actually takes semantics into account).

mgovers (approved):

TonyXiang8787:
@mgovers  can this be merged?

mgovers:
> @mgovers can this be merged?

yes but the branch was out of date so needed to be updated before automerge could work.

TonyXiang8787 (approved):

PR #318, opened 2023-07-26
link to boost::headers

mgovers:
Explicitly link to Boost::headers to prevent confusion about whether we use the header-only library of Boost or not.

Bumps CMake minimum required version to `3.15`

### DIFF
--- a/cmake/export_power_grid_model.cmake
+++ b/cmake/export_power_grid_model.cmake
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-cmake_minimum_required (VERSION 3.9)
+cmake_minimum_required (VERSION 3.15)
 
 # export the license and readme
 configure_file("${CMAKE_SOURCE_DIR}/cmake/README.md.in" "${CMAKE_CURRENT_BINARY_DIR}/README.md")
--- a/cmake/export_power_grid_model.cmake
+++ b/cmake/export_power_grid_model.cmake
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-cmake_minimum_required (VERSION 3.15)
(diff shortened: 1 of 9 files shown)


cmake/export_power_grid_model.cmake
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-cmake_minimum_required (VERSION 3.9)
+cmake_minimum_required (VERSION 3.15)
```
TonyXiang8787, line 5: Do you also need 3.15 for export target? I think for the users of C-API, we don't need this cmake version right?

cmake/pgm_version.cmake
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-cmake_minimum_required (VERSION 3.9)
+cmake_minimum_required (VERSION 3.15)
```
TonyXiang8787, line 5: Same here

tests/package_tests/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model project <dynamic.grid.calculation@alliander.com>
 #
 # SPDX-License-Identifier: MPL-2.0
-cmake_minimum_required (VERSION 3.9)
+cmake_minimum_required (VERSION 3.15)
```
TonyXiang8787, line 4: Same here. We don't need 3.15 for this I guess.

cmake/export_power_grid_model.cmake
mgovers: oh you're right. good call

TonyXiang8787 (approved):

PR #321, opened 2023-07-27
fix validation missing attr

nitbharambe:
- Fixing validation tests:
  - Missing attributes added. Fixes sparse matrix error.
  - Added c factor
- Enable short circuit in C API
- More useful logging when validation tests are failing:
  ```
  : C:\Users\MartijnGovers\Documents\power-grid-model\tests\cpp_validation_tests\test_validation.cpp(312): ERROR: CHECK( match ) is NOT correct!
  62:   values: CHECK( false )
  62:   logged: batch: #56, Component: source #1, attribute: id: actual = 5 vs. expected = 4
  62: 
  62: C:\Users\MartijnGovers\Documents\power-grid-model\tests\cpp_validation_tests\test_validation.cpp(312): ERROR: CHECK( match ) is NOT correct!
  62:   values: CHECK( false )
  62:   logged: batch: #56, Component: source #0, attribute: i: actual = (0.017811, 18.817750, 18.800378) vs. expected = (955.524049, 918.497800, 933.921584)
  ```


TonyXiang8787:
@nitbharambe @mgovers according to address sanitizer we have buffer overflow. Maybe we need to investigate that.

mgovers:
Whoops that sounds like a good starting point. Only on this branch or on the release branch as well?

Op 28 jul. 2023 22:34 schreef Tony Xiang ***@***.***>:

@nitbharambe<https://github.com/nitbharambe> @mgovers<https://github.com/mgovers> according to address sanitizer we have buffer overflow. Maybe we need to investigate that.

mgovers:
interestingly, according to the sanitizer, the buffer overflow happens in the line

```cpp
if (attr.check_nan(reference_result_ptr, obj)) {
```

and not in my new code or in the short circuit analyzer.

mgovers:
> interestingly, according to the sanitizer, the buffer overflow happens in the line
> 
> ```c++
> if (attr.check_nan(reference_result_ptr, obj)) {
> ```
> 
> and not in my new code, in the short circuit solver or even in the actual data in the first place.

i am able to repro locally on WSL with the same failing line, but the amount of other failing checks clutters this result. i'm not entirely sure how to debug before the rest of the validation checks pass

mgovers:
whoops, sorry, i pressed the wrong button.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -257,20 +327,36 @@ void assert_result(ConstDataset const& result, ConstDataset const& reference_res
                         break;
                     }
                 }
+                // for other _angle attribute, we need to find the magnitue and compare together
+                std::regex const angle_regex("(.*)(_angle)");
```
mgovers, line 331: I believe this should allow us to remove the TODO in https://github.com/PowerGridModel/power-grid-model/pull/321/files#diff-729d19abd466fed12f440ac99f0de1c9e94f8008414e7a161cac24b8779f409fR318-R321 , right?

but if I do that, there are 2 failing test cases:
* state_estimation/1os2msr-no-angle-sym-iterative_linear
* state_estimation/dummy-test-sym-sym-iterative_linear

we can also verify this in a follow-up

TonyXiang8787: No, that's a different problem. Adding this would not fix that problem.

TonyXiang8787:
> the transformer issue is out of scope of this PR. this is ready for review

what transformer issue?

nitbharambe:
> > the transformer issue is out of scope of this PR. this is ready for review
> 
> what transformer issue?

I cant validate the entire transformer output for some reason. Likely some issue with test generation.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -752,11 +747,13 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+    // TODO(mgovers): remove this functionality
     template <bool sym, typename Component, std::forward_iterator ResIt>
     ResIt output_result(std::vector<MathOutput<sym>> const& math_output, ResIt res_it) {
         return output_result<Component, MathOutput<sym>, ResIt>(math_output, res_it);
     }
 
+    // TODO(mgovers): remove this functionality
     template <bool sym>
     void output_result(std::vector<MathOutput<sym>> const& math_output, Dataset const& result_data, Idx pos = 0) {
         return output_result<MathOutput<sym>>(math_output, result_data, pos);
```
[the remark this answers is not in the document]
TonyXiang8787: let's do this later.

mgovers (approved):

PR #322, opened 2023-07-27
validation on sensors only when state estimation

mgovers:
when validation input/update data for power flow calculations, sensor data should be ignored


src/power_grid_model/validation/validation.py
```diff
@@ -349,13 +352,13 @@ def validate_values(data: SingleDataset) -> List[ValidationError]:  # pylint: di
         errors += validate_generic_load_gen(data, "asym_gen")
     if "shunt" in data:
         errors += validate_shunt(data)
-    if "sym_voltage_sensor" in data:
+    if "sym_voltage_sensor" in data and calculation_type != CalculationType.power_flow:
```
TonyXiang8787, line 355: calculation_type != CalculationType.state_estimation

TonyXiang8787 (approved):

PR #327, opened 2023-07-31
Feature/readability checks2

Laurynas-Jagutis:

This PR is stopped at the moment. 

Got 267 warnings treated as errors for the readability-identifier-length check (https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-length.html). It regards the lengths of variable and function names, specifically those who are too short. Min length should be at least 3 characters, as we can see we have at least 267 cases where we do not satisfy that requirement. This might be a point of discussion.


.clang-tidy
```diff
@@ -58,20 +58,17 @@ modernize-*,
...
 -readability-magic-numbers,
 -readability-redundant-access-specifiers,
 -readability-redundant-member-init,
 -readability-suspicious-call-argument
 '
 
+MinimumVariableNameLength = 0,
+MinimumParameterNameLength = 0,
+MinimumLoopCounterNameLength = 0,
+
```
mgovers, line 71: this is actually worse than keeping `-readability-identifier-length`. overriding the default options is mainly done when you agree with the check but disagree with its settings (e.g. you want to make the minimum 3 instead of 4, or 2 instead of 3, etc.), but not when you completely disagree with the check

alternatively, you could add a list of exceptions. the issue is that that list becomes quite long as well 😬

mgovers:
please try to keep the PR small. it's already becoming quite big with quite a lot of different checks happening simultaneously

PR #328, opened 2023-08-01
fix several readability issues

Laurynas-Jagutis:

Fixes: 
- readability-redundant-access-specifiers
- readability-redundant-member-init
- readability-suspicious-call-argument

Had to do some minor renaming to fix readability-suspicious-call-argument.


tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -363,12 +363,12 @@ void assert_result(ConstDataset const& result, ConstDataset const& reference_res
 }
 
 // root path
-#ifdef POWER_GRID_MODEL_VALIDATION_TEST_DATA_PATH
+#ifdef POWER_GRID_MODEL_VALIDATION_TEST_data_dir
```
mgovers, line 366: this is a breaking change as well as dangerous since this involves C-style Macros (things prefixed `#`). this definition is a hard preprocessor text replacement that may be defined either somewhere else in the project (other header files **included prior** to this file) or by the CMake project that defines the build targets of this project, e.g. the libraries and executables.

In this case, it is the latter, because `tests/cpp_validation_tests/CMakeLists.txt` contains the following lines:

```cmake
target_compile_definitions(power_grid_model_validation_tests
	PRIVATE POWER_GRID_MODEL_VALIDATION_TEST_DATA_PATH="${CMAKE_SOURCE_DIR}/tests/data"
)
```

which is equivalent to having the following lines at the top of every header file:

```cpp
#ifndef POWER_GRID_MODEL_VALIDATION_TEST_DATA_PATH
#define POWER_GRID_MODEL_VALIDATION_TEST_DATA_PATH="${CMAKE_SOURCE_DIR}/tests/data"`
#endif
```

this change breaks that, which means that the CMake target should be renamed as well.

In addition, there is a convention to have macros in all capitals across the codebase.

mgovers (approved):

PR #331, opened 2023-08-02
add scaling param in params.json

nitbharambe:
The scaling factor was missing in params

### DIFF
--- a/tests/data/short_circuit/single_phase_to_ground/params.json
+++ b/tests/data/short_circuit/single_phase_to_ground/params.json
@@ -1,5 +1,6 @@
 {
   "calculation_method": "iec60909",
+  "voltage_scaling_factor_c": 1.1,
   "rtol": 1e-08,
   "atol": 1e-08
 }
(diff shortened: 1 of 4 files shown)


mgovers:
We currently do not expose that parameter from the API so i'm not sure whether we want to add that value to this file

mgovers (changes requested):
Either also add it to the C and Python APIs or don't mention it to our users at all

TonyXiang8787:
I would propose to not to merge this PR. Until later we add the voltage scaling into the argument, we don't do it.

mgovers:
shall I close as won't fix?

mgovers:
The documentation includes the following lines:

> - To account for the different operational conditions, a voltage scaling factor of `c` is applied to the voltage source while running short circuit calculation function. By default, the value is considered as `1.1` in power-grid-model. 
> A detailed table for its selection is mentioned in IEC 60909.

should we update that with this design choice now or in a follow-up PR?

TonyXiang8787:
This one should be closed. Adding a factor in the `params.json` does not make sense at this moment.

PR #333, opened 2023-08-03
Reconfigure clang-format

mgovers:
* Switch to LLVM style for `clang-format`
* Bump `clang-format` to `15.0.7` (current default for `clang-format-15` on Ubuntu20.04)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -9,7 +9,6 @@ IndentWidth: 4
 
 # New settings
 
-AlwaysBreakTemplateDeclarations: Yes
-SpacesBeforeTrailingComments: 2
+AlwaysBreakTemplateDeclarations: MultiLine
 PackConstructorInitializers: NextLine
 PointerAlignment: Left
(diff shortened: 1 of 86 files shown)


mgovers:
i found a way to force VSCode to use a specific version. will update docs + deps

TonyXiang8787:
@mgovers  please also format the `main.c` file.

mgovers:
it is unfortunately not possible to fix the clang-format version in VSCode because we have checked-in the `.vscode/settings.json`. There is no command to get the current virtual environment directory, so we cannot set the `C_Cpp.Clang_format_path` setting to something like `${workspaceFolder}/*/clang-format`. it also does not support regex - only full paths - so visual studio has to be opened with all correct environment settings, but on windows that's tricky because both the MSVC native environment and the venv we use define `clang-format`

mgovers:
i believe not providing documentation at all is better than providing bad documentation

pyproject.toml
```diff
@@ -38,6 +38,8 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 dev = [
+    "black",
+    "clang-format==15.0.7",
```
TonyXiang8787, line 42: I would not put here. Let's install clang-format from `apt` or Visual Studio. Not via pip.

mgovers: ok, will remove again

TonyXiang8787 (approved):

Issue #320, opened 2023-07-27
[FEATURE] Implement PGM serialization


TonyXiang8787:
# Introduction

We have currently a serialization format in JSON to import and export PGM data. It is mainly for testing purpose. However, it is being increasingly used in production data. We have two problems:

1. The format does not store the meta information, which makes deserialization context sensitive.
2. The current Python implementation of serialization/deserialization is extremely slow.

# Proposal

It is proposed to define a new PGM serialization format, and to implement the serialization/deserialization in C++ efficiently.

## PGM serialization format

The defined PGM serialization format is both in JSON and msgpack. The implementation will focus on efficiency for msgpack. JSON files will be converted from/to msgpack. See below the format for single and batch data. The datasets are shown in JSON for readability.

#### Not a number

* `NaN` and `ID_NaN` shall be represented by `null`, in both `msgpack` and `json`.

#### Infinity

Infinities should be represented as is in `msgpack` as IEEE 754 format.

In `json`, there is no way to represent infinities as numbers. We use string instead:

* positive infinities shall be handled by `"+inf"` and `"inf"` (all must be supported)
* negative infinities shall be handled by `"-inf"` (all must be supported)
* for serialization, let's go with `"inf"` due to NumPy / C++ STL convention and also less data

### Single dataset

```json
{
  "version": "1.0",
  "type": "input",
  "is_batch": false,
  "attributes": {
    "node": [
      "id",
      "u_rated"
    ],
    "sym_load": [
      "id",
      "node",
      "status",
      "type",
      "p_specified",
      "q_specified"
    ],
    "source": [
      "id",
      "node",
      "status",
      "u_ref",
      "sk"
    ]
  },
  "data": {
    "node": [
      [
        1,
        10.5e3
      ],
      [
        2,
        10.5e3
      ],
      [
        3,
        10.5e3
      ],
      [
        100,
        "inf"
      ],
      [
        101,
        "+inf"
      ],
      [
        102,
        "-inf"
      ]
    ],
    "line": [
      {
        "id": 4,
        "from_node": 1,
        "to_node": 2,
        "from_status": 1,
        "to_status": 1,
        "r1": 0.11,
        "x1": 0.12,
        "c1": 4e-05,
        "tan1": 0.1,
        "i_n": 500.0
      },
      {
        "id": 5,
        "from_node": 2,
        "to_node": 3,
        "from_status": 1,
        "to_status": 1,
        "r1": 0.15,
        "x1": 0.16,
[...]

PR #336, opened 2023-08-04
Implement PGM Serialization

TonyXiang8787:
Implement PGM deserializer, as in #320. 

Steps

- [x] C++ Deserializer
- [x] C++ Serializer #349 
- [x] C API #353 
- [x] Python API: https://github.com/PowerGridModel/power-grid-model/pull/367
- [x] Validation test #388 #382 
- [x] Handle infinity #387 
- [x] Use dataset logic for C-API #391 
- [x] Documentation: 
  - [x] Serialization Doc https://github.com/PowerGridModel/power-grid-model/pull/373
  - [x] Update example of test case notebook
  - [x] New example of serialization, modify test case notebook #393 
  - [x] Build guide (new dependencies)


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,377 @@
...
+    std::map<std::string, std::vector<MetaAttribute const*>> attributes_;
+    Idx batch_size_{}; // for single dataset, the batch size is one
+    std::vector<Buffer> buffers_;
+    // attributes to track the movement of the position
+    // for error report purpose
+    mutable std::string_view root_key_;
+    mutable std::string_view component_key_;
+    mutable std::string_view attribute_key_;
+    mutable Idx scenario_number_{-1};
+    mutable Idx element_number_{-1};
```
mgovers, line 121: i'm not really a fan of `mutable` keywords. they may result in unexpected behaviour when dealing with `const` functions. I'd rather provide some kind of error metadata target stream and log error messages there than doing it the other way around

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    void count_data() {
+        msgpack::object const& obj = get_value_from_root("data", is_batch_ ? msgpack::type::ARRAY : msgpack::type::MAP);
+        buffers_ = {};
+        // pointer to array (or single value) of msgpack objects to the data
+        std::span<msgpack::object const> batch_data;
+        if (is_batch_) {
+            batch_size_ = (Idx)obj.via.array.size;
```
mgovers, line 205: please use `static_cast`s where possible

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,377 @@
...
+    template <class T>
+    void parse_attribute_per_type(void* element_pointer, msgpack::object const& obj,
+                                  MetaAttribute const& attribute) const {
+        obj >> *reinterpret_cast<T*>(reinterpret_cast<char*>(element_pointer) + attribute.offset);
+    }
+};
+
+} // namespace power_grid_model::meta_data
+
+#endif
```
mgovers, line 389: newline
```suggestion
#endif

```

mgovers (commented):
not a real review but some quick remarks while i have my break from studying 😬

CMakeLists.txt
```diff
@@ -32,6 +32,8 @@ include(GNUInstallDirs)
 
 find_package(Eigen3 CONFIG REQUIRED)
 find_package(Boost REQUIRED)
+find_package(nlohmann_json CONFIG REQUIRED)
+find_package(msgpack-cxx REQUIRED)
```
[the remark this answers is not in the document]
TonyXiang8787: I am strongly against optional dependencies. As also described by the `vcpkg` maintainer. 
https://youtu.be/sBP17HQAQjk?t=1074

Making a separate library could be possible. But I still doubt what would be the added value here. Both `nlohmann_json` and `msgpack-cxx` are header only, it requires minimal effort for users who are building PGM from source. For Python users and C-API pre-built binary users, this has no extra dependencies.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: These member are the error metadata target stream.

The error metadata target stream has to live somewhere, either as mutable class members, or passing around by all the functions. I prefer the first one. We can even remove the `mutable` keywords and make all the functions non-const. For deserializer class we don't actually need const function.

TonyXiang8787:
@mgovers please have a look at the C++ deserializer.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+    } // namespace adaptor
+} // MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS)
+} // namespace msgpack
+
+namespace power_grid_model::meta_data {
+
+constexpr struct from_msgpack_t {
+} from_msgpack;
+constexpr struct from_json_t {
+} from_json;
```
mgovers, line 67: Maybe more readable:

* split it up into two
* don't specify `from_` so that it can be re-used between the serializer and deserializer, so that we can use the same tag to do a full cycle test. in stead, specify that it's a `_tag`

```cpp
struct json_tag_t{};
struct msgpack_tag_t{};

constexpr auto json_tag = json_tag_t{};
constexpr auto msgpack_tag = msgpack_tag_t{};
```

in addition: i'm surprised that the formatter doesn't make a single line statement of this

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: we can maybe make the best of both worlds by adding a private `struct Cache` that we can pass around.

then we only need to pass around a single object with multiple attributes instead of a separate object for every attribute that bloats the argument list. That way we can keep the functions `const` but still make very clear to the users that there's a cache that captures the latest state

otherwise, i'm in favour of removing the `const` specifier.

mgovers (commented):
first thoughts. will continue the detailed review of the deserializer immediately but wanted to split between high-level and detailed

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+
+    bool check_uniform(IdxVector const& counter) {
+        if (batch_size_ < 2) {
+            return true;
+        }
+        return std::transform_reduce(counter.cbegin(), counter.cend() - 1, counter.cbegin() + 1, true,
+                                     std::logical_and{}, std::equal_to{});
+    }
+
+    Idx get_elements_per_scenario(IdxVector const& counter) {
```
mgovers, line 311: since you're hard-requiring that the data `is_uniform`, maybe you can rename this to `get_uniform_elements_per_scenario`. i thought I understood the code but when I read the comment above that refers to `is_uniform`, I got confused and had to look here to understand what was happening. the comment actually made it more confusing instead of less

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    void parse_component(Buffer const& buffer) const {
+        component_key_ = buffer.component->name;
+        // handle indptr
+        if (buffer.elements_per_scenario < 0) {
+            // first always zero
+            buffer.indptr.front() = 0;
+            // accumulate sum
+            // TODO Apple Clang cannot compile transform_inclusive_scan correctly
```
mgovers, line 329: I thought we use this in the `container.hpp` as well? how do we solve it there? or is it because we need `transform_inclusive_scan` here instead of just regular `inclusive_scan`?

```suggestion
            // TODO(TonyXiang8787) Apple Clang cannot compile transform_inclusive_scan correctly
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+                                            : scenario_number_ * buffer.elements_per_scenario;
+#ifndef NDEBUG
+            if (buffer.elements_per_scenario < 0) {
+                assert(buffer.indptr[scenario_number_ + 1] - buffer.indptr[scenario_number_] ==
+                       static_cast<Idx>(buffer.msg_data[scenario_number_].size()));
+
+            } else {
+                assert(buffer.elements_per_scenario == static_cast<Idx>(buffer.msg_data[scenario_number_].size()));
+            }
+#endif
```
mgovers, line 361: you don't need the `#ifndef NDEBUG`. `assert` already implies `#ifndef NDEBUG` and the compiler **should** optimize the empty `if` and `else` blocks away in `Release`/`RelWithDebInfo` mode, and even if it doesn't, a single boolean check is trivial compared to all the other logic happening

```suggestion
            if (buffer.elements_per_scenario < 0) {
                assert(buffer.indptr[scenario_number_ + 1] - buffer.indptr[scenario_number_] ==
                       static_cast<Idx>(buffer.msg_data[scenario_number_].size()));

            } else {
                assert(buffer.elements_per_scenario == static_cast<Idx>(buffer.msg_data[scenario_number_].size()));
            }
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+            void* scenario_pointer = reinterpret_cast<char*>(buffer.data) + scenario_offset * buffer.component->size;
```
mgovers, line 362: can you move this line to another function? i would prefer to isolate `reinterpret_cast`s as much as possible

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+            parse_scenario(*buffer.component, scenario_pointer, buffer.msg_data[scenario_number_], attributes);
+        }
+        scenario_number_ = -1;
+        component_key_ = "";
+    }
+
+    void parse_scenario(MetaComponent const& component, void* scenario_pointer, ArraySpan msg_data,
+                        std::span<MetaAttribute const* const> attributes) const {
+        for (element_number_ = 0; element_number_ != static_cast<Idx>(msg_data.size()); ++element_number_) {
+            void* element_pointer = reinterpret_cast<char*>(scenario_pointer) + element_number_ * component.size;
```
mgovers, line 372: can you move this line to another function? i would prefer to isolate `reinterpret_cast`s as much as possible

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+            return parse_attribute_per_type<int32_t>(element_pointer, obj, attribute);
+        default:
+            throw SerializationError{"Unknown data type for attriute!\n"};
+        }
+    }
+
+    template <class T>
+    void parse_attribute_per_type(void* element_pointer, msgpack::object const& obj,
+                                  MetaAttribute const& attribute) const {
+        obj >> *reinterpret_cast<T*>(reinterpret_cast<char*>(element_pointer) + attribute.offset);
```
mgovers, line 434: if we re-write this as `*reinterpret_cast<T*>(reinterpret_cast<char*>(element_pointer) + attribute.offset) = ...` (either by creating a wrapper function or by implementing the `operator>>` in the `MetaAttribute`), can we get rid of the double `reinterpret_cast`?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+
+    void handle_error(std::exception& e) const {
+        std::stringstream ss;
+        ss << e.what();
+        if (!root_key_.empty()) {
+            ss << "Position of error: " << root_key_;
+            root_key_ = "";
+        }
+        if (is_batch_ && scenario_number_ >= 0) {
+            ss << "/" << scenario_number_;
```
mgovers, line 445: maybe explicitly mention that this is scenario `xyz`. Same below.

Alternatively instead of using `"/" << xyz`, you can use `"[" << xyz << "]"`. Users will understand that it represents a key/index (since it's already a JSON-like format anyway), and actually querying that key/index in e.g. python will yield the exact result the users need to find the issue

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+    // not movable
+    Deserializer(Deserializer&&) = delete;
+    Deserializer& operator=(Deserializer&&) = delete;
+    // destructor
+    ~Deserializer() = default;
+
+    Deserializer(from_json_t, std::string_view json_string)
+        : Deserializer{from_msgpack, json_to_msgpack(json_string)} {}
+
+    Deserializer(from_msgpack_t, std::span<char const> msgpack_data) {
```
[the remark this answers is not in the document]
TonyXiang8787: It makes more sense that a `string_view` represents a string, and a `span` represents a range of `char` which is not a string.

.clang-tidy
```diff
@@ -26,6 +26,7 @@ cppcoreguidelines-*,
 -cppcoreguidelines-pro-type-member-init,
 -cppcoreguidelines-pro-bounds-pointer-arithmetic,
 -cppcoreguidelines-pro-type-reinterpret-cast,
+-cppcoreguidelines-pro-type-union-access,
```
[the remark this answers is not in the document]
TonyXiang8787: To access `msgpack` object per type we need to use the `via` attribute which is a union. `clang-tidy` does not like union. But we need union in this case.

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -0,0 +1,411 @@
...
+#include <power_grid_model/auxiliary/serialization/deserializer.hpp>
+#include <power_grid_model/auxiliary/update.hpp>
+
+#include <doctest/doctest.h>
+
+namespace power_grid_model::meta_data {
+
+// single data
+namespace {
+constexpr char const* json_single = R"(
```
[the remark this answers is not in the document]
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: 1. A tag type does not need to end with `_tag`. `std::piecewise_construct` does not have `_tag` at the end.
2. I think `from_json` is a good name. You don't need this for serializer as you can have actual named functions. For deserializer you need the tag type in the constructor.
3. I will define the struct and value separately.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+    Idx batch_size() const { return batch_size_; }
+
+    Idx n_components() const { return static_cast<Idx>(buffers_.size()); }
+
+    Buffer const& get_buffer_info(Idx i) const { return buffers_[i]; }
+
+    void set_buffer(char const** components, void** data, Idx** indptrs) {
+        for (Idx i = 0; i != n_components(); ++i) {
+            auto const found = std::find_if(buffers_.begin(), buffers_.end(), [components, i](Buffer const& buffer) {
+                return buffer.component->name == components[i];
```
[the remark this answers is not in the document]
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+        void* data;                      // set by user
+        std::span<Idx> indptr;           // set by user
+    };
+
+    // not copyable
+    Deserializer(Deserializer const&) = delete;
+    Deserializer& operator=(Deserializer const&) = delete;
+    // not movable
+    Deserializer(Deserializer&&) = delete;
+    Deserializer& operator=(Deserializer&&) = delete;
```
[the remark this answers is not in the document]
TonyXiang8787: Fair point, changed to default.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+        } catch (msgpack::type_error&) {
+            throw SerializationError{"Keys in the dictionary should always be a string!\n"};
+        }
+    }
+
+    void post_serialization() {
+        if (handle_.get().type != msgpack::type::MAP) {
+            throw SerializationError{"The root level object should be a dictionary!\n"};
+        }
+        get_value_from_root("version", msgpack::type::STR) >> version_;
```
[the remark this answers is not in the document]
TonyXiang8787: The whole serializer/deserializer would have code not very consistent with rest of the repository, just like `meta_data.hpp`. And the whole serializer/deserializer is about using `msgpack`. To make a separate wrapper header is for me difficult to reason.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+
+    static Idx find_key_from_map(msgpack::object const& map, std::string_view key) {
+        auto const kv_map = map.as<MapSpan>();
+        for (Idx i = 0; i != static_cast<Idx>(kv_map.size()); ++i) {
+            if (key == key_to_string(kv_map[i])) {
+                return i;
+            }
+        }
+        return -1;
+    }
```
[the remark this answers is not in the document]
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: In `container.hpp` we use `inclusive_scan`. That is fine with Apple Clang.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: I do not understand your point. @mgovers can you elaborate?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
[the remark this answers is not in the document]
mgovers: to me, the (de)serializer is independent of the format used - it might as well be XML or plain JSON. the fact that we choose msgpack is not a design decision but an implementation detail. in the same link you shared about the optional dependencies (i agree with you on that), they also comment on the dependencies are used (and how). usage of a function like `extract_attribute(serialization_type_tag, data, component)` could for instance be completely independent of msgpack, even though the function itself strongly serialization format-dependent.

in extension, I like the design of this class the way you have written it right now precisely because it's relatively easy to extract the msgpack implementation details away from this file

.clang-tidy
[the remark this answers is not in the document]
mgovers: fair point. maybe you can restrict this usage to just that line using `// NOLINT(cppcoreguidelines-pro-type-union-access)` (or `NOLINTNEXTLINE` if the line gets too long otherwise)? That way, it becomes very clear that there's only one occurrence and it has is intentional (e.g.: when refactoring that line, you can see something like "the latest version now _does_ support it so we can refactor that"

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: some error messagelike

```txt
an error occurred while parsing data["foo"][1]["bar"]
```

or

```txt
an error occurred while parsing data: component 'foo', scenario 1, attribute 'bar'
```

may be more readable to the user than

```txt
an error occurred while parsing data/foo/1/bar
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: in addition, if the user wants to debug, that person can open e.g. in python like

```py
import json

with open('foo.json', encoding='utf-8') as f:
    data = json.load(f)

print(data["foo"][1]["bar"])
```

and see what went wrong

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: I have added this in the attribute meta data.

The whold file is about using `msgpack`. I would keep using `operator>>` here as recommended in `msgpack`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: Ah, now I understand. I copy this design from `nlohmann_json`. They use `json_pointer` with `/` as separator:
https://json.nlohmann.me/features/json_pointer/#introduction

What do you think?

.clang-tidy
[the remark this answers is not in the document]
TonyXiang8787: Do you know how to ignore one line in clang-tidy?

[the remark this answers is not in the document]
mgovers: yes you can do either

```cpp
// NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
foo = bar;
```

or

```cpp
foo = bar;  // NOLINT(cppcoreguidelines-pro-type-union-access)
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: OHHH i get it now, you're right, they're using HTML-like query syntax: https://www.npmjs.com/package/json-pointer

TonyXiang8787: So shall we also keep like this?

.clang-tidy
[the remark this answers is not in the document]
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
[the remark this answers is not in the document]
TonyXiang8787: I understand your point. On the other hand, I don't want to go for pre-mature abstraction. So let's leave it as is and abstract `msgpack` away to separate headers if we indeed later start to support another serialization format.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+            } else {
+                span = {o.via.map.ptr, o.via.map.size};
+            }
+            return o;
+        }
+    };
+
+    } // namespace adaptor
+} // MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS)
+} // namespace msgpack
```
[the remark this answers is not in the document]
TonyXiang8787: Same here, let's leave it as is and abstract `msgpack` away to separate headers if we indeed later start to support another serialization format.

CMakeLists.txt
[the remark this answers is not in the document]
TonyXiang8787: Same here, let's leave it as is and abstract `msgpack` away to separate headers if we indeed later start to support another serialization format.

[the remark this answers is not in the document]
TonyXiang8787: And to be practical, have a C++ API users is very unlikely. We do not in any means guarantee API stability of C++ core.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: yes

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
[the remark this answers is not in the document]
mgovers: ok for now, but i did want to start the discussion (for background: i am speaking from experience here. i've had to try to remove `opencv-3.4` from the codebase to upgrade to `opencv-4.x` and even just that basically required reimplementing everything even though only core functionality was used)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,471 @@
...
+        auto const kv_map = map.as<MapSpan>();
+        for (Idx i = 0; i != static_cast<Idx>(kv_map.size()); ++i) {
+            if (key == key_to_string(kv_map[i])) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    void read_predefined_attributes() {
```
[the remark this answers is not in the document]
TonyXiang8787: I am quite sure what do you mean here. Please make a suggestion.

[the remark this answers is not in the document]
mgovers:
```cpp
    void read_predefined_attributes() {
        msgpack::object const& attribute_map = get_value_from_root("attributes", msgpack::type::MAP);
        for (auto const& kv : attribute_map.as<MapSpan>()) {
            component_key_ = key_to_string(kv);
            MetaComponent const& component = dataset_->get_component(component_key_);
            read_component_attributes(component, kv.val);
        }
        component_key_ = "";
    }

    void read_component_attributes(MetaComponent const& component, msgpack::object const& attribute_list) {
        if (attribute_list.type != msgpack::type::ARRAY) {
            throw SerializationError{
                "Each entry of attribute dictionary should be a list for the corresponding component!\n"};
        }

        auto const list_span = attribute_list.as<ArraySpan>();

        auto& attributes = attributes_[component.name];
        for (element_number_ = 0; element_number_ != static_cast<Idx>(list_span.size()); ++element_number_) {
            attributes.push_back(
                &component.get_attribute(list_span[element_number_].as<std::string_view>()));
        }
        element_number_ = -1;
    }
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: See
https://github.com/PowerGridModel/power-grid-model/actions/runs/5785220968/job/15677448351?pr=336

Clang-Tidy gives warning in Release build.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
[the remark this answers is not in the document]
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: `const` functions removed, and `mutable` removed.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -234,13 +248,13 @@ struct MetaData {
...
 
-    MetaDataset const& get_dataset(std::string const& dataset_name) const {
+    MetaDataset const& get_dataset(std::string_view dataset_name) const {
         for (auto const& dataset : datasets) {
             if (dataset.name == dataset_name) {
                 return dataset;
             }
         }
-        throw std::out_of_range{"Cannot find dataset with name: " + dataset_name + "!\n"};
+        throw std::out_of_range{"Cannot find dataset with name: " + std::string(dataset_name) + "!\n"};
```
mgovers, line 257:
```suggestion
        throw std::out_of_range{"Cannot find dataset with name: " + std::string{dataset_name} + "!\n"};
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,485 @@
...
+    Buffer const& get_buffer_info(Idx i) const { return buffers_[i]; }
+
+    void set_buffer(char const** components, void** data, Idx** indptrs) {
+        for (Idx i = 0; i != n_components(); ++i) {
+            auto const found =
+                std::find_if(buffers_.begin(), buffers_.end(), [component = components[i]](Buffer const& buffer) {
+                    return buffer.component->name == component;
+                });
+            if (found == buffers_.end()) {
+                throw SerializationError{"Unkown component: " + std::string(components[i]) +
```
mgovers, line 128:
```suggestion
                throw SerializationError{"Unkown component: " + std::string{components[i]} +
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -0,0 +1,485 @@
...
+            msgpack::object const& scenario = batch_data[scenario_number_];
+            Idx const found_component_idx = find_key_from_map(scenario, component.name);
+            if (found_component_idx >= 0) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+                msgpack::object const& element_array = scenario.via.map.ptr[found_component_idx].val;
+                if (element_array.type != msgpack::type::ARRAY) {
+                    throw SerializationError{"Each entry of component per scenario should be a list!\n"};
+                }
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+                counter[scenario_number_] = static_cast<Idx>(element_array.via.array.size);
```
mgovers, line 299: optional: move the accessors to separate functions so that the `NOLINT` is only mentioned in standalone locations, e.g. a function that the equivalent of

```cpp
static auto as_array(msgpack::object& element_array) { return element_array.via.array; }
static auto as_map(msgpack::object& element_array) { return element_array.via.map; }
```

or use full functions to get the size and component value

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace power_grid_model::meta_data {
+
+class Serializer {
+  public:
+};
+
+} // namespace power_grid_model::meta_data
+
+#endif
```
mgovers, line 23:
```suggestion
#endif

```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -0,0 +1,411 @@
...
+            CHECK(is_nan(asym_load[2].p_specified));
+            CHECK(asym_load[2].q_specified(0) == doctest::Approx(70.0));
+            CHECK(asym_load[2].q_specified(1) == doctest::Approx(80.0));
+            CHECK(asym_load[2].q_specified(2) == doctest::Approx(90.0));
+        }
+    }
+}
+
+TEST_CASE("Deserializer with error") {
+
```
mgovers, line 361:
```suggestion
```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -0,0 +1,411 @@
...
+        deserializer.parse();
+    };
+
+    CHECK_THROWS_WITH_AS(run(), doctest::Contains(err_msg), SerializationError);
+}
+
+} // namespace
+
+TEST_CASE("Deserializer") {
+
```
mgovers, line 233:
```suggestion
```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -0,0 +1,411 @@
...
+      ]
+    ]
+  }
+}
+)";
+
+} // namespace
+
+// batch data
+namespace {
```
mgovers, line 131:
```suggestion
// batch data
```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -0,0 +1,411 @@
...
+      ]
+    }
+  ]
+}
+)";
+
+} // namespace
+
+namespace {
+
```
mgovers, line 206:
```suggestion
```

mgovers (commented):
only some nitpicks

mgovers:
please also have a quick look at the code smells. I'm fine with not fixing all of them (some are known issues) but please do double-check

mgovers (approved):

mgovers:
@TonyXiang8787 do you think it's a good idea to pull this back to `Draft` due to the decision to first complete all aspects?

src/power_grid_model/core/serialization.py
```diff
@@ -0,0 +1,319 @@
...
+        A tuple containing:
+            the deserialized dataset in Power grid model input format.
+            the type of the dataset.
+    """
+    result = JsonDeserializer(data).load()
+    assert_no_error()
+    return result
+
+
+def json_serialize(
```
mgovers, line 242: TBD here and for `msgpack` and the `*_to_file` serialization functionality:

> > > > @TonyXiang8787 @petersalemink95 open question on forward-compatibility:
> > > do you think it's a good idea to add an optional `version: str = "1.0"` keyword argument to the new `json_serialize_*` and `msgpack_serialize_*` interfaces? that way, users can decide whether they want to hard fix their format version if they themselves use a very specific output query or that they allow the PGM to serialize using the most recent version
> > 
> > 
> > I don't think that's needed. In a given version of PGM, we only serialize to the latest supported format.
> 
> A valid remark that @petersalemink95 had is that some people may use the serialized data as their temporary format and that they could load into another application/pipeline step. in that case, people may already be using specific queries.
> 
> If ever, I do believe that now is a good time to add such a feature (probably it's better to use an enum now that i think about it), because we would want to bump the default value when we want to. It also would enable us to develop on `main` instead of on a feature branch like we're doing now to prevent conflicts.
> 
> let's discuss on Monday. I do believe it to be out of scope of this PR, though

TonyXiang8787: @mgovers after discussion with @petersalemink95 we have decided not to support this. This would mean maintaining two or multiple versions of serializer in the future which is not feasible.

What we need to guarantee is the forward/backward compatibility of serialization format. If we release version `1.1` of serialization format, the new `json/msgpack` file should be readable with the old PGM serializer, only with the new features missing.

petersalemink95 (approved):

Issue #113, opened 2022-08-02
Resolve code smells from Sonar Cloud in C++ side


TonyXiang8787:
As suggested in Sonar Cloud

https://sonarcloud.io/project/overview?id=alliander-opensource_power-grid-model

- Solve the code smell 
- Ignore it in the `sonar-project.properties` if that's the intended choice. Document in this file about the design choice.

PR #338, opened 2023-08-08
Improvement/code sonar fixes

Laurynas-Jagutis:
Related to https://github.com/PowerGridModel/power-grid-model/issues/113

Fixes:
- Some "Use the transparent comparator "std::less<>" with this associative string container" code smells. There are a couple of these code smells left in Dataset.hpp
- All "Use std::format instead of concatenating pieces manually." code smells

Also had some "Replace the redundant type with "auto"" fixes, but as of now reverted them

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -100,8 +100,8 @@ template <bool is_const> class DataPointer {
 using MutableDataPointer = DataPointer<false>;
 using ConstDataPointer = DataPointer<true>;
 
-using Dataset = std::map<std::string, MutableDataPointer, std::less<std::string>>;
-using ConstDataset = std::map<std::string, ConstDataPointer, std::less<std::string>>;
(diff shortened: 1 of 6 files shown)


mgovers:
> All "Use std::format instead of concatenating pieces manually." code smells

we can't do this one yet because the versions of GCC, Clang and XCode (Apple Clang) we have do not have it implemented yet. you can find it if you look for text formatting ([P0645R10](https://wg21.link/P0645R10)) in https://en.cppreference.com/w/cpp/compiler_support/20

mgovers:
@Laurynas-Jagutis is this ready to merge? if so, please mark it ready for review

Laurynas-Jagutis:
> @Laurynas-Jagutis is this ready to merge? if so, please mark it ready for review

I usually wait till the github checks pass to do that. Is that bad practice?

mgovers:
> > @Laurynas-Jagutis is this ready to merge? if so, please mark it ready for review
> 
> I usually wait till the github checks pass to do that. Is that bad practice?

nope not at all 😄 just was busy reviewing, stumbled upon this and thought it was ready

mgovers (approved):

PR #344, opened 2023-08-08
Code smell for "auto" use fix

Laurynas-Jagutis:
Related to https://github.com/PowerGridModel/power-grid-model/issues/113

Fixes "Replace the redundant type with "auto""


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -48,7 +48,7 @@ template <bool is_const> class DataPointer {
 
     template <class T> std::pair<ptr_t<T>, ptr_t<T>> get_iterators(Idx pos) const {
         assert(pos < batch_size_);
-        ptr_t<T> const ptr = reinterpret_cast<ptr_t<T>>(ptr_);
+        auto* const ptr = reinterpret_cast<ptr_t<T>>(ptr_);
```
mgovers, line 51: did the code smell explicitly say that you should use `auto*` here? (if so, that's great but i would expect `auto` to be sufficient)

mgovers (commented):
since you're already touching all these lines anyway, can you maybe replace all C-style casts `(type)foo` by `static_cast`s `static_cast<type>(foo)` here as well?

tests/cpp_unit_tests/test_three_phase_tensor.cpp
```diff
@@ -85,7 +85,7 @@ TEST_CASE("Three phase tensor") {
         CHECK((mat4 == mat).all());
         CHECK(diag_mult(1.0, 2.0, 3.0) == 6.0);
         // test layout
-        double* arr = (double*)&mat1;
+        auto* arr = (double*)&mat1;
```
mgovers, line 88: it is possible that it wants you to use `reinterpret_cast` or `static_ptr_cast` here, not sure. if it starts complaining, just ignore it for now

tests/cpp_unit_tests/test_three_phase_tensor.cpp
```diff
@@ -85,7 +85,7 @@ TEST_CASE("Three phase tensor") {
         CHECK((mat4 == mat).all());
         CHECK(diag_mult(1.0, 2.0, 3.0) == 6.0);
         // test layout
-        double* arr = (double*)&mat1;
+        auto* arr = reinterpret_cast<double*>(&mat1);
```
mgovers, line 88: so `static_cast` wasn't working here?

Laurynas-Jagutis: It was giving "invalid type conversion" error

mgovers: please have a look at https://stackoverflow.com/questions/14603546/invalid-type-conversion-using-static-cast-what-proper-casting-should-i-use .

you can try and see if you can write it like

```cpp
        // double* arr = (double*)&mat1;  // remove this line
        CHECK(static_cast<double>(mat[0]) == 1.0);
```

if it doesn't work, please revert this change and we'll need to revisit this later because I honestly don't know why we're doing it this way.

in particular, this is exactly the reason why we should use `static_cast` where possible and `reinterpret_cast` only where absolutely necessary (here, it is not absolutely necessary but it hides the fact that we do this because of the C-style cast).

mgovers (approved):

PR #345, opened 2023-08-09
Feature/reduce clang tidy build

mgovers:
Not fond of it but this is the only way to improve the compile time for clang tidy without matrix testing for build target

* only build the C API target for clang tidy by default
* add manual trigger option to override the target

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -62,7 +62,7 @@ jobs:
           brew install boost eigen nlohmann-json doctest
 
       - name: Set build target in case of push
-        if: github.event_name == 'push'
+        if: github.event_name != 'workflow_dispatch'
         run: |
           echo "TARGET=power_grid_model_c" >> $GITHUB_ENV
 
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -17,8 +17,12 @@ on:
   workflow_dispatch:
     inputs:
       target:
-        type: string
-        default: power_grid_model_c  # only build the C API by default
+        type: choice
+        description: The CMake target to run
+        default: power_grid_model_c
+        options:
+          - all
+          - power_grid_model_c
         required: true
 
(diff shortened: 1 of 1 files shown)


TonyXiang8787 (dismissed):
Nice! Now it is down to 6 minutes.

mgovers:
c.c. @Laurynas-Jagutis I added a `workflow-dispatch` trigger in which you can override the build target to `all`.

Unfortunately, I cannot test this until it's in `main` because of how github actions work

.github/workflows/clang-tidy.yml
```diff
@@ -32,18 +43,30 @@ jobs:
...
+
+      - name: Override build target in case of workflow dispatch
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          echo "TARGET=${{ github.event.inputs.target }}"
+
+      - name: Configure
+        run: |
+          cmake --preset ${{ env.PRESET }}
+          cmake --build --preset ${{ env.PRESET }} --target ${{ env.TARGET }}
```
TonyXiang8787, line 72: I do not understand how you override the default target. In your current way, `${{ env.TARGET }}` is always `power_grid_model_c`

mgovers: unfortunately, I can only properly test that this works correctly after this is merged to main because github actions only uses the default branch to check for workflow dispatch triggers

TonyXiang8787: I doubt this will work as intended. Using the syntax you have.

You can test it in another way, Just do echo in one step and print the result in another step.

mgovers: ohhh i see i forgot the `>> $GITHUB_ENV`.

i will do a couple more tests

mgovers: ok it definitely works now. thanks

TonyXiang8787 (approved):

TonyXiang8787 (approved):

PR #347, opened 2023-08-10
Feature/updateable shunt params

### LINKED ISSUE #298: [FEATURE] Allow shunt's values to be updated by batch power-flow calculation.
The batch run feature allows user to update certain network parameters when running power-flow calculations multiple times. At this moment, the shunt values (R and X) are not possible to be updated. I would like to request this feature. It is quite useful for various network analysis. For example, tuning the relationship between the line/cable capacitance and the reactive power compensation level to find the optimal operational settings. Thanks.

mgovers:
* make shunt parameters updateable
* fix bug in python utils where the construction of the update data set from json would fail if the first entry is empty
  * note: this would be OK if the second or later entry is empty

fixes #298


src/power_grid_model/utils.py
```diff
@@ -88,10 +88,10 @@ def convert_list_to_batch_data(list_data: BatchList) -> BatchDataset:
...
                 indptr.append(indptr[-1])
 
-            # Convert the index pointers to a numpy array and combine the list of object numpy arrays into a singe
-            # numpy array. All objects of all batches are now stores in one large array, the index pointers define
-            # which elemets of the array (rows) belong to which batch.
-            batch_data[component] = {"indptr": np.array(indptr, dtype=np.int64), "data": np.concatenate(data, axis=0)}
+        # Convert the index pointers to a numpy array and combine the list of object numpy arrays into a singe
+        # numpy array. All objects of all batches are now stores in one large array, the index pointers define
+        # which elemets of the array (rows) belong to which batch.
+        batch_data[component] = {"indptr": np.array(indptr, dtype=np.int64), "data": np.concatenate(data, axis=0)}
```
mgovers, line 94: This is the python utils fix

cause: `np.concatenate` fails when `data` is empty, so the call would fail if the first entry is empty

open question TBD: what to do when all data entries are empty. current implementation will fail.

Possible alternative implementation that **does** allow all entries to be empty: ignore component

```suggestion
        if data:
            batch_data[component] = {"indptr": np.array(indptr, dtype=np.int64), "data": np.concatenate(data, axis=0)}
```

tests/data/power_flow/dummy-test-batch-shunt/asym_output_batch.json.license
mgovers, line 1: the original `sym_output.json` file did not exist anymore, so that `license` file was redundant.

nitbharambe (approved):

src/power_grid_model/utils.py
nitbharambe: I think ignoring component would be nice. Whats your preference?

mgovers: whoops sorry, already clicked merge. let's discuss this during the next knowledge session. the change is very small anyway

PR #348, opened 2023-08-10
remove topology construction from main model

mgovers:
This is prep work for the discussions next Tuesday. It continues the refactoring of moving detailed core functionality from the main model. No deeper architectural design choices are made


mgovers:
a couple more minor improvements can still be made

mgovers:
code smell is caused by same issue in `main`

PR #349, opened 2023-08-10
Implement C++ Serializer

TonyXiang8787:


TonyXiang8787:
@mgovers the SonarCloud bug seems to be a false positive.

mgovers:
> @mgovers the SonarCloud bug seems to be a false positive.

is there a way you can rewrite it in terms of vectors, iterators, `span`s or any other way so that you can prove that the size of `elements_per_scenario` is really large enough?

with all the validation checks and C API tests that we do, it would surprise me that we now suddenly have a new memory access error in new code

what I can think of is that it flags this as an issue:

```cpp
        Idx const n_elements = 3;
        std::array data{(void const*)asym_load_gen.data()};
        Serializer serializer{"update", false, 1, n_components, components.data(), &n_elements, nullptr, data.data()};
```

`n_elements` is a value, not an array. if it flags because you use array access `[0]` to access its value there doesn't surprise me actually.

can you try to check this by making n_elements  an array like `std::vector<Idx> const n_elements{3}` and then passing `n_elements.data()` and see if that solves the issue

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -6,16 +6,306 @@
...
+          batch_size_{batch_size},
+          n_components_{n_components},
+          packer_{msgpack_buffer_} {
+        if (!is_batch_ && (batch_size_ != 1)) {
+            throw SerializationError{"For non-batch dataset, batch size should be one!\n"};
+        }
+        store_buffers(components, elements_per_scenario, indptrs, data);
+    }
+
+    void serialize(bool use_compact_list) {
```
mgovers, line 99: as a user, i would expect `serialize` to return a value (like in C# or JavaScript). Since the class is already named `Serializer`, I expect this function to be called in the constructor, or otherwise in the `get`ters. maybe this function should be private?

i understand that you want to give the user the option to `use_compact_list` but in that case, i guess the `get_msgpack` and `get_json` functions should expose that option

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -6,16 +6,306 @@
...
+    }
+
+    void pack_root_dict() {
+        packer_.pack_map(size_top_dict);
+        packer_.pack("version");
+        packer_.pack(version);
+        packer_.pack("type");
+        packer_.pack(dataset_->name);
+        packer_.pack("is_batch");
+        packer_.pack(is_batch_);
```
mgovers, line 198: maybe add enters to group by topic. not really necessary tho

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -6,16 +6,306 @@
...
+        attributes_ = {};
+        for (auto const& buffer : component_buffers_) {
+            std::vector<MetaAttribute const*> attributes;
+            for (auto const& attribute : buffer.component->attributes) {
+                // if not all the values of an attribute are nan
+                // add this attribute to the list
+                if (!attribute.check_all_nan(buffer.data, buffer.size)) {
+                    attributes.push_back(&attribute);
+                }
+            }
```
mgovers, line 186: do you think we should also provide a sparse packer? (like with how we used to use the dependent datasets)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -6,16 +6,306 @@
...
+                component_buffer.size = indptrs[component][end_scenario] - indptrs[component][begin_scenario];
+            } else {
+                component_buffer.data = component_buffer.component->advance_ptr(
+                    data[component], elements_per_scenario[component] * begin_scenario);
+                component_buffer.size = elements_per_scenario[component] * (end_scenario - begin_scenario);
+            }
+            // only store the view if it is non-empty
+            if (component_buffer.size > 0) {
+                scenario_buffer.component_buffers.push_back(component_buffer);
+            }
```
mgovers, line 171: theoretically, an empty scenario could exist as well, right? (although it wouldn't make sense to add more than one to an update dataset)

mgovers (commented):
very readable. will approve when the sonar cloud issue is investigated more thoroughly

mgovers:
> > @mgovers the SonarCloud bug seems to be a false positive.
> 
> is there a way you can rewrite it in terms of vectors, iterators, `span`s or any other way so that you can prove that the size of `elements_per_scenario` is really large enough?
> 
> with all the validation checks and C API tests that we do, it would surprise me that we now suddenly have a new memory access error in new code
> 
> what I can think of is that it flags this as an issue:
> 
> ```c++
>         Idx const n_elements = 3;
>         std::array data{(void const*)asym_load_gen.data()};
>         Serializer serializer{"update", false, 1, n_components, components.data(), &n_elements, nullptr, data.data()};
> ```
> 
> `n_elements` is a value, not an array. if it flags because you use array access `[0]` to access its value there that wouldn't surprise me actually.
> 
> can you try to check this by making n_elements an array like `std::vector<Idx> const n_elements{3}` and then passing `n_elements.data()` and see if that solves the issue?

this fixed the bug. i propose keeping it like this, **OR** making a separate constructor that ingests only a single integer instead of a C-style array

tests/cpp_unit_tests/test_serializer.cpp
```diff
@@ -6,8 +6,61 @@
...
+    asym_load_gen[1].id = 6;
+    asym_load_gen[2].id = 7;
+    asym_load_gen[0].p_specified = {10.0, 11.0, 12.0};
+    asym_load_gen[1].p_specified = {15.0, nan, 16.0};
+    // nan for asym_load_gen[2].p_specified
+
+    SUBCASE("Single dataset") {
+        std::array components{"asym_load"};
+        Idx const n_components = 1;
+        std::vector<Idx> const n_elements{3};
```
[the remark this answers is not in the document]
TonyXiang8787: Good

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
TonyXiang8787: Empty scenario will be correctly serialized. This is about empty component. An empty component should not exist in the dict of one scenario. If there are no components in a scenario, the scenario self should exist as an empty dict.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
TonyXiang8787: What do you mean exactly? Sparse regarding different number of elements per scenario? That we already support.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -6,16 +6,306 @@
...
+        for (auto const& component_buffer : scenario_buffer.component_buffers) {
+            pack_component(component_buffer);
+        }
+    }
+
+    void pack_component(ComponentBuffer const& component_buffer) {
+        packer_.pack(component_buffer.component);
+        packer_.pack_array(static_cast<uint32_t>(component_buffer.size));
+        bool const use_compact_list = use_compact_list_;
+        auto const attributes = [&]() -> std::span<MetaAttribute const* const> {
```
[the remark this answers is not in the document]
TonyXiang8787: It is not optional if you have trailing return type.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
TonyXiang8787: done

tests/cpp_unit_tests/test_serializer.cpp
```diff
@@ -6,8 +6,60 @@
...
+    asym_load_gen[2].id = 7;
+    asym_load_gen[0].p_specified = {10.0, 11.0, 12.0};
+    asym_load_gen[1].p_specified = {15.0, nan, 16.0};
+    // nan for asym_load_gen[2].p_specified
+
+    SUBCASE("Single dataset") {
+        std::array components{"asym_load"};
+        Idx const n_components = 1;
+        Idx const n_elements = 3;
+        std::array data{(void const*)asym_load_gen.data()};
```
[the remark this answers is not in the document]
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -6,16 +6,306 @@
...
+        case CType::c_int32:
+            return check_nan_impl<int32_t>(element_ptr, attribute);
+        default:
+            throw SerializationError{"Unknown data type for attriute!\n"};
+        }
+    }
+
+    template <class T> static bool check_nan_impl(void const* element_ptr, MetaAttribute const& attribute) {
+        return is_nan(attribute.get_attribute<T const>(element_ptr));
+    }
```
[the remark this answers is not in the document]
TonyXiang8787: I have implemented a selector to instantiate functor template based on `CType`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
mgovers: oh sorry i misunderstood. that was indeed what i meant

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -51,6 +51,24 @@ template <class T>
...
+    case CType::c_double3:
+        return f.template operator()<RealValue<false>>(std::forward<Args>(args)...);
+    case CType::c_int8:
+        return f.template operator()<int8_t>(std::forward<Args>(args)...);
+    case CType::c_int32:
+        return f.template operator()<int32_t>(std::forward<Args>(args)...);
+    default:
+        throw MissingCaseForEnumError{"CType selector", ctype};
+    }
+}
```
mgovers, line 70: nice solution!

I know we already use this kind of `switch` logic more in the code. Do we want to change that now as well?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -51,6 +51,24 @@ template <class T>
 struct ctype_t<T> : ctype_t<std::underlying_type_t<T>> {};
 template <class T> constexpr CType ctype_v = ctype_t<T>::value;
 
+// function selector based on ctype
+// the operator() of the functor should have a single template parameter
+// the selector will instantiate the operator() with relevant type
+template <class Functor, class... Args> decltype(auto) ctype_func_selector(CType ctype, Functor f, Args&&... args) {
```
mgovers, line 57: should we add `requires` clauses to enforce this behaviour? I know `requires` clauses for functors are often more difficult than for 'regular' template parameters, so i'm also fine with keeping it as is

mgovers (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
TonyXiang8787: The other places are in the validation test. There is a separate ticket to refactor validatin test.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
TonyXiang8787: Let's keep it as is.

PR #358, opened 2023-08-22
Improvement/cl codesmells2

Laurynas-Jagutis:

Fixes:
- readability-implicit-bool-conversion

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
@@ -195,7 +195,9 @@ struct PGM_MetaComponent {
         return -1;
     }
 
-    Idx has_attribute(std::string const& attribute_name) const { return find_attribute(attribute_name) >= 0; }
+    Idx has_attribute(std::string const& attribute_name) const {
+        return static_cast<PGM_MetaComponent::Idx>(find_attribute(attribute_name) >= 0);
+    }
 };
 
 namespace power_grid_model::meta_data {
--- a/tests/benchmark_cpp/benchmark.cpp
+++ b/tests/benchmark_cpp/benchmark.cpp
@@ -36,18 +36,18 @@ struct PowerGridBenchmark {
         NodeInput const source_node{{id_source_node}, 150.0e3};
         NodeInput const cycle_node_1{{id_source_node + 1}, 150.0e3};
         NodeInput const cycle_node_2{{id_source_node + 2}, 150.0e3};
-        LinkInput const link_1{{{3}, 0, 1, true, true}};
-        LinkInput const link_2{{{4}, 1, 2, true, true}};
-        LinkInput const link_3{{{5}, 2, 0, true, true}};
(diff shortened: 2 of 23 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -195,7 +195,9 @@ struct PGM_MetaComponent {
         return -1;
     }
 
-    Idx has_attribute(std::string const& attribute_name) const { return find_attribute(attribute_name) >= 0; }
+    Idx has_attribute(std::string const& attribute_name) const {
+        return static_cast<PGM_MetaComponent::Idx>(find_attribute(attribute_name) >= 0);
```
mgovers, line 199:
```suggestion
        return static_cast<Idx>(find_attribute(attribute_name) >= 0);
```

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -140,4 +140,4 @@ int main(int /* argc */, char** /* argv */) {
     benchmarker.run_benchmark<false>(option, linear);
     // benchmarker.run_benchmark<false>(option, iterative_current);
     return 0;
-}
+}
```
mgovers, line 143:
```suggestion
}

```

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -60,7 +60,7 @@ struct Buffer {
 void parse_single_object(RawDataPtr ptr, json const& j, MetaComponent const& meta, Idx position) {
     for (auto const& it : j.items()) {
         // Allow and skip unknown attributes
-        if (!meta.has_attribute(it.key())) {
+        if (static_cast<int>(meta.has_attribute(it.key())) == 0) {
```
mgovers, line 63:
```suggestion
        if (static_cast<Idx>(meta.has_attribute(it.key())) == 0) {
```

in addition: why do you need to cast to `int`? isn't the data type already `Idx` (which is `int`-type)

mgovers (approved):

tests/cpp_validation_tests/test_validation.cpp
Laurynas-Jagutis: Yes, that's true, it was a suggestion by clang tidy, let's see if it accepts it your way

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -60,7 +60,7 @@ struct Buffer {
 void parse_single_object(RawDataPtr ptr, json const& j, MetaComponent const& meta, Idx position) {
     for (auto const& it : j.items()) {
         // Allow and skip unknown attributes
-        if (!meta.has_attribute(it.key())) {
+        if (static_cast<Idx>(meta.has_attribute(it.key())) == 0) {
```
mgovers, line 63: please also try to build locally with out the `static_cast`. i believe it should work just fine. if it doesn't then that's fine as well but please try it first

```suggestion
        if (meta.has_attribute(it.key()) == 0) {
```

mgovers (approved):

PR #360, opened 2023-08-22
infinite sigmas

mgovers:


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -410,23 +414,31 @@ template <bool sym> class MeasuredValues {
     template <bool only_magnitude = false>
     static SensorCalcParam<sym> combine_measurements(std::vector<SensorCalcParam<sym>> const& data, Idx begin,
                                                      Idx end) {
-        double accumulated_inverse_variance = 0.0;
+        double accumulated_inverse_variance{};
         ComplexValue<sym> accumulated_value{};
         for (Idx pos = begin; pos != end; ++pos) {
-            accumulated_inverse_variance += 1.0 / data[pos].variance;
+            auto const& measurement = data[pos];
+            auto const inv_variance = std::isfinite(measurement.variance) ? 1.0 / measurement.variance : 0.0;
```
mgovers, line 421: part 1 of the fix

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -410,23 +414,31 @@ template <bool sym> class MeasuredValues {
...
             } else {
                 // accumulate value
-                accumulated_value += data[pos].value / data[pos].variance;
+                accumulated_value += measurement.value * inv_variance;
             }
         }
+
+        if (!std::isnormal(accumulated_inverse_variance)) {
+            return SensorCalcParam<sym>{accumulated_value, std::numeric_limits<double>::infinity()};
+        }
```
mgovers, line 440: part 2 of the fix

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -301,75 +301,79 @@ template <bool sym> class MeasuredValues {
...
+        if (appliance_idx == disconnected) {
+            return;
+        }
+
+        auto const& appliance_measurement = extra_value_[appliance_idx];
+        if (std::isinf(appliance_measurement.variance)) {
+            ++n_unmeasured;
+            return;
+        }
```
mgovers, line 372: part 3 of the fix

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
TonyXiang8787: Do we need this? `1.0/inf` is well defined in IEEE 754.

mgovers: yes because it returns `std::nan`, not `0.0`

TonyXiang8787: Are you sure? As far as I know, if you have `x / inf`, when `x` is not `inf` or `nan`, the result is garanteed to be `0.0`.

TonyXiang8787: just try to compile and run the following in different compilers

```c++
// Example program
#include <iostream>
#include <limits>


int main()
{
    std::cout << 1.0 / std::numeric_limits<double>::infinity();
}
```

tests/data/state_estimation/single-line-load/input.json
```diff
@@ -62,7 +62,15 @@
       "measured_terminal_type": 4,
       "p_measured": 2e6,
       "q_measured": 0.0,
-      "power_sigma": 1e20
+      "power_sigma": 1e3
```
nitbharambe, line 65: What was the reason to modify the existing case too?
Was there an interaction with high sigma 1e20 and numerical limits sigma?

nitbharambe (approved):

tests/data/state_estimation/single-line-load/input.json
mgovers: as discussed yesterday: the original test didn't make sense - especially not for a test named `single-line-load`. we want to actually test the behaviour in edge cases (e.g. when two measurements are not in agreement, then we still have to handle it correctly - especially because "not in agreement" is also a statistically indetermined question without specifying confidence intervals which we don't have)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
TonyXiang8787: @mgovers is this merged without resolve this comment?

mgovers: sorry i forgot to comment (i did resolve the comment though; race condition???)

Nitish and I attempted it yesterday and it appeared to work as you said. Nitish applied the change in https://github.com/PowerGridModel/power-grid-model/pull/360/commits/90a16856e0d9a12a75cd8c3a59e6a5224c969597

PR #367, opened 2023-08-30
Feature/pgm serialization python

mgovers:
Implements the Python API for PGM (de-)serialization

* [x] Implement const dataset
* [x] Implement writable dataset
* [x] Implement deserializer
* [x] Implement serializer


tests/unit/test_dataset.py
mgovers, line 1: i still need to add unit tests for the `CWritableDataset` but it will be very difficult to verify that everything goes correctly, because the fact that it is read-only and that we cannot mock things within the C API causes the contents of the buffers to be untestable (only things like batch size are queryable)

.github/conda_pgm_env.yml
```diff
@@ -7,8 +7,10 @@ dependencies:
...
   - setuptools
   - boost-cpp
+  - nlohmann_json
+  - msgpack-cxx
   - eigen
   - numpy
   - pytest
   - pytest-cov
-  - msgpack-cxx
+  - msgpack
```
TonyXiang8787, line 16: `msgpack-cxx` is the correct name, you don't need `msgpack`

mgovers: i believe we need both because the `python` tests make use of `msgpack` as well

mgovers: i see what i did wrong. should be `msgpack-python` for the unit tests

mgovers:
@TonyXiang8787 this is ready afaiac

src/power_grid_model/utils.py
mgovers, line 1: there are quite a lot of functions here that are not documented in the Python API Reference (in fact, only the now-deprecated `import_json_data` and `export_json_data` are documented. should i remove all other functionality that is not included by `power_grid_model.__init__`?

TonyXiang8787: Yes, except `import_json_data` and `export_json_data`, other stuffs can be removed if it is no long used by other internal functions/tests.

docs/advanced_documentation/build-guide.md
```diff
@@ -73,12 +73,13 @@ This repository uses [CMake](https://cmake.org/) (version 3.23 or later) and [Ni
...
+| [boost](https://www.boost.org/)                                     | Will be installed automatically      | CMake needs to be able find `boost`         | header-only | [Boost Software License - Version 1.0](https://www.boost.org/LICENSE_1_0.txt)                                |
+| [eigen3](https://eigen.tuxfamily.org/)                              | Will be installed automatically      | CMake needs to be able find `eigen3`        | header-only | [Mozilla Public License, version 2.0](https://www.mozilla.org/en-US/MPL/2.0/)                                |
+| [doctest](https://github.com/doctest/doctest)                       | None                                 | CMake needs to be able find `doctest`       | header-only | [MIT](https://github.com/doctest/doctest/blob/master/LICENSE.txt)                                            |
+| [nlohmann-json](https://github.com/nlohmann/json)                   | None                                 | CMake needs to be able find `nlohmann_json` | header-only | [MIT](https://github.com/nlohmann/json/blob/develop/LICENSE.MIT)                                             |
```
TonyXiang8787, line 81: json and msgpack is now also build dependencies for `Requirements to build Python package`.

src/power_grid_model/core/serialization.py
```diff
@@ -0,0 +1,319 @@
...
+    Raises:
+        KeyError if the dataset_type was not provided and could not be deduced from the dataset (i.e. it was empty).
+        PowerGridError if there was an internal error.
+
+    Returns:
+        a serialized string containing the dataset
+    """
+    result = MsgpackSerializer(data=data, dataset_type=dataset_type).dump(use_compact_list=use_compact_list)
+    assert_no_error()
+    return result
```
TonyXiang8787, line 319: please add helper functions to load/dump json/msgpack from/to files.

src/power_grid_model/core/serialization.py
```diff
@@ -0,0 +1,319 @@
...
+
+        instance._dataset = CWritableDataset(instance._dataset_ptr)
+        assert_no_error()
+
+        return instance
+
+    def __del__(self):
+        pgc.destroy_deserializer(self._deserializer)
+
+    def load(self) -> Tuple[Mapping[str, Union[np.ndarray, Mapping[str, np.ndarray]]], str]:
```
TonyXiang8787, line 61: I think we agreed that we only return the dictionary to the user. Not a tuple. If the user is not sure about the dataset type, the user can call the dataset detection function.

src/power_grid_model/core/serialization.py
mgovers: what would you call them? we now have `(json|msgpack)_(deserialize|serialize)`

would you use `(json|msgpack)_(load|dump)`? for `msgpack` it's not really loading/dumping, but packing/unpacking, right?

src/power_grid_model/core/serialization.py
mgovers: the user can only call the dataset detection function if the data type is not empty. what would he have to do when it is empty?

src/power_grid_model/utils.py
mgovers: to prevent this discussion in the future, I would recommend adding the `_` prefix to any objects or functions in libraries we expect users to be able to include

src/power_grid_model/core/serialization.py
TonyXiang8787: `(json|msgpack)_(deserialize|serialize)_(to|from)_file`

src/power_grid_model/core/serialization.py
TonyXiang8787: This is a very niche corner case, the user have an empty dataset input, and he does not know the dataset type. I would not just for this case create a function with tuple return. Tuple return is in general (I think) confusing.

If the user really has this situation, he/she can always load/unpack the data self and read the `type` key.

TonyXiang8787: I would say in most use cases, the user already know the dataset type of a certain file.

src/power_grid_model/utils.py
mgovers: i moved all internally used utils to `_utils` and all test-only utils to `tests.unit.utils`

src/power_grid_model/core/serialization.py
mgovers: i exposed the functionality to the user and it raises when it cannot be determined

TonyXiang8787: I do not understand. Are you still returning a tuple?

mgovers: nope

mgovers: instead, i exposed the functionality `get_dataset_type(data: Dataset)` to the user

src/power_grid_model/_utils.py
```diff
@@ -0,0 +1,280 @@
+# SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model project <dynamic.grid.calculation@alliander.com>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+"""
+This file contains helper functions for library-internal use only.
+
+Disclaimer!
+
+We do not officially support support this functionality and may remove features in this library at any given time!
```
mgovers, line 10:
```suggestion
We do not officially support this functionality and may remove features in this library at any given time!
```

mgovers:
@TonyXiang8787 @petersalemink95 open question on forward-compatibility:

do you think it's a good idea to add an optional `version: str = "1.0"` keyword argument to the new `json_serialize_*` and `msgpack_serialize_*` interfaces? that way, users can decide whether they want to hard fix their format version if they themselves use a very specific output query or that they allow the PGM to serialize using the most recent version

petersalemink95 (approved):

TonyXiang8787:
> @TonyXiang8787 @petersalemink95 open question on forward-compatibility:
> 
> 
> 
> do you think it's a good idea to add an optional `version: str = "1.0"` keyword argument to the new `json_serialize_*` and `msgpack_serialize_*` interfaces? that way, users can decide whether they want to hard fix their format version if they themselves use a very specific output query or that they allow the PGM to serialize using the most recent version

I don't think that's needed. In a given version of PGM, we only serialize to the latest supported format.

TonyXiang8787 (approved):

mgovers:
> > @TonyXiang8787 @petersalemink95 open question on forward-compatibility:
> > do you think it's a good idea to add an optional `version: str = "1.0"` keyword argument to the new `json_serialize_*` and `msgpack_serialize_*` interfaces? that way, users can decide whether they want to hard fix their format version if they themselves use a very specific output query or that they allow the PGM to serialize using the most recent version
> 
> I don't think that's needed. In a given version of PGM, we only serialize to the latest supported format.

A valid remark that @petersalemink95 had is that some people may use the serialized data as their temporary format and that they could load into another application/pipeline step. in that case, people may already be using specific queries.

If ever, I do believe that now is a good time to add such a feature (probably it's better to use an enum now that i think about it), because we would want to bump the default value when we want to. It also would enable us to develop on `main` instead of on a feature branch like we're doing now to prevent conflicts.

PR #368, opened 2023-08-30
add fault current validation

nitbharambe:


tests/data/short_circuit/single_phase_to_ground_c_maximum/sc_output_batch.json
```diff
@@ -2,23 +2,23 @@
...
+        {"id": 0, "energized": 1, "u_pu": [1.1000039840999545, 1.1000016506895227, 1.1000042703666313], "u": [95263.13944946544, 95262.9373701943, 95263.16424088688], "u_angle": [-1.6752923662861285e-06, -2.094395252466148, 2.0943960267868786]},
+        {"id": 1, "energized": 1, "u_pu": [1.6612656886172617e-16, 1.9056219941306511, 1.9056225158197297], "u": [9.591321925186651e-13, 11002.113712850029, 11002.116724823329], "u_angle": [-0.5733177432085695, 3.1415256616159075, 2.0944613189550867]},
+        {"id": 2, "energized": 1, "u_pu": [0.00015708279091403854, 1.905581420575748, 1.9057370030880025], "u": [0.9069179161927785, 11001.879461321576, 11002.77771737489], "u_angle": [0.8872079931204931, 3.14145189401517, 2.0943744277646714]},
+        {"id": 3, "energized": 1, "u_pu": [0.0001767182112095378, 1.90557634935507, 1.9057513147761516], "u": [1.020283068125358, 11001.850182615339, 11002.860345944942], "u_angle": [0.8872063197433703, 3.1414426729060394, 2.0943635670026213]}
```
mgovers, line 8: you're not calculating things differently, right? why do you get rounding errors?

TonyXiang8787: I don't this this would be an issue. Even if you upgrade your numpy library can result in a different solution. Floating pointer is never exact.

mgovers: it shouldn't be but for the sake of keeping this PR clean i would suggest to try and keep the rounding differences to a minimum

TonyXiang8787: @mgovers I would not spend too much time on rounding differences.

TonyXiang8787:
@nitbharambe what is the status here? Can we merge it?

TonyXiang8787 (approved):

Issue #365, opened 2023-08-29
[IMPROVEMENT] Input/output/update data structures as standard layout


mgovers:
## Background

The underlying [input, output and update data structures](https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html) behind the data types retrieved by [initialize_array](https://power-grid-model.readthedocs.io/en/stable/api_reference/python-api-reference.html#power_grid_model.initialize_array) are generated using [code generation](https://github.com/PowerGridModel/power-grid-model/tree/main/code_generation).

The current implementation uses inheritance to define the relationship between components and the types they implement and are derived from. This keeps the data structures short and readable.

## Problem

The downside is that inheritance causes the data types are not guaranteed to obey standard layout. This may result in less efficient data structures, and also makes the code more bug-prone. As an example, #241 was caused by the fact that the data type does not have standard layout.

### Plain text with implicit reference operator conversion

```c++
// Example program
#include <iostream>
#include <limits>


using ID = int;

struct BaseInput { ID id; };
struct NodeInput{
    ID id;
    double u_rated;
    
    // Sonar Cloud warnings about `explicit` keyword and reinterpret_cast should be disabled here
    operator BaseInput&() { return reinterpret_cast<BaseInput&>(*this); }
    operator BaseInput const&() const { return reinterpret_cast<BaseInput const&>(*this); }

};

int main()
{
    NodeInput const node_input{ 5, 10.0 };
    BaseInput const& base = node_input;
    std::cout << base.id << '\n';
}
```

### Concepts

The standard layout solution can be achieved using a plain struct types without references to any base classes. For written code, that would cause a lot of code duplication, but since the code under consideration is generated, this is not an issue.

The notion of the "base" types are enabled by declaring them `concept`s, which can be tested using `static_assert`.

```cpp
template <typename T>
concept base_input_c = requires (T t) {
    { t.id } -> std::same_as<ID&>;
};

template <typename T>
concept node_input_c = base_input_c <T> && requires (T t) {
    { t.u_rated} -> std::same_as<double&>;
};

template <bool sym>
struct NodeInput {
    ID id{};
    double u_rated{};
};

static_assert(base_input_c<NodeInput<true>>);
static_assert(node_input_c<NodeInput<true>>);
static_assert(base_input_c<NodeInput<false>>);
static_assert(node_input_c<NodeInput<false>>);
```

Constructors of components will then take the following form (and equivalent for their update and output functions):

```cpp
class Base {
[...]

PR #376, opened 2023-09-11
[DO NOT MERGE] test implicit conversion

mgovers:
Tests for #365


mgovers:
This was a POC on #365 . Since the POC is done and documented in #365 , this will now be closed

PR #379, opened 2023-09-12
Refactor indptr / sparse structure

nitbharambe:


power_grid_model_c/power_grid_model/include/power_grid_model/main_core/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,50 @@
+// SPDX-FileCopyrightText: 2022 Contributors to the Power Grid Model project <dynamic.grid.calculation@alliander.com>
+//
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+#ifndef POWER_GRID_MODEL_MAIN_CORE_TOPOLOGY_INDEX_HPP
+#define POWER_GRID_MODEL_MAIN_CORE_TOPOLOGY_INDEX_HPP
```
mgovers, line 7:
```suggestion
#ifndef POWER_GRID_MODEL_MAIN_CORE_SPARSE_IDX_VECTOR_HPP
#define POWER_GRID_MODEL_MAIN_CORE_SPARSE_IDX_VECTOR_HPP
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "../power_grid_model.hpp"
+
+#include <boost/range.hpp>
+
+namespace power_grid_model::main_core {
+
+class SparseIdxVector {
+
```
mgovers, line 16:
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,50 @@
...
+#ifndef POWER_GRID_MODEL_MAIN_CORE_TOPOLOGY_INDEX_HPP
+#define POWER_GRID_MODEL_MAIN_CORE_TOPOLOGY_INDEX_HPP
+
+#include "../power_grid_model.hpp"
+
+#include <boost/range.hpp>
+
+namespace power_grid_model::main_core {
+
+class SparseIdxVector {
```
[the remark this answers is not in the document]
nitbharambe: I thought SparseIdxVector in place of (eg. source_bus_indptr)
And SparseDataVector in Mathmodel's source param. (Actual data of source)

Is that logic okay?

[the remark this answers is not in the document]
nitbharambe: Did you mean this way? (I made a commit.)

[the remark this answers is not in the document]
nitbharambe: @mgovers  I shall merge the two. One class with two vector members. one of indptr and data, both internal to it. 
And externally we only retrieve the data as we see fit.

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,42 @@
...
+#ifndef POWER_GRID_MODEL_SPARSE_IDX_VECTOR_HPP
+#define POWER_GRID_MODEL_SPARSE_IDX_VECTOR_HPP
+
+#include "power_grid_model.hpp"
+
+#include <boost/range.hpp>
+
+namespace power_grid_model::detail {
+
+template <class T> class SparseIdxVector {
```
mgovers, line 15: this becomes a `SparseVector`

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/sparse_idx_vector.hpp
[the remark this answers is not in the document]
mgovers: Apologies, I misinterpreted what you wanted to do initially. But I also see what you implemented now and it looks good to me. (see below what I misunderstood)

## New implementation with only a `SparseVector`

This is your current implementation (commit [e08ff3b](https://github.com/PowerGridModel/power-grid-model/pull/379/commits/e08ff3bca7f70e31d4d64fdb2ca1ad55094b1933)).

As per how it will be used, maybe i wasn't concrete enough on what the usage would look like. i would suggest is something like:

```cpp
struct SourceParam {
  double a{};
  double b{};
};

template <typename T> class SparseVector {
  private:
    IdxVector indptr;
    std::vector<T> data;

  public:
    // vector interface
    // map interface
}

struct CalculationParams {
  SparseVector<SourceParam> source_param;
};
```

The advantage is that there's no `SparseIdxVector` exposed to the outside world (range-based over index-based for loops). The downside is that the data storage is row-based, not columnar, as for the current implementation, although I am not convinced yet that that is a problem.

I now also understand why you initially went with your alternative approach. See below.

## Your initial approach

The main advantage of your initial approach (2 data types with no explicit relation between the two) is that the data is kept separate from the indices - as with the current existing implementation. you can also keep the columnar data storage format.

The downside of your original approach is that your map-like interface returns `std::pair`, which the data should not have to care about. I would expect in your approach that the `SparseIdxVector` returns ONLY an `Idx`, so that your `SparseDataVector` can just be the `std::vector`

## `SparseVector` with columnar storage

To implement the new implementation with columnar storage format would require variadic templates (that would almost become something like the `Container` class, but it would be near-impossible to deal with duplicate data types like two fields that are both `double`) or something like your original implementation

```cpp
class SparseIdxVector {
  // blabla
};

template <typename T> class SparseDataVector{
  public:
    SparseDataVector(SparseIdxVector const& indices) : indices_{indices} {};
    // or with std::shared_ptr<SparseIdxVector const>

    // vector-like interface
    // map-like interface

  private:
    SparseIdxVector const& indices_;
    // or with std::shared_ptr<SparseIdxVector const>

    std::vector<T> data_;
};
```

otherwise you would end up with a lot of data duplication (a `SparseIdxVector` per `SparseDataVector`). Your initial approach but with only `IdxVector` specified would be far superior.

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,199 @@
...
+        using iterator_category = std::forward_iterator_tag;
+
+      public:
+        Iterator(ValueVector const& indptr, size_t const& group, size_t const& element)
+            : indptr_(indptr), group_(group), element_(element) {}
+
+        friend bool operator==(Iterator const& lhs, Iterator const& rhs) {
+            return lhs.group_ == rhs.group_ && lhs.element_ == rhs.element_;
+        }
+        friend bool operator!=(Iterator const& lhs, Iterator const& rhs) { return !(lhs == rhs); }
```
mgovers, line 68: not sure but i think you should be able to use `= default` here

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,199 @@
...
+            return static_cast<KeyType>(group_);
+        }
+
+        auto operator*()
+            requires ItemsView<ViewTag>
+        {
+            return std::make_tuple(group_, indptr_[group_] + element_);
+        }
+
+        auto size() { return static_cast<KeyType>(indptr_.back()); }
```
mgovers, line 82: here and to some of the other functions: please try to make them `constexpr`

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,199 @@
...
+              group_begin_(group_begin),
+              group_end_(group_end),
+              element_begin_(element_begin),
+              element_end_(element_end) {}
+
+        friend bool operator==(ElementRange const& lhs, ElementRange const& rhs) {
+            return lhs.group_begin_ == rhs.group_begin_ && lhs.element_begin_ == rhs.element_begin_ &&
+                   lhs.group_end_ == rhs.group_end_ && lhs.element_end_ == rhs.element_end_;
+        }
+        friend bool operator!=(ElementRange const& lhs, ElementRange const& rhs) { return !(lhs == rhs); }
```
mgovers, line 151: here as well: maybe `= default` does the job

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,199 @@
...
+      public:
+        using ElementRangeType = ElementRange<ViewTag>;
+
+        GroupIterator(ValueVector const& indptr, size_t group_begin, size_t group_end, size_t element_begin,
+                      size_t element_end)
+            : ElementRangeType{indptr, group_begin, group_end, element_begin, element_end} {}
+
+        GroupIterator& operator++() {
+            ++this->group_begin_;
+            ++this->group_end_;
```
mgovers, line 174: for readability, can you either remove `this->` or add braces?

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,199 @@
...
+            return *this;
+        }
+
+        auto operator*() { return static_cast<ElementRangeType>(*this); }
+
+        auto operator[](size_t const group) { return ElementRangeType{this->indptr_, group, group + 1, 0, 0}; }
+
+        size_t size() { return this->indptr_.size() - 1; }
+        auto begin() { return GroupIterator{this->indptr_, 0, 1, 0, 0}; }
+        auto end() { return GroupIterator{this->indptr_, this->indptr_.size() - 1, this->indptr_.size(), 0, 0}; }
```
mgovers, line 184: here and in other places: i don't think you need `this`

tests/cpp_unit_tests/test_sparse_idx_vector.cpp
```diff
@@ -0,0 +1,82 @@
...
+        IdxVector actual_groups{};
+        for (Idx i : elm_iter) {
+            actual_groups.push_back(i);
+        }
+        CHECK(actual_groups == expected_groups);
+
+        CHECK(elm_iter.size() == 7);
+    }
+
+    // TODO Implement items
```
mgovers, line 35: remaining TODO

mgovers (commented):
overall looks good

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,112 @@
...
+        friend class boost::iterator_core_access;
+
+        boost::iterator_range<IdxCount> dereference() const {
+            return boost::counting_range(indptr_[idx_], indptr_[idx_ + 1]);
+        }
+        bool equal(Iterator const& other) const { return idx_ == other.idx_; }
+        void increment() { ++idx_; }
+        void decrement() { --idx_; }
+        void advance(Idx n) { idx_ += n; }
+        Idx distance_to(Iterator const& other) const { return other.idx_ - idx_; }
```
mgovers, line 55: shouldn't it be weighted by group?

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
mgovers, line 1: the name of this file is no longer up to tae

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    auto get_element_range(Idx group) { return boost::iterator_range<IdxCount>(indptr_[group], indptr_[group + 1]); }
+
+    Idx get_group(Idx element) {
+        return static_cast<Idx>(std::upper_bound(indptr_.begin(), indptr_.end(), element) - indptr_.begin());
```
mgovers, line 61: i think it should throw an `std::out_of_range` error when reading outside the allowed elements

https://en.cppreference.com/w/cpp/error/out_of_range

tests/cpp_unit_tests/test_sparse_idx_vector.cpp
```diff
@@ -0,0 +1,92 @@
...
+            if (range_size == 0) {
+                CHECK(actual_idx_counts.empty());
+            } else {
+                CHECK(actual_idx_counts.front() == IdxCount{sample_indptr[group_number]});
+                CHECK(actual_idx_counts.back() == IdxCount{sample_indptr[group_number + 1] - 1});
+            }
+        }
+    }
+
+    SUBCASE("Zip iterator") {
```
mgovers, line 75: since you wrote a variadic `zip`, can you also verify that it works correctly for:

* 0, 1 and 3 vectors
* dense index vectors

tests/cpp_unit_tests/test_sparse_idx_vector.cpp
```diff
@@ -0,0 +1,92 @@
...
+            auto group_i = dense_idx_vector.get_element_range(group_number);
+            std::copy(group_i.begin(), group_i.end(), actual_idx_counts.begin());
+            if (range_size == 0) {
+                CHECK(actual_idx_counts.empty());
+            } else {
+                CHECK(actual_idx_counts.front() == IdxCount{sample_indptr[group_number]});
+                CHECK(actual_idx_counts.back() == IdxCount{sample_indptr[group_number + 1] - 1});
+            }
+        }
+    }
```
mgovers, line 73: the test cases for the `SparseIdxVector` and `DenseIdxVector` should IMO be the same aside from the input values. Maybe you can reflect that by making a template test function for that?

tests/cpp_unit_tests/test_sparse_idx_vector.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    // 2nd sparse vector
+    IdxVector const sample_indptr_2{0, 0, 1, 3, 6, 6, 6};
+    IdxVector const expected_elements{0, 1, 2, 3, 4, 5};
+    std::vector<IdxCount> expected_idx_counts_groups_2{expected_elements.begin(), expected_elements.end()};
+    SparseIdxVector sparse_idx_vector_2{sample_indptr_2};
+
+    // Dense Vector
+    IdxVector sample_dense_vector{1, 1, 1, 3, 3, 3, 4};
+    DenseIdxVector dense_idx_vector{sample_dense_vector};
```
mgovers, line 26: shorter

```suggestion
    DenseIdxVector dense_idx_vector{{1, 1, 1, 3, 3, 3, 4}};
```

tests/cpp_unit_tests/test_sparse_idx_vector.cpp
```diff
@@ -0,0 +1,92 @@
...
+    IdxVector const sample_indptr{0, 0, 3, 3, 6, 7, 7};
+    IdxVector expected_elements{0, 1, 2, 3, 4, 5, 6};
+    std::vector<IdxCount> expected_idx_counts_groups{expected_elements.begin(), expected_elements.end()};
+    SparseIdxVector sparse_idx_vector{sample_indptr};
+
+    // 2nd sparse vector
+    IdxVector const sample_indptr_2{0, 0, 1, 3, 6, 6, 6};
+    IdxVector const expected_elements{0, 1, 2, 3, 4, 5};
+    std::vector<IdxCount> expected_idx_counts_groups_2{expected_elements.begin(), expected_elements.end()};
+    SparseIdxVector sparse_idx_vector_2{sample_indptr_2};
```
mgovers, line 22: shorter

```suggestion
    std::vector<IdxCount> expected_idx_counts_groups_2{0, 1, 2, 3, 4, 5};
    SparseIdxVector sparse_idx_vector_2{{0, 0, 1, 3, 6, 6, 6}};
```

tests/cpp_unit_tests/test_sparse_idx_vector.cpp
```diff
@@ -0,0 +1,92 @@
...
+TEST_CASE("Sparse idx data strucuture for topology") {
+    // Sparse vector to test
+    IdxVector const sample_indptr{0, 0, 3, 3, 6, 7, 7};
+    IdxVector expected_elements{0, 1, 2, 3, 4, 5, 6};
+    std::vector<IdxCount> expected_idx_counts_groups{expected_elements.begin(), expected_elements.end()};
+    SparseIdxVector sparse_idx_vector{sample_indptr};
+
+    // 2nd sparse vector
+    IdxVector const sample_indptr_2{0, 0, 1, 3, 6, 6, 6};
+    IdxVector const expected_elements{0, 1, 2, 3, 4, 5};
```
mgovers, line 20: you cannot have the same variable name twice in the same context

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,112 @@
...
+    // using all_true = std::is_same<bool_pack<bs..., true>, bool_pack<true, bs...>>;
+
+    // static_assert(all_true<(Numbers == 0 || Numbers == 1)...>::value, "");
+    // Check if all are of same size
+    // Add index at the start position
+
+    auto zip_begin = boost::make_zip_iterator(boost::make_tuple(sparse_vectors.begin()...));
+    auto zip_end = boost::make_zip_iterator(boost::make_tuple(sparse_vectors.end()...));
+    return boost::make_iterator_range(zip_begin, zip_end);
+}
```
mgovers, line 108: * also for dense

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_idx_vector.hpp
```diff
@@ -0,0 +1,112 @@
...
+auto zip_idx_vector(T&... sparse_vectors) {
+    // assert(all(sparse_vectors.size()...))
+
+    // template<bool...> struct bool_pack;
+    // template<bool... bs> 
+    // using all_true = std::is_same<bool_pack<bs..., true>, bool_pack<true, bs...>>;
+
+    // static_assert(all_true<(Numbers == 0 || Numbers == 1)...>::value, "");
+    // Check if all are of same size
+    // Add index at the start position
```
mgovers, line 103: remove uncommented code

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+        constexpr bool equal(GroupIterator const& other) const { return group_ == other.group_; }
+        constexpr void increment() { ++group_; }
+        constexpr void decrement() { --group_; }
+        constexpr void advance(Idx n) { group_ += n; }
+        constexpr Idx distance_to(GroupIterator const& other) const { return other.group_ - group_; }
+    };
+
+  public:
+    SparseIdxVector() = default;
+    explicit SparseIdxVector(IdxVector indptr) : indptr_(indptr.empty() ? IdxVector{0} : indptr) {
```
mgovers, line 57:
```suggestion
    explicit SparseIdxVector(IdxVector indptr) : indptr_(indptr.empty() ? IdxVector{0} : std::move(indptr)) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+    auto all_equal_sizes = ((first.size() == rest.size()) && ...);
+    assert(all_equal_sizes);
+    auto zip_begin = boost::make_zip_iterator(boost::make_tuple(first.begin(), rest.begin()...));
+    auto zip_end = boost::make_zip_iterator(boost::make_tuple(first.end(), rest.end()...));
+    return boost::make_iterator_range(zip_begin, zip_end);
+}
+
+} // namespace power_grid_model
+
+#endif
```
mgovers, line 148:
```suggestion
#endif

```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+template <typename T>
+concept grouped_idx_vector_type =
+    std::is_same<T, const SparseIdxVector>::value || std::is_same<T, const DenseIdxVector>::value;
+
+template <grouped_idx_vector_type First, grouped_idx_vector_type... Rest>
+auto zip_sequence(First& first, Rest&... rest) {
+    auto all_equal_sizes = ((first.size() == rest.size()) && ...);
+    assert(all_equal_sizes);
+    auto zip_begin = boost::make_zip_iterator(boost::make_tuple(first.begin(), rest.begin()...));
+    auto zip_end = boost::make_zip_iterator(boost::make_tuple(first.end(), rest.end()...));
```
mgovers, line 142:
```suggestion
    assert((first.size() == rest.size()) && ...);
    auto const zip_begin = boost::make_zip_iterator(boost::make_tuple(first.begin(), rest.begin()...));
    auto const zip_end = boost::make_zip_iterator(boost::make_tuple(first.end(), rest.end()...));
```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+    auto get_element_range(Idx group) const { return *group_iterator(group); }
+
+  private:
+    IdxVector dense_vector_;
+    Idx groups_size_;
+};
+
+template <typename T>
+concept grouped_idx_vector_type =
+    std::is_same<T, const SparseIdxVector>::value || std::is_same<T, const DenseIdxVector>::value;
```
mgovers, line 135: shorter + it's good practice to have concepts require other concepts instead of their respective template struct test

```suggestion
concept grouped_idx_vector_type =
    std::same_as<T, const SparseIdxVector> || std::same_as<T, const DenseIdxVector>;
```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+*/
+
+namespace power_grid_model {
+
+class SparseIdxVector {
+  private:
+    class GroupIterator : public boost::iterator_facade<GroupIterator, Idx, boost::random_access_traversal_tag,
+                                                        boost::iterator_range<IdxCount>, Idx> {
+      public:
+        explicit constexpr GroupIterator(IdxVector const& indptr, Idx group) : indptr_(indptr), group_(group) {}
```
mgovers, line 38: it's probably a good idea to have constructors with tags (similar to how it's done for the `Deserializer`). that way, the converter we wrote for the tests can be used to easily construct one from the other

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+Another intuitive way to look at this for python developers is like list of lists: [[0, 1, 2], [3, 4, 5], [6]].
+
+DenseIdxVector is a vector of element to group. ie. [0, 1, 1, 4] would denote that [[0], [1, 2], [], [], [3]]
+The input, ie. [0, 1, 3] should be strictly increasing
+
+*/
+
+namespace power_grid_model {
+
+class SparseIdxVector {
```
mgovers, line 33: out of scope for this PR but i'm starting to think that it's a good idea to have a data class that dynamically uses one or the other. maybe i'll do it in #402 but we'll see. let's first actually do the refactoring before making it more detailed

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,148 @@
...
+    IdxVector dense_vector_;
+    Idx groups_size_;
+};
+
+template <typename T>
+concept grouped_idx_vector_type =
+    std::is_same<T, const SparseIdxVector>::value || std::is_same<T, const DenseIdxVector>::value;
+
+template <grouped_idx_vector_type First, grouped_idx_vector_type... Rest>
+auto zip_sequence(First& first, Rest&... rest) {
```
mgovers, line 138:
```suggestion
template <grouped_idx_vector_type First, grouped_idx_vector_type... Rest>
auto zip_sequence(grouped_idx_vector_type auto const& first, grouped_idx_vector_type  auto const& rest) {
```

mgovers, line 138: global functions should always be explicitly marked inline or be in a nameless namespace.

To be more precise:

* for headers: mark either `constexpr` or `inline` (preferably `constexpr`)
* for cpp-files: put in a nameless namespace:
  ```cpp
  namespace {
    // impl
  }`
  ```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
mgovers, line 135: to test this concept, you can add `static_asserts`:

```cpp
static_assert(grouped_idx_vector_type<SparseIdxVector>);
static_assert(grouped_idx_vector_type<DenseIdxVector>);
```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
nitbharambe: This was giving a compile error in gcc. It said the fold doesnt work directly within braces of assert.

mgovers: then with an additional pair of braces

```suggestion
    assert(((first.size() == rest.size()) && ...));
    auto const zip_begin = boost::make_zip_iterator(boost::make_tuple(first.begin(), rest.begin()...));
    auto const zip_end = boost::make_zip_iterator(boost::make_tuple(first.end(), rest.end()...));
```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,150 @@
...
+    constexpr auto get_group(Idx element) const { return dense_vector_[element]; }
+    auto get_element_range(Idx group) const { return *group_iterator(group); }
+
+  private:
+    IdxVector dense_vector_;
+    Idx groups_size_;
+};
+
+template <typename T>
+concept grouped_idx_vector_type = std::same_as<T, SparseIdxVector> || std::same_as<T, DenseIdxVector>;
```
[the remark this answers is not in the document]
mgovers: fixed myself

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,204 @@
...
+        constexpr void advance(Idx n) { group_ += n; }
+        constexpr Idx distance_to(GroupIterator const& other) const { return other.group_ - group_; }
+    };
+
+    constexpr auto group_iterator(Idx group) const { return GroupIterator{indptr_, group}; }
+
+  public:
+    SparseIdxVector() = default;
+    explicit SparseIdxVector(IdxVector indptr) : indptr_{indptr.empty() ? IdxVector{0} : std::move(indptr)} {
+        assert(!indptr_.empty());
```
mgovers, line 114: also assert that it is sorted

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,204 @@
...
+        }
+        constexpr Idx distance_to(GroupIterator const& other) const { return other.group_ - group_; }
+    };
+
+    constexpr auto group_iterator(Idx group) const { return GroupIterator{dense_vector_, group}; }
+
+  public:
+    DenseIdxVector() = default;
+    explicit DenseIdxVector(IdxVector dense_vector, Idx groups_size)
+        : dense_vector_(std::move(dense_vector)), groups_size_(groups_size) {}
```
mgovers, line 176: also assert that it is sorted

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+                                                                         { t.element_size() } -> std::same_as<Idx>;
+                                                                         { t.get_group(idx) } -> std::same_as<Idx>;
+                                                                     };
+} // namespace detail
+
+template <typename T>
+concept grouped_idx_vector_type = detail::grouped_index_vector_type<T>;
+
+class SparseIdxVector {
```
mgovers, line 87: since the contents are grouped, maybe it's a good idea to reflect that in the class name

mgovers (approved):
i think this is about as far as we can go for this PR. i have not run into any new issues on my refactor branch so so far, so good

tests/cpp_unit_tests/test_grouped_index_vector.cpp
```diff
@@ -0,0 +1,161 @@
...
+    SUBCASE("Zipped 3 same grouped idx vectors") {
+        // Create additional 2 grouped idx vectors
+        auto const idx_vector_2 = construct_from<IdxVectorType, from_natural_t>(groups, num_groups);
+        auto const idx_vector_3 = construct_from<IdxVectorType, from_natural_t>(groups, num_groups);
+
+        // Test 3 zipped iterations
+        std::vector<boost::iterator_range<IdxCount>> actual_ranges_1{};
+        std::vector<boost::iterator_range<IdxCount>> actual_ranges_2{};
+        std::vector<boost::iterator_range<IdxCount>> actual_ranges_3{};
+        for (auto [element_range_1, element_range_2, element_range_3] :
```
TonyXiang8787, line 103: In the zip you should also have a group number right?

```c++
for (auto [group_number, element_range_1, element_range_2, element_range_3] :zip_sequence(idx_vector, idx_vector_2, idx_vector_3))
```

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,255 @@
...
+  public:
+    auto size() const { return static_cast<Idx>(indptr_.size()) - 1; }
+    auto begin() const { return group_iterator(0); }
+    auto end() const { return group_iterator(size()); }
+    auto get_element_range(Idx group) const { return *group_iterator(group); }
+
+    auto element_size() const { return indptr_.back(); }
+    auto get_group(Idx element) const -> Idx {
+        assert(element < element_size());
+        return std::distance(std::begin(indptr_), std::upper_bound(std::begin(indptr_), std::end(indptr_), element)) -
```
TonyXiang8787, line 154: what's the reason to use `upper_bound` everywhere instead of `lower_bound`? I guess using `lower_bound` you don't need to use the `-1`?

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,255 @@
...
+struct from_dense_t {};
+
+constexpr auto from_sparse = from_sparse_t{};
+constexpr auto from_dense = from_dense_t{};
+
+class DenseGroupedIdxVector;
+
+class SparseGroupedIdxVector {
+  private:
+    class GroupIterator : public boost::iterator_facade<GroupIterator, Idx, boost::random_access_traversal_tag,
```
TonyXiang8787, line 123: I do not foresee the need of using random access iterator for the `GroupIterator`. A forward iterator should be enough. You only need to define `increment` function.

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,255 @@
...
+    SparseGroupedIdxVector(from_dense_t /* tag */, IdxVector const& dense_group_elements, Idx num_groups)
+        : SparseGroupedIdxVector{detail::sparse_encode(dense_group_elements, num_groups)} {}
+
+  private:
+    IdxVector indptr_;
+};
+
+class DenseGroupedIdxVector {
+  private:
+    class GroupIterator : public boost::iterator_facade<GroupIterator, Idx, boost::random_access_traversal_tag,
```
TonyXiang8787, line 176: Same here
I do not foresee the need of using random access iterator for the `GroupIterator`. A forward iterator should be enough. You only need to define `increment` function.

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,255 @@
...
+        }
+        constexpr bool equal(GroupIterator const& other) const { return group_ == other.group_; }
+        constexpr void increment() { advance(1); }
+        constexpr void decrement() { advance(-1); }
+        constexpr void advance(Idx n) {
+            auto const start = n > 0 ? group_range_.second : std::cbegin(dense_vector_);
+            auto const stop = n > 0 ? std::cend(dense_vector_) : group_range_.first;
+
+            group_ += n;
+            group_range_ = std::equal_range(start, stop, group_);
```
TonyXiang8787, line 208: This is exactly we don't want to implement `advance` to satisfy random access iterator.

If we only implement `increment` to satistfy forware iterator. The logic here can be much simpler.

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
```diff
@@ -0,0 +1,255 @@
...
+  private:
+    class GroupIterator : public boost::iterator_facade<GroupIterator, Idx, boost::random_access_traversal_tag,
+                                                        boost::iterator_range<IdxCount>, Idx> {
+      public:
+        using iterator = boost::iterator_range<IdxCount>;
+
+        explicit constexpr GroupIterator(IdxVector const& indptr, Idx group) : indptr_(indptr), group_(group) {}
+
+      private:
+        IdxVector const& indptr_;
```
TonyXiang8787, line 131: Iterator needs to be default constructible and assignable. I think that's the reason you do not satisify the requirements of standard iterator concept.

Please define this as a pointer. The default constructor will assign this pointer to `nullptr`.

mgovers: i did an investigation for this and thought i found that the boost `counting_iterator` does not satisfy the concept, but i will double-check

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
mgovers: the problem is when there is are repeating values (no elements in a group).

e.g.:

the indptr `{0, 0, 2}` will result in `get_group(0) == 0` instead of the correct value of `get_group(0) == 1`

power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
mgovers: as discussed: i fixed the `nullptr` part. the `counting_iterator` does not satisfy the `std::random_access_iterator` constraints. I added a TODO with the following static assert:

```cpp
// TODO(mgovers): replace the below relevant subset here ourselves with the STD equivalent when we have std::ranges.
// boost::counting_iterator does not satisfy all requirements std::*_iterator concepts:
static_assert(!std::random_access_iterator<IdxCount>);
// we have to declare the relevant subset here ourselves.
```

mgovers:
@TonyXiang8787 this is ready for review

TonyXiang8787 (approved):

PR #382, opened 2023-09-18
Feature/pgm serialization validation tests

mgovers:
* Bump the validation tests to the new serialization format
* Keep a few tests with the original format in the `deprecated_format/` subdirectory

Refactoring of the C++ validation tests explicitly is postponed to a separate PR

Upgrade script:

```py
from pathlib import Path


def get_type(file_path: Path):
    return "".join(file_path.with_suffix("").name.split("_batch"))


def get_batch(file_path: Path):
    return "_batch" in str(file_path.name)


def update(file_path):
    with open(file_path, mode="r", encoding="utf-8") as fp:
        orig_lines = fp.readlines()

    with open(file_path, mode="w", encoding="utf-8") as fp:
        fp.write("{\n")
        fp.write('  "version": "1.0",\n')
        fp.write('  "type": "' + get_type(file_path) + '",\n')
        fp.write('  "is_batch": ' + ("true" if get_batch(file_path) else "false") + ",\n")
        fp.write(r'  "attributes": {},' + "\n")
        fp.write('  "data": ' + orig_lines[0])
        for line in orig_lines[1:]:
            fp.write("  " + line.rstrip() + "\n")  # avoid inconsistency between new-line and no new-line
        fp.write('}')
    # print(file_path, "datset_type =", dataset_type, "is_batch =", is_batch)


def main():
    for file_path in Path("tests/data").glob("**/*"):
        if file_path.is_dir():
            continue
        if file_path.suffix != ".json":
            continue
        if file_path.name == "params.json":
            continue
        if "deprecated_format" in str(file_path):
            continue

        print("updating", file_path)
        update(file_path)


if __name__ == "__main__":
    main()

```


tests/data/power_flow/1os2msr/input.json
```diff
@@ -1,69 +1,75 @@
...
-      "type": 0,
-      "p_specified": 1.02e6,
-      "q_specified": 0.22e6
-    }
-  ]
+  "attributes": {},
+  "data": {
+    "line": [
+      {
+        "c1": 4.1380285203892784e-05,
```
petersalemink95, line 6: Does not seem like a logical order to me. Is this generated random?

tests/data/power_flow/1os2msr/input.json
```diff
@@ -1,69 +1,75 @@
...
+        "tan1": 0.10588235294117646,
+        "to_node": 3,
+        "to_status": 1,
+        "x1": 0.16
+      }
+    ],
+    "node": [
+      {
+        "id": 1,
+        "u_rated": 10500
```
petersalemink95, line 33: Used to be `10.5e3`, which seems more readable to me (especially for other values like `p_specified` -> see below)

tests/data/power_flow/deprecated_format/dummy-test-batch-dependent-not-cacheable/sym_output.json
```diff
@@ -0,0 +1,18 @@
...
+      },
+      {
+        "id": 2,
+        "u": 10
+      }
+    ]
+  },
+  "is_batch": false,
+  "type": "sym_output",
+  "version": "1.0"
```
petersalemink95, line 17: Does the deprecated format have a version?

tests/data/power_flow/1os2msr/input.json
mgovers: i can try to fix the test data if you think it's important

tests/data/power_flow/1os2msr/input.json
mgovers: not sure we can do that unfortunately... `msgpack` decides for us. if you think it's important, i will have to manually go over all test files or write a separate script that does this

tests/data/power_flow/1os2msr/input.json
TonyXiang8787: JSON does not define the order in the dictionary. So there is no way of specifying the order officially.

tests/data/power_flow/1os2msr/input.json
TonyXiang8787: We can write a simple script to do string concatenation to transform existing json files.

mgovers: yes that's what i'm doing now after a quick discussion with peter

TonyXiang8787:
@mgovers 

I still do not understand why we need a `deprecated` folder with old formats.

mgovers:
> @mgovers
> 
> I still do not understand why we need a `deprecated` folder with old formats.

Because there could already be people (ab)using the original format and we want evidence that that at still works at the very least until everything is merged into `main`.

my proposal would be to give a heads up at the next PGM meetup that we will remove that old functionality

TonyXiang8787:
> > @mgovers
> > I still do not understand why we need a `deprecated` folder with old formats.
> 
> Because there could already be people (ab)using the original format and we want evidence that that at still works at the very least until everything is merged into `main`.
> 
> my proposal would be to give a heads up at the next PGM meetup that we will remove that old functionality

We do not need this in the validation test right?

We can create a small python unit test with the old format.

mgovers:
> > > @mgovers
> > > I still do not understand why we need a `deprecated` folder with old formats.
> > 
> > 
> > Because there could already be people (ab)using the original format and we want evidence that that at still works at the very least until everything is merged into `main`.
> > my proposal would be to give a heads up at the next PGM meetup that we will remove that old functionality
> 
> We do not need this in the validation test right?
> 
> We can create a small python unit test with the old format. To test the older `import_json` function.

We do but it also depends on whether you interpret "validation" as "the values are correct" or "this is an end-to-end integration test". In the first case, I would agree we don't necessarily need them, but in the second, I would argue that this is part of the workflow - at least for the upcoming months.

Maybe we should define the goal of validation tests better within the team.

Do you really strongly oppose the deprecation tests in the validation tests? c.c. @petersalemink95 any thoughts?

TonyXiang8787:
@mgovers I do strongly oppose the idea of putting deprecated json files in the validation test. This makes test scripts to handle two format and is also confusing for anyone who wants to contribute new validation test.

You have a good point to test the end-to-end flow of the old format to make sure they are still working. We can write a separate test for this deprecated workflow with only one dataset.

mgovers:
> @mgovers I do strongly oppose the idea of putting deprecated json files in the validation test. This makes test scripts to handle two format and is also confusing for anyone who wants to contribute new validation test.
> 
> You have a good point to test the end-to-end flow of the old format to make sure they are still working. We can write a separate test for this deprecated workflow with only one dataset in a separate folder.

ok. i will create a mixed sparse/dense update test case under a new pytest part `tests/unit/deprecated/end_to_end.py`

mgovers:
> > @mgovers I do strongly oppose the idea of putting deprecated json files in the validation test. This makes test scripts to handle two format and is also confusing for anyone who wants to contribute new validation test.
> > You have a good point to test the end-to-end flow of the old format to make sure they are still working. We can write a separate test for this deprecated workflow with only one dataset in a separate folder.
> 
> ok. i will create a mixed sparse/dense update test case under a new pytest part `tests/unit/deprecated/end_to_end.py`

proposed solution in https://github.com/PowerGridModel/power-grid-model/pull/385 . Not sure if that's the right way going forward though

PR #384, opened 2023-09-21
Handle infinity in serialization

Laurynas-Jagutis:
**Fixes infinity handling in issue**: https://github.com/PowerGridModel/power-grid-model/issues/320

"inf", "+inf" and "-inf" will be serialized


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -27,6 +28,28 @@ namespace power_grid_model::meta_data {
...
 // NOLINTEND(cppcoreguidelines-pro-type-union-access)
+
+constexpr void handle_inf(msgpack::object const& obj, auto& attr) {
+    using namespace msgpack;
+    using namespace std::string_view_literals;
+
+    auto infinity = std::numeric_limits<double>::infinity();
+    if (obj.type == type::STR) {
+        auto const& obj_string = obj.as<std::string_view>();
+        if (obj_string == "inf" || obj_string == "-inf" || obj_string == "+inf") {
```
mgovers, line 39: this is redundant. please directly check `if (... "inf" || ... "+inf") { ... } else if (... "-inf") { ... } else { throw ... }`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -27,6 +28,28 @@ namespace power_grid_model::meta_data {
...
+            if (obj_string == "inf" || obj_string == "+inf") {
+                attr = infinity;
+            } else {
+                attr = -infinity;
+            }
+        } else {
+            throw type_error();
+        }
+    } else {
+        obj >> attr;
```
mgovers, line 49: if you use the `return result` approach, you can directly `return infinity` vs `return -infinity` and you don't need as much nesting.

even more so if you reverse the order `if (obj.type != type::STR) { // unpack and return float }` there's no nesting or `else` statements at all anymore

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -421,7 +446,14 @@ class Deserializer {
         }
         // call relevant parser
         ctype_func_selector(attribute.ctype, [element_pointer, &obj, &attribute]<class T> {
-            obj >> attribute.get_attribute<T>(element_pointer);
+            using namespace std::string_view_literals;
+            using namespace std::string_literals;
```
mgovers, line 450: can be removed

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -421,7 +446,14 @@ class Deserializer {
...
+            auto& attr = attribute.get_attribute<T>(element_pointer);
+            if constexpr (std::floating_point<T>) {
+                handle_inf(obj, attr);
+            } else {
+                obj >> attr;
+            }
```
mgovers, line 456: :+1:

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
Laurynas-Jagutis: I can't really visualize this suggestion, so I didn't choose the return option

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -19,6 +19,20 @@
...
+    using namespace power_grid_model::meta_data;
+    constexpr auto infinity = "inf"sv;
+    constexpr auto neg_infinity = "-inf"sv;
+
+    if (std::isinf(attr)) {
+        packer_inf.pack(attr > 0 ? infinity : neg_infinity);
+    } else {
+        packer_inf.pack(attr);
+    }
+}
```
mgovers, line 34:
```suggestion
constexpr void pack_inf(auto& packer, auto& attr) {
    using namespace msgpack;
    using namespace std::string_view_literals;
    using namespace power_grid_model::meta_data;
    constexpr auto infinity = "inf"sv;
    constexpr auto neg_infinity = "-inf"sv;

    if (std::isinf(attr)) {
        packer.pack(attr > 0 ? infinity : neg_infinity);
    } else {
        packer.pack(attr);
    }
}
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -41,12 +55,14 @@ MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
         template <typename Stream>
         msgpack::packer<Stream>& operator()(msgpack::packer<Stream>& p,
                                             power_grid_model::RealValue<false> const& o) const {
+            using namespace std::string_view_literals;
```
mgovers, line 58: this one is now redundant
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -19,6 +19,20 @@
 #include <span>
 #include <string_view>
 
+constexpr void pack_inf(auto& packer_inf, auto& attr) {
+    using namespace msgpack;
+    using namespace std::string_view_literals;
+    using namespace power_grid_model::meta_data;
```
mgovers, line 25: not sure but i think both of these are now redundant

```suggestion
    using namespace std::string_view_literals;
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -19,6 +19,20 @@
 #include <span>
 #include <string_view>
 
+constexpr void pack_inf(auto& packer_inf, auto& attr) {
```
mgovers, line 22: since this function is quite specific, please be explicit

```suggestion
constexpr void pack_inf(msgpack::packer<Stream>& p packer_inf, std::floating_point auto& attr) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -27,6 +28,26 @@ namespace power_grid_model::meta_data {
...
+constexpr void unpack(msgpack::object const& obj, std::floating_point auto& attr) {
+    using namespace msgpack;
+    using namespace std::string_view_literals;
+
+    auto infinity = std::numeric_limits<double>::infinity();
+    if (obj.type == type::STR) {
+        auto const& obj_string = obj.as<std::string_view>();
+        if (obj_string == "inf" || obj_string == "+inf") {
+            attr = infinity;
+        } else if (obj_string == "-inf") {
```
mgovers, line 41:
```suggestion
        if (obj_string == "inf"sv || obj_string == "+inf"sv) {
            attr = infinity;
        } else if (obj_string == "-inf"sv) {
```

or remove `using namespace std::string_view_literals;`. there's no point to say you're `using` it if you don't use it 😬

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -27,6 +28,26 @@ namespace power_grid_model::meta_data {
...
+            attr = -infinity;
+        } else {
+            throw type_error();
+        }
+    } else {
+        obj >> attr;
+    }
+}
```
mgovers, line 49: as i said in one of my previous comments, please switch the logic around: default should be parsing the floating point, fallback should be parsing infinity

please also consider using a return statement rather than passing by reference. especially for floating point types, it's actually more efficient to pass by value than to pass by reference

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -27,6 +28,26 @@ namespace power_grid_model::meta_data {
 constexpr auto const& as_array(msgpack::object const& obj) { return obj.via.array; }
 constexpr auto const& as_map(msgpack::object const& obj) { return obj.via.map; }
 // NOLINTEND(cppcoreguidelines-pro-type-union-access)
+
+constexpr void unpack(msgpack::object const& obj, std::floating_point auto& attr) {
+    using namespace msgpack;
+    using namespace std::string_view_literals;
+
+    auto infinity = std::numeric_limits<double>::infinity();
```
mgovers, line 36:
```suggestion
    constexpr auto infinity = std::numeric_limits<double>::infinity();
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: that would be something of the order of

```cpp
template <std::floating_point Float>
auto unpack(msgpack::object<Stream>& obj) -> Float {
  if (obj.type != msgpack::type::STR) {
    return obj.as<Float>();
  }

  auto const value = obj.as<std::string_view>();
  if (value == "inf"sv || value == "+inf"sv) {
    return infinity;
  }
  if (value == "-inf"sv) {
    return -infinity;
  }

  throw msgpack::type_error();
}

// usage
{
  double attr = unpack<double>(obj);
}
```

mgovers: that way, there's no nested if/else statements and the logic flow.

small sidenote: you unfortunately cannot use `msgpack::type::FLOAT` because there's ambiguity between `msgpack::type::FLOAT`, `msgpack::type::FLOAT32` and `msgpack::type::FLOAT64`. the actual floating point encoding should not affect the floating point value (which may have a different value type)

TonyXiang8787:
Closed in favor of #387

PR #402, opened 2023-10-10
Replace indptr with sparse and dense grouped index vector

mgovers:
Replace indptr with `SparseIdxVector`

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -100,10 +100,10 @@ struct MathModelTopology {
     SparseGroupedIdxVector voltage_sensors_per_bus;
     SparseGroupedIdxVector power_sensors_per_source;
     SparseGroupedIdxVector power_sensors_per_load_gen;
-    IdxVector shunt_power_sensor_indptr;       // indptr of the shunt
-    IdxVector branch_from_power_sensor_indptr; // indptr of the branch
-    IdxVector branch_to_power_sensor_indptr;   // indptr of the branch
-    IdxVector bus_power_sensor_indptr;         // indptr of the bus
+    SparseGroupedIdxVector power_sensors_per_shunt;
+    SparseGroupedIdxVector power_sensors_per_branch_from; // indptr of the branch
+    SparseGroupedIdxVector power_sensors_per_branch_to;   // indptr of the branch
+    IdxVector bus_power_sensor_indptr;                    // indptr of the bus
 
     Idx n_bus() const { return static_cast<Idx>(phase_shift.size()); }
 
@@ -121,11 +121,11 @@ struct MathModelTopology {
 
     Idx n_load_gen_power_sensor() const { return power_sensors_per_load_gen.element_size(); }
 
-    Idx n_shunt_power_power_sensor() const { return shunt_power_sensor_indptr.back(); }
+    Idx n_shunt_power_power_sensor() const { return power_sensors_per_shunt.element_size(); }
 
-    Idx n_branch_from_power_sensor() const { return branch_from_power_sensor_indptr.back(); }
+    Idx n_branch_from_power_sensor() const { return power_sensors_per_branch_from.element_size(); }
 
-    Idx n_branch_to_power_sensor() const { return branch_to_power_sensor_indptr.back(); }
+    Idx n_branch_to_power_sensor() const { return power_sensors_per_branch_to.element_size(); }
 
     Idx n_bus_power_sensor() const { return bus_power_sensor_indptr.back(); }
 };
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -97,7 +97,7 @@ struct MathModelTopology {
     SparseGroupedIdxVector shunt_buses;
     SparseGroupedIdxVector load_gen_buses;
     std::vector<LoadGenType> load_gen_type;
-    IdxVector voltage_sensor_indptr;
+    SparseGroupedIdxVector voltage_sensor_buses;
     IdxVector source_power_sensor_indptr;      // indptr of the source
     IdxVector load_gen_power_sensor_indptr;    // indptr of the load_gen
     IdxVector shunt_power_sensor_indptr;       // indptr of the shunt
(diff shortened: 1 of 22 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -92,7 +93,7 @@ struct MathModelTopology {
     std::vector<double> phase_shift;
     std::vector<BranchIdx> branch_bus_idx;
     std::vector<BranchIdx> fill_in;
-    IdxVector source_bus_indptr;
+    SparseGroupedIdxVector source_buses; // TODO(mgovers) replace with DenseGroupedIdxVector
```
mgovers, line 96: maybe we want to make abstraction of whether it is sparse or dense. see also

https://github.com/PowerGridModel/power-grid-model/pull/379#discussion_r1354195358

tests/cpp_unit_tests/test_typing.cpp
```diff
@@ -0,0 +1,66 @@
...
+    SUBCASE("Different types (in range)") {
+        check_narrow_cast<Idx, ID>();
+        check_narrow_cast<ID, Idx>();
+        check_narrow_cast<uint8_t, Idx>();
+        check_narrow_cast<Idx, uint8_t>();
+        check_narrow_cast<uint8_t, ID>();
+        check_narrow_cast<ID, uint8_t>();
+
+        check_narrow_cast<long, long long>();
+        check_narrow_cast<long long, long>();
```
mgovers, line 63:
```suggestion
        check_narrow_cast<uint64_t>();
    }

    SUBCASE("Different types (in range)") {
        check_narrow_cast<Idx, ID>();
        check_narrow_cast<ID, Idx>();
        check_narrow_cast<uint8_t, Idx>();
        check_narrow_cast<Idx, uint8_t>();
        check_narrow_cast<uint8_t, ID>();
        check_narrow_cast<ID, uint8_t>();

        check_narrow_cast<long, uint64_t>();
        check_narrow_cast<uint64_t, long>();
```

TonyXiang8787:
@mgovers this PR needs more attention. Do not blindly change every `indptr` to Sparse mapping. Some of them should actually be dense mapping. This involves the physical knowledge to see what makes sense for a typical power grid.

mgovers:
> @mgovers this PR needs more attention. Do not blindly change every `indptr` to Sparse mapping. Some of them should actually be dense mapping. This involves the physical knowledge to see what makes sense for a typical power grid.

Yes i agree. But i need your (or someone else's) input for that. Hence also my comment on the common data type.

I made sure to only ever use `auto`, so changing the classes is almost trivial. Changing the input data type is a littlebit more trucky, because it requires modifying e.g test data. But i strongly feel that this PR should only change classes, not test data, to separate dangerous refactorings (it's easy to make a mistake)

mgovers:
> > @mgovers this PR needs more attention. Do not blindly change every `indptr` to Sparse mapping. Some of them should actually be dense mapping. This involves the physical knowledge to see what makes sense for a typical power grid.
> 
> Yes i agree. But i need your (or someone else's) input for that. Hence also my comment on the common data type.
> 
>     * I made sure to only ever use `auto`, so changing the classes is almost trivial. The `from_sparse` tag ensures that no constructor input needs to be changed for that change.
> 
>     * Changing the input data type is a little bit more tricky, because it requires modifying e.g test data. But i strongly feel that this PR should only change classes, not test data, to separate dangerous refactorings (it's easy to make a mistake)

We agreed during standup that this is actually how we should approach this PR to prevent too many changes from happening in one go

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
TonyXiang8787: I am not in favor of determine dense/sparse mapping dynamically. It incurs runtime cost. We need to use power system prior knowledge to determine which mapping is dense or sparse at compile time.

mgovers: ok fair enough.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -92,41 +93,41 @@ struct MathModelTopology {
...
-    IdxVector branch_from_power_sensor_indptr; // indptr of the branch
-    IdxVector branch_to_power_sensor_indptr;   // indptr of the branch
-    IdxVector bus_power_sensor_indptr;         // indptr of the bus
+    SparseGroupedIdxVector voltage_sensors_per_bus;
+    SparseGroupedIdxVector power_sensors_per_source;
+    SparseGroupedIdxVector power_sensors_per_load_gen;
+    SparseGroupedIdxVector power_sensors_per_shunt;
+    SparseGroupedIdxVector power_sensors_per_branch_from;
+    SparseGroupedIdxVector power_sensors_per_branch_to;
+    SparseGroupedIdxVector power_sensors_per_bus;
```
mgovers, line 106: check which ones need to be `DenseGroupedIdxVector` and which ones need to be `SparseGroupedIdxVector`

nitbharambe: In the grids that I see, shunts are never present.
All the sensors are never multiple at each `measured_object`. Hence it can be DenseGroupedIdxVector.

I guess this needs a brainstorm though. There might be application cases in future where this is not so.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -314,7 +315,7 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
                     // should always found
                     assert(found != col_indices.cbegin() + row_indptr[l_row + 1]);
                     assert(*found == u_col);
-                    a_idx = (Idx)std::distance(col_indices.cbegin(), found);
+                    a_idx = narrow_cast<Idx>(std::distance(col_indices.cbegin(), found));
```
nitbharambe, line 318: Why wont static cast be applicable here?

mgovers: because depending on the OS and architecture, it may be a useless cast and `clang` and `gcc` will complain about that. E.g.: `static_cast<int32_t>(int32_t{})`

In particular, MacOS had complaints about the result type of `std::distance`

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: cfr. discussion during standup: out of scope of this PR

nitbharambe (approved):

PR #403, opened 2023-10-13
Feature/full clang tidy after indptrs

mgovers:
* [x] Fix `clang-cl` warning
* [x] Force `clang-tidy` on Windows to use `clang-cl` as basic compiler back-end
* [x] Full build with `clang-tidy` on local device
* [x] Fix any new issues for `clang-tidy-16` not captured by `clang-tidy-15` (only locally verified; not run in CI)
* [x] Full run `clang-tidy` workflow in CI (in https://github.com/PowerGridModel/power-grid-model/actions/runs/6546516397 )
  * [x] ~Repro + report bug in `msgpack`~ moved to #404


tests/c_api_tests/test_c_api_model.cpp
```diff
@@ -46,7 +46,7 @@ TEST_CASE("C API Model") {
     PGM_Options* opt = unique_options.get();
 
     // input data
-    ConstDatasetPtr const unique_input_dataset{PGM_create_dataset_const(hl, "input", false, 1)};
+    ConstDatasetPtr const unique_input_dataset{PGM_create_dataset_const(hl, "input", 0, 1)};
```
petersalemink95, line 49: Why is true/false not correct?

power_grid_model_c/power_grid_model/include/power_grid_model/container.hpp
```diff
@@ -319,12 +319,16 @@ class Container<RetrievableTypes<GettableTypes...>, StorageableTypes...> {
       public:
         explicit Proxy(container_type& container)
             : begin_{&container, 0}, end_{&container, container.template size<base_type>()} {}
-        Iterator<Gettable> begin() { return begin_; }
-        Iterator<Gettable> end() { return end_; }
+
+        auto begin() -> Iterator<Gettable> { return begin_; }
```
petersalemink95, line 323: Why do we need trailing return types here?

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
```diff
@@ -495,41 +495,61 @@ class Topology {
...
     // proxy class to find the coupled object in math model in the coupling process to a single type object
     //    given a particular component index
-    struct SingleTypeObjectFinder {
-        Idx size() const { return static_cast<Idx>(component_obj_idx.size()); }
-        Idx2D find_math_object(Idx component_i) const { return objects_coupling[component_obj_idx[component_i]]; }
-        IdxVector const& component_obj_idx;
-        std::vector<Idx2D> const& objects_coupling;
+    class SingleTypeObjectFinder {
+      public:
+        SingleTypeObjectFinder(IdxVector const& component_obj_idx, std::vector<Idx2D> const& objects_coupling)
```
petersalemink95, line 500: Why didn't we have a constructor in the first place? Was it constructed somewhere else?

petersalemink95:
Looks good to me, had some small questions (mainly for my own understanding)

TonyXiang8787:
@mgovers is this something we need for now? clang-tidy already checks all the core part (everything used by the C-API) right?

tests/c_api_tests/test_c_api_model.cpp
mgovers: implicit casts are often dangerous, especially when the types don't map to eachother bijectionally. In this case, there's a many-to-one relation (surjection) from all nonzero integer values to `true` (including negative values). To avoid confusion and to make it explicit, it's recommended to either use `static_cast` every single time, OR to use integer values.

Although C++ guarantees that `true == 1`, implicit casting may e.g. result in `5 -> true -> 1`, which may result in unexpected behaviour

power_grid_model_c/power_grid_model/include/power_grid_model/container.hpp
mgovers: we don't necessarily need it but it keeps the function names lined up

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
mgovers: initially, this was a `struct` that would be constructed with a `IdxVector const&` to `component_obj_idx`.

The problem is, that this breaks `std::movable` behaviour (`const` and `&` member fields cause `move` constructor to be implicitly deleted, because `const` and l-value `&` objects can't be moved, only copied)

Just replacing `const&` with `const*` would fix this, BUT it would imply that a `nullptr` check would have to be done at every function execution (remember that `struct` by convention implies that the objects of that type do not have so-called invariants, meaning that not all combinations of member variables are valid, while for `class`, some values/combinations of member variables may be invalid by design).

In this case, we do not want to check for `nullptr` every executable, so I:
* made this `struct` a `class`
* added a constructor that takes in a `const&` that ensures that `nullptr` will never be set as a value
* made the members `private` instead of `public`
This ensures that the variable will never ever have a `nullptr` and therefore, the `nullptr` check is not necessary

mgovers:
> @mgovers is this something we need for now? clang-tidy already checks all the core part (everything used by the C-API) right?

i found some performance issues while implementing the other PR that were not caught by the CI version of `clang-tidy` (on `model.cpp`). I don't know why, but doing a full `clang-tidy` run on all files actually did find the issue

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
petersalemink95: Thanks! I missed the fact that it was a struct before.

petersalemink95 (approved):

Issue #369, opened 2023-09-04
Newton-Raphson method for state estimation


TonyXiang8787:
Newton-Raphson method should solve the issue in #237

mgovers:
Update 2024/02/08:

Since #438 , Newton-Raphson for state estimation is now in beta as an experimental feature.

Because we do not officially support it yet, nor guarantee its correctness, it is not yet possible to call this feature via the regular API's [`PowerGridModel.calculate_state_estimation`](https://power-grid-model.readthedocs.io/en/stable/api_reference/python-api-reference.html#power_grid_model.PowerGridModel.calculate_state_estimation)).

Instead, call the protected method `PowerGridModel._calculate_state_estimation` with as arguments:
* `calculation_method=CalculationMethod.newton_raphson`
* `experimental_features=power_grid_model.enum._ExperimentalFeatures.enabled`
* any other arguments which are the same arguments as for the iterative linear case.

E.g. someone who would usually call

```py
from power_grid_model import PowerGridModel
from power_grid_model.enum import CalculationMethod

# create input data
model = PowerGridModel(input_data)
model.calculate_state_estimation(calculation_method=CalculationMethod.iterative_linear)
```

may run the Newton-Raphson state estimation by running

```py
from power_grid_model import PowerGridModel
from power_grid_model.enum import CalculationMethod, ExperimentalFeatures_

# create input data
model = PowerGridModel(input_data)
model.calculate_state_estimation(calculation_method=CalculationMethod.newton_raphson, experimental_features=ExperimentalFeatures_.enabled)
```

Issue #237, opened 2023-04-12
[FEATURE] Possibility to specify different sigma's for P and Q in state estimator


marionsnijders:
Within team Insight into Loads we apply the state estimator to the MS grid. We see an issue in our output related to the restriction (if we understand correctly that this is the case) of one sigma for both the P and Q of a (pseudo-)measurement. We have the following inputs:

- MSRs (loads) containing lots of households: those will have a larger margin of uncertainty for P due to sampling of anonymous user profiles.
- MSRs containing just one large consumer (AC5): those will have a very small margin of uncertainty for P due to an available measurement
- Measurement of P and Q at the substation (afgaande velden): both P and Q have a small margin of uncertainty here.

In general, we expect a low Q for households and a possibly higher Q for the larger customers. Unfortunately, because we specify one sigma for P and Q, the reactive power is largely attributed to the households now whenever the measurement at the substation indicates that there is reactive power on the route. This makes the output of such routes/grids invalid for our application. Would it be possible to specify two different sigma's within one (pseudo-)measurement within the state estimator, so we can start influencing the attribution of Q more accurate?

TonyXiang8787:
Decisions:

1. Expand the current iterative method to decouple the real and imaginary part. So it can handle different sigmas.
2. In the power sensor, add a new optional attribute `reactive_power_sigma`. If it is specified, `power_sigma` becomes only for active power. Otherwise, `power_sigma`  is for complex power.

mgovers:
Update 2023/11/29:

* Since #409 , the user may already use `sigma_p` and `sigma_q` to override `power_sigma`.
  * For certain systems, this will already provide some improvement over using `power_sigma` by itself.
  * See https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html#id28 on details, input requirements, ... .
  * See https://power-grid-model.readthedocs.io/en/stable/user_manual/calculations.html#state-estimation-algorithms on details on the treatment for the implemented calculation type(s)
* Implementation work on #369 has started to add the Newton-Raphson calculation method for state estimation
  * This should unlock the full potential of separating `sigma_p` and `sigma_q`

mgovers:
Update 2024/02/08:

#369 is now in beta. For details, please have a look at https://github.com/PowerGridModel/power-grid-model/issues/369#issuecomment-1934542251

PR #409, opened 2023-10-19
add reactive_power_sigma attribute

mgovers:
Related to #237 
Related to #369 

# Changes

- [x] Add new `RealValue` attributes `p_sigma` and `q_sigma` to power sensor input and update. These are optional attributes.
  - If `p_sigma` and `q_sigma` are both `NaN`, the `power_sigma` is the apparent power sigma and is required (current behaviour) for state estimation.
  - If `p_sigma` and `q_sigma` are both not `NaN`, `p_sigma` is the active power sigma and `q_sigma` is the reactive power sigma.
  - It is not supported to provide only one of `p_sigma` and `q_sigma`, just like how it is not supported to only provide 1 of three phases of an asymmetric `RealValue`. **The behavior is undefined in the core. In the validation function in Python side we raise an error.**
  - [x] Update data definition
  - [x] Update documentation
- [x] Create a new struct in `calculation_parameters.hpp` with `PowerSensorCalcParam`, `VoltageSensorCalcParam`, as the calculation parameter output of relevant sensor components. In `VoltageSensorCalcParam` is the same as old `SensorCalcParam`.  `PowerSensorCalcParam` should have `variance_p` and `variance_q`, these are also `RealValue`. So it is three phase for asymmetric calculation.
- [x] Change the sensor component to store `power_sigma`, `p_sigma` and `q_sigma` as member data. Store them as per unit value as it is now. **Note: the base power for `sym_power_sensor` and `asym_power_sensor` is different.** Change the way to generate `PowerSensorCalcParam`:
  - [x] If `p_sigma` is `NaN`: Set `variance_p = variance_q = power_sigma**2 / 2`
    - [x] For asymmetric calculation (for both `sym_power_sensor` and `asym_power_sensor`), `variance_p = variance_q = power_sigma**2 / 2 * [1, 1, 1]`
  - [x] If `p_sigma` is not `NaN`: Set `variance_p = p_sigma**2` and `variance_q = q_sigma**2`
    - [x] For asymmetric calculation of `sym_power_sensor`, `variance_p = p_sigma**2 * [1, 1, 1]`, same as Q.
    - [x] For symmetric calculation of `asym_power_sensor`, `variance_p = mean(p_sigma**2)`, same as Q.
  - [x] If either `p_sigma` is provided and `q_sigma` is not, or vice versa. The behaviour is undefined. That's why we only check `p_sigma` in the above two cases.
  - [x] If either `p_sigma` or `q_sigma` is `inf`, set both `variance_p` and `variance_q` to `inf`.
  - [x] If `p_sigma` is not provided and `power_sigma` is `inf`, set both `variance_p` and `variance_q` to `inf`.
- [x] Change the `MeasuredValue` logic, to split voltage and power measurement from one single `main_value`. For power measurement, aggregate and deaggregate P/Q separately based on `variance_p` and `variance_q`.
- [x] For the current iterative linear solver, for a power measurement, the weighting factor is `1/(variance_p + variance_q)`. For asymmetric calculation, this is a diagonal matrix.


docs/_static/css/custom.css
```diff
@@ -2,4 +2,7 @@
 
 .wy-nav-content {
     max-width: none;
-}
\ No newline at end of file
+}
+.wy-table-responsive table td, .wy-table-responsive table th {
+    white-space: inherit;
+}
```
mgovers, line 8: set table width dynamically to fit to screen width. See e.g.: the Components docs: https://power-grid-model--409.org.readthedocs.build/en/409/user_manual/components.html

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -487,14 +487,11 @@ template <bool sym> class MeasuredValues {
...
+            accumulated_inverse_p_variance += RealValue<sym>{1.0} / measurement.p_variance;
+            accumulated_inverse_q_variance += RealValue<sym>{1.0} / measurement.q_variance;
 
-            accumulated_inverse_p_variance += inv_p_variance;
-            accumulated_inverse_q_variance += inv_q_variance;
-
-            accumulated_p_value += real(measurement.value) * inv_p_variance;
-            accumulated_q_value += imag(measurement.value) * inv_q_variance;
+            accumulated_p_value += real(measurement.value) / measurement.p_variance;
+            accumulated_q_value += imag(measurement.value) / measurement.q_variance;
```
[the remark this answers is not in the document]
TonyXiang8787: You cannot declare `auto` of the return value of `eigen` expression. `eigen` expression is lazy evaluation, the return value is a proxy object.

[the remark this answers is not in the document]
TonyXiang8787: https://eigen.tuxfamily.org/dox/TopicPitfalls.html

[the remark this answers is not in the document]
mgovers: yep i realized. in hindsight i knew this, but i had forgotten about it

mgovers:
@TonyXiang8787 this is ready for review. Remaining sonar cloud code smells are not caused by this PR

tests/cpp_unit_tests/test_power_sensor.cpp
```diff
@@ -101,13 +113,29 @@ TEST_CASE("Test power sensor") {
...
             CHECK(real(sym_sensor_param.value) == doctest::Approx(3.0 * 1e-3));
             CHECK(imag(sym_sensor_param.value) == doctest::Approx(4.0 * 1e-3));
 
-            CHECK(asym_sensor_param.variance == doctest::Approx(4.0 / 1e2));
+            CHECK(asym_sensor_param.p_variance[0] == doctest::Approx(4.0 / 1e2 / 2));
+            CHECK(asym_sensor_param.q_variance[1] == doctest::Approx(4.0 / 1e2 / 2));
             CHECK(real(asym_sensor_param.value[0]) == doctest::Approx(3.0 * 1e-3));
             CHECK(imag(asym_sensor_param.value[1]) == doctest::Approx(4.0 * 1e-3));
+
+            // Check update with different p and q sigma variances
```
mgovers, line 126: here and below, i did not add any tests for infinite sigmas cause I deemed that trivial. I can add them if you like

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -15,8 +15,8 @@ namespace power_grid_model {
 
 // Entry of YBus, node addmittance matrix
 struct YBusElement {
-    YBusElementType element_type;
-    Idx idx; // index of the component
+    YBusElementType element_type{};
+    Idx idx{}; // index of the component
```
TonyXiang8787, line 19: It this desired? If you make default value for a `struct`, it is not aggregate type anymore. So you cannot do

```c++
YBusElement element{.element_type = 1, .idx=5};
```

power_grid_model_c/power_grid_model/include/power_grid_model/component/sensor.hpp
```diff
@@ -17,12 +17,19 @@
 
 namespace power_grid_model {
 
+template <template <bool symmetric> typename SensorCalcParam>
+    requires sensor_calc_param_type<SensorCalcParam<true>> && sensor_calc_param_type<SensorCalcParam<false>>
```
TonyXiang8787, line 21: Do we really need CRTP here? Mixing CRTP and virtual functions is pretty confusing.

I see the reason you use CRTP is now we have different return type for `sym_calc_param` and `asym_calc_param`. I would propose:

1. remove these two pure virtual functions at `Sensor`.
2. define new pure virtual functions at `GenericVoltageSensor` and `GenericPowerSensor`. At that time the return type is known.

tests/cpp_unit_tests/test_power_sensor.cpp
```diff
@@ -7,6 +7,10 @@
 #include <doctest/doctest.h>
 
 namespace power_grid_model {
+namespace {
+const auto r_nan = RealValue<false>{nan};
```
TonyXiang8787, line 11: use east const

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -538,21 +600,27 @@ template <bool sym> class MeasuredValues {
...
                                              FlowVector& source_flow) const {
         // residual normalized by variance
         // mu = (sum[S_i] - S_cal) / sum[variance]
-        ComplexValue<sym> const mu = (bus_appliance_injection.value - s) / bus_appliance_injection.variance;
+        ComplexValue<sym> const mu = (bus_appliance_injection.value - s) /
+                                     (bus_appliance_injection.p_variance + bus_appliance_injection.q_variance);
+
         // S_i = S_i_mea - var_i * mu
+        auto const calculate_injection = [&mu](auto const& power) {
+            return power.value - (power.p_variance + power.q_variance) * mu;
```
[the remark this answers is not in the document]
TonyXiang8787: Indeed here P and Q should be calculated separately. Please make this change within this PR. This is in the scope

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -695,11 +765,13 @@ template <bool sym> class IterativeLinearSESolver {
                         for (IntS const measured_side : std::array<IntS, 2>{0, 1}) {
                             // has measurement
                             if (std::invoke(has_branch_[measured_side], measured_value, obj)) {
-                                // G += Y{side, b0}^H * Y{side, b1} / variance
+                                // G += Y{side, b0}^H * Y{side, b1} * (variance^-1)
```
TonyXiang8787, line 768: This should be `Y{side, b0}^H * (variance^-1) * Y{side, b1}`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -681,9 +749,11 @@ template <bool sym> class IterativeLinearSESolver {
                     // shunt
                     if (type == YBusElementType::shunt) {
                         if (measured_value.has_shunt(obj)) {
-                            // G += Ys^H * Ys / variance
-                            block.g() += dot(hermitian_transpose(param.shunt_param[obj]), param.shunt_param[obj]) /
-                                         measured_value.shunt_power(obj).variance;
+                            // G += Ys^H * Ys * (variance^-1)
```
TonyXiang8787, line 752: same here
`Ys^H * (variance^-1) * Ys`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -681,9 +749,11 @@ template <bool sym> class IterativeLinearSESolver {
...
+                            auto const& shunt_power = measured_value.shunt_power(obj);
+                            block.g() += dot(hermitian_transpose(param.shunt_param[obj]), param.shunt_param[obj],
+                                             inv(ComplexTensor<sym>{static_cast<ComplexValue<sym>>(
```
TonyXiang8787, line 755: Do not use inverse. This is a known diagonal matrix. 

(
Comment
Do not use inverse. This is a known diagonal matrix. Use something like:

```
to_diag(1.0 / (shunt_power.p_variance + shunt_power.q_variance))
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -695,11 +765,13 @@ template <bool sym> class IterativeLinearSESolver {
...
-                                // G += Y{side, b0}^H * Y{side, b1} / variance
+                                // G += Y{side, b0}^H * Y{side, b1} * (variance^-1)
+                                auto const& power = std::invoke(branch_power_[measured_side], measured_value, obj);
                                 block.g() +=
                                     dot(hermitian_transpose(param.branch_param[obj].value[measured_side * 2 + b0]),
-                                        param.branch_param[obj].value[measured_side * 2 + b1]) /
-                                    std::invoke(branch_power_[measured_side], measured_value, obj).variance;
+                                        param.branch_param[obj].value[measured_side * 2 + b1],
+                                        inv(ComplexTensor<sym>{
+                                            static_cast<ComplexValue<sym>>(power.p_variance + power.q_variance)}));
```
TonyXiang8787, line 774: Same here

```
to_diag(1.0 / (power.p_variance + power.q_variance))
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -768,10 +842,10 @@ template <bool sym> class IterativeLinearSESolver {
                 // shunt
                 if (type == YBusElementType::shunt) {
                     if (measured_value.has_shunt(obj)) {
-                        SensorCalcParam<sym> const& m = measured_value.shunt_power(obj);
+                        PowerSensorCalcParam<sym> const& m = measured_value.shunt_power(obj);
                         // eta -= Ys^H * i_shunt / variance
```
TonyXiang8787, line 846: This should be

`(variance)^-1 * Ys^H * i_shunt`

TonyXiang8787: Oh, for this one it is actually the same. `Ys^H * i_shunt` and `variance` are both vectors, so the order of multiplication does not matter.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -792,7 +866,7 @@ template <bool sym> class IterativeLinearSESolver {
                             rhs_block.eta() +=
```
TonyXiang8787, line 866: Same here
 But the order does not matter. Maybe at list specify the formula in the comment in the correct order.

power_grid_model_c/power_grid_model/include/power_grid_model/component/power_sensor.hpp
```diff
@@ -86,15 +107,31 @@ template <bool sym> class PowerSensor : public GenericPowerSensor {
...
+        if (is_normal(p_sigma_) && is_normal(q_sigma_)) {
+            calc_param.p_variance = mean_val(p_sigma_ * p_sigma_);
+            calc_param.q_variance = mean_val(q_sigma_ * q_sigma_);
+        } else {
+            auto const variance = is_nan(p_sigma_) ? apparent_power_sigma_ * apparent_power_sigma_ / 2
+                                                   : std::numeric_limits<double>::infinity();
+            calc_param.p_variance = variance;
+            calc_param.q_variance = variance;
+        }
         calc_param.value = mean_val(s_measured_);
```
TonyXiang8787, line 121: add a debug assertion here to check if `p_sigma` and `q_sigma` are set with one NaN and one not NaN, as it is undefined.

power_grid_model_c/power_grid_model/include/power_grid_model/component/power_sensor.hpp
```diff
@@ -86,15 +107,31 @@ template <bool sym> class PowerSensor : public GenericPowerSensor {
...
+        if (is_normal(p_sigma_) && is_normal(q_sigma_)) {
+            calc_param.p_variance = piecewise_real_value(p_sigma_ * p_sigma_);
+            calc_param.q_variance = piecewise_real_value(q_sigma_ * q_sigma_);
+        } else {
+            auto const variance = RealValue<false>{is_nan(p_sigma_) ? apparent_power_sigma_ * apparent_power_sigma_ / 2
+                                                                    : std::numeric_limits<double>::infinity()};
+            calc_param.p_variance = variance;
+            calc_param.q_variance = variance;
+        }
         calc_param.value = piecewise_complex_value(s_measured_);
```
TonyXiang8787, line 135: add a debug assertion here to check if `p_sigma` and `q_sigma` are set with one NaN and one not NaN, as it is undefined.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: shouldn't matter: https://godbolt.org/z/o814Kff7q

mgovers: in fact, if you don't have a default initializer, you will receive compilation errors when you do not initialize all fields. Try removing `.b = 2` when constructing `A` in https://godbolt.org/z/3Y7c65WEo

The only argument against default initializing `{}` is that it doesn't default initialize to `nan` (or equivalent per type) but we can decide on adding that now

power_grid_model_c/power_grid_model/include/power_grid_model/component/sensor.hpp
mgovers: that's not possible because we need `calc_param` to exist here, including return type. or we need to step away from the notion that all sensors have calculation params. `calc_param` cannot be virtual without a return type. any thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/three_phase_tensor.hpp
```diff
@@ -104,6 +104,19 @@ concept rk2_tensor = (static_cast<Idx>(T::RowsAtCompileTime) ==
 template <class T>
 concept column_vector_or_tensor = column_vector<T> || rk2_tensor<T>;
 
+// piecewise factory construction for complex vector
+template <bool sym = false> inline RealValue<sym> piecewise_real_value(double x) {
```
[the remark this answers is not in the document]
mgovers: if i remember correctly it's to work around unnecessary static cast but will check

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
TonyXiang8787: Okay good. Then leave it as is.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
mgovers: you're right 🤦 how did i not catch this

power_grid_model_c/power_grid_model/include/power_grid_model/component/sensor.hpp
TonyXiang8787: Do we have to require `calc_param` to exist at `Sensor` level?

We only iterate of all `GenericVoltageSensor` or `GenericPowerSensor`. But never all `Sensor`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
mgovers: `Ys^H * (variance)^-1 * i_shunt`, right?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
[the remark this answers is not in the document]
mgovers: if i do that change, all tests fail, so this change would be incomplete.
i'll have a look to see if i can find the issue in a separate branch and make a PR to this branch if i can solve it

power_grid_model_c/power_grid_model/include/power_grid_model/component/sensor.hpp
mgovers: will remove

mgovers: i did the alternative implementation and it introduces a lot of additional rule-of-five logic. imho, it gets more ugly instead of less. also not sure why but running into some linking errors that seem hard to resolve. are you really strongly against this?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
mgovers: eigen::DiagonalWrapper is a rabbithole that i'll need to dive further into. I've tried going down this route before (also relating to how we implement `dot`) but I will try it again

power_grid_model_c/power_grid_model/include/power_grid_model/component/sensor.hpp
mgovers: instead of Sensor being abstract, I went with a solution where I made all constructors and the destructor `protected`, meaning it's a class that is not abstract but it is something not constructible. statically ensuring that it's not `copyable` and also not constructible/copyable from derived classes

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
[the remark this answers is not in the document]
mgovers: this was indeed a bug. fixed in https://github.com/PowerGridModel/power-grid-model/pull/409/commits/bbc430b492a00ebbd73204a80015064713cd0906

mgovers:
remaining code smells either already on main or related to equivalent on main (see e.g. https://sonarcloud.io/project/issues?resolved=false&rules=cpp%3AS1709&id=PowerGridModel_power-grid-model&open=AYm75yP7TgVATTPnBGJv )

mgovers:
@TonyXiang8787 i think i resolved all comments and it's ready for review again

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
TonyXiang8787: FYI, only under C++11, the type will become non-aggregate if you have default initializer for data member. See
https://godbolt.org/z/87fsr18bx

This constraint has been removed since C++14.
https://en.cppreference.com/w/cpp/language/aggregate_initialization

TonyXiang8787 (approved):

PR #412, opened 2023-10-24
Feature/recombine enumerated zip sequence

mgovers:
Resolve comments from knowledge sharing session:

* `++` and `--` instead of `+= 1` and `-= 1`
* zip-like loop over both `sources` and `load_gens` instead of two loops

@TonyXiang8787 I did a benchmark on (Clang CL) release build comparing `main` and this branch and it appears that `main` is almost consistently better. See also the benchmark outputs in [benchmark.zip](https://github.com/PowerGridModel/power-grid-model/files/13111445/benchmark.zip). My guess is that the zipped loop actually introduces more cache misses because there are more variables in the output. Maybe you can try to reproduce this? I think it would be a shame to replace efficient code by something that is less efficient, but I don't want to block this change if we cannot reproduce the regression

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/grouped_index_vector.hpp
@@ -209,13 +209,13 @@ class DenseGroupedIdxVector {
         constexpr auto distance_to(GroupIterator const& other) const { return other.group_ - group_; }
 
         constexpr void increment() {
-            group_ += 1;
+            ++group_;
             group_range_ = std::make_pair(group_range_.second,
                                           std::find_if(group_range_.second, std::cend(*dense_vector_),
                                                        [group = group_](Idx value) { return value > group; }));
         }
         constexpr void decrement() {
-            group_ -= 1;
+            --group_;
             group_range_ =
                 std::make_pair(std::find_if(std::make_reverse_iterator(group_range_.first), std::crend(*dense_vector_),
                                             [group = group_](Idx value) { return value < group; })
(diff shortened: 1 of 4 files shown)


TonyXiang8787:
@mgovers can you do a benchmark on gcc/clang in Linux?

mgovers:
> @mgovers can you do a benchmark on gcc/clang in Linux?

This looks more like it (but keep in mind that this is on WSL, which is a virtual machine, not a native one) [benchmark_wsl.zip](https://github.com/PowerGridModel/power-grid-model/files/13111669/benchmark_wsl.zip)

Just for the sake of completeness, here's also the benchmark for MSVC, which also looks comparable: [benchmark_msvc.zip](https://github.com/PowerGridModel/power-grid-model/files/13111688/benchmark_msvc.zip)

TonyXiang8787:
@mgovers it seems in Linux the benchmark results are comparable between main and this feature. So we could conclude the performance impact is minimal.

It is very interesting to see in WSL clang has significant performance improvements than gcc.

TonyXiang8787:
@mgovers Another observation about current benchmark target. We are compiling it with address sanitizer enabled, which can have significant performance slow-down.

mgovers:
> @mgovers Another observation about current benchmark target. We are compiling it with address sanitizer enabled, which can have significant performance slow-down.

hmm good point. do i need to re-run the benchmarks without address sanitizer? and do we need to overwrite the compiler flags for the benchmark target?

TonyXiang8787:
> > @mgovers Another observation about current benchmark target. We are compiling it with address sanitizer enabled, which can have significant performance slow-down.
> 
> hmm good point. do i need to re-run the benchmarks without address sanitizer? and do we need to overwrite the compiler flags for the benchmark target?

Yes, use clang/gcc in Linux/WSL as the main judgement. After disabling address-sanitizer, if the performance benchmark does not differ very much, merge this PR as is. If `main` is significantly faster (>10%), we need to talk about it.

mgovers:
new benchmark results:

[benchmark_wsl.zip](https://github.com/PowerGridModel/power-grid-model/files/13165759/benchmark_wsl.zip)

mgovers:
i guess if we only compare the fasted runs for the same compiler then the results are comparable

TonyXiang8787 (approved):

PR #416, opened 2023-10-25
replace some std::find by std::ranges::find

mgovers:
replace `std::find` by `std::ranges::find` where possible

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -160,17 +160,25 @@
       ]
     },
     {
+      "name": "clang-base",
+      "environment": {
+        "COMPILER_CXX_FLAGS": "$env{COMPILER_C_FLAGS} -stdlib=libc++"
+      },
+      "inherits": "unix-sanitizer",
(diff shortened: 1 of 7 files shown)


CMakePresets.json
```diff
@@ -159,18 +159,26 @@
...
       ]
     },
+    {
+      "name": "clang-base",
+      "environment": {
+        "COMPILER_CXX_FLAGS": "$env{COMPILER_C_FLAGS} -stdlib=libc++"
+      },
+      "inherits": "unix-sanitizer",
+      "hidden": true
+    },
```
mgovers, line 169: this is not an issue in our repo but an incompatibility in libstdc++

see also https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1918839.html and https://stackoverflow.com/questions/76859275/error-compiling-a-cpp-containing-stdchrono-errorstatic-constexpr-unsigned-fra

PR #417, opened 2023-10-30
fix multiple different code smells

Laurynas-Jagutis:
Fixed:
- Most of "Remove this redundant cast."
- All "Extract this nested conditional operator into an independent statement"

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
@@ -20,8 +20,10 @@
 #include <string_view>
 
 // custom packers
-namespace msgpack::adaptor {
+namespace msgpack {
 MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
+    namespace adaptor {
+
     // pack name of component and attribute
     template <class T>
         requires(std::same_as<T, power_grid_model::meta_data::MetaComponent> ||
@@ -50,9 +52,9 @@ MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
             return p;
         }
     };
-
+    } // namespace adaptor
 } // MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS)
-} // namespace msgpack::adaptor
+} // namespace msgpack
 
 namespace power_grid_model::meta_data {
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
@@ -52,6 +52,7 @@ MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
             return p;
         }
     };
+
     } // namespace adaptor
 } // MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS)
 } // namespace msgpack
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
@@ -20,10 +20,8 @@
 #include <string_view>
 
 // custom packers
-namespace msgpack {
+namespace msgpack::adaptor {
 MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
-    namespace adaptor {
-
     // pack name of component and attribute
     template <class T>
         requires(std::same_as<T, power_grid_model::meta_data::MetaComponent> ||
@@ -53,9 +51,8 @@ MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
         }
     };
 
-    } // namespace adaptor
 } // MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS)
-} // namespace msgpack
+} // namespace msgpack::adaptor
 
 namespace power_grid_model::meta_data {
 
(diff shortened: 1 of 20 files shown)


mgovers:
FYI: your CI fails because of a known issue in the ubuntu container (more specifically with the combination `clang` + `libstdc++`). Re-running failed jobs should do the trick (maybe twice).

The specific issue causes this log, so if you see this, then you know that's what's wrong:

```cpp
      /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: note: undefined function '_S_fractional_width' cannot be used in a constant expression
      /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2275:2: note: declared here
              _S_fractional_width()
              ^
      1 error generated.
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -20,10 +20,8 @@
 #include <string_view>
 
 // custom packers
-namespace msgpack {
+namespace msgpack::adaptor {
 MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
-    namespace adaptor {
-
```
mgovers, line 26: this is a false positive. you need `MSPACK_API_VERSION_NAMESPACE` or everything will break

power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
```diff
@@ -64,7 +64,14 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
...
     using UpdateType = LoadGenUpdate<sym>;
     using BaseClass = std::conditional_t<is_gen, GenericGenerator, GenericLoad>;
-    static constexpr char const* name = sym ? (is_gen ? "sym_gen" : "sym_load") : (is_gen ? "asym_gen" : "asym_load");
+    static constexpr char const* get_name() {
+        if constexpr (sym) {
+            return is_gen ? "sym_gen" : "sym_load";
+        } else {
+            return is_gen ? "asym_gen" : "asym_load";
+        }
+    }
```
mgovers, line 73: if you're not going to make this a lambda-function, please make this function private. no other user should look ever at this function and instead, the `name` constexpr variable should be used

power_grid_model_c/power_grid_model/include/power_grid_model/timer.hpp
```diff
@@ -57,7 +57,7 @@ class Timer {
         if (info_ != nullptr) {
             auto const now = Clock::now();
             auto const duration = Duration(now - start_);
-            info_->operator[](Timer::make_key(code_, name_)) += (double)duration.count();
+            info_->operator[](Timer::make_key(code_, name_)) += duration.count();
```
mgovers, line 60: please double-check that this indeed results in exactly the same behaviour, and that no compiler/OS-dependency is introduced

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
Laurynas-Jagutis: Yep, I noticed that and was reverting it just now :)

Laurynas-Jagutis: Although all the builds and tests pass

mgovers: what do you mean? do you mean all builds in CI? or the C++ ones? or the ones on your PC? there may be version-, compiler- and OS-dependencies here, hence the differences

Laurynas-Jagutis:
> FYI: your CI fails because of a known issue in the ubuntu container (more specifically with the combination `clang` + `libstdc++`). Re-running failed jobs should do the trick (maybe twice).
> 
> The specific issue causes this log, so if you see this, then you know that's what's wrong:
> 
> ```c++
>       /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: note: undefined function '_S_fractional_width' cannot be used in a constant expression
>       /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2275:2: note: declared here
>               _S_fractional_width()
>               ^
>       1 error generated.
> ```

Tried rerunning the failed jobs multiple times now, did not help.

mgovers:
> > FYI: your CI fails because of a known issue in the ubuntu container (more specifically with the combination `clang` + `libstdc++`). Re-running failed jobs should do the trick (maybe twice).
> > The specific issue causes this log, so if you see this, then you know that's what's wrong:
> > ```c++
> >       /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: note: undefined function '_S_fractional_width' cannot be used in a constant expression
> >       /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2275:2: note: declared here
> >               _S_fractional_width()
> >               ^
> >       1 error generated.
> > ```
> 
> Tried rerunning the failed jobs multiple times now, did not help.

can you please change the base to `feature/llvm-15-in-ci` and merge that branch into this one? the fix in https://github.com/PowerGridModel/power-grid-model/pull/418 should do the trick

mgovers:
@Laurynas-Jagutis please change the base branch as well (edit the PR), because otherwise, your PR will contain those changes as well, which is undesirable

power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
```diff
@@ -60,11 +60,20 @@ class GenericGenerator : public GenericLoadGen {
...
+
   public:
     using InputType = LoadGenInput<sym>;
     using UpdateType = LoadGenUpdate<sym>;
     using BaseClass = std::conditional_t<is_gen, GenericGenerator, GenericLoad>;
-    static constexpr char const* name = sym ? (is_gen ? "sym_gen" : "sym_load") : (is_gen ? "asym_gen" : "asym_load");
+    static constexpr char const* name = get_name();
```
mgovers, line 76: sonar cloud doesn't like this because it's a function that's not called at runtime (cause it's `constexpr`) and therefore "untested". Please change this to

```cpp
    static constexpr char const* name = []{
        if constexpr (sym) {
            return is_gen ? "sym_gen" : "sym_load";
        } else {
            return is_gen ? "asym_gen" : "asym_load";
        }
    }();
```

mgovers (approved):

PR #419, opened 2023-11-01
Feature/refactor test power sensor

mgovers:
This ensures that code changes in #409 remain small, reviewable and isolated

I could still refactor even further, but I believe that we can do that at a later stage if needed. At least most of the near-identical code duplications are now removed

### DIFF
--- a/tests/cpp_unit_tests/test_power_sensor.cpp
+++ b/tests/cpp_unit_tests/test_power_sensor.cpp
@@ -15,550 +15,195 @@ TEST_CASE("Test power sensor") {
     // ------------------------------------------------------
     // --------------- Symmetric power sensor ---------------
     // ------------------------------------------------------
-    SUBCASE("Symmetric Power Sensor - Generator") {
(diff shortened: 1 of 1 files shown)


TonyXiang8787:
@mgovers  So this is a direct PR to main and ready for review?

petersalemink95:
Is it correct that there are more tests removed then added? Seems like you're testing less now

mgovers:
> @mgovers So this is a direct PR to main and ready for review?

yes indeed



> Is it correct that there are more tests removed then added? Seems like you're testing less now

i combined the tests for power sensor on `{Shunt, Load}` as well as `{Generator, Branch_from, Branch_to, Source}`, because they were exactly the same (aside from typos in the comments). The way I tested this was by creating a `for` loop that loops over the terminal types for each of those sets, and then do a regex match on the full test case and i fixed the difference between the comments - and that worked as expected.

TonyXiang8787 (approved):

PR #425, opened 2023-11-10
Feature addition to check if `p_sigma` and `q_sigma` co-(in)exist.

Jerry-Jinfeng-Guo:

Draft PR


src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +697,33 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
     return errors
 
 
+def exist_in_pairs(data: SingleDataset, component: str, field_1: str, field_2: str) -> bool:
+    """
+    Check if anyone of the specific pair of fields of a particular type of component exist.
```
mgovers, line 702: here and below: `anyone` is reserved for people. `any one` or just `any` would be correct since it concerns inanimate objects

in addition, the body doesn't do what the name suggests: `exist in pairs` to me suggests the `AND` operation, while the body and description suggest `OR`

src/power_grid_model/validation/validation.py
```diff
@@ -307,13 +308,19 @@ def validate_required_values(
...
         required["voltage_sensor"] += ["u_sigma", "u_measured"]
         required["power_sensor"] += ["power_sigma", "p_measured", "q_measured"]
     required["sym_voltage_sensor"] = required["voltage_sensor"].copy()
     required["asym_voltage_sensor"] = required["voltage_sensor"].copy()
     required["sym_power_sensor"] = required["power_sensor"].copy()
     required["asym_power_sensor"] = required["power_sensor"].copy()
+    if sigma_required and exist_in_pairs(data, "sym_voltage_sensor", "p_sigma", "q_sigma"):
+        required["sym_power_sensor"] += ["p_sigma", "q_sigma"]
+    if sigma_required and exist_in_pairs(data, "asym_voltage_sensor", "p_sigma", "q_sigma"):
+        required["asym_power_sensor"] += ["p_sigma", "q_sigma"]
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: I recall from our conversation that either i) both `p_sigma` and `q_sigma` exist at the same time or ii) neither of them exist?

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,33 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+def valid_p_q_sigma(data: SingleDataset, component: str) -> List[MultiFieldValidationError]:
+    """
+    Check if for 'sym_power_sensor' and 'asym_power_sensor' the p_sigma and q_sigma are valid.
+
+    Args:
+        data: The input/update data set for all components
+        component: The component of interest, in this case only 'sym_power_sensor' or 'asym_power_sensor'
+
+    Returns:
+        A list containing zeor or one MultiFieldValidationError, listing the p_sigma and q_sigma mismatch
```
mgovers, line 710:
```suggestion
        A list containing zero or one MultiFieldValidationError, listing the p_sigma and q_sigma mismatch
```

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,33 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+    """
+    errors = []
+    p_sigma = data[component]["p_sigma"]
+    q_sigma = data[component]["q_sigma"]
+    ids = [data[component]["id"].flatten().tolist()]
+    p_nan = np.isnan(p_sigma)
+    q_nan = np.isnan(q_sigma)
+    p_inf = np.isinf(p_sigma)
+    q_inf = np.isinf(q_sigma)
+    if np.logical_xor(p_nan.any(), q_nan.any()) or np.logical_or(p_inf.any(), q_inf.any()):
```
mgovers, line 720: i think you'll have to specify `axis=-1` to the inner `any` to ensure that it is OK to have two different components in the same scenario with different conventions (one may have `p_sigma` and `q_sigma` specified while the other may not). same for the same component across multiple scenarios

what i don't know is whether `axis=-1` will cause different behaviour between `asym_power_sensor` and `sym_power_sensor`. that would obviously be wrong

these are some nice test cases though

tests/unit/validation/test_validation_functions.py
```diff
@@ -431,6 +433,34 @@ def test_validate_values__infinite_sigmas(sensor_type, parameter):
...
+    def random_id(dim=2):
+        return np.random.randint(0, dim)
+
+    def random_fill(array, sensor_type, parameter):
+        if sensor_type == "sym_power_sensor":
+            array[parameter[random_id()]] = np.random.rand()
+        else:
+            array[parameter[random_id()]][0][0] = np.random.rand()
+            array[parameter[random_id()]][0][1] = np.random.rand()
+            array[parameter[random_id()]][0][2] = np.random.rand()
```
mgovers, line 453: we tend to not do random testing since what we do is seldom sufficiently generalizable and the same seed may result in different random values (due to differences in built-in random engine, especially in Python) between platforms, making it hard to reproduce.

instead, usually we provide dummy values (based on the `N > 1` argument that if you test `-1, 0, 1, 2` and the implementation is reasonably good, then that is pretty much equivalent to testing `-1, 0, 1, 2, ..., many`).

I do agree that it's not ideal, so i guess we're open to new input, but either fixing and/or clearly logging the `pytest.seed` would be key. also, i know different people have different opinions within the team, so if you want to introduce this, please request some time in the next knowledge sharing session

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,33 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
     return errors
 
 
+def valid_p_q_sigma(data: SingleDataset, component: str) -> List[MultiFieldValidationError]:
+    """
+    Check if for 'sym_power_sensor' and 'asym_power_sensor' the p_sigma and q_sigma are valid.
```
mgovers, line 703:
```suggestion
    Check whether for 'sym_power_sensor' and 'asym_power_sensor' the pairs (p_sigma, q_sigma) are valid.
```

`if` is for a conditional statement (i.e.: only run the `check` in case that the pairs are valid), while `whether` implies that the statement is always run but has different result/behaviour depending on the validity of the condition.

compare e.g.:

```py
# if
if bool(condition):
   return statement()

# whether
return statement(bool(condition))
```

or in this particular case

```py
# if
if valid(p_sigma, q_sigma):
    check()

# whether
check(valid(p_sigma, q_sigma))
```

I know this is incorrect in many locations in the code and we're slowly improving, but it's better to do such a change when touching the file rather than going over all docstrings in one go, because we're bound to miss some

tests/unit/validation/test_validation_functions.py
Jerry-Jinfeng-Guo: Hmm. Although here the values are randomly set, but following the same erroneous pattern.

src/power_grid_model/validation/rules.py
Jerry-Jinfeng-Guo: Here I agree that we don't agree on if-or-not `if` is a good choice; but for the sake of eliminating ambiguity, a different way of putting it is needed. However, I don't quite agree on the second part where you distinguish `if` and `whether`. After all, English can be rather vague, and ambiguity is largely subjective...

Jerry-Jinfeng-Guo:
How on the actual earth can I get rid of the `Check Code Quality` error? 💢

src/power_grid_model/validation/rules.py
mgovers: https://paperpal.com/blog/academic-writing-guides/language-grammar/if-vs-whether

tests/unit/validation/test_validation_functions.py
```diff
@@ -431,6 +433,31 @@ def test_validate_values__infinite_sigmas(sensor_type, parameter):
...
+    ],
+)
+def test_validate_values__bad_p_q_sigma(sensor_type, parameter):
+    def arbitrary_fill(array, sensor_type, parameter):
+        if sensor_type == "sym_power_sensor":
+            array[parameter[0]] = 0.5
+        else:
+            array[parameter[1]][0][0] = 0.6
+            array[parameter[0]][0][1] = 0.7
+            array[parameter[1]][0][2] = 0.8
```
mgovers, line 450: please also add bad-weather scenarios, e.g.:

* single invalid `sym_power_sensor` with the `p_sigma` is nan but `q_sigma` is set
* single invalid `sym_power_sensor` with the `p_sigma` is set but `q_sigma` is nan
* 1 valid and one invalid `sym_power_sensor` (only the invalid one should be flagged)
* single invalid `asym_power_sensor` with all `p_sigma` are nan but all `q_sigma` are set
* single invalid `asym_power_sensor` with all `p_sigma` are set but all `q_sigma` are nan
* single invalid `asym_power_sensor` with both `p_sigma` and `q_sigma` set for some phases but nan for others (e.g.: `{"p_sigma" = [nan, 1.0, nan], "q_sigma = [nan, 1.0, nan]}`)
* single invalid `asym_power_sensor` with some combination of set and nan `p_sigma` and `q_sigma` (e.g.: `{"p_sigma" = [nan, 1.0, nan], "q_sigma = [1.0, nan, nan]}`)
* single invalid `asym_power_sensor` with all `p_sigma` are set but all `q_sigma` are nan
* 1 valid and one invalid `asym_power_sensor` as before

of course, both all set **and** all nan should be checked as well

I know it's a lot of test cases but it's a lot of edge cases as well and we need to catch them all

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,31 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+    """
+    errors = []
+    p_sigma = data[component]["p_sigma"]
+    q_sigma = data[component]["q_sigma"]
+    ids = [data[component]["id"].flatten().tolist()]
+    p_nan = np.isnan(p_sigma)
+    q_nan = np.isnan(q_sigma)
+    p_inf = np.isinf(p_sigma)
+    q_inf = np.isinf(q_sigma)
+    mis_match = p_nan != q_nan  # same logic as np.logical_xor(p_nan.any(axis=-1), q_nan.any(axis=-1)).any()
```
mgovers, line 720: this is not sufficient for asymmetric components, because in this case, there might be some component that has e.g. `p_sigma = [nan, 1.0, nan], q_sigma = [nan, 1.0, nan]`. Your logic will yield `any([False, False, False]) = False` but the result should be `True` (mismatch).

That is due to the fact that `power_sigma = sqrt(((p_sigma[0] + p_sigma[1] + p_sigma[2])/3)^2 + ((q_sigma[0] + q_sigma[1] + q_sigma[2])/3)^2)` requires all elements to be not nan

In addition, please also verify that it works as intended when there are multiple components (only invalid components are flagged)

docs/user_manual/components.md
```diff
@@ -470,7 +470,7 @@ Because of this distribution, at least one appliance is required to be connected
...
-| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Complete Types`. | &#10024; only for state estimation. | &#10004; |                        `> 0`                         |
+| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Concrete Types`. | &#10024; only for state estimation. | &#10004; |                        `> 0`                         |
```
mgovers, line 473: link is verified to be fixed ✅ see also https://power-grid-model--425.org.readthedocs.build/en/425/user_manual/components.html#id21

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,36 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+    errors = []
+    p_sigma = data[component]["p_sigma"]
+    q_sigma = data[component]["q_sigma"]
+    ids = [data[component]["id"].flatten().tolist()]
+    p_nan = np.isnan(p_sigma)
+    q_nan = np.isnan(q_sigma)
+    p_inf = np.isinf(p_sigma)
+    q_inf = np.isinf(q_sigma)
+    mis_match = p_nan != q_nan
+    mis_match |= np.logical_xor(p_nan, q_nan).any()
```
mgovers, line 722: the `!=` is the same as `np.logical_xor`

```suggestion
    mis_match = p_nan != q_nan
```

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,36 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+    if component == "asym_power_sensor":
+        mis_match |= p_nan.any() == (~p_nan).any()
+        mis_match |= q_nan.any() == (~q_nan).any()
+    if mis_match.any() or np.logical_or(p_inf.any(), q_inf.any()):
+        errors.append(MultiFieldValidationError(component, ["p_sigma", "q_sigma"], ids))
```
mgovers, line 727: you need to replace all instances of `.any()` with `.any(axis=-1)` for `"asym_power_sensor"`, and remove it entirely for `sym_power_sensor`.

`.any()` causes the entire list across all id's to be squashed into a single boolean value. We only want to return the ids that were incorrect. `mis_match` should therefore be an array of shape `data[component].shape` (note that for asymmetric sensors, `data[component]` still has the same shape, but `data[component]['p_sigma']` and `data[component]['q_sigma']` again are numpy arrays but with an additional dimension.

The following is just for illustration. The input is a `SingleDataset`, not a `BatchDataset`, so the first dimension drops, and just `axis=1` should do the trick as well, but `axis=-1` is more clear in this case

```py
>>> import power_grid_model as pgm
>>> asym_ps = pgm.initialize_array("input", "asym_power_sensor", (3, 2))
>>> asym_ps.shape
(3, 2)
>>> asym_ps["p_sigma"].shape
(3, 2, 3)
>>> asym_ps["q_sigma"].shape
(3, 2, 3)
```

`np.logical_xor` makes an element-wise comparison between two arrays of the same shape:

```py
>>> import numpy as np
>>> np.logical_xor(asym_ps["p_sigma"], asym_ps["q_sigma"]).shape
(3, 2, 3)
>>> np.logical_xor(asym_ps["p_sigma"], asym_ps["q_sigma"]).any()
False
```

Instead, the `mis_match` should become an array of the same shape (as you already do earlier):

```py
>>> mismatch = asym_ps["p_sigma"] != asym_ps["q_sigma"]
>>> mismatch.shape
(3, 2, 3)
```

I think the complete logic should be something like this (taking inspiration from and following the conventions of the function in the above: https://github.com/PowerGridModel/power-grid-model/pull/425/files#diff-216fb00ddcbaef2c504f2b9244c35a587fe21ec8b3fda59130ec57d58e49d3e5R688-R697 )

```suggestion
    p_nan = np.isnan(p_sigma)
    q_nan = np.isnan(q_sigma)
    p_inf = np.isinf(p_sigma)
    q_inf = np.isinf(q_sigma)
    if p_sigma.ndim > 1:
        p_sigma = p_sigma.any(axis=-1)
        q_sigma = p_sigma.any(axis=-1)
        p_inf = p_inf.any(axis=-1)
        q_inf = q_inf.any(axis=-1)

    invalid = (p_nan != q_nan) or p_inf or q_inf
    if invalid.any():
        ids = data[component]["id"][invalid].flatten().tolist()
        errors.append(MultiFieldValueError(component, ["p_sigma", "q_sigma"], ids))
```

tests/unit/validation/test_validation_functions.py
```diff
@@ -431,6 +433,81 @@ def test_validate_values__infinite_sigmas(sensor_type, parameter):
         assert not isinstance(error, InfinityError)
 
 
+@pytest.mark.parametrize(
+    ("sensor_type", "parameter", "values", "error_types"),
+    [
+        ("sym_power_sensor", ["p_sigma", "q_sigma"], [NaN, NaN], [InvalidIdError]),
```
mgovers, line 439: as discussed, please also add scenarios with 2 components (dw, i'm not trying to put pressure onto you, i'm just adding this comment so that I don't forget about it)

src/power_grid_model/validation/rules.py
mgovers: why did you `unresolve` this? i don't see any changes yet

Jerry-Jinfeng-Guo: It's on my local files, changes not yet pushed...

Jerry-Jinfeng-Guo: I updated `invalid = (p_nan != q_nan) or p_inf or q_inf`, as it causes issues with ambiguity. Rest incoming.

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,38 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+    q_nan = np.isnan(q_sigma)
+    p_inf = np.isinf(p_sigma)
+    q_inf = np.isinf(q_sigma)
+    if p_sigma.ndim > 1:  # if component == 'asym_power_sensor':
+        p_nan = p_nan.any(axis=-1)
+        q_nan = q_nan.any(axis=-1)
+        p_inf = p_inf.any(axis=-1)
+        q_inf = q_inf.any(axis=-1)
+    mis_match = p_nan != q_nan
+    mis_match |= p_inf.any() or q_inf.any()
```
mgovers, line 726:
```suggestion
    mis_match = p_nan != q_nan
    mis_match |= p_inf or q_inf
```
cause otherwise your ids are messed up again for inf as well

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,38 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+    if mis_match.any():
+        ids = [data[component]["id"].flatten().tolist()]
```
mgovers, line 728:
```suggestion
        ids = [data[component]["id"][mis_match].flatten().tolist()]
```

please double-check with my proposed change from my earlier comment

tests/unit/validation/test_validation_functions.py
```diff
@@ -431,6 +434,140 @@ def test_validate_values__infinite_sigmas(sensor_type, parameter):
...
+        node = initialize_array("input", "node", 1)
+        node["id"] = 123
+        sym_power_sensor = initialize_array("input", "sym_power_sensor", 1)
+        sym_power_sensor["p_sigma"] = values[0][0]
+        sym_power_sensor["q_sigma"] = values[0][1]
+        sym_power_sensor["id"] = 456
+        asym_power_sensor = initialize_array("input", "asym_power_sensor", 1)
+        asym_power_sensor["p_measured"] = values[1][0]
+        asym_power_sensor["q_measured"] = values[1][1]
+        asym_power_sensor["id"] = 789
```
mgovers, line 556: i actually meant 2 different components of the same type, e.g.

```suggestion
        sym_power_sensor = initialize_array("input", "sym_power_sensor", 2)
        sym_power_sensor["p_sigma"] = values[0][0]
        sym_power_sensor["q_sigma"] = values[0][1]
        sym_power_sensor["id"] = [456, 789]
```

such that e.g. the first one is OK but the second one is invalid (or vice versa). the error should contain the correct `id` of only the invalid one

src/power_grid_model/validation/rules.py
Jerry-Jinfeng-Guo: Except this will cause the ambiguity I mentioned earlier
`
 ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
`

docs/user_manual/components.md
mgovers: yes indeed, that's why i resolved the comment

src/power_grid_model/validation/rules.py
```diff
@@ -697,6 +698,39 @@ def none_missing(data: SingleDataset, component: str, fields: Union[str, List[st
...
+    q_inf = np.isinf(q_sigma)
+    if p_sigma.ndim > 1:  # if component == 'asym_power_sensor':
+        p_nan = p_nan.any(axis=-1)
+        q_nan = q_nan.any(axis=-1)
+        p_inf = p_inf.any(axis=-1)
+        q_inf = q_inf.any(axis=-1)
+    mis_match = p_nan != q_nan
+    mis_match |= np.logical_or(p_inf, q_inf)
+    if mis_match.any():
+        # ids = data[component]["id"].flatten().tolist()
```
mgovers, line 728:
```suggestion
```

tests/unit/validation/test_validation_functions.py
```diff
@@ -431,6 +434,177 @@ def test_validate_values__infinite_sigmas(sensor_type, parameter):
...
+            "node": node,
+            "sym_power_sensor": sym_power_sensor,
+            "asym_power_sensor": asym_power_sensor,
+        }
+
+    data = two_component_data(values)
+    all_errors = validate_values(data)
+    for error in all_errors:
+        assert any(isinstance(error, error_type) for error_type in error_types)
+        assert data[error.component]["id"][0] == error.ids[0]
```
mgovers, line 583: should be like this, right? or am i misunderstanding something?

```suggestion
        assert (data[error.component]["id"] == error.ids).all()
```
or
```suggestion
        for actual, expected in zip(error.ids, data[error.component]["id"]):
            assert actual == expected
```
or
```suggestion
        assert data[error.component]["id"][1] == error.ids[1]
```

tests/unit/validation/test_validation_functions.py
```diff
@@ -431,6 +434,177 @@ def test_validate_values__infinite_sigmas(sensor_type, parameter):
...
+        ),
+        (
+            "asym_power_sensor",
+            ["p_sigma", "q_sigma"],
+            [[0.1, 0.1, 0.1], [0.1, 0.1, 0.1]],
+            [InvalidIdError, NotUniqueError],
+        ),
+    ],
+)
+def test_validate_values__bad_p_q_sigma(sensor_type, parameter, values, error_types):
```
mgovers, line 510: i don't think you need the `parameter` field. instead you can provide `p_sigma` and `q_sigma` values as separate input fields. might also make things a little bit more clear

```suggestion
def test_validate_values__bad_p_q_sigma(sensor_type, p_sigma, q_sigma, error_types):
```

mgovers (commented):
i think it's pretty much done. just a couple final nitpicks.

please also don't forget to update the branch with `origin/main` (either locally or using the `Update Branch` button in the PR

src/power_grid_model/validation/rules.py
Jerry-Jinfeng-Guo: done

tests/unit/validation/test_validation_functions.py
Jerry-Jinfeng-Guo: done

mgovers (approved):

PR #426, opened 2023-11-13
update topology to not necessarily go through sparse mapping

Laurynas-Jagutis:
**Fixes issue**: update topology to not necessarily go through sparse mapping


Laurynas-Jagutis:
Current task: update the sorting, so that the complexity of the algorithm would be linear

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -7,6 +7,7 @@
 #define POWER_GRID_MODEL_SPARSE_MAPPING_HPP
 
 #include "power_grid_model.hpp"
+using namespace std;
```
mgovers, line 10: please never ever ever do `using namespace std;` in a header file 😬 it may mess up your entire environment

the only exception may be if it is used in another namespace (and even then, be veeeerrrryyyyy careful when you do)

if you really need some functionality from `std` and want to expose it globally as if it were your own function, you can do `using std::func;` but please only do this if you know what you're doing

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -50,35 +51,82 @@ struct SparseMapping {
...
+
+    DenseMapping dense_mapping;
+
+    dense_mapping.indvector.resize(n_A);
+    // size of reorder will be 6
+    dense_mapping.reorder.resize(n_A);
 
+    IdxVector vec = idx_B_in_A;
+
+    sort(vec.begin(), vec.end());
```
mgovers, line 124: just use `std::sort` instead of `using namespace std;`

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
Laurynas-Jagutis: Okay, this will be changed together with the sorting algorithm, since I only wanted to use it to test the sorting as well as the unit test. Thank you for letting me know!

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -50,35 +50,109 @@ struct SparseMapping {
...
+    std::copy(xndvector.cbegin(), xndvector.cend(), counter.begin());
+
+    for (auto i = n_A - 1; i >= 0; i--) {
+        dense_mapping.indvector[xndvector[idx_B_in_A[i]] - 1] = idx_B_in_A[i];
+        xndvector[idx_B_in_A[i]]--;
+    }
+
+    for (auto it_entry = entries.crbegin(); it_entry != entries.crend(); ++it_entry) {
+        dense_mapping.reorder[--counter[it_entry->second]] = it_entry->first;
+    }
```
mgovers, line 151: there's also a different approach:

what we seek is the sorted list `[a_i]` of the input values `{a_i | i in [0, ..., n]}` together with the (reordered) indices of the original list `[0, 1, ..., n]`.

There are 2 different algorithms that may be efficient, depending on the size `n` and the highest entry `r = max{a_i | i in [0, ..., n]}` (i use the convention in https://en.wikipedia.org/wiki/Sorting_algorithm#Non-comparison_sorts ).

if `n` is of the order of or greater than `r`, then just using `build_sparse_mapping` and then using `sparse_decode` on the `indptrs` suffices. This is the [Counting Sort](https://en.wikipedia.org/wiki/Counting_sort) you implemented. To shorted the code, you can also just do

```cpp
auto const sparse_mapping = build_sparse_mapping(idx_B_in_A, n_B);
return DenseMapping {
    .values = sparse_decode(sparse_mapping.indptr),
    .reorder = std::move(sparse_mapping.reorder)
};
```

The additional computational overhead is of the order of `r` and not any worse than the existing implementation (because that's exactly what happens when you do `DenseGroupedIdxVector{from_sparse, sparse_mapping.indptr}`).

Otherwise, `n << r`. This means that if you `sort` a vector of pairs containing both the original index and the input value to sort based on the value, then you automatically obtain the correct ordering. E.g. (removing any reserves, moving, etc.; just for illustration)

```cpp
std::vector<std::pair<Idx, Idx>> mapping_to_from;
std::ranges::transform(idx_B_in_A,
                       std::vies::iota(0, static_cast<Idx>(idx_B_in_A.size()),
                       std::back_inserter(mapping_to_from),
                       [](Idx value, Idx orig_idx){ return std::pair{value, orig_idx}; });
std::ranges::sort(mapping_to_from);  // runs in O(n log n)

DenseMapping result;
std::ranges::transform(mapping_to_from, std::back_inserter(result.indvector),
                       [](std::pair<Idx, Idx> const& to_from) { return to_from.first; } );
std::ranges::transform(mapping_to_from, std::back_inserter(result.reorder),
                       [](std::pair<Idx, Idx> const& to_from) { return to_from.second; });
```

Now it just so happens that in most cases where we use `DenseGroupedIdxVector`, `n << r` (e.g.: there will usually be only one or a couple sources in a power grid). In that case, the first algorithm becomes `O(r)` and the second one remains `O(n log n)`. Since `n << r`, the second algorithm will be much faster.

I would recommend doing both implementations, keeping both of them as short as possible, with an `if` statement selecting one or the other based on whether `n log n < c * r` (but to do so, please calculate for each method the amount of operations, so that you have the correct threshold `c`)

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -50,35 +50,109 @@ struct SparseMapping {
...
+
+    std::copy(xndvector.cbegin(), xndvector.cend(), counter.begin());
+
+    for (auto i = n_A - 1; i >= 0; i--) {
+        dense_mapping.indvector[xndvector[idx_B_in_A[i]] - 1] = idx_B_in_A[i];
+        xndvector[idx_B_in_A[i]]--;
+    }
+
+    for (auto it_entry = entries.crbegin(); it_entry != entries.crend(); ++it_entry) {
+        dense_mapping.reorder[--counter[it_entry->second]] = it_entry->first;
```
mgovers, line 150: these 2 for loops may be combined into one single for loop

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -50,35 +50,164 @@ struct SparseMapping {
...
+        // n_A operations
+        auto index_range = std::ranges::views::iota(0, static_cast<Idx>(idx_B_in_A.size()));
+
+        // n_A * 2 operations?
+        std::ranges::transform(idx_B_in_A, index_range, std::back_inserter(mapping_to_from),
+                               [](Idx value, Idx orig_idx) {
+                                   return std::pair{value, orig_idx};
+                               });
+        // n_A log n_A?
+        std::ranges::sort(mapping_to_from); // runs in O(n log n)
```
mgovers, line 135: you'll have to measure the prefactor before `n log n` by providing worst-case scenarios to the algorithm for different values of `n_A`, e.g. `100, 1000, 10000, 100000, 1000000`)

(please refer to wikipedia; in particular `introsort` (which is used by the GNU STL and MSVC STL implementations, as mentioned by https://en.cppreference.com/w/cpp/algorithm/ranges/sort , as well as the LLVM STL as mentioned by wikipedia. the constant term is negligible for large `n_A` which are interesting for us)

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -50,35 +50,164 @@ struct SparseMapping {
...
+
+        // n_A * 2 operations?
+        std::ranges::transform(mapping_to_from, std::back_inserter(dense_mapping.indvector),
+                               [](DenseEntry const& to_from) { return to_from.first; });
+        // n_A * 2 operations?
+        std::ranges::transform(mapping_to_from, std::back_inserter(dense_mapping.reorder),
+                               [](DenseEntry const& to_from) { return to_from.second; });
+    } else {
+        // 1
+        std::vector<DenseEntry> entries(n_A);
```
mgovers, line 145: nope, allocation of a vector of size `n_A` is O(n_A)

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -50,35 +50,164 @@ struct SparseMapping {
...
+        dense_mapping.indvector.resize(n_A);
+        // n_A or can we say that resize is constant 1?
+        dense_mapping.reorder.resize(n_A);
+
+        IdxVector xndvector;
+        // n_B or can we say that resize is constant 1?
+        xndvector.resize(n_B);
+        // n_B
+        IdxVector counter(n_B, 0);
+
```
mgovers, line 163: `n_B` (approximately; depending on STL version, compiler and distribution could be better (linux uses block heap memory, meaning it can choose to allocate e.g. 1kB with one allocation call if it is lucky and there's such a block free) or worse (Windows has contiguous memory so it has to look for a free spot) but i would just count `n_B` or maybe 2 or 3 times, but we're talking orders here - not exact values

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -50,35 +50,164 @@ struct SparseMapping {
...
+    IdxVector reorder;
+};
+
+inline DenseMapping build_dense_mapping(IdxVector const& idx_B_in_A, Idx const n_B) {
+    auto const n_A = static_cast<Idx>(idx_B_in_A.size());
+    auto const r = std::max_element(idx_B_in_A.begin(), idx_B_in_A.end())[0];
+    // for counting sort nr_of_operations = 1 + 3*n_A + 2*n_B + 2*n_A + 2*n_B - 1 + 1 + n_A + 1 + 14 * n_A
+    // nr_of_operations = 20*n_A + 4*n_B + 2
+    auto const c = 20 * n_A + 4 * n_B + 2;
+    // for n log n sort nr of operations = 1 + 7*n_A + n_A*log(n_A)?
```
mgovers, line 118: this seems reasonable. it's nice to see the approximate form of the formula being so simple as they are.

based on this, the criterion reduces to something approximately like `n_B < ((prefactor/4) * log(n_A) - 13/4) n_A`, or (approximating and simplifying even further) `n_B / n_A < c * log(n_A) - 3`

the important thing now is to measure the prefactor (or, equivalently, `c`) before `n_A log n_A`.

Because we can assume that, usually, either `n_A = 0, 1, 2, 3` (**MUCH** smaller than `n_B`), or `n_A` is of the order of `n_B / 10` or `n_B / 100`, or otherwise otherwise `n_A ~ n_B` or `n_A ~ 10 n_B`

* In the first case, then `n_B / n_A` will be large and the prefactor doesn't matter.
* In the other cases, the prefactor is important, because the threshold depends on it. If it is large, basically for any `n_A` for which `n_B / n_A` is small, the counting sort will be faster, but if it is small, that will only be true for large `n_A`

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
mgovers: see also the CPP reference https://en.cppreference.com/w/cpp/container/vector/vector

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -7,7 +7,7 @@
 #define POWER_GRID_MODEL_SPARSE_MAPPING_HPP
 
 #include "power_grid_model.hpp"
-
+#include <ranges>
```
mgovers, line 10: please add whitelines before and after `#include <ranges>`

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -54,31 +54,92 @@ inline SparseMapping build_sparse_mapping(IdxVector const& idx_B_in_A, Idx const
...
+            return DenseEntry{i_A, j_B};
+        });
+
+        dense_mapping.indvector.resize(n_A);
+
+        dense_mapping.reorder.resize(n_A);
+
+        IdxVector xndvector;
+
+        xndvector.resize(n_B + 1);
```
mgovers, line 125: i've commented that before, but what's `xnd`? please replace with `ind` or even `index_vector`

tests/cpp_unit_tests/test_sparse_mapping.cpp
```diff
@@ -5,6 +5,7 @@
 #include <power_grid_model/sparse_mapping.hpp>
 
 #include <doctest/doctest.h>
+#include <random>
```
mgovers, line 8:
```suggestion

#include <random>
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -54,31 +54,92 @@ inline SparseMapping build_sparse_mapping(IdxVector const& idx_B_in_A, Idx const
...
+        for (auto it_entry = entries.crbegin(); it_entry != entries.crend(); ++it_entry) {
+            dense_mapping.indvector[xndvector[idx_B_in_A[it_entry->first]] - 1] = idx_B_in_A[it_entry->first];
+            dense_mapping.reorder[--xndvector[it_entry->second]] = it_entry->first;
+        }
+    }
+
+    return dense_mapping;
+}
+
+} // namespace power_grid_model
```
mgovers, line 144:
```suggestion
} // namespace power_grid_model

```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -54,31 +54,92 @@ inline SparseMapping build_sparse_mapping(IdxVector const& idx_B_in_A, Idx const
...
+        for (auto const& entry : entries) {
+            ++counter[entry.second];
+        }
+
+        std::inclusive_scan(counter.cbegin(), counter.cend(), xndvector.begin());
+
+        for (auto it_entry = entries.crbegin(); it_entry != entries.crend(); ++it_entry) {
+            dense_mapping.indvector[xndvector[idx_B_in_A[it_entry->first]] - 1] = idx_B_in_A[it_entry->first];
+            dense_mapping.reorder[--xndvector[it_entry->second]] = it_entry->first;
+        }
```
mgovers, line 138: for this entire block, i believe you can also just use the `build_sparse_mapping` implementation before. i don't see why you need the sparse mapping 2x. alternatively (if you think this implementation is better than the other one), then please improve that implementation instead and call it from this function.

the reason is: try to remove code duplications

tests/cpp_unit_tests/test_sparse_mapping.cpp
```diff
@@ -17,4 +18,58 @@ TEST_CASE("Test sparse mapping") {
...
+    int j = 0;
+    for (Idx i = 999999; i >= 0; i--) {
+        idx_B_in_A[j] = i;
+        j++;
+    }
+
+    IdxVector sorted_idx_B_in_A = idx_B_in_A;
+    std::ranges::sort(sorted_idx_B_in_A);
+
+    DenseMapping mapping_2 = build_dense_mapping(idx_B_in_A, 1000001);
```
mgovers, line 67:
```suggestion
```suggestion
    auto constexpr count = Idx{1'000'000};

    IdxVector idx_B_in_A(count);

    int j = 0;
    for (Idx i = count - 1; i >= 0; i--) {
        idx_B_in_A[j] = i;
        j++;
    }

    IdxVector sorted_idx_B_in_A = idx_B_in_A;
    std::ranges::sort(sorted_idx_B_in_A);

    DenseMapping mapping_2 = build_dense_mapping(idx_B_in_A, count + 1);
```

tests/cpp_unit_tests/test_sparse_mapping.cpp
```diff
@@ -17,4 +18,58 @@ TEST_CASE("Test sparse mapping") {
...
+
+    CHECK(mapping_2.indvector == sorted_idx_B_in_A);
+    CHECK(*(mapping_2.indvector.begin()) == *std::min_element(mapping_2.indvector.begin(), mapping_2.indvector.end()));
```
mgovers, line 70:
```suggestion
    CHECK(mapping_2.indvector.front() == *std::ranges::min_element(mapping_2.indvector));
```

tests/cpp_unit_tests/test_sparse_mapping.cpp
```diff
@@ -17,4 +18,58 @@ TEST_CASE("Test sparse mapping") {
...
+    CHECK(*(mapping_2.indvector.end() - 1) ==
+          *std::max_element(mapping_2.indvector.begin(), mapping_2.indvector.end()));
```
mgovers, line 72:
```suggestion
    CHECK(mapping_2.indvector.back() == *std::ranges::max_element(mapping_2.indvector));
```

tests/cpp_unit_tests/test_sparse_mapping.cpp
```diff
@@ -17,4 +18,58 @@ TEST_CASE("Test sparse mapping") {
...
+//     std::ranges::sort(sorted_idx_B_in_A);
+
+//     DenseMapping mapping_2 = build_dense_mapping(idx_B_in_A, 1000001);
+
+//     CHECK(mapping_2.indvector == sorted_idx_B_in_A);
+//     CHECK(*(mapping_2.indvector.begin()) == *std::min_element(mapping_2.indvector.begin(),
+//     mapping_2.indvector.end())); CHECK(*(mapping_2.indvector.end() - 1) ==
+//           *std::max_element(mapping_2.indvector.begin(), mapping_2.indvector.end()));
+// }
+
```
mgovers, line 53: for unit tests, please don't randomly generate, OR fix the seed. random unit tests introduce something called flaky tests, which means that they may fail once in a while, and it's very hard to backtrack in which particular situation it fails

in addition, using default random number generator is in a sense OK for unit tests, but it's a biased number generator, which means that there is a higher probability to get some (series of) values than it is to get others. https://en.wikipedia.org/wiki/Random_number_generation#Uniform_distributions provides a good read-up on that.
 
this is seen as a security risk, especially when used in production code. It is less so for unit tests, but it's considered good practice to treat your test code as production code for various reasons. hence, great caution is necessary

therefore, the other test case is better

```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -54,31 +54,92 @@ inline SparseMapping build_sparse_mapping(IdxVector const& idx_B_in_A, Idx const
...
+};
+
+inline DenseMapping build_dense_mapping(IdxVector const& idx_B_in_A, Idx const n_B) {
+    auto const n_A = static_cast<Idx>(idx_B_in_A.size());
+
+    using DenseEntry = std::pair<Idx, Idx>;
+    DenseMapping dense_mapping;
+
+    if (static_cast<double>(n_A) * log(static_cast<double>(n_A)) <=
+        static_cast<double>(n_A) + static_cast<double>(n_B)) {
```
mgovers, line 96: this is unreadable. please replace with something like

```suggestion
    auto const comparison_sort_complexity = [](Idx n_a, Idx /* n_B */) {
        auto const double_n_a = static_cast<double>(n_a);
        return double_n_a * std::log(double_n_a);
    };
    auto const counting_sort_complexity = [](Idx n_a, Idx n_b) { return static_cast<double>(n_a + n_b); };
    if (comparison_sort_complexity(n_A, n_B) <= counting_sort_complexity(n_A, n_B)) {
```

or (without lambdas)

```suggestion
    const auto comparison_sort_complexity = static_cast<double>(n_A) * std::log(static_cast<double>(n_B));
    const auto counting_sort_complexity = static_cast<double>(n_A + n_B);
    if (comparison_sort_complexity < counting_sort_complexity) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_mapping.hpp
```diff
@@ -54,31 +54,92 @@ inline SparseMapping build_sparse_mapping(IdxVector const& idx_B_in_A, Idx const
...
+
+        std::ranges::transform(mapping_to_from, std::back_inserter(dense_mapping.indvector),
+                               [](DenseEntry const& to_from) { return to_from.first; });
+
+        std::ranges::transform(mapping_to_from, std::back_inserter(dense_mapping.reorder),
+                               [](DenseEntry const& to_from) { return to_from.second; });
+    } else {
+        std::vector<DenseEntry> entries(n_A);
+
+        std::transform(idx_B_in_A.cbegin(), idx_B_in_A.cend(), IdxCount{0}, entries.begin(), [](Idx j_B, Idx i_A) {
```
mgovers, line 115: i believe this could also be

```suggestion
        std::ranges::transform(idx_B_in_A, IdxCount{0}, entries.begin(), [](Idx j_B, Idx i_A) {
```

mgovers: this is not possible due to `boost::counting_iterator` not satisfying `std::forward_iterator`. We can't switch to `std::views::iota` due to `boost::iterators::zip_iterator` not supporting the correct implementation. deadlock much.

TonyXiang8787 (approved):

Issue #74, opened 2022-05-27
[FEATURE] Automatic adjustment of transformer tap changer


TonyXiang8787:
# Feature

- Add control logic for transformer tap changer.
- In power flow calculation, add an option to change the tap changer of transformer automatically.

# Considerations

- How to model tap changer controller? A separate class or inside transformer?
- How to implement it in the power flow solver?

Jerry-Jinfeng-Guo:
For the tap changer, we implemented a ranking algorithm for transformers (with tap) in the following PRs:
- [STEP 1]
- https://github.com/PowerGridModel/power-grid-model/pull/545
- https://github.com/PowerGridModel/power-grid-model/pull/558
- [UNIT TEST]

PR #427, opened 2023-11-15
inverse update approach

mgovers:
* Add update using inverse update (which updates the component with the original component attributes to restore)
* Remove cached update functionality

Enabler for #74 

## Performance

During development, it appeared as if there was a significant performance regression. I improved the implementation and now the performance is actually better than on `main`

### Main

image

### This branch

image


mgovers:
left TODO: test cases

power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
```diff
@@ -87,14 +87,26 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
...
     }
 
+    LoadGenUpdate<sym> inverse(LoadGenUpdate<sym> update_data) const {
+        double const scalar = direction_ * base_power<sym>;
+
+        assert(update_data.id == this->id());
+
+        set_if_not_nan(update_data.status, static_cast<IntS>(this->status()));
+        set_if_not_nan(update_data.p_specified, real(s_specified_) * scalar);
+        set_if_not_nan(update_data.q_specified, imag(s_specified_) * scalar);
```
TonyXiang8787, line 105: I do not understand why this will work. You try to generate an inverse update by reading the current value in the component class. But the data members in the component class are not one-to-one mapping to the values in the update struct. In case of `LoadGen`, the values stored inside the class is p.u. value. But the value in the update struct is using `W` and `var` unit. You need to than back converts the values when you generate the inverse update. 

This is straightforward for `LoadGen`, but might be very complicated for other values.

mgovers: that's what the `scalar` is used for. that's the responsibility of the in-class implementation. I added explicit tests or every component to check that the initially provided value in the input data has the same value as the result data.

I personally also believe that it's the classes' own responsibility to ensure that the output matches closely enough that there is no value drift due to floating-point rounding errors, in which case both values would need to be stored. That would not introduce an unreasonably large memory overhead, because the components are no longer copied. I did not implement that yet, though. Do you think that I should implement that now?

TonyXiang8787: I don't think that's needed unless it is proven that certain operations hit numerical issues.

mgovers: well there's a potential drift for every scenario. if there's a batch of 10k scenarios, then it could become an issue

TonyXiang8787: We are not doing incremental stuff here. It only a multiplication. The drift is really minimum.

As for batch calculations, you always return the original value by `p_specified_org_ * scale`. That is always one drift, regardless of hoe many scenario you have.

mgovers: the drift may occur when doing `input(p_specified / scale) -> inverse(p_specified / scale * scale) -> update(p_specified / scale * scale / scale) -> inverse(p_specified / scale * scale / scale * scale) -> update(p_specified / scale * scale / scale * scale / scale) -> ...` but i guess indeed the difference is minimal

TonyXiang8787 (approved):

PR #429, opened 2023-11-17
Feature/dgc 2055 get batch size function

Jerry-Jinfeng-Guo:
Added the interface function that returns the size of batch data, or raises error when the data is inconsistent.


tests/unit/test_utils.py
```diff
@@ -41,6 +44,23 @@ def test_get_dataset_scenario():
         get_dataset_scenario(data, 2)
 
 
+@pytest.fixture
+def batch_data() -> Dict[str, np.ndarray]:
```
mgovers, line 48: please also add a sparse batch test case

src/power_grid_model/utils.py
```diff
@@ -14,6 +14,7 @@
 
 import numpy as np
 
+from power_grid_model._utils import get_and_verify_batch_sizes
```
mgovers, line 17: the docstring is OK

```suggestion
from power_grid_model._utils import get_and_verify_batch_sizes as get_dataset_batch_size
```

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,22 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
...
+
+    Raises:
+        ValueError: if the batch dataset is inconsistent.
+
+    Returns:
+        The size of the batch dataset. Making use of existing _utils function.
+    """
+    return get_and_verify_batch_sizes(dataset)
+
+
```
mgovers, line 73: the docstring is actually OK. you can just do

```py
from

File: tests/unit/test_utils.py [hunk 4]
```diff
@@ -41,6 +46,46 @@ def test_get_dataset_scenario():
...
+        "asym_load": {
+            "data": np.zeros(shape=4, dtype=power_grid_meta_data["input"]["asym_load"]),
+            "indptr": np.array([0, 2, 3, 4]),
+        },
+        "link": np.zeros(shape=(batch_size, 4), dtype=power_grid_meta_data["input"]["link"]),
+    }
+
+    dataset = CConstDataset(data, ["input"])
+
+    assert get_dataset_batch_size(dataset) == 3
```
mgovers, line 86: please also add a bad weather test case that checks that the correct behavior happens when the batch size is inconsistent. Most importantly, the error should be consistent with the get_scenario function.

If i'm correct, then that function raises a `KeyError`, meaning this function should raise a `ValueError` (due to inconsistent batch size)

raising tests can be tested using

```py
with pytest.raises(ValueError, msg):
    call()
```
if you look for `pytest.raises` you'll find plenty of examples

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,24 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
...
+    Get the number of scenarios in the batch dataset.
+
+    Args:
+        dataset: the batch dataset
+
+    Raises:
+        ValueError: if the batch dataset is inconsistent.
+
+    Returns:
+        The size of the batch dataset. Making use of existing _utils function.
```
mgovers, line 69: the user shouldn't care
```suggestion
        The size of the batch dataset.
```

src/power_grid_model/utils.py
mgovers:
```suggestion
from power_grid_model._utils import get_and_verify_batch_sizes as _get_batch_size
```

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,24 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
...
+    """
+    if isinstance(dataset, CConstDataset):
+        return dataset.get_info().batch_size()
+    return get_and_verify_batch_sizes(dataset)
```
mgovers, line 73:
```suggestion
    return _get_batch_sizes(dataset)
```

src/power_grid_model/utils.py
```diff
@@ -14,7 +14,8 @@
 
 import numpy as np
 
-from power_grid_model.core.power_grid_dataset import get_dataset_type
+from power_grid_model._utils import get_and_verify_batch_sizes
+from power_grid_model.core.power_grid_dataset import CConstDataset, get_dataset_type
```
mgovers, line 18: the `CConstDataset` is a low-level wrapper of the C API functionality and is not something the user should care about

```suggestion
from power_grid_model.core.power_grid_dataset import get_dataset_type
```

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,24 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
     return {component: _get_component_scenario(component_data) for component, component_data in dataset.items()}
 
 
+def get_dataset_batch_size(dataset: BatchDataset) -> int:
```
mgovers, line 58: a `BatchDataset` type is not a `CConstDataset`. in fact, as I mentioned before, the `CConstDataset` (or in general, any class prefixed with `C`) is low-level and should not be exposed to the user.

If there's a really good reason to keep it, please update the type as such:

```suggestion
def get_dataset_batch_size(dataset: Union[BatchDataset, CConstDataset]) -> int:
```

mgovers (changes requested):

src/power_grid_model/utils.py
Jerry-Jinfeng-Guo: What is the official way of constructing a sparse dataset? I imported CConstDataset only for the sake of sparse dataset. If there is a proper class to use, it indeed shouldn't appear.

tests/unit/test_utils.py
Jerry-Jinfeng-Guo: I am not sure by this way of constructing sparse dataset, an `inconsistent` complaint would be given. I assume you are referring to construct a inconsistent `Dict[str, np.ndarray]`  instead of a sparse dataset?

mgovers: well, both actually:

the input for `get_dataset_scenario` has type `dataset: Dict[str, ndarray | Dict[str, ndarray]]` (see also https://power-grid-model.readthedocs.io/en/stable/api_reference/python-api-reference.html#power_grid_model.utils.get_dataset_scenario ).

The different types are given in https://github.com/PowerGridModel/power-grid-model/blob/main/src/power_grid_model/data_types.py (it may be good to add that file to the docs in https://github.com/PowerGridModel/power-grid-model/blob/main/docs/api_reference/python-api-reference.md )

In particular, there are 2 batch array types that both may have consistent and inconsistent batch size. E.g. (modulo correct data types):

* Fully dense inconsistent update dataset: `{"sym_power_sensor": np.array([[]]), "asym_power_sensor": np.array([[], []])}`
  * `sym_power_sensor` has batch size 1 while `asym_power_sensor` has batch size 2
* Fully sparse inconsistent update dataset: `{"sym_power_sensor": {"indptr": np.array([0, 0]), "data": np.array([])}, "asym_power_sensor": {"indptr": np.array([0, 0, 0]), "data": np.array([])}}`
  * idem
* Mixed inconsistent update dataset: `{"sym_power_sensor": np.array([[]]),  "asym_power_sensor": {"indptr": np.array([0, 0, 0]), "data": np.array([])}}}`

for easier creation, you can also use the `initialize_array` function. the `shape` argument may help there

src/power_grid_model/utils.py
mgovers: there's no official way but probably, it's useful to create it as follows:

```py
"sym_power_sensor": {
  "indptr": np.ndarray([0, 0, 1]),
  "data": initialize_array("update", "sym_power_sensor", (1)),
}
```
as opposed to dense
```py
"sym_power_sensor": initialize_array("update", "sym_power_sensor", (2, 1))
```

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,35 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
...
+    Args:
+        dataset: the batch dataset
+
+    Raises:
+        ValueError: if the batch dataset is inconsistent.
+
+    Returns:
+        The size of the batch dataset. Making use of existing _utils function.
+    """
+    return get_and_verify_batch_sizes(dataset)
```
mgovers, line 71: why not do `... import get_and_verify_batch_sizes as get_dataset_batch_size`? is the docstring crappy?

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,35 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
...
+    """
+    Determine the number of batches and verify the data structure
+
+    Args:
+        data_array: a batch array for power-grid-model
+
+    Returns:
+        The number of batches in data_array
+    """
+    return get_batch_size(data_array)
```
mgovers, line 84: idem

src/power_grid_model/utils.py
Jerry-Jinfeng-Guo: Because 
`pylint:
src\power_grid_model\utils.py:17:0: W0611: Unused get_and_verify_batch_sizes imported from power_grid_model._utils as get_data_set_batch_size (unused-import)
`
after ` # pylint: disable=unused-import`

Jerry-Jinfeng-Guo: I'll stick to the way that's in the last commit. The following code piece causes issues between isort and pylint.
Original:
`
from power_grid_model._utils import get_and_verify_batch_sizes as get_data_set_batch_size,  # pylint: disable=unused-import
from power_grid_model._utils import get_batch_size as get_data_array_batch_size  # pylint: disable=unused-import
`
isort would fix this to:
`
from power_grid_model._utils import (
    get_and_verify_batch_sizes as get_data_set_batch_size,  # pylint: disable=unused-import
)
from power_grid_model._utils import get_batch_size as get_data_array_batch_size  # pylint: disable=unused-import
`
and then pylint would complain about the unused first import.
On the other hand, user would easily go to _utils.py when checking for definition of `get_data_set_batch_size()` and `get_data_array_batch_size ()`

src/power_grid_model/utils.py
Jerry-Jinfeng-Guo: idem

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,35 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
     return {component: _get_component_scenario(component_data) for component, component_data in dataset.items()}
 
 
+def get_data_set_batch_size(dataset: BatchDataset) -> int:
```
mgovers, line 58: i know it's linguistically incorrect but across the code base we use
```suggestion
def get_dataset_batch_size(dataset: BatchDataset) -> int:
```

src/power_grid_model/utils.py
```diff
@@ -54,6 +55,35 @@ def _get_component_scenario(component_scenarios: BatchArray) -> np.ndarray:
...
+    Raises:
+        ValueError: if the batch dataset is inconsistent.
+
+    Returns:
+        The size of the batch dataset. Making use of existing _utils function.
+    """
+    return get_and_verify_batch_sizes(dataset)
+
+
+def get_data_array_batch_size(data_array: BatchArray) -> int:
```
mgovers, line 74: maybe rename to

```suggestion
def get_component_batch_size(data_array: BatchArray) -> int:
```

src/power_grid_model/utils.py
```diff
@@ -14,6 +14,7 @@
 
 import numpy as np
 
+from power_grid_model._utils import get_and_verify_batch_sizes, get_batch_size  # pylint: disable=unused-import
```
mgovers, line 17: since they are now used: remove `# pylint: disable=unused-import`

in addition, since we actually don't want to expose these functions (we want users to use the `get_dataset_batch_size` instead of `get_and_verify_batch_sizes`:

```suggestion
from power_grid_model._utils import (
    get_and_verify_batch_sizes as _get_and_verify_batch_size,
    get_batch_size as _get_batch_size
)
```

to make them "private". i don't know whether it will work correctly with pylint and isort. If it doesn't work, that's fine as well

Jerry-Jinfeng-Guo: Isort again 'fixed' this, but to an acceptable state

src/power_grid_model/utils.py
```diff
@@ -14,6 +14,8 @@
 
 import numpy as np
 
+from power_grid_model._utils import get_and_verify_batch_sizes as _get_and_verify_batch_sizes
+from power_grid_model._utils import get_batch_size as _get_batch_size
```
mgovers, line 18: ahhh i know what this is. @nitbharambe also found this. The solution is to do this change in the `pyproject.toml`:

https://github.com/Alliander/power-grid-model-conversions/pull/104/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R62

please do that change instead of trying to find against a sucky `isort` default setting

mgovers (approved):

PR #430, opened 2023-11-17
Feature/standard layout input

### LINKED ISSUE #365: [IMPROVEMENT] Input/output/update data structures as standard layout

mgovers:
Fixes #365 

Implements all input, output and update data structures as standard layout by using flat attribute expansion in the generated code. conversions to former base types is now done using `reinterpret_cast`

Relates to #241 and #276


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/input.hpp
```diff
@@ -21,36 +21,78 @@ struct BaseInput {
...
-struct NodeInput : BaseInput {
+static_assert(std::is_standard_layout_v<BaseInput>);
+
+struct NodeInput {
+    ID id;  // ID of the object
     double u_rated;  // rated line-line voltage
+
+    // implicit conversions to BaseInput
+    operator BaseInput&() { return reinterpret_cast<BaseInput&>(*this); }
+    operator BaseInput const&() const { return reinterpret_cast<BaseInput const&>(*this); }
```
mgovers, line 32: every implicit conversion introduces 2 new sonar cloud code smells. that's 4 per base type, and 250 new ones in total.

I'm personally not a fan of adding `// NOSONAR` to every single line.

I've also read that it's also possible to do `*static_cast<BaseInput*>(static_cast<void*>(this))`. if the alignment is the same, it should actually be possible to do directly `*static_cast<BaseInput*>(this)`, but for some reason that's not working here. see also https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1120 and https://godbolt.org/z/E884oTEqr

In addition, it also complains that we should use `std::bit_cast` when dealing with types that have no additional attributes (e.g. `SensorShortCircuitOutput`). In those cases, shouldn't we just use e.g. `using SensorShortCircuitOutput = BaseOutput` instead?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/input.hpp
```diff
@@ -21,36 +21,78 @@ struct BaseInput {
...
 };
 
-struct BranchInput : BaseInput {
+static_assert(std::is_standard_layout_v<NodeInput>);
```
mgovers, line 35: should i also check that the pointer to all base attribute members and/or alignment are identical? e.g.

```suggestion
static_assert(std::is_standard_layout_v<NodeInput>);
static_assert(std::alignment_of_v<NodeInput> == std::alignment_of_v<BaseInput>);
static_assert(offsetof(NodeInput, id) == offsetof(BaseInput, id));
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/input.hpp
TonyXiang8787: Let's not make thing more difficult. We know what we are doing with `reintepret_cast` here. Just disable globally the warning from SonarCloud.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/input.hpp
TonyXiang8787: It should be check on `>=`. As long as the alignment of "child" class is bigger or equal than de "parent" class, it should be fine.

TonyXiang8787: assert of `offsetof` is good.

TonyXiang8787: @mgovers  is this comment resolved and the PR is ready for review again?

mgovers: no not yet, i've implemented, but i'm still struggling to find a good way to deal with the template case without getting 1000000 static checks per component type and it makes the files unreadable.

i can also add those static asserts to a separate file, e.g. `input_static_checks.hpp` that is then included in `update.hpp` so as to not affect the readability of the dataset declarations

any thoughts?

mgovers: put into separate directory `<...>/auxiliary/static_asserts/<...>.hpp`

mgovers:
@TonyXiang8787 this is ready for review

mgovers (commented):
should i also revert parts of https://github.com/PowerGridModel/power-grid-model/pull/276/files#diff-2bac4fa15db710784f8450b130e98864e112b2b5e4b104c0470f876d5894f4eeL138-R141 ? or is that not necessary? or maybe we can use `offsetof` also in those functions?

mgovers:
@TonyXiang8787 did everything as discussed, including conversion to base types.

The following is created in #435 

> should i also revert parts of https://github.com/PowerGridModel/power-grid-model/pull/276/files#diff-2bac4fa15db710784f8450b130e98864e112b2b5e4b104c0470f876d5894f4eeL138-R141 ? or is that not necessary? or maybe we can use `offsetof` also in those functions?

TonyXiang8787 (approved):

PR #437, opened 2023-11-23
Fix/fix inverse drift

mgovers:
there was a bug in the inverse for the power sensor where the direction was not provided correctly.

@nitbharambe i've got a reduced version of the data you provided. Is that something that we can share? Or is that confidential and I need to make another validation test case manually?


TonyXiang8787 (approved):

nitbharambe:
I guess the exact same is not good. Maybe multiply all power by a random number each would do

mgovers:
> exact same is not good. Maybe multiply all power by a random nu

ok will do

mgovers:
> > exact same is not good. Maybe multiply all power by a random nu
> 
> ok will do

not random but arbitrary. should do the trick

nitbharambe (approved):

PR #438, opened 2023-11-27
Newton Raphson State Esimation Implementation

nitbharambe:
Relates to #369 

The tests `distribution-case` and `sensor-update` are marked as xfail.
The fails are minor differences and pattern is difficult to find. Needs deeper dive.


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,443 @@ namespace power_grid_model {
...
+
+    GetterType<0, 0> theta() { return this->template get_val<0, 0>(); }
+    GetterType<1, 0> v() { return this->template get_val<1, 0>(); }
+    GetterType<2, 0> phi_theta() { return this->template get_val<2, 0>(); }
+    GetterType<3, 0> phi_v() { return this->template get_val<3, 0>(); }
+
+    GetterType<0, 0> eta_theta() { return this->template get_val<0, 0>(); }
+    GetterType<1, 0> eta_v() { return this->template get_val<1, 0>(); }
+    GetterType<2, 0> tau_theta() { return this->template get_val<2, 0>(); }
+    GetterType<3, 0> tau_v() { return this->template get_val<3, 0>(); }
```
petersalemink95, line 41: There is a mistake here for tau and phi, please see the suggestion below. Note that this might affect the rest of the codebase, so that should be updated as well. I've added it to the mathematical workout as clarification

```suggestion
    GetterType<0, 0> theta() { return this->template get_val<0, 0>(); }
    GetterType<1, 0> v() { return this->template get_val<1, 0>(); }
    GetterType<2, 0> phi_p() { return this->template get_val<2, 0>(); }
    GetterType<3, 0> phi_q() { return this->template get_val<3, 0>(); }

    GetterType<0, 0> eta_theta() { return this->template get_val<0, 0>(); }
    GetterType<1, 0> eta_v() { return this->template get_val<1, 0>(); }
    GetterType<2, 0> tau_p() { return this->template get_val<2, 0>(); }
    GetterType<3, 0> tau_q() { return this->template get_val<3, 0>(); }
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,443 @@ namespace power_grid_model {
...
+};
+
+// block class for the right hand side in state estimation equation
+template <bool sym> using NRSERhs = NRSEUnknown<sym>;
+
+// class of 4*4 (12*12) se gain block
+/*
+[
+   [G, QH]
```
petersalemink95, line 50:
```suggestion
   [G, QT]
```

We do the transpose of Q, not hermitian transpose, since we're working with real values here

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,443 @@ namespace power_grid_model {
...
+
+    GetterType<0, 0> g_P_theta() { return this->template get_val<0, 0>(); }
+    GetterType<0, 1> g_P_v() { return this->template get_val<0, 1>(); }
+    GetterType<1, 0> g_Q_theta() { return this->template get_val<1, 0>(); }
+    GetterType<1, 1> g_Q_v() { return this->template get_val<1, 1>(); }
+
+    GetterType<0, 2> qh_P_theta() { return this->template get_val<0, 2>(); }
+    GetterType<0, 3> qh_P_v() { return this->template get_val<0, 3>(); }
+    GetterType<1, 2> qh_Q_theta() { return this->template get_val<1, 2>(); }
+    GetterType<1, 3> qh_Q_v() { return this->template get_val<1, 3>(); }
```
petersalemink95, line 70:
```suggestion
    GetterType<0, 2> qt_P_theta() { return this->template get_val<0, 2>(); }
    GetterType<0, 3> qt_P_v() { return this->template get_val<0, 3>(); }
    GetterType<1, 2> qt_Q_theta() { return this->template get_val<1, 2>(); }
    GetterType<1, 3> qt_Q_v() { return this->template get_val<1, 3>(); }
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,443 @@ namespace power_grid_model {
...
 template <bool sym> class NewtonRaphsonSESolver {
+
+    struct NRSEJacobian {
+        struct i_side_block {};
+        struct j_side_block {};
+
+        RealTensor<sym> dP_dt{};
+        RealTensor<sym> dP_dv{};
+        RealTensor<sym> dQ_dt{};
+        RealTensor<sym> dQ_dv{};
```
petersalemink95, line 93: Maybe it whould be nice to be explicit here, i.e. use `theta` instead of `t`. What do you think?

```suggestion
        RealTensor<sym> dP_dtheta{};
        RealTensor<sym> dP_dv{};
        RealTensor<sym> dQ_dtheta{};
        RealTensor<sym> dQ_dv{};
```

nitbharambe: It looked good when all things were multiplying together as it aligned well :D

mgovers: blue pill/red pill 💀

petersalemink95: Okay, let's keep it `dt` then

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,443 @@ namespace power_grid_model {
...
-    MathOutput<sym> run_state_estimation(YBus<sym> const& /* y_bus */, StateEstimationInput<sym> const& /* input */,
-                                         double /* err_tol */, Idx /* max_iter */,
-                                         CalculationInfo& /* calculation_info */) {
-        // hide implementation
-        return {};
+    NewtonRaphsonSESolver(YBus<sym> const& y_bus, std::shared_ptr<MathModelTopology const> topo_ptr)
+        : n_bus_{y_bus.size()},
+          math_topo_{std::move(topo_ptr)},
+          data_gain_(y_bus.nnz_lu()),
+          del_x_rhs_(y_bus.size()),
```
petersalemink95, line 101: I am not finished with looking at the entire implementation so maybe it might be correct now, just writing down doubts on the way. Shouldn't this be:
```suggestion
          del_x_rhs_(2 * y_bus.size()),
```
Since `delta_x` is a `2 * N_b` vector

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,443 @@ namespace power_grid_model {
...
+        sub_timer = Timer(calculation_info, 2221, "Pre-process measured value");
+        MeasuredValues<sym> const measured_values{y_bus.shared_topology(), input};
+
+        // initialize voltage with initial angle
+        sub_timer = Timer(calculation_info, 2223, "Initialize voltages");
+        RealValue<sym> const mean_angle_shift = measured_values.mean_angle_shift();
+        for (Idx bus = 0; bus != n_bus_; ++bus) {
+            x_[bus].v() = 1.0;
+            x_[bus].theta() = mean_angle_shift + math_topo_->phase_shift[bus];
+            output.u[bus] = exp(1.0i * x_[bus].theta());
```
petersalemink95, line 128: Do we need to have the complex value `U` here? Shouldn't we just calculate with `V` and `theta`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
nitbharambe: The input is the number of blocks in total, hence size of ybus

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,443 @@ namespace power_grid_model {
...
+
+        // loop data index, all rows and columns
+        for (Idx row = 0; row != n_bus_; ++row) {
+            auto const& ui = current_u[row];
+            RealValue<sym> const& abs_ui = x_[row].v();
+            auto const unit_ui = exp(1.0i * x_[row].theta());
+            NRSERhs<sym>& rhs_block = del_x_rhs_[row];
+            rhs_block = NRSERhs<sym>{};
+            NRSEGainBlock<sym>& diag_block = data_gain_[y_bus.lu_diag()[row]];
+            diag_block = NRSEGainBlock<sym>{};
```
petersalemink95, line 193: Why do you initialize the `diag_block` and then immediately make it an empty block?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
petersalemink95: So you have `delta_x` of length `N_b`, where each entry is a `2x1` vector?

nitbharambe: Yeah sort of. Each block already has `4x1` vector.: The theta, v, phi_p, phi_q
The total entries would then be this multiplied by number of buses

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
nitbharambe: It seemed like that there would be an extra multiple calculation if it is not stored. Hence it is stored in both forms, just like NR PF solver.
(Maybe it is not so in the new workout. Will keep/remove as per that.)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
nitbharambe: The above is a reference, a block was needed to write there. It gets initialized only once as there is a condition statement on Line 199-200

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,21 +167,277 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+            // skip for fill-in
+            if (y_bus.map_lu_y_bus()[data_idx_lu] == -1) {
+                continue;
+            }
+            Idx const data_idx_tranpose = y_bus.lu_transpose_entry()[data_idx_lu];
+            data_gain_[data_idx_lu].qt_P_theta() = data_gain_[data_idx_tranpose].q_P_theta();
+            data_gain_[data_idx_lu].qt_P_v() = data_gain_[data_idx_tranpose].q_Q_theta();
+            data_gain_[data_idx_lu].qt_Q_theta() = data_gain_[data_idx_tranpose].q_P_v();
+            data_gain_[data_idx_lu].qt_Q_v() = data_gain_[data_idx_tranpose].q_Q_v();
+        }
```
mgovers, line 362: you can create a separate function for this, e.g. `make_symmetric_from_lower_triangle`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,21 +167,277 @@ template <bool sym> class NewtonRaphsonSESolver {
     SparseLUSolver<NRSEGainBlock<sym>, NRSERhs<sym>, NRSEUnknown<sym>> sparse_solver_;
     typename SparseLUSolver<NRSEGainBlock<sym>, NRSERhs<sym>, NRSEUnknown<sym>>::BlockPermArray perm_;
 
-    void prepare_matrix_and_rhs(YBus<sym> const& /*y_bus*/, MeasuredValues<sym> const& /*measured_value*/,
-                                ComplexValueVector<sym> const& /*current_u*/) {
+    void prepare_matrix_and_rhs(YBus<sym> const& y_bus, MeasuredValues<sym> const& measured_value,
```
mgovers, line 170: the cognitive complexity of this function is out of the park. please split it up

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,21 +167,277 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+
+        // matrix multiplication of F_k^T . w_k . F_k
+        block.g_P_theta() += dot(w_p, block_1.dP_dt, block_2.dP_dt) + dot(w_q, block_1.dQ_dt, block_2.dQ_dt);
+        block.g_P_v() += dot(w_p, block_1.dP_dv, block_2.dP_dv) + dot(w_q, block_1.dQ_dv, block_2.dQ_dv);
+        block.g_Q_theta() += dot(w_p, block_1.dP_dt, block_2.dP_dv) + dot(w_q, block_1.dQ_dt, block_2.dQ_dv);
+        block.g_Q_v() += dot(w_p, block_1.dP_dv, block_2.dP_dt) + dot(w_q, block_1.dQ_dv, block_2.dQ_dt);
+
+        // matrix multiplication of F_k^T . w_k . (z - f(x))
+        rhs_block.eta_theta() += dot(w_p, block_1.dP_dt, del_power_p) + dot(w_q, block_1.dQ_dt, del_power_q);
+        rhs_block.eta_v() += dot(w_p, block_1.dP_dv, del_power_p) + dot(w_q, block_1.dQ_dv, del_power_q);
```
mgovers, line 405: while i feel like this could be implemented differently to prevent repeating yourself, this is very readable and directly translates to the overleaf document :+1: i understand what's happening and it looks clear

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,21 +167,277 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+    auto g_cos_plus_b_sin(ComplexTensor<sym> const& yij, ComplexValue<sym> const& ui, ComplexValue<sym> const& uj) {
+        return real(yij) * ui_uj_cos_ij(ui, uj) + imag(yij) * ui_uj_sin_ij(ui, uj);
+    }
+
     double iterate_unknown(ComplexValueVector<sym>& u) {
         double max_dev = 0.0;
 
         for (Idx bus = 0; bus != n_bus_; ++bus) {
-            // angle
+            // accumulate unknown
```
mgovers, line 435: "unknown" sounds weird, especially when talking about "accumulate unknown". is there another name you can use for this? or maybe make it plural: "unknowns" vs "accumulate unknowns". That actually has a mathematical meaning

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,21 +167,277 @@ template <bool sym> class NewtonRaphsonSESolver {
...
-                                ComplexValueVector<sym> const& /*current_u*/) {
+    void prepare_matrix_and_rhs(YBus<sym> const& y_bus, MeasuredValues<sym> const& measured_value,
+                                ComplexValueVector<sym> const& current_u) {
+        MathModelParam<sym> const& param = y_bus.math_model_param();
+        IdxVector const& row_indptr = y_bus.row_indptr_lu();
+        IdxVector const& col_indices = y_bus.col_indices_lu();
+        // get generated (measured/estimated) voltage phasor
+        // with current result voltage angle
+        // check if this is  the right way or not
+        ComplexValueVector<sym> measured_u = measured_value.voltage(current_u);
```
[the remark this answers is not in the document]
nitbharambe: It is indeed needed everywhere. Spotted a mistake at oncoming lines where i used `ui = current_u[row]`. That should be `ui=measured_u[row]`. Renamed it as well now.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
nitbharambe: Changed to `accumulate the unknown variable`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,21 +167,277 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+
+        // loop data index, all rows and columns
+        for (Idx row = 0; row != n_bus_; ++row) {
+            auto const& ui = current_u[row];
+            auto const& abs_ui_inv = diagonal_inverse(x_[row].v());
+            NRSERhs<sym>& rhs_block = del_x_rhs_[row];
+            rhs_block = NRSERhs<sym>{};
```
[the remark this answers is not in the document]
nitbharambe: Added a clear function to block. Can you check if its appropriate?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/block_matrix.hpp
```diff
@@ -62,6 +62,8 @@ class Block : public block_trait<T, sym, is_tensor, n_sub_block>::ArrayType {
             return (*this)(get_asym_row_idx<r>(), get_asym_col_idx<c>());
         }
     }
+
+    void clear() { *this = Block<T, sym, is_tensor, n_sub_block>{}; }
```
mgovers, line 66: doesn't this work?

```suggestion
    void clear() { *this = {}; }
```

if not, then this is OK

power_grid_model_c/power_grid_model/include/power_grid_model/three_phase_tensor.hpp
```diff
@@ -198,8 +199,9 @@ template <column_vector DerivedA> inline double max_val(Eigen::ArrayBase<Derived
 
 // function to sum rows of tensor
 template <rk2_tensor DerivedA> inline auto sum_row(Eigen::ArrayBase<DerivedA> const& m) { return m.rowwise().sum(); }
-// overload for double
+// overload for double and double complex
 inline double sum_row(double d) { return d; }
+inline DoubleComplex sum_row(DoubleComplex d) { return d; }
```
mgovers, line 204: the docstring only points at the first function after the comment. either have a comment on the newline, or none at all (actually, the comment in the tensor implementation is already redundant), or (inferior) make it into something like

```suggestion
template <typename T>
    requires std::floating_point<T> || std::same_as<T, DoubleComplex>
inline T sum_row(T d) { return d; }
```

mgovers, line 204: maybe should be `const&`

```suggestion
inline DoubleComplex const& sum_row(DoubleComplex const& d) { return d; }
```

but i'm not sure cause it actually "computes" a result and if the user changes the input, this result should not change... of course, it would be the user's responsibility to use `auto` instead of `auto const&` anyways in that case, but... difficult. let's discuss

power_grid_model_c/power_grid_model/include/power_grid_model/three_phase_tensor.hpp
```diff
@@ -340,6 +342,31 @@ inline void set_if_not_nan(RealValue<false>& target, RealValue<false> const& val
...
+    }
+}
+
+template <bool sym> inline RealValue<sym> real_val(ComplexValue<sym> const& c) {
+    if constexpr (sym) {
+        return real(c);
+    } else {
+        return c.real();
+    }
+}
```
mgovers, line 361: remove (`imag(c)` resp `real(c)` should always be used)

power_grid_model_c/power_grid_model/include/power_grid_model/three_phase_tensor.hpp
```diff
@@ -340,6 +342,31 @@ inline void set_if_not_nan(RealValue<false>& target, RealValue<false> const& val
...
+
+template <bool sym> inline RealValue<sym> cabs_or_real(ComplexValue<sym> const& value) {
+    if (is_nan(imag(value))) {
+        return real(value); // only keep real part
+    }
+    return cabs(value); // get abs of the value
+}
+
```
mgovers, line 369: as discussed: move to `iterative_linear_se_solver.hpp`

tests/cpp_unit_tests/test_main_model_se.cpp
```diff
@@ -10,9 +10,17 @@ namespace power_grid_model {
 namespace {
 constexpr double s3 = sqrt3;
 constexpr double ph = 2.0 / 3.0 * pi;
+struct IterativeLinearCalculationMethod {
```
mgovers, line 13: readability
```suggestion

struct IterativeLinearCalculationMethod {
```

tests/cpp_unit_tests/test_main_model_se.cpp
```diff
@@ -10,9 +10,17 @@ namespace power_grid_model {
...
+    static constexpr auto calculation_method = CalculationMethod::newton_raphson;
+};
 } // namespace
 
-TEST_CASE("Test main model - state estimation") {
+TEST_CASE_TEMPLATE("Test main model - state estimation", CalculationMethod, IterativeLinearCalculationMethod,
+                   NewtonRaphsonCalculationMethod) {
+    constexpr auto calculation_method = CalculationMethod::calculation_method;
     MainModel main_model{50.0};
     SUBCASE("State Estimation") {
```
mgovers, line 25: readability:

```suggestion
    constexpr auto calculation_method = CalculationMethod::calculation_method;

    MainModel main_model{50.0};

    SUBCASE("State Estimation") {
```

tests/data/state_estimation/1os2msr-no-angle/params.json
```diff
@@ -6,8 +6,7 @@
     "newton_raphson": {
       "experimental_features": "enabled",
       "fail": {
-        "reason": "Experimental feature under development",
-        "raises": "PowerGridError"
+        "reason": "Experimental feature under development"
```
mgovers, line 9: FYI: @nitbharambe and I agreed that the current implementation is good enough to already review, refactor and merge. Some of the NRSE validation cases were fixed.

A follow-up PR will fix the remaining `xfail`s like this one. That also means that this NRSE implementation is not complete. This is OK because of the `experimental_features` flag that we implemented in #471 : a partial implementation exists but it is still `experimental` and not `stable`.

tests/data/state_estimation/residual-test/params.json
```diff
@@ -7,11 +7,7 @@
   },
   "extra_params": {
     "newton_raphson": {
-      "experimental_features": "enabled",
-      "fail": {
-        "reason": "Experimental feature under development",
-        "raises": "PowerGridError"
-      }
+      "experimental_features": "enabled"
```
mgovers, line 10: :tada:

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +177,325 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                // fill block with voltage measurement, only diagonal
+                if (row == col) {
+                    add_voltage_measurements(block, rhs_block, measured_value, row);
+                }
+                // fill block with branch, shunt measurement
+                for (Idx element_idx = y_bus.y_bus_entry_indptr()[data_idx];
+                     element_idx != y_bus.y_bus_entry_indptr()[data_idx + 1]; ++element_idx) {
+                    Idx const obj = y_bus.y_bus_element()[element_idx].idx;
+                    YBusElementType const type = y_bus.y_bus_element()[element_idx].element_type;
+                    // shunt
```
mgovers, line 227: redundant (the next line literally says `type == YBusElementType::shunt`)
```suggestion

```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +177,325 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                    if (type == YBusElementType::shunt && measured_value.has_shunt(obj)) {
```
mgovers, line 228: can you please rewrite this as a `switch`/`case` statement?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +177,325 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                            multiply_add_jacobian_blocks(block, rhs_block, block_j, block_j, measured_power,
+                                                         calculated_power);
+
+                        } else if (type == YBusElementType::btf) {
+                            auto const jac_complex_unit_other_tf = dot(jac_complex_tf, abs_uj_inv);
+                            auto const block_j = calculate_jacobian(jac_complex_tf, jac_complex_unit_other_tf);
+                            auto block_i = jacobian_diagonal_component(jac_complex_unit_other_tf, calculated_power);
+                            block_i += block_j;
+                            multiply_add_jacobian_blocks(block, rhs_block, block_j, block_i, measured_power,
+                                                         calculated_power);
```
mgovers, line 288: a lot of this amounts to just `i <-> j`, right? can we rewrite this?

mgovers: let's do this in a follow-up PR

TonyXiang8787:
@nitbharambe please do a down merge. There are many things changed in `main`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
```diff
@@ -64,7 +64,7 @@ template <bool sym> class MathSolver {
 
         switch (calculation_method) {
         case default_method:
-            [[fallthrough]]; // use Newton-Raphson by default
+            [[fallthrough]]; // use iterative linear by default
```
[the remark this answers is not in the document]
TonyXiang8787: Can we change the default method to Newton-Raphson? As this is the safest method in general.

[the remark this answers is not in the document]
mgovers: To be honest, I don't know whether we can. I know some people use (sometimes even binary) equivalence tests to verify results in power-grid-model output. We definitely need to communicate and investigate with the stakeholders whether they have any objections.

Regardless of the answer, we definitely should not do this change now as it is still experimental.

nitbharambe:
As this PR was getting too big, we decided to restrict it to everything except line measurements. The validation tests fail for line measurements too. In the coming PR, there will be a unit test on line measurement and the corresponding logic flow.

TonyXiang8787:
> As this PR was getting too big, we decided to restrict it to everything except line measurements. The validation tests fail for line measurements too. In the coming PR, there will be a unit test on line measurement and the corresponding logic flow. Newton Raphson is marked as experimental till then

@mgovers, @nitbharambe, please state in the beginning of this PR what is working and not working after merging of this PR.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +178,312 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                any_zero(measured_values.bus_injection(bus).q_variance)) {
+                x_[bus].phi_p() += del_x_rhs_[bus].phi_p();
+                x_[bus].phi_q() += del_x_rhs_[bus].phi_q();
+            }
+
+            auto const old_u = u[bus];
+            u[bus] = x_[bus].v() * exp(1.0i * x_[bus].theta());
             // get dev of last iteration, get max
-            double const dev = max_val(cabs(u_tmp - u[bus]));
+            double const dev = max_val(cabs((u[bus] - old_u)));
```
mgovers, line 470:
```suggestion
            double const dev = max_val(cabs(u[bus] - old_u));
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +178,312 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                                  MeasuredValues<sym> const& measured_value, Idx const& bus) {
+        if (!measured_value.has_voltage(bus)) {
+            return;
+        }
+
+        // G += 1.0 / variance
+        // for 3x3 tensor, fill diagonal
+        auto const w_v = RealTensor<sym>{1.0 / measured_value.voltage_var(bus)};
+        auto const abs_measured_v = detail::cabs_or_real<sym>(measured_value.voltage(bus));
+        auto const del_v = abs_measured_v - x_[bus].v();
```
mgovers, line 374: again `delta` instead of `del`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +178,312 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+
+        auto w_angle = RealTensor<sym>{1.0};
+        auto del_theta = RealValue<sym>{-x_[bus].theta()};
```
mgovers, line 377: shouldn't this also be `delta` instead of `del`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +178,312 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+        // assign the transpose of the transpose entry of Q
+        make_symmetric_from_lower_triangle(y_bus);
 
         // prefactorize
         sparse_solver_.prefactorize(data_gain_, perm_);
     }
 
-    double iterate_unknown(ComplexValueVector<sym>& u) {
-        double max_dev = 0.0;
+    void make_symmetric_from_lower_triangle(YBus<sym> const& y_bus) {
```
mgovers, line 350: great name!

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +178,312 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                        rhs_block.tau_q() += injection.value.imag();
+
+                        block.r_P_theta() -=
+                            RealTensor<sym>{RealValue<sym>{RealValue<sym>{1.0} / injection.p_variance}};
+                        block.r_Q_v() -= RealTensor<sym>{RealValue<sym>{RealValue<sym>{1.0} / injection.q_variance}};
+                    }
+
+                }
+                // injection measurement not exist
+                else {
```
mgovers, line 325: the if statement already suggests this and the comment already explains what should happen when there's no injection measurement

```suggestion
                } else {
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +178,312 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                            auto const block_i = calculate_jacobian(jac_complex_tf, jac_complex_unit_other_tf);
+                            multiply_add_jacobian_blocks(block, rhs_block, block_j, block_i, measured_power,
+                                                         calculated_power);
+                        }
+                    }
+                }
+
+                // fill block with injection measurement
+                // injection measurement exist
+                if (measured_value.has_bus_injection(row)) {
```
mgovers, line 293: redundant
```suggestion
                // fill block with injection measurement constraints
                if (measured_value.has_bus_injection(row)) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +178,312 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                    // R_ii = -variance, only diagonal
+                    if (row == col) {
+                        // assign variance to diagonal of 3x3 tensor, for asym
+                        auto const& injection = measured_value.bus_injection(row);
+                        rhs_block.tau_p() += injection.value.real();
+                        rhs_block.tau_q() += injection.value.imag();
+
+                        block.r_P_theta() -=
+                            RealTensor<sym>{RealValue<sym>{RealValue<sym>{1.0} / injection.p_variance}};
+                        block.r_Q_v() -= RealTensor<sym>{RealValue<sym>{RealValue<sym>{1.0} / injection.q_variance}};
```
mgovers, line 320: i believe i have written a function for this called `diagonal_inverse` in `iterative_linear_se`. that one has to be `ComplexTensor` though instead of `RealTensor` so it may not fit our needs, but i would suggest to make a similar function in NRSE.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
mgovers: comments need to be made on that PR as well before closing this one

mgovers (commented):
i will probably have time to resolve or migrate the comments tomorrow morning. I will also review mathematical correctness then.

mgovers:
* Code review OK as far as I'm concerned
* All remaining comments that we decided to move out of scope are re-commented on #478 .
* Mathematics still need to be cross-checked

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -117,12 +118,12 @@ template <bool sym> class IterativeLinearSESolver {
             sub_timer = Timer(calculation_info, 2225, "Solve sparse linear equation (pre-factorized)");
             sparse_solver_.solve_with_prefactorized_matrix(data_gain_, perm_, x_rhs_, x_rhs_);
             sub_timer = Timer(calculation_info, 2226, "Iterate unknown");
-            max_dev = iterate_unknown(output.u, measured_values.has_angle_measurement());
+            max_dev = iterate_unknown(output.u, measured_values.has_angle());
```
petersalemink95, line 121: Why is this renamed to `has_angle`, when the function is to check if it has an angle measurements (implementation: `constexpr bool has_angle() const { return n_voltage_angle_measurements_ > 0; }`)

mgovers: to keep the similarity with `has_voltage`. we can also bump both to `has_*_measurement`

petersalemink95: Both are fine by me, so we can keep it as is. Good idea to keep similarity :)

TonyXiang8787:
@nitbharambe how far is this PR? I see the CI still fails.

nitbharambe:
> @nitbharambe how far is this PR? I see the CI still fails.

Its still WIP. Some validation tests are failing which werent before fixes.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -454,9 +472,23 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                return 0.0;
+            }
+            auto const& theta = x_[math_topo_->slack_bus].theta();
+            if constexpr (sym) {
+                return theta;
+            } else {
+                return theta(0);
+            }
+        }();
+
```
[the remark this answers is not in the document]
mgovers: nvm, i wrote created it XD

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -170,31 +171,352 @@ template <bool sym> class NewtonRaphsonSESolver {
...
-                                ComplexValueVector<sym> const& /*current_u*/) {
+    void initialize_unknown(ComplexValueVector<sym>& initial_u, MeasuredValues<sym> const& measured_values) {
+        reset_unknown();
+        RealValue<sym> const mean_angle_shift = measured_values.mean_angle_shift();
+        for (Idx bus = 0; bus != n_bus_; ++bus) {
+            x_[bus].theta() = mean_angle_shift + math_topo_->phase_shift[bus];
+            if (measured_values.has_voltage(bus)) {
+                if (measured_values.has_angle_measurement(bus)) {
+                    x_[bus].theta() = arg(measured_values.voltage(bus));
+                }
```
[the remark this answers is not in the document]
mgovers: I would indeed expect (as in the code) that
$$
\theta_i^{(0)} = \theta_i
$$
but would like a double-check

TonyXiang8787:
@mgovers are validation tests passing? According to @nitbharambe there are still failures?

mgovers:
> @mgovers are validation tests passing? According to @nitbharambe there are still failures?

There are 2 validation cases still failing for which @nitbharambe claims are likely due to different convergence points. In both cases, the algorithm converges and some residuals are less while others are more, hence his conclusion.

@nitbharambe , @petersalemink95 and I agreed on the following approach:
1. Only do code review on existing functionality + code quality on this branch
2. Merge this branch - release the NRSE as an experimental `beta` feature. This can be communicated to the stakeholders as such
3. Follow-up PRs to add 2 separate validation cases on the currently failing ones that have the new saturation point as expected output
4. Mathematical review on `main` after merge.
5. After all these steps are done, the `experimental` flag can be removed and the full feature can be released as `v1.7.0`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
Jerry-Jinfeng-Guo, line 1: This document primarily focuses on mathematical content of the solver, an area in which I am not completely knowledgeable. I don’t have any specific remarks regarding the code itself.

Jerry-Jinfeng-Guo (commented):
My two cents on the code only.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/measured_values.hpp
```diff
@@ -15,6 +15,15 @@ Collect all measured Values
 
 namespace power_grid_model::math_solver {
 
+namespace detail {
+template <bool sym> inline RealValue<sym> cabs_or_real(ComplexValue<sym> const& value) {
+    if (is_nan(imag(value))) {
+        return real(value); // only keep real part
+    }
+    return cabs(value); // get abs of the value
```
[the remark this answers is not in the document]
mgovers: this logic should go OK because there's no division involved. `imag == 0` happens so rarely that even just evaluating the `if` statement every call would be overhead (branch prediction). since there's no division before taking the square of the norm, this will always be positive (by IEEE-754 floating point standard) regardless of architecture

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/measured_values.hpp
```diff
@@ -149,6 +139,26 @@ template <bool sym> class MeasuredValues {
...
+    // for magnitude and angle measurement, the measured phasor is used
+    ComplexValueVector<sym>
+    combine_voltage_iteration_with_measurements(ComplexValueVector<sym> const& current_u) const {
+        ComplexValueVector<sym> u(current_u.size());
+
+        for (Idx bus = 0; bus != static_cast<Idx>(current_u.size()); ++bus) {
+            if (!has_voltage(bus)) { // no measurement
+                u[bus] = current_u[bus];
+            } else if (!has_angle_measurement(bus)) {
+                u[bus] = real(voltage(bus)) * current_u[bus] / cabs(current_u[bus]); // U / |U| to get angle shift
```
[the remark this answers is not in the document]
mgovers: fixed

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -117,12 +118,12 @@ template <bool sym> class IterativeLinearSESolver {
...
         };
 
         // calculate math result
         sub_timer = Timer(calculation_info, 2227, "Calculate Math Result");
```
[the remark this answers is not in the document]
mgovers: fixed (this was copy-pasted from other solvers and it's solved there too

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+                    }
+                    default:
+                        assert(type == YBusElementType::bff || type == YBusElementType::btt);
+                        break;
+                    }
+                }
+
+                // fill block with injection measurement constraints
+                if (measured_values.has_bus_injection(row)) {
+                    auto const& yij = y_bus.admittance()[data_idx];
```
Jerry-Jinfeng-Guo, line 302: This line accesses the element `(i,j)` in the y_bus admittance matrix, is `data_idx` then the correct index to use here?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+
+                block.qt_P_theta() = data_gain_[data_idx_transpose].q_P_theta();
+                block.qt_P_v() = data_gain_[data_idx_transpose].q_Q_theta();
+                block.qt_Q_theta() = data_gain_[data_idx_transpose].q_P_v();
+                block.qt_Q_v() = data_gain_[data_idx_transpose].q_Q_v();
+            },
+            y_bus);
+    }
+
+    /// @brief Process Lagrange Multiplier eta_i(i) = sum( Q^T(j,i) * phi(j) ) for j = 1 to n_bus
```
Jerry-Jinfeng-Guo, line 464:
```suggestion
    /// @brief Process Lagrange multiplier eta_i(i) = sum( Q^T(j,i) * phi(j) ) for j = 1 to n_bus
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
     }
 
-    double iterate_unknown(ComplexValueVector<sym>& u) {
+    /// Q_ij = 0
+    /// R_ii = -1.0, only diagonal
+    /// assign -1.0 to diagonal of 3x3 tensor, for asym
+    void virtually_remove_constraints(NRSEGainBlock<sym>& block) {
+        block.r_P_theta() = RealTensor<sym>{-1.0};
+        block.r_Q_v() = RealTensor<sym>{-1.0};
+    }
```
Jerry-Jinfeng-Guo, line 326: Not sure if this function actually does what's claimed: doesn't seem to distinguish `asym` from `sym` and it does not seem to do anything to diagonal

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+
+    /// R_ii = -variance, only diagonal
+    /// assign variance to diagonal of 3x3 tensor, for asym
+    void process_injection_diagonal(NRSEGainBlock<sym>& block, NRSERhs<sym>& rhs_block, auto const& injection) {
+        rhs_block.tau_p() += injection.value.real();
+        rhs_block.tau_q() += injection.value.imag();
+        block.r_P_theta() = RealTensor<sym>{RealValue<sym>{-injection.p_variance}};
+        block.r_Q_v() = RealTensor<sym>{RealValue<sym>{-injection.q_variance}};
+    }
```
Jerry-Jinfeng-Guo, line 335: Similar complaint about this function as the function above ^

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+        auto const nl_ui_ui_ys = dot(hm_ui_ui_ys, u_state.abs_ui_inv);
+        auto const f_x_complex = sum_row(hm_ui_ui_ys);
+        auto const f_x_complex_abs_ui_inv = sum_row(nl_ui_ui_ys);
+
+        auto jac_block = calculate_jacobian(hm_ui_ui_ys, nl_ui_ui_ys);
+        jac_block += jacobian_diagonal_component(f_x_complex_abs_ui_inv, f_x_complex);
+        auto const& block_F_T_k_w = transpose_multiply_weight(jac_block, measured_power);
+        multiply_add_jacobian_blocks_lhs(block, block_F_T_k_w, jac_block);
+        multiply_add_jacobian_blocks_rhs(rhs_block, block_F_T_k_w, measured_power, f_x_complex);
+    }
```
Jerry-Jinfeng-Guo, line 384: I think the `i` suffix added to variable names can be dropped:
 - the context is clear that it is for diagonal blocks
 - there is no `j` to differrentiate from

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+    ///              xi = f, mu = t, chi = col, psi = row
+    ///          if to_measurement,
+    ///              xi = t, mu = f, chi = row, psi = col
+    ///
+    /// f_P_(x) = -M(U_chi, U_chi, Y_xi_xi) - M(U_chi, U_psi, Y_xi_mu)
+    /// f_Q_(x) = H(U_chi, U_chi, Y_xi_xi) + H(U_chi, U_psi, Y_xi_mu)
+    ///
+    ///
+    /// @param block G_(r, c)
+    /// @param diag_block G_(r, r)
```
Jerry-Jinfeng-Guo, line 406: Here, what is the purpose of adding `_` to `G` in  math context and basically everywhere else uses `G` directly?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+    /// @param rhs_block RHS(r)
+    /// @param y_xi_xi shunt admittance near to branch measurement
+    /// @param y_xi_mu admittance from the branch measurement to other bus
+    /// @param u_state Voltage state of iteration voltage state vector
+    /// @param order bool to determine if (chi, psi) = (row, col) or (col, row)
```
Jerry-Jinfeng-Guo, line 411: Would it make sense to make this bool `order` an enum to be more descriptive? Say `row_major` and `column_major`? Or perhaps make the bool variable name more descriptive, e.g., `row_ordered`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+
+                Idx const col = y_bus.col_indices_lu()[data_idx];
+                Idx const data_idx_transpose = y_bus.lu_transpose_entry()[data_idx];
+
+                func(row, col, data_idx, data_idx_transpose);
+            }
+        }
+    }
+
+    /// @brief G(row,row) += w_k
```
Jerry-Jinfeng-Guo, line 500:
```suggestion
    /// @brief G(row, row) += w_k
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+    /// eta(row) += w_k . (z_k - f_k(x))
```
Jerry-Jinfeng-Guo, line 501: Here and below: the dot product is denoted with `.`, which is reasonable. However, in other existing documents, e.g.,  `iterative_linear_se_solver.hpp`, the dot product is denoted simply with `*`. Not sure if either one is better (in my view it should be \cdot) but better to be consistent?

Jerry-Jinfeng-Guo (commented):
Reviews on the `nr_se_solver` document.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
mgovers: imho out of scope of this PR. it's going to be a huge change to change one or the other and it's only for internal documentation anyways

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
mgovers: i think for consistency keeping it as is is OK. we can always change this in the future.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
mgovers: the constructor of `RealTensor<sym>` creates a diagonal object. E.g. `1` has the same behaviour as the unit matrix. `RealTensor` by definition is a `double` for `sym == true` and a tensor for `sym == false`. this is just us making our lives easier

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
mgovers: yes cause `data_idx` by design is the data element that accesses the sparse matrix `y_bus.admittance()` in the correct data entry to obtain the element representing the element at `(i,j)`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -163,38 +178,506 @@ template <bool sym> class NewtonRaphsonSESolver {
...
+            }
+            initial_u[bus] = estimated_result.v() * exp(1.0i * estimated_result.theta());
+        }
+    }
+
+    void reset_unknown() {
+        auto const default_unknown = [] {
+            NRSERhs<sym> x;
+            x.v() = 1.0;
+            x.theta() = 0.0;
```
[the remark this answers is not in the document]
mgovers: indeed. see the `note` in https://power-grid-model.readthedocs.io/en/stable/user_manual/calculations.html#state-estimation

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -136,14 +151,14 @@ template <bool sym> class NewtonRaphsonSESolver {
...
         };
 
         // calculate math result
-        sub_timer = Timer(calculation_info, 2227, "Calculate Math Result");
-        // calculate_result(y_bus, measured_values, output);
+        sub_timer = Timer(calculation_info, 2227, "Calculate math result");
+        detail::calculate_se_result<sym>(y_bus, measured_values, output);
 
         // Manually stop timers to avoid "Max number of iterations" to be included in the timing.
         sub_timer.stop();
```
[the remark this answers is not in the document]
mgovers: arbitrarily set and yes, i agree, we should come up with a better way to do this. I consider this out of scope of this PR. no benchmarking is being done yet.

Most importantly, timing happens differently per solver

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
Jerry-Jinfeng-Guo: Agreed. Let's keep it for later discussion.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
Jerry-Jinfeng-Guo: Awesomo.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Agreed. Let's keep this for later discussion.

Jerry-Jinfeng-Guo (approved):
I have no more complaints, please approve my approval.

mgovers (approved):

mgovers (approved):

mgovers (dismissed):

mgovers (approved):

PR #440, opened 2023-11-27
Feature/minimum degree topology

### LINKED ISSUE #433: [FEATURE] In-house minimum degree reordering
# Introduction

Currently we use [`boost::minimum_degree_ordering`](https://www.boost.org/doc/libs/1_83_0/libs/graph/doc/minimum_degree_ordering.html) to re-order the nodes in the cyclic part of the grid. It has some shortcomings:

1. The library is long not maintained with many compilation warnings. We have to tweak a lot on the compiler.
2. The documentation is not very clear about some settings.
3. We also need the list of fill-ins after the reordering. The function does not products that output. So we need to build the graph again with the re-ordered vertices and manually track the fill-ins.

# Proposal

We would like to implement our in-house minimum degree ordering to replace the `boost::minimum_degree_ordering`. Essentially, replace the following code snippert to a separate function module of our own in-house solution.

https://github.com/PowerGridModel/power-grid-model/blob/95212803cd9af486066deeea4d334b558ccf6e1a/power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp#L313-L404

mgovers:
Fixes #433


power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
```diff
@@ -11,13 +11,16 @@
...
+#include "sparse_ordening.hpp"
 
 #include <boost/graph/adjacency_list.hpp>
 #include <boost/graph/compressed_sparse_row_graph.hpp>
 #include <boost/graph/depth_first_search.hpp>
 #include <boost/graph/iteration_macros.hpp>
 #include <boost/graph/minimum_degree_ordering.hpp>
 
+#include <iostream>
+
```
mgovers, line 23: TODO: remove

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+inline void remove_element_vector_pair(Idx& u, std::vector<std::pair<Idx, Idx>>& dgd) {
+    Idx i = 0;
+    while (i < dgd.size()) {
+        if (dgd[i].first == u) {
+            dgd.erase(dgd.begin() + i);
+            break;
+        } else {
+            i++;
+        }
+    }
```
mgovers, line 29: improve implementation, e.g.

```suggestion
    std::erase_if(dgd, [&u](auto const& v) { v.first == u});
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+inline void set_element_vector_pair(Idx& u, Idx v, std::vector<std::pair<Idx, Idx>>& dgd) {
+    Idx i = 0;
+    while (i < dgd.size()) {
+        if (dgd[i].first == u) {
+            dgd[i].second = v;
+            break;
+        } else {
+            i++;
+        }
+    }
```
mgovers, line 41:
```suggestion
    if (auto& it = std::ranges::find_if(dgd, [&u](auto const& value) { value.first == u; }; auto != dgd.end()) {
        it->second = v;
    }
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+inline std::vector<Idx> adj(Idx& u, std::map<Idx, std::vector<Idx>>& d) {
+    std::vector<Idx> l;
+
+    for (const auto& it : d) {
+        if (it.first == u)
+            l.insert(l.end(), it.second.begin(), it.second.end());
```
mgovers, line 49:
```suggestion
        if (it.first == u) {
            l.insert(l.end(), it.second.begin(), it.second.end());
        }
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+        if (find(it.second.begin(), it.second.end(), u) != it.second.end()) {
+            std::vector<Idx> v{it.first};
+            l.insert(l.end(), v.begin(), v.end());
```
mgovers, line 53:
```suggestion
            std::vector<Idx> v{it.first};
            l.push_back(it.first);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+inline std::vector<Idx> adj(Idx& u, std::map<Idx, std::vector<Idx>>& d) {
+    std::vector<Idx> l;
+
+    for (const auto& it : d) {
+        if (it.first == u)
+            l.insert(l.end(), it.second.begin(), it.second.end());
+
+        if (find(it.second.begin(), it.second.end(), u) != it.second.end()) {
```
mgovers, line 51:
```suggestion
        if (std::ranges::find(it.second, u) != it.second.cend()) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+            std::vector<Idx> v{it.first};
+            l.insert(l.end(), v.begin(), v.end());
+        }
+    }
+
+    return l;
+}
+
+inline bool compair_ids(std::pair<Idx, Idx>& a, std::pair<Idx, Idx>& b) { return a.first < b.first; }
+
```
mgovers, line 61:
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+inline std::vector<std::pair<Idx, std::vector<std::pair<Idx, Idx>>>>
+comp_size_degrees_graph(std::map<Idx, std::vector<Idx>>& d) {
+    std::vector<std::pair<Idx, Idx>> dd;
+    std::vector<Idx> v;
+    Idx n = 0;
+
+    for (const auto& it : d) {
+        Idx k = it.first;
+        if (find(v.begin(), v.end(), k) == v.end()) {
```
mgovers, line 70:
```suggestion
        if (std::ranges::find(v, k) == v.cend()) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+            std::vector<Idx> vk{k};
+            v.insert(v.end(), vk.begin(), vk.end());
+            n += 1;
+            dd.push_back({k, adj(k, d).size()});
```
mgovers, line 74: try:

```suggestion
            dd.emplace_back(k, adj(k, d).size());
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+comp_size_degrees_graph(std::map<Idx, std::vector<Idx>>& d) {
+    std::vector<std::pair<Idx, Idx>> dd;
+    std::vector<Idx> v;
+    Idx n = 0;
+
+    for (const auto& it : d) {
+        Idx k = it.first;
+        if (find(v.begin(), v.end(), k) == v.end()) {
+            std::vector<Idx> vk{k};
+            v.insert(v.end(), vk.begin(), vk.end());
```
mgovers, line 72:
```suggestion
            v.push_back(k);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+        }
+        for (const Idx& el : it.second) {
+            Idx e = el;
+            if (find(v.begin(), v.end(), e) == v.end()) {
+                std::vector<Idx> ve{e};
+                v.insert(v.end(), ve.begin(), ve.end());
+                n += 1;
+                dd.push_back({e, adj(e, d).size()});
+            }
+        }
```
mgovers, line 84:
```suggestion
        for (Idx e : it.second) {
            if (std::ranges::find(v, e) == v.cend()) {
                v.push_back(e);
                n += 1;
                dd.emplace_back(e, adj(e, d).size());
            }
        }
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    sort(dd.begin(), dd.end());
```
mgovers, line 87:
```suggestion
    std::ranges::sort(dd);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+    l = adj(u, d);
+    lu = l;
+    lu.insert(lu.end(), vu.begin(), vu.end());
+
+    for (auto& v : l) {
+        lv = adj(v, d);
+        vv = {v};
+        lv.insert(lv.end(), vv.begin(), vv.end());
+        sort(lu.begin(), lu.end());
+        sort(lv.begin(), lv.end());
```
mgovers, line 104:
```suggestion
        std::ranges::sort(lu);
        std::ranges::sort(lv);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+    return {{n, dd}};
+}
+
+inline std::vector<std::pair<std::vector<Idx>, std::vector<Idx>>>
+check_indistguishable(Idx& u, std::map<Idx, std::vector<Idx>>& d) {
+    std::vector<Idx> l, rl, lu, lv, vu{u}, vv;
+    l = adj(u, d);
+    lu = l;
+    lu.insert(lu.end(), vu.begin(), vu.end());
```
mgovers, line 97:
```suggestion
    std::vector<Idx> rl;

    auto l = adj(u, d);
    auto lu = l;
    lu.push_back(lu, u);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+    for (auto& v : l) {
+        lv = adj(v, d);
```
mgovers, line 100:
```suggestion
        auto lv = adj(v, d);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+        vv = {v};
+        lv.insert(lv.end(), vv.begin(), vv.end());
```
mgovers, line 102:
```suggestion
        lv.push_back(v);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+        sort(lu.begin(), lu.end());
+        sort(lv.begin(), lv.end());
+        if (lu == lv) {
+            rl.insert(rl.end(), vv.begin(), vv.end());
```
mgovers, line 106:
```suggestion
            rl.push_back(v);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+inline std::map<Idx, std::vector<Idx>> make_clique(std::vector<Idx>& l) {
+    std::map<Idx, std::vector<Idx>> d;
+    Idx b = l.size() - 1;
+
+    for (int i = 0; i < b; i++) {
+        Idx index = i + 1;
+        auto start = l.begin() + index;
+        std::vector<Idx> sl(l.size() - index);
+        copy(start, l.end(), sl.begin());
```
mgovers, line 121:
```suggestion
        auto const idx = i + 1
        std::vector<Idx> sl(l.size() - idx);
        std::copy(l.begin() + idx, l.end(), sl.begin());
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+    dd = make_clique(nbs);
+
+    for (auto& it : dd) {
+        Idx k = it.first;
+        for (const Idx& e : it.second) {
+            std::pair<Idx, Idx> t{k, e};
+            if (!in_graph(t, d)) {
+                if (d.find(k) != d.end()) {
+                    std::vector<Idx> ve{e};
+                    d[k].insert(d[k].end(), ve.begin(), ve.end());
```
mgovers, line 184:
```suggestion
                    d[k].push_back(e);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+                    fills.push_back({k, e});
+                } else if (d.find(e) != d.end()) {
+                    std::vector<Idx> vk{k};
+                    d[e].insert(d[e].end(), vk.begin(), vk.end());
```
mgovers, line 188:
```suggestion
                    std::vector<Idx> vk{k};
                    d[e].push_back(k);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+                    fills.push_back({e, k});
+                } else {
+                    std::vector<Idx> ve{e};
+                    d[k].insert(d[k].end(), ve.begin(), ve.end());
```
mgovers, line 192:
```suggestion
                    d[k].push_back(e);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+                    fills.push_back({k, e});
```
mgovers, line 193:
```suggestion
                    fills.emplace_back(k, e);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+            std::pair<Idx, Idx> t{k, e};
+            if (!in_graph(t, d)) {
+                if (d.find(k) != d.end()) {
+                    std::vector<Idx> ve{e};
+                    d[k].insert(d[k].end(), ve.begin(), ve.end());
+                    fills.push_back({k, e});
+                } else if (d.find(e) != d.end()) {
+                    std::vector<Idx> vk{k};
+                    d[e].insert(d[e].end(), vk.begin(), vk.end());
+                    fills.push_back({e, k});
```
mgovers, line 189:
```suggestion
                    fills.emplace_back(e, k);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+    for (auto& it : dd) {
+        Idx k = it.first;
+        for (const Idx& e : it.second) {
+            std::pair<Idx, Idx> t{k, e};
+            if (!in_graph(t, d)) {
+                if (d.find(k) != d.end()) {
+                    std::vector<Idx> ve{e};
+                    d[k].insert(d[k].end(), ve.begin(), ve.end());
+                    fills.push_back({k, e});
```
mgovers, line 185:
```suggestion
                    fills.emplace_back(k, e);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+        if ((d.size() == 1) and d.begin()->second.size() == 1) {
+            Idx a = d.begin()->first;
+            Idx b = d.begin()->second[0];
+            if (alpha.back() == a) {
+                std::vector<Idx> vb{b};
+                alpha.insert(alpha.end(), vb.begin(), vb.end());
+            } else {
+                std::vector<Idx> va{a};
+                alpha.insert(alpha.end(), va.begin(), va.end());
+            }
```
mgovers, line 229:
```suggestion
            alpha.push_back(alpha.back() == a ? b : a);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+    Idx& n = data[0].first;
+    std::vector<std::pair<Idx, Idx>>& dgd = data[0].second;
+    std::vector<Idx> alpha;
+    std::vector<std::pair<Idx, Idx>> fills;
+
+    for (int k = 0; k < n; k++) {
+        Idx u =
+            get<0>(*min_element(begin(dgd), end(dgd), [](auto lhs, auto rhs) { return get<1>(lhs) < get<1>(rhs); }));
+        std::vector<Idx> vu{u};
+        alpha.insert(alpha.end(), vu.begin(), vu.end());
```
mgovers, line 219:
```suggestion
        alpha.push_back(u);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+                std::vector<Idx> vb{b};
+                alpha.insert(alpha.end(), vb.begin(), vb.end());
+            } else {
+                std::vector<Idx> va{a};
+                alpha.insert(alpha.end(), va.begin(), va.end());
+            }
+            break;
+        } else {
+            std::vector<Idx> va = detail::remove_vertices_update_degrees(u, d, dgd, fills);
+            alpha.insert(alpha.end(), va.begin(), va.end());
```
mgovers, line 233:
```suggestion
            std::ranges::copy(detail::remove_vertices_update_degrees(u, d, dgd, fills), std::back_inserter(alpha));
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+            it.second.erase(remove(it.second.begin(), it.second.end(), uu), it.second.end());
+            if (it.second.empty()) {
+                Idx k = it.first;
+                std::vector<Idx> vk{k};
+                el.insert(el.end(), vk.begin(), vk.end());
+            }
+        }
+
+        std::vector<Idx> vuu{uu};
+        el.insert(el.end(), vuu.begin(), vuu.end());
```
mgovers, line 169:
```suggestion
        el.push_back(uu);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+        for (auto& it : el)
+            d.erase(it);
```
mgovers, line 172:
```suggestion
        for (auto& it : el) {
            d.erase(it);
        }
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+            nbs.erase(remove(nbs.begin(), nbs.end(), uu), nbs.end());
+
+        remove_element_vector_pair(uu, dgd);
+        std::vector<Idx> el;
+        for (auto& it : d) {
+            it.second.erase(remove(it.second.begin(), it.second.end(), uu), it.second.end());
+            if (it.second.empty()) {
+                Idx k = it.first;
+                std::vector<Idx> vk{k};
+                el.insert(el.end(), vk.begin(), vk.end());
```
mgovers, line 164:
```suggestion
                el.push_back(k);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+    rl.insert(rl.begin(), vu.begin(), vu.end());
+
+    for (auto& uu : rl) {
+        if (uu != u)
+            nbs.erase(remove(nbs.begin(), nbs.end(), uu), nbs.end());
+
+        remove_element_vector_pair(uu, dgd);
+        std::vector<Idx> el;
+        for (auto& it : d) {
+            it.second.erase(remove(it.second.begin(), it.second.end(), uu), it.second.end());
```
mgovers, line 160:
```suggestion
            std::erase(it.second, uu);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,243 @@
...
+inline std::vector<Idx> remove_vertices_update_degrees(Idx& u, std::map<Idx, std::vector<Idx>>& d,
+                                                       std::vector<std::pair<Idx, Idx>>& dgd,
+                                                       std::vector<std::pair<Idx, Idx>>& fills) {
+    std::vector<std::pair<std::vector<Idx>, std::vector<Idx>>> nbsrl = check_indistguishable(u, d);
+    std::vector<Idx>& nbs = nbsrl[0].first;
+    std::vector<Idx>& rl = nbsrl[0].second;
+    std::vector<Idx> alpha = rl, vu{u};
+    std::map<Idx, std::vector<Idx>> dd;
+
+    rl.insert(rl.begin(), vu.begin(), vu.end());
```
mgovers, line 151:
```suggestion
    std::map<Idx, std::vector<Idx>> dd;

    rl.push_back(u);
```

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        el.push_back(uu);
+
+        for (auto const& it : el) {
+            d.erase(it);
+        }
+    }
+
+    dd = make_clique(nbs); // TODO check data type
```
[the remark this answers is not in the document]
mgovers: Marc says that `std::unordered_map` closer is to his current implementation in Python

tests/cpp_unit_tests/test_sparse_ordening.cpp
TonyXiang8787, line 1: is ordering, not ordening.

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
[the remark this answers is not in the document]
TonyXiang8787: Given the number of edges + fill-ins is relatively low (3-5 for most of the vertices, max 20-30 in big nodes), my proposal is to use `vector`. The theoretical big-O performance of `map` or `unordered_map` cannot compensate for their complexity and cache unfriendly.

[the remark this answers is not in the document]
mgovers: part of the difficulty also lies in the fact that this induces yet another remapping of the edges before computation can start. will think about how this impacts comprehensiveness.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -608,12 +608,15 @@ TEST_CASE("Validation test single") {
 TEST_CASE("Validation test batch") {
     std::vector<CaseParam> const& all_cases = get_all_batch_cases();
     for (CaseParam const& param : all_cases) {
-        SUBCASE(param.case_name.c_str()) {
-            try {
-                validate_batch_case(param);
-            } catch (std::exception& e) {
-                auto const msg = std::string("Unexpected exception with message: ") + e.what();
-                FAIL_CHECK(msg);
+        if (param.case_name ==
```
TonyXiang8787, line 611: Why do we have this for a specific validation case?

mgovers: test logic. to be disabled

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordering.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+    for (auto const& e : nbs) {
+        set_element_vector_pair(e, static_cast<Idx>(adj(e, d).size()), dgd);
+    }
+
+    return alpha;
+}
+} // namespace detail
+
+inline std::pair<IdxVector, std::vector<std::pair<Idx, Idx>>> minimum_degree_ordering(std::map<Idx, IdxVector>& d) {
```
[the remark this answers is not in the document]
mgovers: i'm working on changing `map` to `vector`. i'm not too sure about the `boost` graph structure, though. i've noticed that the `boost` graph structure is not nice to do basic operations like looping. i can try to re-add but that will introduce quite some significant code changes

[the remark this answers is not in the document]
TonyXiang8787: If it becomes too difficult to use `boost` graph, then not.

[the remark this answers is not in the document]
mgovers: due to mapping from `cyclic_node` to indices (required for the `vector` approach), that change will take a little bit more time. also not sure how nice the final solution will be due to the amount of intermediate objects, which probably will require some refactoring as well to make more efficient

TonyXiang8787:
@mgovers how far are we close to merge this?

mgovers:
> @mgovers how far are we close to merge this?

Cfr. our discussions, we run the benchmark case and if the amount of fill-ins is the same for both implementations, it's ready for merge.

Together with @petersalemink95 and the rest of the team, this work has been refined, estimated and prioritized and is waiting to be picked up. However, we have a couple high-prio things that need to be picked up first.

mgovers:
> Cfr. our discussions, we run the benchmark case and if the amount of fill-ins is the same for both implementations, it's ready for merge.

The amount of fill-ins differs on the benchmark case: for the meshed grid cases, all runs return 91 fill-ins for the current `main` and 98 fill-ins for this branch.

Further investigation needed

mgovers:
cfr. discussion: `boost` and this branch use slightly different choices regarding the chosen point, but they scale similarly:

* sometimes, `boost` is better (e.g. benchmark case for large grids),
* at other times, this branch provides less fill-ins (e.g. benchmark case for small grids)
* they both are not necessarily optimal solutions

An efficient exact optimal minimum degree algorithm is proposed in https://epubs.siam.org/doi/epdf/10.1137/1.9781611976465.45 that could provide speed-ups.
Simpler solutions also exist, like randomly trying n vertices with the same minimal degree instead of only a single arbitrary vertex of minimum degree (e.g. the first, which is the one implemented in this branch)

For now, this solution seems to be OK and it is ready for review and merge as is.

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordening.hpp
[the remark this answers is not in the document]
mgovers: the current state seems to be OK

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordering.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+    auto l = adj(u, d);
+    auto lu = l;
+    lu.push_back(u);
+
+    for (auto const& v : l) {
+        auto lv = adj(v, d);
+        lv.push_back(v);
+        std::ranges::sort(lu);
+        std::ranges::sort(lv);
```
Jerry-Jinfeng-Guo, line 92: This `sort` right after appending looks less than ideal

mgovers: i think they have the same complexity, but since the amount of nodes may be quite large, appending before sorting is usually better. see e.g. https://stackoverflow.com/questions/168891/is-it-faster-to-sort-a-list-after-inserting-items-or-adding-them-to-a-sorted-lis .

if we really want to go for it here, using a heap (as also mentioned there) might be even more optimal, but in this case, that's not where the complexity is

Jerry-Jinfeng-Guo: Sorting after appending each item results in an arithmetic series, i.e., N*(N+1)/2 many in the best case.

mgovers: after investigating, this is actually not that bad, but you're right in that the logic is convoluted. I will investigate to see whether I can make it more readable (and maybe more efficient, as well).

As to why it is not immediately terrible: the vector only contains adjacent edges, and the list of adjacent edges is expected to be small. In addition, the vector is newly created, so `push_back` and then sorting is, in fact, the most efficient way. That said, I don't see any reason why the full list of adjacent edges for every adjacent edge needs to be recreated every single time, but I will look into that.

mgovers: the only trivially straightforward optimization i see is to do the sorting of `lu` outside the loop. all the rest would probably result in more convoluted code

Jerry-Jinfeng-Guo: Indeed

mgovers:
FINALLY
image
image

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordering.hpp
```diff
@@ -0,0 +1,221 @@
...
+    // make symmetric
+    auto d = d_;
+    for (auto& [k, adjacent] : d_) {
+        for (auto a : adjacent) {
+            d[a].push_back(k);
+        }
+    }
+    for (auto& [k, adjacent] : d) {
+        std::ranges::sort(adjacent);
+    }
```
mgovers, line 195: i can also do this in the data prep cause this is useless overhead, but it will change the input and therefore the tests

Jerry-Jinfeng-Guo: Nice improvement so far! I think we can merge this PR once ci is fixed

Jerry-Jinfeng-Guo (commented):
I will approve the PR once ci is fixed and code cleaned

power_grid_model_c/power_grid_model/include/power_grid_model/sparse_ordering.hpp
mgovers: i decided to remove the copy here, and instead go with dedup using `set` to fix this. that basically removes the precondition that this piece of code introduced without a large weird copy on the input

Jerry-Jinfeng-Guo (approved):

PR #441, opened 2023-11-29
Feature/nrse interface

mgovers:
add Newton-Raphson for state estimation as an experimental internal-only calculation method.

NOTE: this is the interface only. actual implementations in #438 

Relates to #369


nitbharambe (approved):

mgovers:
Merging even though Sonar Cloud quality gate fails because it's literally (deliberately) empty functionality. Implementation will come in #438

PR #442, opened 2023-12-01
Feature/nrse validation case

mgovers:
add test validation case for state estimation inspired by #237 . For this specific case, the iterative linear and Newton-Raphson calculation methods are expected to result in different global optima. In particular, the following different 

* multiple power sensors on same appliance are aggregated using Kalman filter
* power sensors across several appliances on the same node are added as independent random variables
* power sensors across several nodes are treated differently:
  * iterative linear: only apparent power is treated as a single symmetric complex measurement with only (real) variance
  * Newton Raphson: active and reactive power are treated as independent real measurements with separate variances

this means that when there is mixing of total active and reactive powerflow across nodes, iterative linear will produce the same result when `sigma_p` and `sigma_q` are varied, as long as the variance of the total aggregated apparent power remains the same for each node.

Newton-Raphson, on the other hand, should produce results that have a smaller `p_residual` when the `sigma_p` decreases (and `sigma_q` is increased), and vice versa.


TonyXiang8787:
@mgovers If I understand correctly, we can merge this. And create a new test case for NRSE with the same input data but different expected result data?

TonyXiang8787 (approved):

PR #455, opened 2023-12-19
NRSE low level design

nitbharambe:

Structure of NRSE solver to be added


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/measured_values.hpp
```diff
@@ -0,0 +1,586 @@
...
+#define POWER_GRID_MODEL_MATH_SOLVER_MEASURED_VALUES_HPP
+
+/*
+Collect all measured Values
+*/
+
+#include "../calculation_parameters.hpp"
+#include "../three_phase_tensor.hpp"
+
+// hide implementation in inside namespace
```
mgovers, line 16: this is redundant

```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -21,17 +22,180 @@ namespace power_grid_model {
...
+// block class for the right hand side in state estimation equation
+template <bool sym> using NRSERhs = NRSEUnknown<sym>;
+
+// class of 4*4 (12*12) se gain block
+/*
+[
+   [G, QH]
+   [Q, R ]
+]
+*/
```
mgovers, line 53: i forgot to check whether there are other instances of this, but please only use `/* ... */` for in-line comments. please make this
```suggestion
// [
//    [G, QH]
//    [Q, R ]
// ]
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -26,7 +27,7 @@ namespace power_grid_model {
 namespace math_model_impl {
 
 // block class for the unknown vector and/or right-hand side in state estimation equation
-template <bool sym> struct SEUnknown : public Block<DoubleComplex, sym, false, 2> {
+template <bool sym> struct ILSEUnknown : public Block<DoubleComplex, sym, false, 2> {
```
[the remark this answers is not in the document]
nitbharambe: So it would be `power_grid_model::math_model_impl::se_solver_impl`?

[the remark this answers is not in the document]
mgovers: or just

`power_grid_model::math_model_impl::ilse`

alternatively, you could also do

```cpp
namespace power_grid_model::math_model_impl {
namespace iterative_linear_se {
// ...
} // namespace iterative_linear_se

namespace ilse = iterative_linear_se;
} // namespace power_grid_model::math_model_impl
```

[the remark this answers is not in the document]
mgovers: you can change the name back, right?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -355,9 +355,9 @@ template <bool sym> class IterativeLinearSESolver {
 template class IterativeLinearSESolver<true>;
 template class IterativeLinearSESolver<false>;
 
-} // namespace math_model_impl
+} // namespace math_model_impl::ilse_solver_impl
 
-template <bool sym> using IterativeLinearSESolver = math_model_impl::IterativeLinearSESolver<sym>;
+template <bool sym> using IterativeLinearSESolver = math_model_impl::ilse_solver_impl::IterativeLinearSESolver<sym>;
```
mgovers, line 360: if you want to, you can create a namespace alias using `namespace ilse = ilse_solver_impl`

in addition, I do not like the duplication in `math_model_impl` and then ALSO `ilse_solver_impl`.

I would just replace the entire `math_model_impl` namespace with `math_model::impl` and then the `ilse_solver_impl` just becomes `math_model::impl::ilse`. Anything that resides in something called `impl` by definition is also `impl`, so no repetition is necessary

Idem for nrse of course

nitbharambe: So 3 namespaces? What things do math_model cover? Or directly nest?

mgovers: you can see a namespace as a sort of json-like list that can contain function names and other namespaces. having multiple nested ones (as long as the list doesn't become TOO large) is perfectly OK.

i see that `math_model` actually wouldn't contain anything (i thought the `math_solver` was part of that namespace as well but it isn't)

i personally would have a namespace per directory in which the headers reside (e.g.: `power_grid_model`, `power_grid_model::main_core`, `power_grid_model::component`, `power_grid_model::math_solver` and `power_grid_model::auxiliary`), and anything that does not need to be exposed outside of certain headers or the directory may reside in `<directory>::detail` (helper functions) or `<directory>::impl` (implementation details of something that IS exposed)

In that line, it would theoretically be fine to have just `power_grid_model::math_solver` but there are so many implementation details per solver that it starts to make sense to put them in separate namespaces per header as well. I have a preference to not have too many class member functions (even if they are `static`). It promotes reasonings like "oh we have it available anyways, so I don't need to make this function `static`" which in turn results in spaghetti-code. Instead, I prefer to mimic the same behaviour as `ClassName::member_function` by `ns::static_function`. For both the compiler and the user, the behaviour is exactly the same (aside from `public`/`protected`/`private`, of course).

If you rename `math_model_impl` to just `math_solver`, the rest should be OK and you indeed don't need `math_model::impl`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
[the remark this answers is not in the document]
mgovers: this one still remains

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
```diff
@@ -42,6 +42,8 @@ if there are sources
 
 namespace power_grid_model {
 
+namespace math_solver {
```
mgovers, line 45: shouldn't this become `math_solver::lpf`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
```diff
@@ -113,13 +115,16 @@ template <bool sym> class LinearPFSolver {
...
+        detail::calculate_result(y_bus, input, *sources_per_bus_, *load_gens_per_bus_, output,
+                                 [](Idx /*i*/) { return LoadGenType::const_y; });
     }
 };
 
 template class LinearPFSolver<true>;
 template class LinearPFSolver<false>;
+} // namespace math_solver
+
+template <bool sym> using LinearPFSolver = math_solver::LinearPFSolver<sym>;
```
mgovers, line 127: for this one and the others: do you think it's a good idea to expose these not to `namespace power_grid_model` but only to `namespace power_grid_model::math_solver`? the `MathSolver` itself does have unrestricted access to anything in the `namespace power_grid_model::math_solver`, so there should be no issue there

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
```diff
@@ -171,6 +173,9 @@ template <bool sym> class MathSolver {
 
 template class MathSolver<true>;
 template class MathSolver<false>;
+} // namespace math_solver
+
+template <bool sym> using MathSolver = math_solver::MathSolver<sym>;
```
mgovers, line 178: please double-check but i think that just doing

```suggestion
using math_solver::MathSolver;
```

may also work...

nitbharambe: It does work

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
nitbharambe: I changed it so that each solver has its own namespace now

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_pf_solver.hpp
```diff
@@ -20,7 +20,7 @@
 #include "../three_phase_tensor.hpp"
 #include "../timer.hpp"
 
-namespace power_grid_model::math_model_impl {
+namespace power_grid_model::math_solver {
```
mgovers, line 23: optionally:

```suggestion
namespace power_grid_model::math_solver::iterative_pf {
```

with according `using` statement ofcourse (optional because there's no other functions here)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
```diff
@@ -113,13 +115,16 @@ template <bool sym> class LinearPFSolver {
...
     }
 };
 
 template class LinearPFSolver<true>;
 template class LinearPFSolver<false>;
+} // namespace math_solver::linear_pf
+
+using math_solver::linear_pf::LinearPFSolver;
 
 } // namespace power_grid_model
```
mgovers, line 129: this is in the wrong namespace. should be:

```suggestion
using linear_pf::LinearPFSolver;

} // namespace power_grid_model::math_solver
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
```diff
@@ -42,6 +42,8 @@ if there are sources
 
 namespace power_grid_model {
 
+namespace math_solver::linear_pf {
```
mgovers, line 45: this should be

```suggestion
namespace power_grid_model::math_solver {

namespace linear_pf {
```

please also apply to all other files in this directory

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver.hpp
```diff
@@ -171,6 +173,9 @@ template <bool sym> class MathSolver {
 
 template class MathSolver<true>;
 template class MathSolver<false>;
+} // namespace math_solver
+
+using math_solver::MathSolver;
 
 } // namespace power_grid_model
```
mgovers, line 180: for this PR not really an issue but usually, it's good practice to not clog the parent namespace except for very rare cases. The `MathSolver` is an edge case in which it doesn't matter, but the individual solvers should be restricted to the `math_solver` namespace, only

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -435,9 +435,9 @@ template <bool sym> class NewtonRaphsonPFSolver : public IterativePFSolver<sym,
 template class NewtonRaphsonPFSolver<true>;
 template class NewtonRaphsonPFSolver<false>;
 
-} // namespace math_model_impl
+} // namespace math_solver::newton_raphson_pf
 
-template <bool sym> using NewtonRaphsonPFSolver = math_model_impl::NewtonRaphsonPFSolver<sym>;
+using math_solver::newton_raphson_pf::NewtonRaphsonPFSolver;
 
 } // namespace power_grid_model
```
mgovers, line 442: as mentioned before

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -19,28 +20,189 @@ iterative linear state estimation solver
...
 template class NewtonRaphsonSESolver<true>;
 template class NewtonRaphsonSESolver<false>;
 
-} // namespace math_model_impl
+} // namespace math_solver::newton_raphson_se
 
-template <bool sym> using NewtonRaphsonSESolver = math_model_impl::NewtonRaphsonSESolver<sym>;
+using math_solver::newton_raphson_se::NewtonRaphsonSESolver;
 
 } // namespace power_grid_model
```
mgovers, line 207: as before

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/short_circuit_solver.hpp
```diff
@@ -442,9 +441,9 @@ template <bool sym> class ShortCircuitSolver {
 template class ShortCircuitSolver<true>;
 template class ShortCircuitSolver<false>;
 
-} // namespace math_model_impl
+} // namespace math_solver::short_circuit
 
-template <bool sym> using ShortCircuitSolver = math_model_impl::ShortCircuitSolver<sym>;
+using math_solver::short_circuit::ShortCircuitSolver;
 
 } // namespace power_grid_model
```
mgovers, line 448: as before

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -335,10 +335,9 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
 };
 
-} // namespace math_model_impl
+} // namespace math_solver
 
-template <class Tensor, class RHSVector, class XVector>
-using SparseLUSolver = math_model_impl::SparseLUSolver<Tensor, RHSVector, XVector>;
+using math_solver::SparseLUSolver;
 
 } // namespace power_grid_model
```
mgovers, line 342: as before

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
```diff
@@ -433,11 +433,11 @@ template <bool sym> class YBus {
...
-} // namespace math_model_impl
+} // namespace math_solver
 
-template <bool sym> using YBus = math_model_impl::YBus<sym>;
+using math_solver::YBus;
 
-using YBusStructure = math_model_impl::YBusStructure;
+using math_solver::YBusStructure;
 
 } // namespace power_grid_model
```
mgovers, line 442: as before.

actually, since everything is restricted to the `math_solver`, i think these `using` statements can be removed

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
mgovers: ok but it should be the other way around:

all `using` declarations should be in the `namespace power_grid_model::math_solver` namespace, not the parent `namespace power_grid_model`

mgovers (commented):
the rest of the PR looks good, but i still have a couple remarks (mostly the same one repeated)

nitbharambe:
@mgovers I have now moved all inside math_solver namespace except Ybus and MathSolver

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
```diff
@@ -432,11 +432,11 @@ template <bool sym> class YBus {
...
 template class YBus<false>;
 
-} // namespace math_model_impl
+} // namespace math_solver
 
-template <bool sym> using YBus = math_model_impl::YBus<sym>;
+using math_solver::YBus;
 
-using YBusStructure = math_model_impl::YBusStructure;
+using math_solver::YBusStructure;
```
mgovers, line 440: `YBusStructure` is never used outside this file except in the `test_y_bus.cpp` tests. it should not be exposed

```suggestion
```

mgovers (commented):
one very small nitpick

mgovers (approved):

PR #456, opened 2023-12-20
Feature/remove needless id lookup

### LINKED ISSUE #432: [Improvement] Remove needless id lookup
In #427 , the inverse update was implemented as an improved way to restore components after a batch update. The current implementation queries the id to sequence idx map once for each update step, and therefore once for the regular update, and once for the inverse update. That query is expensive.

The unneeded query may be removed by also caching the sequence idx in addition to the inverse update data.

mgovers:
Fixes #432


power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -162,10 +162,18 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
 
         if constexpr (CacheType::value) {
+            constexpr auto comp_index = AllComponents::template index_of<CompType>();
+
+            auto& cached_sequence = std::get<comp_index>(cached_updated_sequence_indices_);
+            if (sequence_idx.empty()) {
+                main_core::get_component_sequence<CompType>(state_, begin, end, std::back_inserter(cached_sequence));
+            } else {
+                assert(cached_sequence.empty());
+                cached_sequence = sequence_idx;
```
mgovers, line 172: i don't like this copy but it was the only way i could think of that does not involve significant code changes to the batch calculation implementation.

The issue lies in the fact that we basically calculate the same thing (a cache of the ids lookup as a sequence that is by definition the same for all ids in a scenario) twice for independent batch update datasets: once for the entire batch, and once for the restore components.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -162,10 +162,18 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+            }
+
             main_core::update_inverse<CompType>(
-                state_, begin, end,
-                std::back_inserter(std::get<AllComponents::template index_of<CompType>()>(cached_inverse_update_)),
-                sequence_idx);
+                state_, begin, end, std::back_inserter(std::get<comp_index>(cached_inverse_update_)), cached_sequence);
         }
 
         UpdateChange const changed = main_core::update_component<CompType>(state_, begin, end, sequence_idx);
```
mgovers, line 179: this line should also make use of the `cached_updated_sequence_indices_` if created

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -13,29 +13,48 @@
...
-    bool const has_sequence_id = !sequence_idx.empty();
     Idx seq = 0;
 
     // loop to to update component
     for (auto it = begin; it != end; ++it, ++seq) {
-        // get component
-        // either using ID via hash map
-        // either directly using sequence id
-        Idx2D const sequence_single =
-            has_sequence_id ? sequence_idx[seq] : state.components.template get_idx_by_id<Component>(it->id);
```
mgovers, line 32: for all update types, we now determine the sequence beforehand.

The rationale is:

* for independent batch updates, the sequence is determined once
  * this is the same behaviour as before
* for non-independent batch updates, the sequence is determined for every scenario separately
  * this behaviour is equivalent to the situation before, but with half the amount of ID lookups
* for permanent updates, the sequence is determined once
  * this only causes a minor memory overhead compared to the situation before, but that is not the bottleneck.
  * in addition, permanent updates are less performance-critical, so it's not an issue

this allows this piece of complexity to be removed.

mgovers:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=456) **Quality Gate passed**
> 
> The SonarCloud Quality Gate passed, but some issues were introduced.
> 
> [2 New issues](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=456&resolved=false&inNewCodePeriod=true) [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=PowerGridModel_power-grid-model&pullRequest=456&resolved=false&inNewCodePeriod=true) [100.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=456&metric=new_coverage&view=list) [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=456&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=456)

sonar cloud issues were already present in `main`

TonyXiang8787 (approved):

PR #460, opened 2024-01-04
Revert "Revert "Feature/remove needless id lookup""

mgovers:
Re-apply PowerGridModel/power-grid-model#459

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
@@ -13,28 +13,47 @@
 namespace power_grid_model::main_core {
 
 namespace detail {
-template <std::derived_from<Base> Component, class ComponentContainer, std::forward_iterator ForwardIterator,
-          typename Func>
+template <std::derived_from<Base> Component, class ComponentContainer, typename UpdateType>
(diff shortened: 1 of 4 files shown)


TonyXiang8787:
@mgovers can you elaborate the problem here?

mgovers:
> @mgovers can you elaborate the problem here?

I am still digging further into it, but the `SIGBUS` error suggests that there is either an alignment issue. (see also e.g. https://www.geeksforgeeks.org/segmentation-fault-sigsegv-vs-bus-error-sigbus/ ).

Steffan mentioned that on virtual devices (e.g. a container) even the stack memory is not necessarily contiguous. Especially when dealing with multi-threading, virtualization may result in unexpected behaviour when accessing stack memory of another thread (due to potentially different offsets). Workarounds involve either allocating memory on the heap (which has built-in support for non-contiguous memory access), or giving each thread its own stack.

From this information, I deduced that issue likely arose in the original implementation because of a copy of a reference to stack memory initialized in another thread. The new implementation does not have that issue.

This PR is still in draft because investigation whether this change will solve the issue is very much in progress. I am not 100% sure that what I was told about virtualization is correct in the first place, let alone that this part of the change was the issue, and that my new change solves that.

petersalemink95 (approved):

Issue #461, opened 2024-01-09
[BUG] non-independent scenarios when batch scenario fails


mgovers:
**Describe the bug**

A failing batch scenario may affect consecutive batch scenario

**To Reproduce**

* Create valid input data
* Create update batch with 2 scenarios
  * Ensure that the first scenario fails during calculation (e.g. sparse matrix error)
  * Ensure that the second scenario passes (e.g. by running it separately)
* Create a model with the input data
* Run powerflow calculations with the update batch with `treading=-1`
* Observe the result
  * The first scenario fails with the correct batch error
  * The second scenario has a different output than if it is run separately as an individual scenario (e.g. in a batch without the first scenario)

**Expected behavior**

* The first scenario fails with the correct batch error
* The second scenario produces the same output 

**Additional context**

* The restore functionality of the main model at the end of a scenario is only called when a scenario passes. This results in a drift of the state of the main model. We do not have validation cases for that yet. The offending lines are https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp#L472-L475
* The erroneous behavior was first observed in `v1.6.44` due to a change in the restore behavior. This change was rolled back to the previous (also erroneous) behavior in `v1.6.45`, which has the same behavior as `v1.6.43` and previous. See also #459 . 
* Threading might isolate some scenarios from others, and therefore might result in different outputs.

c.c. @petersalemink95 @TonyXiang8787 This is quite severe. I'm surprised no-one has run into this issue yet.

PR #462, opened 2024-01-09
[Hotfix] fix restore after failing batch scenario

mgovers:
**Fixes issue**: #461 

This is a hotfix that does not care about performance.

Approprate sustainable solution will be done in #460 .

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
@@ -475,8 +475,16 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
                     }
                 } catch (std::exception const& ex) {
                     exceptions[batch_number] = ex.what();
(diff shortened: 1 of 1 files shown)


mgovers:
Failing CI is due to boost artifactory bug https://github.com/boostorg/boost/issues/844

TonyXiang8787 (approved):
Approved.

Just to be sure on the reasoning. Because we do not need to worry about performance for error scenario, so we choose to copy the model (less efficient), other than try to restore the components (more complicated).

mgovers:
> Approved.
> 
> Just to be sure on the reasoning. Because we do not need to worry about performance for error scenario, so we choose to copy the model (less efficient), other than try to restore the components (more complicated).

yes indeed. In the near future, a sustainable solution will improve on this. Since this is a hotfix, fixing functionality is more important than performance

mgovers:
> Failing CI is due to boost artifactory bug [boostorg/boost#842](https://github.com/boostorg/boost/issues/842)

re-running after https://github.com/TonyXiang8787/pybuild-header-dependency/pull/3 fixed the boost URL

PR #463, opened 2024-01-10
Feature improve sparse matrix behaviour

mgovers:
* improve the throwing of sparse matrix errors.
* improve the behaviour when required values `p_specified` and/or `q_specified` are not provided

The original behaviour is not necessarily an issue because:

* sparse matrix errors will already always be thrown in supported but underdetermined
* not providing both P and Q was already undefined behaviour (UB) according to the PGM docs. The data validator already forbids the edge case where P and/or Q are not specified.

this is **not** a breaking change.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
@@ -83,7 +83,7 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
         update_real_value<sym>(new_p_specified, ps, scalar);
         update_real_value<sym>(new_q_specified, qs, scalar);
 
-        s_specified_ = ps + 1.0i * qs;
+        s_specified_ = ComplexValue<sym>{ps, qs};
     }
 
     // update for load_gen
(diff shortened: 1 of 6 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
```diff
@@ -108,14 +108,28 @@ class LoadGen final : public std::conditional_t<is_gen, GenericGenerator, Generi
     }
 
   private:
-    ComplexValue<sym> s_specified_{}; // specified power injection
+    ComplexValue<sym> s_specified_{std::complex<double>{nan, nan}}; // specified power injection
```
mgovers, line 111: this is the important part of the fix of the P/Q not provided part

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -203,7 +203,8 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
                     block_perm_array[pivot_row_col] = {lu_factor.permutationP(), lu_factor.permutationQ()};
                     return block_perm_array[pivot_row_col];
                 } else {
-                    if (lu_matrix[pivot_idx] == 0.0) {
+                    auto const& lu_factor = lu_matrix[pivot_idx];
+                    if (!is_normal(lu_factor)) {
```
mgovers, line 207: this is the important part of the raising of sparse matrix errors

power_grid_model_c/power_grid_model/include/power_grid_model/three_phase_tensor.hpp
```diff
@@ -275,7 +280,16 @@ inline bool is_nan(Enum x) {
...
 inline auto is_normal(std::floating_point auto value) { return std::isnormal(value); }
-inline auto is_normal(RealValue<false> const& value) {
+template <std::floating_point T> inline auto is_normal(std::complex<T> const& value) {
+    if (value.real() == T{0}) {
+        return is_normal(value.imag());
+    }
+    if (value.imag() == T{0}) {
+        return is_normal(value.real());
+    }
+    return is_normal(value.real()) && is_normal(value.imag());
```
mgovers, line 290: i don't think we have to take care of underflows (subnormal numbers). please let me know if you think that is necessary, then i'll improve this function.

TonyXiang8787 (approved):

PR #464, opened 2024-01-10
DGC 2106 Changes support python version from 3.8 to 3.9

Jerry-Jinfeng-Guo:
Support Python >3.9 in PGM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,7 +107,7 @@ test-command = "pytest {package}/tests"
 #    PyPy
 #    musllinux when python < 3.9
 #    musllinux in aarch64
-skip = ["pp*", "cp38-musllinux*", "*-musllinux_aarch64"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -107,7 +107,7 @@ test-command = "pytest {package}/tests"
 #    PyPy
 #    musllinux when python < 3.9
 #    musllinux in aarch64
-skip = ["pp*", "cp38-musllinux*", "*-musllinux_aarch64"]
+skip = ["pp*", "cp39-musllinux*", "*-musllinux_aarch64"]
```
mgovers, line 110: please read the comment 😬

TonyXiang8787: We can now remove `"cp38-musllinux*"`. The `skip` will be joined with `python >=`. If we have `python >= 3.9`, any `3.8` build will be skipped anyway.

TonyXiang8787 (approved):

PR #468, opened 2024-01-15
Feature/sustainable solution for bug

mgovers:
Sustainable solution for #461 .
Replaces quick fix in #462 .
Tests in #460 .

A large portion of the complexity in the batch update is refactored away to separate lambda functions:

* The construction of the sub batch calculation is moved to a separate function, as well as its dispatching
* Each scenario itself has a separate setup, winddown and run and also has an exception handler and a recovery method for otherwise unfixable errors
* the Lippincott pattern is used to handle all scenario exceptions in one go https://en.cppreference.com/w/cpp/error/current_exception . That way, the actual implementation for the running of the scenario only requires one try/catch block

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
@@ -463,8 +463,8 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
 
         return [&base_model, &exceptions, &infos, &calculation_fn, &result_data, &update_data,
                 is_independent](Idx start, Idx stride, Idx n_batch) {
-            assert(exceptions.size() >= n_batch);
-            assert(infos.size() >= n_batch);
+            assert(n_batch <= narrow_cast<Idx>(exceptions.size()));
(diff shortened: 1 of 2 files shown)


TonyXiang8787:
should this be merged before #460?

TonyXiang8787:
@mgovers If this is still working in progress, please put the PR in draft.

mgovers:
> @mgovers If this is still working in progress, please put the PR in draft.

everything is done except for sonar cloud/CI. i thought it was already ready for review until i saw the code smells. CI should pass now. If not, i'll pull it back to draft

mgovers:
remaining code smells were already introduced in original `main` but just were moved around

TonyXiang8787 (changes requested):
We were using some confusing names before. Since we are refactoring here, please change the names of variables correctly using the terminology defined here.

https://power-grid-model.readthedocs.io/en/stable/user_manual/dataset-terminology.html

Especially, we use the loop variable name `batch_number` , it should be actually `scenario_number`. We only have one batch, which consists of multiple sub-batches for multithreading, which consist again multiple scenarios per sub-batch.

TonyXiang8787 (approved):

PR #475, opened 2024-01-22
Use constexpr for all meta data

### LINKED ISSUE #435: [FEATURE] Compile-time meta attribute offsets
Add the offsets for each component attribute to the [generated metadata attribute setters](https://github.com/PowerGridModel/power-grid-model/tree/main/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen) using [`offsetof`](https://en.cppreference.com/w/cpp/types/offsetof) (defined in `<cstddef>` in the STL) to replace the run-time determination using [`get_offset`](https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp#L20-L28).

However, that is a C Macro - not a template, meaning that the attribute needs to be obtained directly from the attribute name - not as a template. The template getters used later in the code should be generated by the code generator for each meta attribute.

**Background**
In #276 , the way metadata attribute offsets were calculated was modified to account for the fact that derived objects are not guaranteed to be `std::standard_layout_v`(see in particular #241 and [this change](https://github.com/PowerGridModel/power-grid-model/pull/276/files#diff-2bac4fa15db710784f8450b130e98864e112b2b5e4b104c0470f876d5894f4eeL138-R141)). In #430 , the generated component meta data structures were changed to plain structures, removing the need for runtime `inline` offset getters.

> should i also revert parts of https://github.com/PowerGridModel/power-grid-model/pull/276/files#diff-2bac4fa15db710784f8450b130e98864e112b2b5e4b104c0470f876d5894f4eeL138-R141 ? or is that not necessary? or maybe we can use `offsetof` also in those functions?

_Originally posted by @mgovers in https://github.com/PowerGridModel/power-grid-model/pull/430#pullrequestreview-1742138943_

TonyXiang8787:
# Intro

This should fix #435. All meta data are now known at compile time. The whole meta data is a `constexpr` struct.

# Approach

Because the variable length of components and attributes. We define some meta-data getter struct which contains `static constexpr std::array` of attributes and components. Then we expose this array into the real meta-data in a `std::span`.

# Todo

- [x] Remove the `meta_data` function. Make it directly as a `constexpr` global member.
- [x] Fix the checks.


code_generation/templates/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/attribute_classes.hpp.jinja
```diff
@@ -29,14 +30,12 @@ struct get_attributes_list<{{ attribute_class.name }}<sym>> {
 template<>
 struct get_attributes_list<{{ attribute_class.name }}> {
 {%- endif %}
-    std::vector<MetaAttribute> operator() () const {
-        // all attributes including base class
-        return {
+    static constexpr std::array<MetaAttribute, {{ attribute_class.full_attributes|length }}> value{
```
mgovers, line 33: the compiler is able to deduce the length of the array and it will be much more beautiful

```suggestion
    static constexpr std::array value{
```

if that doesn't work, then you'll have to do something like

```cpp
    static constexpr std::array value{
            MetaAttribute{meta_data_gen::get_meta_attribute<{{ attribute_class.full_name }}, &{{ attribute_class.full_name }}::{{ attribute.names }}, offsetof({{ attribute_class.full_name }}, {{ attribute.names }}), []{ return "{{ attribute.names }}"; }>::value},
```

code_generation/templates/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/attribute_classes.hpp.jinja
```diff
@@ -29,14 +30,12 @@ struct get_attributes_list<{{ attribute_class.name }}<sym>> {
...
+            // all attributes including base class
             {% for attribute in attribute_class.full_attributes %}
-            {MetaAttributeImpl<{{ attribute_class.full_name }}, &{{ attribute_class.full_name }}::{{ attribute.names }}>{}, "{{ attribute.names }}"},
+            meta_data_gen::get_meta_attribute<{{ attribute_class.full_name }}, &{{ attribute_class.full_name }}::{{ attribute.names }}, offsetof({{ attribute_class.full_name }}, {{ attribute.names }}), []{ return "{{ attribute.names }}"; }>::value,
```
mgovers, line 36: this line becomes incredibly long for some components. can you maybe for readability put the attributes on new lines?

power_grid_model_c/power_grid_model_c/src/dataset.cpp
```diff
@@ -31,7 +29,7 @@ PGM_Idx PGM_dataset_info_n_components(PGM_Handle* /*unused*/, PGM_DatasetInfo co
 
 char const* PGM_dataset_info_component_name(PGM_Handle* /*unused*/, PGM_DatasetInfo const* info,
                                             PGM_Idx component_idx) {
-    return info->component_info[component_idx].component->name.c_str();
+    return info->component_info[component_idx].component->name;
```
mgovers, line 32: as mentioned before: I would like to maintain the `constexpr string_view` version if possible

tests/c_api_tests/test_c_api_meta_data.cpp
```diff
@@ -17,24 +17,24 @@ TEST_CASE("C API Meta Data") {
     // get handle
     HandlePtr const unique_handle{PGM_create_handle()};
     PGM_Handle* hl = unique_handle.get();
-    auto const& meta = meta_data();
+    auto const& meta = meta_data;
```
mgovers, line 20: why not just make it `meta_data` everywhere?

tests/c_api_tests/test_c_api_meta_data.cpp
```diff
@@ -17,24 +17,24 @@ TEST_CASE("C API Meta Data") {
...
 
     SUBCASE("Datasets") {
         // check dataset
         CHECK(PGM_meta_n_datasets(hl) == meta.n_datasets());
         for (Idx idx_dataset = 0; idx_dataset != meta.n_datasets(); ++idx_dataset) {
             PGM_MetaDataset const* const dataset = PGM_meta_get_dataset_by_idx(hl, idx_dataset);
-            char const* const dataset_name = PGM_meta_dataset_name(hl, dataset);
-            CHECK(PGM_meta_get_dataset_by_name(hl, dataset_name) == dataset);
+            std::string const dataset_name = PGM_meta_dataset_name(hl, dataset);
```
mgovers, line 27: or just `auto`

```suggestion
            auto const dataset_name = PGM_meta_dataset_name(hl, dataset);
```

(same below)

tests/cpp_unit_tests/test_serializer.cpp
```diff
@@ -49,11 +49,11 @@ constexpr std::string_view single_dataset_list_indent =
   "type": "update",
   "is_batch": false,
   "attributes": {
-    "asym_load": [
+    "sym_load": [
       "id",
       "p_specified"
     ],
-    "sym_load": [
+    "asym_load": [
```
mgovers, line 56: i guess this is because the new implementation uses a `constexpr std::array` instead of a `std::map` with sorted strings as keys. i don't mind keeping it like this but i perfer the sorted behaviour (since most json formatters also sort their entries).

if we want to keep the sorted behaviour, then we need to change it in the serializer because changing it in the code generator would cause a breaking change for every time we add a new attribute

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -225,7 +225,7 @@ void assert_result(ConstDataset const& result, ConstDataset const& reference_res
             // loop all attribute
             for (MetaAttribute const& attr : component_meta.attributes) {
                 // TODO skip u angle, need a way for common angle
-                if (attr.name == "u_angle") {
+                if (attr.name == std::string("u_angle")) {
```
mgovers, line 228:
```suggestion
                if (attr.name == "u_angle"s) {
```

mgovers (commented):
very nice improvement!

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -86,50 +89,9 @@ using RawDataPtr = void*;            // raw mutable data ptr
...
-          component_size{sizeof(StructType)},
-          check_nan{MetaAttributeImpl<StructType, member_ptr>::check_nan},
-          check_all_nan{MetaAttributeImpl<StructType, member_ptr>::check_all_nan},
-          set_value{MetaAttributeImpl<StructType, member_ptr>::set_value},
-          get_value{MetaAttributeImpl<StructType, member_ptr>::get_value},
-          compare_value{MetaAttributeImpl<StructType, member_ptr>::compare_value} {}
-
     // meta data
-    std::string name;
+    char const* name{};
```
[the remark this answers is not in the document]
TonyXiang8787: You cannot do with `string_view`. In the C-API we return `char const*` for the name of attribute, component, dataset. In `string_view` there is no zero termination.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/gen_getters.hpp
```diff
@@ -0,0 +1,100 @@
...
+// get meta data
+template <auto dataset_name_getter, template <class> class struct_getter> struct dataset_mark;
+template <class comp_list, class... T> struct get_meta_data;
+template <class comp_list, auto... dataset_name_getter, template <class> class... struct_getter>
+struct get_meta_data<comp_list, dataset_mark<dataset_name_getter, struct_getter>...> {
+    static constexpr std::array<MetaDataset, sizeof...(dataset_name_getter)> datasets{
+        get_meta_dataset<dataset_name_getter, struct_getter, comp_list>::value...};
+    static constexpr MetaData value{
+        .datasets = datasets,
+    };
```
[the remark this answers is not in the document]
TonyXiang8787: A temporary `std::array` will be destroyed and the `span` will be dangling. We need to define the `std::array` in a `static constexpr` to ensure the lifetime.

code_generation/templates/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/attribute_classes.hpp.jinja
TonyXiang8787: Compiler must deduce type and size at the same time. It is possible here. But we are using code generation. So to generate it explicitly should not hurt.

code_generation/templates/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/attribute_classes.hpp.jinja
TonyXiang8787: Again. This is generated code. Most of the time, no one (including PGM developers) will read it.

tests/c_api_tests/test_c_api_meta_data.cpp
TonyXiang8787: You can't. We have to instantiate a `string`, otherwise it is `char const*`. In the later check, we need to compare the string, not the address of `char const*`.

tests/cpp_unit_tests/test_serializer.cpp
TonyXiang8787: Actually, I prefer to sort based on given order by `AllComponents`. In PGM not all components are equally important. We define the order in `AllComponents` as their importance to the physical model.

https://github.com/PowerGridModel/power-grid-model/blob/daa46892da9793536d7537f137da66ff51cfecd6/power_grid_model_c/power_grid_model/include/power_grid_model/all_components.hpp#L28-L30

code_generation/templates/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/attribute_classes.hpp.jinja
```diff
@@ -13,12 +13,13 @@
 #include "../../power_grid_model.hpp" // NOLINT
 #include "../../three_phase_tensor.hpp" // NOLINT
 #include "../meta_data.hpp" // NOLINT
+#include "gen_getters.hpp" // NOLINT
```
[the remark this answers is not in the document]
TonyXiang8787: If I try to format the generated code. It format exactly the other way around as opposed to what you mentioned, see below. Is this correct?

```c++
#include "../../enum.hpp"               // NOLINT
#include "../../power_grid_model.hpp"   // NOLINT
#include "../../three_phase_tensor.hpp" // NOLINT
#include "../input.hpp"                 // NOLINT
#include "../meta_data.hpp"             // NOLINT
#include "gen_getters.hpp"              // NOLINT
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/gen_getters.hpp
```diff
@@ -0,0 +1,100 @@
...
+};
+
+} // namespace power_grid_model::meta_data::meta_data_gen
+
+#endif
```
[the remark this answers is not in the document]
TonyXiang8787: done

tests/cpp_validation_tests/test_validation.cpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data_gen.hpp
```diff
@@ -23,52 +24,38 @@
...
+template <class T> struct struct_getter_update {
+    using type = typename T::UpdateType;
 };
+template <class T> struct struct_getter_sym_output {
+    using type = typename T::template OutputType<true>;
+};
+template <class T> struct struct_getter_asym_output {
+    using type = typename T::template OutputType<false>;
+};
+template <class T> struct struct_getter_sc_output {
```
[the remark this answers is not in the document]
TonyXiang8787: done

tests/c_api_tests/test_c_api_meta_data.cpp
TonyXiang8787: I was lazy to change.

TonyXiang8787:
@mgovers I have tacked your comments with two open questions.

Please feel free to start a full clang tidy run.

code_generation/templates/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_gen/attribute_classes.hpp.jinja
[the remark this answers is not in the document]
mgovers: what we usually do is add an additional whiteline.

The underlying thing is that the second you put something in another directory, you kind of create a structure that's in some way similar to creating a different interface library. In that regard, anything in that parent directory can be treated as being in a separate library. Because of the same reason, I personally tend to use a separate namespace for every subdirectory.

See e.g. https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model/include/power_grid_model/component/branch.hpp#L9-L17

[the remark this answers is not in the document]
TonyXiang8787: done

mgovers:
> Please feel free to start a full clang tidy run.

full clang-tidy run in https://github.com/PowerGridModel/power-grid-model/actions/runs/7611650762

mgovers (commented):
code review OK. will approve when clang tidy passes

mgovers (approved):
full clang-tidy run successful
sonar cloud issues are not caused by this PR

Issue #466, opened 2024-01-11
[BUG] ValueErrors on MacBook on >=v1.6.22


mgovers:
**Describe the bug**

The data types on MacBook changed in `v1.6.22`. As a result, `pandas.DataFrame` representations of the PGM output data may not be correct and/or consistent.

Further investigation will be done in pair programming session.

**Input Data Validity**

(private data)

**Data Dump**

(private data)

**To Reproduce**

```py
# (private data)
df = pd.merge(output_data, ...)
df.reindex(...)  # <- raises ValueError: cannot reindex on an axis with duplicate labels
```

**Expected behavior**

(private data)

**Additional context**

It affects MacBooks with M1 processors. It seems to be closely related to https://github.com/PowerGridModel/power-grid-model/compare/v1.6.21..v1.6.22

TonyXiang8787:
@mgovers there is no description of the bug?

mgovers:
Will be added to in the near future. I asked the reporter to add to this issue

Op 11 jan. 2024 10:20 schreef Tony Xiang ***@***.***>:

@mgovers<https://github.com/mgovers> there is no description of the bug?

mgovers:
unit tests + validation tests pass on Jean's MacBook. Working on a minimal repro case

mgovers:
Both independent reporters @Thijss and Jean confirmed the issue to be resolved by #477 (released in `v1.6.55`).

PR #477, opened 2024-01-24
Fix rounding error of transformer calculation in arm64

mgovers:
Likely culprit of #466 seems to be floating point rounding errors.

In this particular case, `i0 = 1.0e-6` and `p0 = 0.1` and `sn = 100000`. As per the documentation on the transformer input (https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html#id7), `i0 >= po / sn`. In this case `i0 == p0 / sn` (equality edge case). is reached.

## Additional interesting details

`assert_valid_input_data` is python-only and does not fail because indeed `i0 * sn >= p0`.

setting `i0 = 0.9e-6` or `i0 = 1.1e-6` cause the errors to go away (as reported by @Thijss ).
Idem for `i0 = 0.0` and `p0 = 0.0`

### ARM processors

* `<=v1.6.21`: valid output
* `==v1.6.22`: `nan` output
* recent versions (e.g. `==v1.6.54`: `SparseMatrixError`, most likely related to #463 ). This is actually the desired behavior over the behavior of `v1.6.22`.

### Other architectures

all valid output


TonyXiang8787:
@mgovers let's see if `x64` mac will also fail or not.

TonyXiang8787:
@mgovers, this is the problem I think. We compare `y_shunt.real()` and `y_shunt_abs` in the `if` statement. But in the calculation we do `std::sqrt(y_shunt_abs * y_shunt_abs - y_shunt.real() * y_shunt.real())`. We should compare `y_shunt.real() * y_shunt.real()` and `y_shunt_abs * y_shunt_abs`.

https://github.com/PowerGridModel/power-grid-model/blob/9922340451ca7152d6218b7ca327e97c207dfc0b/power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp#L190-L194

mgovers:
> @mgovers, this is the problem I think. We compare `y_shunt.real()` and `y_shunt_abs` in the `if` statement. But in the calculation we do `std::sqrt(y_shunt_abs * y_shunt_abs - y_shunt.real() * y_shunt.real())`. We should compare `y_shunt.real() * y_shunt.real()` and `y_shunt_abs * y_shunt_abs`.
> 
> https://github.com/PowerGridModel/power-grid-model/blob/9922340451ca7152d6218b7ca327e97c207dfc0b/power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp#L190-L194

how do you want to proceed? i do think we should make this change regardless. I can add it to this PR and then when the new version is released, ask @Thijss  to check whether it worked, but that doesn't sound like the right workflow in this case.

TonyXiang8787:
`arm64` on Linux fails too. So we can already reproduce the error. Let's push a fix in this PR.

TonyXiang8787 (approved):

petersalemink95 (approved):

PR #485, opened 2024-02-02
bump cibuildwheel to `v2.16.5

mgovers:
Attempt to fix Python CI


mgovers:
it's running instead of failing within 45sec so it's ready

Jerry-Jinfeng-Guo (approved):
Really can't complain here ;)

PR #487, opened 2024-02-06
replace inline comments by multiline docstrings

mgovers:


TonyXiang8787:
@mgovers why are we doing this? I would argue that using `/**/` is much better for long formulas in the comment.

mgovers:
> @mgovers why are we doing this? I would argue that using `/**/` is much better for long formulas in the comment.

different IDE's and formatters could have different behavior when dealing with `/**/`. The current paradigm is that `//` (comment) or `///` (docstrings) are much more stable except for the edge case where you actually have inline comments, e.g. of the form of `void func(Foo /* unused_variable */){}`

mgovers:
>     * you are dealing with ASCII art with `\` characters end-line. all formatters actually do allow whitespaces after such a character in such a case (and only then)

turns out that Apple Clang does not obey the conventions. We can either proceed and just ignore the ASCII art for the three winding transformer case (i.e. keep the `/* */` there), or close this PR

TonyXiang8787:
My proposal is to park this (close PR).

PR #488, opened 2024-02-07
NRSE documentation

mgovers:
part of #369 

this PR is deliberately independent of #438

Preview in https://power-grid-model--488.org.readthedocs.build/en/488/user_manual/calculations.html

### DIFF
--- a/docs/user_manual/calculations.md
+++ b/docs/user_manual/calculations.md
@@ -114,10 +114,10 @@ The table below can be used to pick the right algorithm. Below the table a more 
 
 | Algorithm                                                         | Default  | Speed    | Accuracy | Algorithm call                                                                                                        |
 | ----------------------------------------------------------------- | -------- | -------- | -------- | --------------------------------------------------------------------------------------------------------------------- |
-| [Newton-Raphson](calculations.md#newton-raphson-power-flow)       | &#10004; |          | &#10004; | [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#CalculationMethod.newton_raphson)       |
-| [Iterative current](calculations.md#iterative-current-power-flow) |          |          | &#10004; | [`CalculationMethod.iterative_current`](../api_reference/python-api-reference.md#CalculationMethod.iterative_current) |
-| [Linear](calculations.md#linear-power-flow)                       |          | &#10004; |          | [`CalculationMethod.linear`](../api_reference/python-api-reference.md#CalculationMethod.linear)                       |
-| [Linear current](calculations.md#linear-current-power-flow)       |          | &#10004; |          | [`CalculationMethod.linear_current`](../api_reference/python-api-reference.md#CalculationMethod.linear_current)       |
+| [Newton-Raphson](calculations.md#newton-raphson-power-flow)       | &#10004; |          | &#10004; | [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#newton_raphson)       |
+| [Iterative current](calculations.md#iterative-current-power-flow) |          |          | &#10004; | [`CalculationMethod.iterative_current`](../api_reference/python-api-reference.md#iterative_current) |
+| [Linear](calculations.md#linear-power-flow)                       |          | &#10004; |          | [`CalculationMethod.linear`](../api_reference/python-api-reference.md#linear)                       |
+| [Linear current](calculations.md#linear-current-power-flow)       |          | &#10004; |          | [`CalculationMethod.linear_current`](../api_reference/python-api-reference.md#linear_current)       |
 
 ```{note}
 By default, the [Newton-Raphson](#newton-raphson-power-flow) method is used.
@@ -154,7 +154,7 @@ and then obtaining the real and reactive power flow through the branches. The fo
 
 #### Newton-Raphson power flow
 
-Algorithm call: [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#CalculationMethod.newton_raphson)
+Algorithm call: [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#newton_raphson)
 
 This is the traditional method for power flow calculations. This method uses a Taylor series, ignoring the higher order
 terms, to solve the nonlinear set of equations iteratively:
@@ -241,7 +241,7 @@ For each iteration the following steps are executed:
 
 #### Iterative current power flow
 
-Algorithm call: [`CalculationMethod.linear_current`](../api_reference/python-api-reference.md#CalculationMethod.linear_current)
+Algorithm call: [`CalculationMethod.linear_current`](../api_reference/python-api-reference.md#linear_current)
 
 This algorithm is a Jacobi-like method for powerflow analysis.
 It has linear convergence as opposed to quadratic convergence in the Newton-Raphson method. This means that the number of iterations will be greater. Newton-Raphson will also be more robust in achieving convergence in case of greater meshed configurations. However, the iterative current algorithm will be faster most of the time.
@@ -268,7 +268,7 @@ The $Y_{bus}$ matrix also remains unchanged in certain batch calculations like t
 
 #### Linear power flow
 
-Algorithm call: [`CalculationMethod.linear`](../api_reference/python-api-reference.md#CalculationMethod.linear)
+Algorithm call: [`CalculationMethod.linear`](../api_reference/python-api-reference.md#linear)
 
 This is an approximation method where we assume that all loads and generations are of constant impedance type regardless of their actual `LoadGenType`.
 By doing so, we obtain huge performance benefits as the computation required is equivalent to a single iteration of the iterative methods.
@@ -365,8 +365,8 @@ At the moment, only iterative state estimation algorithms are implemented.
 
 | Algorithm                                                             | Default  | Speed    | Accuracy | Algorithm call                                                                                                      |
 | --------------------------------------------------------------------- | -------- | -------- | -------- | ------------------------------------------------------------------------------------------------------------------- |
-| [Iterative linear](calculations.md#iterative-linear-state-estimation) | &#10004; | &#10004; |          | [`CalculationMethod.iterative_linear`](../api_reference/python-api-reference.md#CalculationMethod.iterative_linear) |
-| [Newton-Raphson](calculations.md#newton-raphson-state-estimation)     |          |          | &#10004; | [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#CalculationMethod.newton_raphson)     |
+| [Iterative linear](calculations.md#iterative-linear-state-estimation) | &#10004; | &#10004; |          | [`CalculationMethod.iterative_linear`](../api_reference/python-api-reference.md#iterative_linear) |
+| [Newton-Raphson](calculations.md#newton-raphson-state-estimation)     |          |          | &#10004; | [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#newton_raphson)     |
 
 ```{note}
 By default, the [iterative linear](#iterative-linear) method is used.
@@ -399,7 +399,7 @@ Where $S_k$ and $\sigma_{P,k}$ and $\sigma_{Q,k}$ are the measured value and the
 
 #### Iterative linear state estimation
 
-Algorithm call: [`CalculationMethod.iterative_linear`](../api_reference/python-api-reference.md#CalculationMethod.iterative_linear)
+Algorithm call: [`CalculationMethod.iterative_linear`](../api_reference/python-api-reference.md#iterative_linear)
 
 Linear WLS requires all measurements to be linear. This is only possible if all measurements are phasor unit measurements,
 which is not realistic in a distribution grid. Therefore, traditional measurements are linearized before the algorithm is performed:
@@ -466,7 +466,7 @@ The algorithm will assume angles to be zero by default (see the details about vo
 
 #### Newton-Raphson state estimation
 
-Algorithm call: [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#CalculationMethod.newton_raphson)
+Algorithm call: [`CalculationMethod.newton_raphson`](../api_reference/python-api-reference.md#newton_raphson)
 
 The Newton-Raphson approach solves state estimation by considering it as a system of real, non-linear equations.
 It then iteratively solves the first order Taylor expansion of that system.
@@ -501,7 +501,7 @@ The algorithm will assume angles to be zero by default (see the details about vo
 
 ### Short circuit algorithms
 
-Algorithm call: [`CalculationMethod.iec60909`](../api_reference/python-api-reference.md#CalculationMethod.iec60909)
+Algorithm call: [`CalculationMethod.iec60909`](../api_reference/python-api-reference.md#iec60909)
 
 In the short circuit calculation, the following equations are solved with border conditions of faults added as constraints. 
 
--- a/docs/user_manual/calculations.md
+++ b/docs/user_manual/calculations.md
@@ -34,7 +34,7 @@ Output:
 #### State estimation
 
 State estimation is a statistical calculation method that determines the most probable state of the grid, based on
-network data and measurements. Here, measurements may be power flow or voltage values with some kind of uncertainty, which were 
+network data and measurements. Here, measurements can be power flow or voltage values with certain kind of uncertainty, which were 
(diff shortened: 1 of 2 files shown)


docs/user_manual/calculations.md
```diff
@@ -48,10 +48,10 @@ Output:
...
 - Deviation between measurement values and estimated state
 
-In order to perform a state estimation the system should be observable. If the system is not observable the calculation will 
-raise a singular matrix error. Simply said, observability means that the number of measurements
-should be larger than or equal to the number of unknowns. For each node there are two unknowns, `u` and `u_angle`, so the following
-equations should be met:
+In order to perform a state estimation, the system should be observable. If the system is not observable, the calculation will 
+raise a sparse matrix error, since the matrix reprensentation of the system is too sparse to solve - in particular when it is
+singular. Simply said, observability means that the number of measurements should be larger than or equal to the number of
+unknowns. For each node, there are two unknowns, `u` and `u_angle`, so the following equations should be met:
```
Jerry-Jinfeng-Guo, line 54:
```suggestion
singular. In short, meeting the requirement of observability indicates that the system is either an overdetermined system (when the number of measurements is larger than the number of
unknowns) or a balanced system (when the number of measurements is equal to the number of unknowns). For each node, there are two unknowns, `u` and `u_angle`, so the following conditions should be met:
```

docs/user_manual/calculations.md
```diff
@@ -76,13 +76,16 @@ The number of measurements can be found by the sum of the following:
 - two times the number of branches with a power sensor
 
 ```{note}
-Having enough measurements does not necessarily mean that the system is observable. The location of the measurements is also
-of importance. Also, there should be at least one voltage measurement. The [iterative linear](#iterative-linear) 
-state estimation algorithm assumes voltage angles to be zero when not given. This might result in the calculation succeeding, but giving 
-a faulty outcome instead of raising a singular matrix error.
+Having enough measurements does not necessarily mean that the system is observable. The location of the measurements is
+also of importance. Also, there should be at least one voltage measurement.
```
Jerry-Jinfeng-Guo, line 80:
```suggestion
also of importance. Additionally, there should be at least one voltage measurement.
```

docs/user_manual/calculations.md
```diff
@@ -76,13 +76,16 @@ The number of measurements can be found by the sum of the following:
...
 ```
 
-#### Short Circuit Calculations
+```{warning}
+The [iterative linear](#iterative-linear-state-estimation) and [Newton-Raphson](#newton-raphson-state-estimation) state estimation algorithms will assume angles to be zero by default (see the details about voltage sensors).
+This produces more correct outputs when the system is observable, but will prevent the calculation from raising an exception, even if it is unobservable, therefore giving faulty results.
```
Jerry-Jinfeng-Guo, line 85:
```suggestion
In observable systems this helps better outputting correct results. On the other hand with unobservable systems, exceptions raised from calculations due to faulty results will be prevented.
```

docs/user_manual/calculations.md
```diff
@@ -106,20 +109,22 @@ Iterative methods are more accurate and should thus be selected when an accurate
...
 ```{note}
-By default, the Newton-Raphson method is used.
+By default, the [Newton-Raphson](#newton-raphson-power-flow) method is used.
 ```
 
 ```{note}
-When all the load/generation types are of constant impedance, power-grid-model uses the [Linear](#linear) method regardless of the input provided by the user. 
-This is because this method will then be accurate and fastest.
+When all the load/generation types are of constant impedance, power-grid-model uses the [Linear](#linear-power-flow) method regardless of the input provided by the user. 
+This is because this method will then be both accurate and the fastest.
```
Jerry-Jinfeng-Guo, line 127:
```suggestion
When all the load/generation types are of constant impedance, the [Linear](#linear-power-flow) method will be the fastest without loss of accuracy. Therefore power-grid-model will use this method regardless of the input provided by the user in this case. 
```

docs/user_manual/calculations.md
```diff
@@ -231,10 +238,13 @@ For each iteration the following steps are executed:
...
 - Compute $x(i+1)$ from $\Delta x(i) =  x(i+1) - x(i)$
 
-#### Iterative Current
+#### Iterative current power flow
+
+Algorithm call: {py:class}`CalculationMethod.iterative_current <power_grid_model.enum.CalculationMethod.iterative_current>`
 
 This algorithm is a Jacobi-like method for powerflow analysis.
-It has linear convergence as opposed to quadratic convergence in the Newton-Raphson method. This means that the number of iterations will be greater. Newton-Raphson will also be more robust in achieving convergence in case of greater meshed configurations. However, the iterative current algorithm will be faster most of the time.
+It has linear convergence as opposed to quadratic convergence in the [Newton-Raphson](#newton-raphson-power-flow) method. This means that the number of iterations will be greater.
```
Jerry-Jinfeng-Guo, line 246:
```suggestion
It has a linear convergence rate as opposed to the quadratic convergence rate in the [Newton-Raphson](#newton-raphson-power-flow) method. This means that more iterations is needed to achieve similar convergence.
```

docs/user_manual/calculations.md
```diff
@@ -231,10 +238,13 @@ For each iteration the following steps are executed:
...
+[Newton-Raphson](#newton-raphson-power-flow) will also be more robust in achieving convergence in case of greater meshed configurations. However, the iterative current algorithm will be faster most of the time.
```
Jerry-Jinfeng-Guo, line 247:
```suggestion
Additionally, [Newton-Raphson](#newton-raphson-power-flow) will be more robust converging in case of greater meshed configurations. Nevertheless, the iterative current algorithm will be faster most of the time.
```

Jerry-Jinfeng-Guo: This is a problematic sentence: `meshed configuration` is a term but why `greater`? Unless this is a domain terminology, I suggest we come up with clearer wording; the `iterative current` faster in what sense? In terms of the runtime performance per iteration or in achieving similar convergence?

docs/user_manual/calculations.md
```diff
@@ -256,30 +266,35 @@ Factorizing the matrix of linear equation is the most computationally heavy task
...
+
+Algorithm call: {py:class}`CalculationMethod.linear <power_grid_model.enum.CalculationMethod.linear>`
 
-This is an approximation method where we assume that all loads and generations are of constant impedance type regardless of their actual `LoadGenType`.
+This is an approximation method where we assume that all loads and generations are of constant impedance type regardless of their actual {py:class}`LoadGenType <power_grid_model.enum.LoadGenType>`.
 By doing so, we obtain huge performance benefits as the computation required is equivalent to a single iteration of the iterative methods.
 It will be more accurate when most of the load/generation types are of constant impedance or the actual node voltages are close to 1 p.u.
-When all the load/generation types are of constant impedance, power-grid-model uses Linear method regardless of the input provided by the user. This is because this method will then be accurate and fastest. 
+When all the load/generation types are of constant impedance, power-grid-model uses Linear method regardless of the input provided by the user.
+This is because this method will then be both accurate and the fastest.
```
Jerry-Jinfeng-Guo, line 277: See comment above.

docs/user_manual/calculations.md
```diff
@@ -256,30 +266,35 @@ Factorizing the matrix of linear equation is the most computationally heavy task
...
-This approximation method will give better results when most of the load/generation types resemble constant current. 
-Similar to [Iterative Current](calculations.md#iterative-current), batch calculations like timeseries will also be faster. 
-The reason is the same: the $Y_{bus}$ matrix does not change across batches and the same factorization would be used.
+Algorithm call: {py:class}`CalculationMethod.linear_current <power_grid_model.enum.CalculationMethod.linear_current>`
+
+**This algorithm is essentially a single iteration of [iterative current power flow](#iterative-current-power-flow).**
 
+This approximation method will give better results when most of the load/generation types resemble constant current.
+Similar to [iterative current](#iterative-current-power-flow), batch calculations like timeseries will also be faster.
+The reason is the same: the $Y_{bus}$ matrix does not change across batches and the same factorization would be used.
```
Jerry-Jinfeng-Guo, line 294:
```suggestion
The reason is the same: the $Y_{bus}$ matrix does not change across batches and as a result the same factorization could be used.
```

docs/user_manual/calculations.md
```diff
@@ -374,12 +398,14 @@ $$
 
 Where $S_k$ and $\sigma_{P,k}$ and $\sigma_{Q,k}$ are the measured value and the standard deviation of the individual appliances.
 
-#### Iterative linear
+#### Iterative linear state estimation
+
+Algorithm call: {py:class}`CalculationMethod.iterative_linear <power_grid_model.enum.CalculationMethod.iterative_linear>`
 
-Linear WLS requires all measurements to be linear. This is only possible is all measurements are phasor unit measurements, 
+Linear WLS requires all measurements to be linear. This is only possible if all measurements are phasor unit measurements,
```
Jerry-Jinfeng-Guo, line 405:
```suggestion
Linear WLS requires all measurements to be linear. This is only possible when all measurements are phasor unit measurements,
```

docs/user_manual/calculations.md
```diff
@@ -374,12 +398,14 @@ $$
...
 which is not realistic in a distribution grid. Therefore, traditional measurements are linearized before the algorithm is performed:
```
Jerry-Jinfeng-Guo, line 406:
```suggestion
which is not realistic in distribution grids. Therefore, traditional measurements are linearized prior to running the algorithm:
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
-prevent the calculation from raising an exception if it is unobservable, therefore giving faulty results.
+The algorithm will assume angles to be zero by default (see the details about voltage sensors). This produces more correct outputs when the system is observable, but will prevent the calculation from raising an exception, even if it is unobservable, therefore giving faulty results.
 ```
 
-Algorithm call: `CalculationMethod.iterative_linear`
+#### Newton-Raphson state estimation
 
-### Short circuit algorithms
+```warning
+This feature is still experimental and is at the time of writing not yet publicly available.
```
Jerry-Jinfeng-Guo, line 471:
```suggestion
 At the time of writing this feature is still experimental and is not yet publicly available.
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+```
+
+Algorithm call: {py:class}`CalculationMethod.newton_raphson <power_grid_model.enum.CalculationMethod.newton_raphson>`
+
+The Newton-Raphson approach solves state estimation by considering it as a system of real, non-linear equations.
```
Jerry-Jinfeng-Guo, line 476:
```suggestion
The Newton-Raphson state estimation considers the problem as a system of real, non-linear equations.
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+It then iteratively solves the first order Taylor expansion of that system. It does not make any linearization assumptions.
```
Jerry-Jinfeng-Guo, line 477:
```suggestion
It iteratively solves the first order Taylor expansion of the system. It does not make any assumptions on linearization.
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+It may therefore be more accurate and is likely to achieve convergence in less iterations than the [iterative linear](#iterative-linear-state-estimation) approach.
```
Jerry-Jinfeng-Guo, line 478:
```suggestion
It could be more accurate and converge in less iterations than the [iterative linear](#iterative-linear-state-estimation) approach.
```

Jerry-Jinfeng-Guo: The reason that it could be more accurate is unclear from context, hence removing `Therfore`.

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+However, the Jacobian matrix needs to be calculated every iteration and cannot be prefactorized, which, on average, causes it to be slower.
```
Jerry-Jinfeng-Guo, line 479:
```suggestion
However, the Jacobian matrix needs to be calculated every iteration and cannot be prefactorized, which, on average, makes it slower.
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+
+The Newton-Raphson methode considers all measurements to be independent real measurements.
```
Jerry-Jinfeng-Guo, line 481:
```suggestion
The Newton-Raphson method considers all measurements to be independent real measurements.
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+I.e., $\sigma_P$, $\sigma_Q$ and $\sigma_U$ are all independent values.
+The rationale behind to calculation is similar to that of the [Newton-Raphson for power flow](#newton-raphson-power-flow).
+Because of that, the iteration process differs slightly from the one for [iterative linear state estimation](#iterative-linear-state-estimation), as shown below.
```
Jerry-Jinfeng-Guo, line 484:
```suggestion
Consequently, the iteration process differs slightly from that of [iterative linear state estimation](#iterative-linear-state-estimation), as shown below.
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+  - Construct and (LU-)factorize the Jacobian matrix for $\boldsymbol{x}^{(k)}$, using the network parameters.
+  - Compute the temporary new voltage magnitude and angle result, $\widetilde{\boldsymbol{U}}^{(k)}$ and $\widetilde{\boldsymbol{\theta}}^{(k)}$,
+  using the prefactorized matrix. See also [Matrix-prefactorization](./performance-guide.md#matrix-prefactorization)
+  - Normalize the result voltage phasor angle by setting angle of slack bus to zero:
+  $\underline{U}_i^{(k)} = \widetilde{\underline{U}}_i^{(k)} \times |\widetilde{\underline{U}}_s^{(k)}| / \widetilde{\underline{U}}_s^{(k)}$.
+  - If the maximum deviation between $\underline{\boldsymbol{U}}^{(k)}$ and $\underline{\boldsymbol{U}}^{(k-1)}$ is smaller than the tolerance $\epsilon$,
+  stop the iteration, otherwise continue until the maximum number of iterations is reached.
+  Note: we're using the phasor here: $\underline{\boldsymbol{U}} = \boldsymbol{U}e^{j\boldsymbol{\theta}}$.
+
+As for the [iterative linear](#iterative-linear-state-estimation) approach, in the iteration process, the phase angle of voltages at each bus is updated using the last iteration.
```
Jerry-Jinfeng-Guo, line 503:
```suggestion
As for the [iterative linear](#iterative-linear-state-estimation) approach, during iterations, phase angles of voltage at each bus are updated using ones from the previous iteration.
```

docs/user_manual/calculations.md
```diff
@@ -419,31 +445,69 @@ for the next iteration:
...
+The system error of the phase shift converges to zero.
+
+```{warning}
+The algorithm will assume angles to be zero by default (see the details about voltage sensors). This produces more correct outputs when the system is observable, but will prevent the calculation from raising an exception, even if it is unobservable, therefore giving faulty results.
+```
```
Jerry-Jinfeng-Guo, line 508: See above comment.

Jerry-Jinfeng-Guo (commented):
Few comments. Mostly suggestions on wording.

docs/user_manual/calculations.md
```diff
@@ -34,7 +34,7 @@ Output:
 #### State estimation
 
 State estimation is a statistical calculation method that determines the most probable state of the grid, based on
-network data and measurements. Measurements meaning power flow or voltage values with some kind of uncertainty, which were 
+network data and measurements. Here, measurements may be power flow or voltage values with some kind of uncertainty, which were
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Now that I re-read:
```suggestion
network data and measurements. Here, measurements can be power flow or voltage values with certain kind of uncertainty, which were 
```

Jerry-Jinfeng-Guo:
Upon discussion with @TonyXiang8787 , the reviews are incorporated into the documentation and all related conversations are hereby resolved.

TonyXiang8787 (approved):

PR #494, opened 2024-02-14
Feature/test unbalanced power measurement

mgovers:
proof that NRSE is much more accurate than iterative linear.

@TonyXiang8787 the residuals are `0.0` within the `atol` for `.+_residual`. i did a quick investigation to see whether the resulting residuals indeed are correct, and i confirmed that to be the case for both the voltage sensor and the power sensors, across all scenarios

Relates to #237 and #369


tests/data/state_estimation/unbalanced-power-measurements-newton-raphson/input.json
```diff
@@ -0,0 +1,98 @@
...
+        "p_measured": 1010000.0,
+        "q_measured": 210000.0,
+        "power_sigma": 10316.0
+      },
+      {
+        "id": 801,
+        "measured_object": 8,
+        "measured_terminal_type": 4,
+        "p_measured": 1020000.0,
+        "q_measured": 220000.0,
```
Jerry-Jinfeng-Guo, line 93: Since how many digits do we use scientific notation? Lots of `0`'s for `p_measured`and `q_measured` ;)

Jerry-Jinfeng-Guo (commented):
Only one complaint

tests/data/state_estimation/unbalanced-power-measurements-newton-raphson/input.json
mgovers: for now, this doesn't really matter, especially because this was 1-to-1 copied from the original iterative linear validation test. you're right that it could've been nicer, but imho it's clear enough as is. in addition: keeping the tests as identical as possible has its own merits

Jerry-Jinfeng-Guo: Haha, fair enough XD

Jerry-Jinfeng-Guo (approved):

PR #495, opened 2024-02-15
Finalize NRSE

mgovers:
* [x] prepare_matrix_and_rhs:
  * [x] BUG: (L250) also clear fill-in
  * [x] (L328) no prefactorize is needed.
* [x] The NRSEGainBlock (L64-L67):
  * [x] qt should be (hermitian) transposed
  * [x] qt should be directly obtained from hermitian transpose instead of per element
* [x] Run_state_estimation:
  * [x] (L163) run solve directly using `prefactorize_and_solve`
* [x] fix dense grouped index vector not initialized by default
* [x] sensor-update case for NRSE
* [x] distribution-case for NRSE and iterative linear
* [x] single-line-load case

Relates to #369


TonyXiang8787:
After examination I have concluded that the case `sensor-update` has indeed another local minimum for NRSE. Thus a new test case created.

mgovers:
> After examination I have concluded that the case `sensor-update` has indeed another local minimum for NRSE. Thus a new test case created.

ok, clear. i'm working on the block matrix thingy. should i do that in this PR or a separate one?

TonyXiang8787:
> > After examination I have concluded that the case `sensor-update` has indeed another local minimum for NRSE. Thus a new test case created.
> 
> ok, clear. i'm working on the block matrix thingy. should i do that in this PR or a separate one?

Just push to here. We try to finish everything.

mgovers:
> > > After examination I have concluded that the case `sensor-update` has indeed another local minimum for NRSE. Thus a new test case created.
> > 
> > 
> > ok, clear. i'm working on the block matrix thingy. should i do that in this PR or a separate one?
> 
> Just push to here. We try to finish everything.

i would like to do the version bump + remove experimental flag in a separate PR, just to really make sure that that version bump is extremely minimal and that we don't end up finding something **again** before pushing

TonyXiang8787 (dismissed):

tests/data/state_estimation/distribution-case/params.json
```diff
@@ -1,16 +1,8 @@
 {
-  "calculation_method": ["iterative_linear", "newton_raphson"],
+  "calculation_method": "iterative_linear",
```
mgovers, line 2: you disabled `newton_raphson`. that was not intended, right?

TonyXiang8787 (approved):

mgovers:
I see no more weird/unexpected things. CI appears to go OK both locally and remotely. Hereby enabling automerge

PR #496, opened 2024-02-15
fix Documentation of calculation of y_base

Welthulk:

> Here you can elaborate on the chosen solution strategy, which changes did you make and which goal do they serve. Perhaps also which things are you still unsure of.

-
-
- ..

> Here you can point out modules or complex implementation that require special attention, .e.g have a look at module `foo.py` and `bar.py`.

- 
- 
- ..


mgovers:
Hi @sudo-ac and @Welthulk , can you please fix the DCO? You can follow the steps in https://github.com/PowerGridModel/power-grid-model/pull/496/checks?check_run_id=21620188944 .

TonyXiang8787 (approved):

Issue #493, opened 2024-02-13
[FEATURE] Add custom attributes for serialization


TonyXiang8787:
# Background

Currently the (de-)serialization only supports attributes which are defined by PGM self. Users wish to add custom attributes to the (de-)serialization format and parser. Following attribute types can be 

- Attributes per element
- Aggregated attributes along all elements for a given type and scenario
- Aggregated attributes along all scenarios for a given type and element
- Global attributes for a given type

# Approach

To be continued

IrenaDk:
This would be very useful for our product because we do a lot of manipulations on the initial PGM network that we use and it would be useful to (de)serialize it with the additional attributes. Such output can be used as input in the next steps of our workflow, final output, and has the additional benefit of making debugging easier.

mgovers:
Initial thoughts: we could go a number of ways:

1. always aggregate. the results would end up in the meta data of the output. this marks a compatible change tot the serialization schema. the difficult decision here is whether it should be ignored or checked for `strict=true` when deserializing
2. add optional flags to aggregate across multiple dimensions and output.
3. similarly to the `extra_info` attribute to the PgmJsonConverter: make this optional and output "all the rest" as an new string of the same data type as the input. This would not break current functionality and with the visitor pattern, it should be relatively straight forward to produce, but i do not think this enables all of the above requested features

PR #497, opened 2024-02-16
Release v1.7.0: Newton-Raphson State Estimation

### LINKED ISSUE #237: [FEATURE] Possibility to specify different sigma's for P and Q in state estimator

### LINKED ISSUE #369: Newton-Raphson method for state estimation

mgovers:
Closes #369
Closes #237 

* Bump version
* Make Newton-Raphson state estimation stable (instead of experimental)

* [ ] full clang-tidy run


TonyXiang8787:
@mgovers also remove the deprecated import export json functions.

mgovers:
> @mgovers also remove the deprecated import export json functions.

can we do that? i think the tests in PGM-IO still use the old function calls as a fallback. Would imply a version bump on PGM-IO as well.

I would propose to wait until #493 is implemented, no?

TonyXiang8787:
@mgovers please do not create a release branch

TonyXiang8787:
> > @mgovers also remove the deprecated import export json functions.
> 
> 
> 
> can we do that? i think the tests in PGM-IO still use the old function calls as a fallback. Would imply a version bump on PGM-IO as well.
> 
> 
> 
> I would propose to wait until #493 is implemented, no?

Then keep the json functions for now.

mgovers:
> @mgovers please do not create a release branch

yeah i noticed. will recreate the PR from a different branch

PR #498, opened 2024-02-16
Release v1.7.0: Newton-Raphson State Estimation

### LINKED ISSUE #237: [FEATURE] Possibility to specify different sigma's for P and Q in state estimator

### LINKED ISSUE #369: Newton-Raphson method for state estimation

mgovers:
Closes #369
Closes #237 

* Bump version
* Make Newton-Raphson state estimation stable (instead of experimental)

* [ ] full clang-tidy run

Re-creation of https://github.com/PowerGridModel/power-grid-model/pull/497 due to `release/**` prefix


mgovers:
failing CI job should disappear after the other CI passes. it's a remnant from the other PR.

TonyXiang8787:
Your branch name cannot contain release...

mgovers:
> Your branch name cannot contain release...

ohhhh... yeah that's a bug in the CI 😬 there's a reason prefixes are used (e.g. `release/**` instead of `constains("release")`)

let's not fix that now but in a separate PR. Will re-create PR again

PR #503, opened 2024-02-20
Feature/Updated validation to allow for missing sigma

Jerry-Jinfeng-Guo:
Updated the validation, test and documentation.


src/power_grid_model/validation/validation.py
```diff
@@ -221,6 +221,20 @@ def validate_ids_exist(update_data: Dict[str, np.ndarray], input_data: SingleDat
...
+    """
+    Helper function to process the required list when both `p_sigma` and `q_sigma` exist
+    and valid but `power_sigma` is missing.
+    """
+    if sensor in data and isinstance(data[sensor], np.ndarray):
+        sensor_data = data[sensor]
+        if "p_sigma" in sensor_data.dtype.names and "q_sigma" in sensor_data.dtype.names:
+            p_sigma = sensor_data["p_sigma"]
+            q_sigma = sensor_data["q_sigma"]
+            if not (np.isnan(p_sigma).any() or np.isnan(q_sigma).any()):
```
TonyXiang8787, line 234: We cannot do this for all the elements. This has to be validated per element. So element 1 can have a valid `power_sigma`, and element 2 can have no valid `power_sigma`, but a valid `p_sigma` and `q_sigma`.

TonyXiang8787: The current code checks `p_sigma` and `q_sigma` for **all** the `sym_power_sensor`. We want to check this for each inidividual `sym_power_sensor`. So `sym_power_sensor_1` might have a valid `power_sigma` but no `p_sigma` or `q_sigma`. And `sym_power_sensor_1` might have a no `power_sigma` but valid `p_sigma` or `q_sigma`.

Jerry-Jinfeng-Guo: I see what you mean. I'll take a look and make some changes accordingly (where necessary).

docs/user_manual/components.md
```diff
@@ -615,6 +615,18 @@ the meaning of `RealValueInput` is different, as shown in the table below.
 | `p_sigma`    | `RealValueInput` | watt (W)                   | standard deviation of the active power measurement error. Usually this is the absolute measurement error range divided by 3.   | &#10060; see the explanation below. | &#10004; | `> 0` |
 | `q_sigma`    | `RealValueInput` | volt-ampere-reactive (var) | standard deviation of the reactive power measurement error. Usually this is the absolute measurement error range divided by 3. | &#10060; see the explanation below. | &#10004; | `> 0` |
 
+Valid combinations of `power_sigma`, `p_sigma` and `q_sigma` are:
+| `power_sigma` | `p_sigma` | `q_sigma` |  result  |
```
mgovers, line 619:
```suggestion

| `power_sigma` | `p_sigma` | `q_sigma` |  result  |
```

docs/user_manual/components.md
```diff
@@ -615,6 +615,18 @@ the meaning of `RealValueInput` is different, as shown in the table below.
...
+|:-------------:|:---------:|:---------:|:--------:|
+|    &#10004;   |  &#10004; |  &#10004; |  valid   |
+|    &#10004;   |  &#10004; | &#10060;  | invalid  |
+|    &#10004;   | &#10060;  |  &#10004; | invalid  |
+|    &#10004;   | &#10060;  | &#10060;  |  valid   |
+|   &#10060;    |  &#10004; |  &#10004; |  valid   |
+|   &#10060;    |  &#10004; | &#10060;  | invalid  |
+|   &#10060;    | &#10060;  |  &#10004; | invalid  |
+|   &#10060;    | &#10060;  | &#10060;  | invalid  |
```
mgovers, line 628: maybe instead provide just a `x`-mark for "provided" and then a &#10004; or a &#10060; for the `valid/invalid` column. i think that looks easier

docs/user_manual/components.md
```diff
@@ -594,7 +594,7 @@ Because of this distribution, at least one appliance is required to be connected
...
-| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Concrete Types`. | &#10024; only for state estimation. | &#10004; |                        `> 0`                         |
+| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Concrete Types`. | &#10024; only for state estimation. See the explanation below. | &#10004; |                        `> 0`                         |
```
mgovers, line 597: we also have to think about the possiblity that in the future we provide e.g. only a single-phase asym power sensor or maybe even only an active power sensor or only a reactive power sensor. is just the comment `See the explanation below.` sufficient? maybe make this something like

```suggestion
| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Concrete Types`. | &#10024; in certain cases for state estimation. See the explanation below. | &#10004; |                        `> 0`                         |
```

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: Taken into account in the latest change.

docs/user_manual/components.md
Jerry-Jinfeng-Guo: Changes included in the latest commit.

docs/user_manual/components.md
Jerry-Jinfeng-Guo: Changes included in the latest commit.

docs/user_manual/components.md
```diff
@@ -594,7 +594,7 @@ Because of this distribution, at least one appliance is required to be connected
...
-| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Concrete Types`. | &#10024; only for state estimation. | &#10004; |                        `> 0`                         |
+| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Concrete Types`. | &#10024; in certain cases for state estimation. See the explanation below. | &#10004; |                        `> 0`                         |
```
mgovers, line 597: now that i read this back, it's in a different (sub)section, so maybe this makes it more clear:

```suggestion
| `power_sigma`            | `double`                                                                      | volt-ampere (VA) | standard deviation of the measurement error. Usually this is the absolute measurement error range divided by 3. See {hoverxreftooltip}`user_manual/components:Power Sensor Concrete Types`. | &#10024; in certain cases for state estimation. See the explanation for [concrete types](#power-sensor-concrete-types) below. | &#10004; |                        `> 0`                         |
```

src/power_grid_model/validation/validation.py
```diff
@@ -221,6 +221,23 @@ def validate_ids_exist(update_data: Dict[str, np.ndarray], input_data: SingleDat
...
+def process_power_sigma_and_p_q_sigma(data: SingleDataset, sensor: str) -> None:
+    """
+    Helper function to process the required list when both `p_sigma` and `q_sigma` exist
+    and valid but `power_sigma` is missing. The field `power_sigma` is set to 1 in this case.
+    The argument is that when both `p_sigma` and `q_sigma` are valid, the value pf `power_sigma`
+    is always ignored.
+    """
+    if sensor in data and isinstance(data[sensor], np.ndarray):
+        sensor_data = data[sensor]
+        for i in range(len(sensor_data)):
```
mgovers, line 233: is there a way to do this using `np.logical_*` or `np.vectorize` to avoid the Python `for`-loop? or is that being done in plenty of other locations in the code already?

tests/unit/validation/test_validation_functions.py
```diff
@@ -646,3 +648,69 @@ def test_validate_generic_power_sensor__terminal_types(
...
+    # power sensor
+    power_sensor = initialize_array("input", "sym_power_sensor", 2)
+    power_sensor["id"] = [6, 7]
+    power_sensor["measured_object"] = [2, 4]
+    power_sensor["measured_terminal_type"] = [MeasuredTerminalType.branch_from, MeasuredTerminalType.load]
+    power_sensor["p_measured"] = [1e6, -1e6]
+    power_sensor["q_measured"] = [1e6, -1e6]
+    power_sensor["power_sigma"] = [np.nan, 1e9]
+    power_sensor["p_sigma"] = [1e4, np.nan]
+    power_sensor["q_sigma"] = [1e9, np.nan]
```
mgovers, line 704: maybe add another sensor that has both specified

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: Included in the latest commit.

tests/unit/validation/test_validation_functions.py
```diff
@@ -9,8 +9,10 @@
 import numpy as np
 import pytest
 
-from power_grid_model import MeasuredTerminalType, initialize_array, power_grid_meta_data
+# from power_grid_model import PowerGridModel, LoadGenType, MeasuredTerminalType, initialize_array, CalculationType, CalculationMethod
```
mgovers, line 12:
```suggestion
```

docs/user_manual/components.md
mgovers, line 619: nitpick
```suggestion
| `power_sigma` | `p_sigma` | `q_sigma` |  result  |

```

src/power_grid_model/validation/validation.py
```diff
@@ -221,6 +221,28 @@ def validate_ids_exist(update_data: Dict[str, np.ndarray], input_data: SingleDat
     return list(chain(*errors))
 
 
+def process_power_sigma_and_p_q_sigma(data: SingleDataset, sensor: str) -> None:
+    """
+    Helper function to process the required list when both `p_sigma` and `q_sigma` exist
+    and valid but `power_sigma` is missing. The field `power_sigma` is set to 1 in this case.
```
mgovers, line 227: i think this comment is out of date, right?

src/power_grid_model/validation/validation.py
```diff
@@ -221,6 +221,28 @@ def validate_ids_exist(update_data: Dict[str, np.ndarray], input_data: SingleDat
     return list(chain(*errors))
 
 
+def process_power_sigma_and_p_q_sigma(data: SingleDataset, sensor: str) -> None:
```
mgovers, line 224: this may actively make changes to the user input data set. that might not be desirable, right? why not just create a validation check?

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: Indeed, "...set to the norm of..."

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: > you may in that case remove it from the standard list of required fields, and just make a custom check next to that check. In more formal terms: you make the fields optional and impose another sanity check, similar to how is done for the valid range of e.g. s0 for transformers
This of course can be done. I would argue for simplicity here. It was made clear to users that with such input combination, one is always ignored. By logic they shouldn't  expect that we leave that nan and not put a reasonable value in its place. Besides, this will also require changes to many test cases that explicitly check `assert (MissingValueError("sym_power_sensor", "power_sigma", [NaN]) in required_values_errors) == [something something]`

petersalemink95: I agree with @mgovers The input validation is just a validation. We should never change the input itself here.

docs/user_manual/components.md
mgovers: Whoops sorry no an empty line before the table begins, apologies

src/power_grid_model/validation/validation.py
mgovers: As a user, i own the data. Who knows if i attached some kind of database logic to any changes in my data. If i call a function that does not explicitly mention side effects, i do not expect any, as it may have extensive consequenses. It also makes usage of this functionality (validate and then investigate why it happened) difficult

Jerry-Jinfeng-Guo: @mgovers @petersalemink95 Would the idea of doing a check on proxy appeal to you? That way the original input data will stay intact/untouched.

Jerry-Jinfeng-Guo: @mgovers @petersalemink95 Input data is left untouched in the latest changes.

mgovers: If it works, then it's fine, but please add a test that proves that

src/power_grid_model/validation/validation.py
```diff
@@ -338,13 +338,14 @@ def validate_required_values(
     required["sym_power_sensor"] = required["power_sensor"].copy()
     required["asym_power_sensor"] = required["power_sensor"].copy()
 
+    _data = data # proxy
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: You are absolutely right here, I am working on the solution ;)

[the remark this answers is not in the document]
mgovers: 💀 rip python

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: Changes pushed. Each power sensor got its individual validation w.r.t. `power_sigma`.

src/power_grid_model/validation/validation.py
```diff
@@ -221,6 +221,43 @@ def validate_ids_exist(update_data: Dict[str, np.ndarray], input_data: SingleDat
     return list(chain(*errors))
 
 
+def process_power_sigma_and_p_q_sigma(
```
mgovers, line 224: since it actually has some behaviour that a user should never directly call, please make this function hidden from the module interface

```suggestion
def _process_power_sigma_and_p_q_sigma(
```

src/power_grid_model/validation/validation.py
```diff
@@ -331,7 +379,25 @@ def validate_required_values(
...
+            *(
+                none_missing(data, component, item, index)
+                for component in data
+                for index, item in enumerate(
+                    (sublist for sublist in required.get(component, []))
+                    if isinstance(required.get(component, []), list)
+                    and all(isinstance(i, list) for i in required.get(component, []))
+                    else [required[component]]
+                )
+                if data[component] is not None and isinstance(data[component], Sized) and index < len(data[component])
```
mgovers, line 397: this statement is so complex that it for me suggests that this is probably not the right solution. maybe earlier you should've just created a `np.array` instead of a `list` to enable this?

IMO, the original return statement was still readable but now it's gotten a bit too much for me.

while i am very much for chaining in the modern programming sense (or similarly in the Lisp sense), not everything is more readable in one statement. remember the fork bomb? `:(){:|:&};:`

src/power_grid_model/validation/validation.py
```diff
@@ -331,7 +379,25 @@ def validate_required_values(
         required["line"] += ["r0", "x0", "c0", "tan0"]
         required["shunt"] += ["g0", "b0"]
 
-    return list(chain(*(none_missing(data, component, required.get(component, [])) for component in data)))
+    # Mark `power_sigma` for each power_sensor based on the state of `p_sigma` and `q_sigma`
+    process_power_sigma_and_p_q_sigma(data, "sym_power_sensor", required)
+    process_power_sigma_and_p_q_sigma(data, "asym_power_sensor", required)
+
```
mgovers, line 385: i'm still wondering why the alternative implementation of making all three of `power_sigma`, `p_sigma` and `q_sigma` optional and then just doing a sanity check such that

```py
s_nan = np.nan(power_sigma)
p_nan = np.nan(p_sigma)
q_nan = np.nan(q_sigma)
if np.any(np.logical_xor(p_nan, q_nan)):
    raise ValidationError("Only one p_sigma and q_sigma is provided")
if np.any(np.logical_and(s_nan, p_nan)):
    raise ValidationError("Neither power_sigma nor p_sigma and q_sigma is provided")
```

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: Well, when it comes to this sort of topic, it is always a matter of degree. 
While 
> not the right solution

is clearly not the right wording, I agree with you that it might be too much for one statement. However, For its purpose of tailoring the `required` field according to `data`, it might not need further modification.

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: This approach would mean that for every future such check we would have to write a specific function. While if we make the `required` field tailored to the data, which we are doing anyways, we might as well just do it the right way. Wouldn't you agree?

mgovers: `required` really means required without any changes. if you look at the implementation for the trio (`sn`, `uk`, `pk`), then there are additional requirements imposed to the explicit structure.

in this specific case, neither `power_sigma` nor `p_sigma` nor `q_sigma` are `required` in the classical sense (since if you only look at `power_sigma`, then it is technically `optional`). Instead, there's an additional requirement that either `power_sigma` and/or both `p_sigma` and `q_sigma` need to be provided. It may raise the same exception as the `required` checks, but it should not follow the same code path

Jerry-Jinfeng-Guo: > required really means required without any changes.

Are you referring to "changes to the `required` field" or "changes to the data"?

> in this specific case, neither power_sigma nor p_sigma nor q_sigma are required in the classical sense

`power_sigma` most definitely have lived in `required::*sym_power_sensor` for quite a while and not `optional`.

In short, either the previous logic is problematic or misleading OR it's not the way you personally understood/liked it.

Jerry-Jinfeng-Guo:
I'll happily make the changes and switch to another incremental approach if the whole team agrees @mgovers @petersalemink95 @TonyXiang8787 . Else I'll stick to my approach, and perhaps simplify some of the code to reduce reading complexity. 

Summary:
The issue relates to the check regarding the `power_sigma` inside `*sym_power_sensor`. Previously it lives inside `required::*sym_power_sensor`. Since data input can contain multitude of such sensors, I proposed the solution to tailor the `required` field to `data`. Martijn suggested that they should be checked next to the regular `required` check, by first removing them from `required` entirely and then followed by a dedicated `power_sigma`/`p_sigma`/`q_sigma` check. 

While I agree with Martijn's comment on part of the code's reading complexity, I do not see this incremental check for every new feature as a better solution. I stand corrected. Could use some input here, thanks in advance!

TonyXiang8787:
@mgovers @Jerry-Jinfeng-Guo My proposal is to push this with minimal necessary refactoring.

mgovers:
> @mgovers @Jerry-Jinfeng-Guo My proposal is to push this with minimal necessary refactoring. Do not change the logic of how validation function works unless it is really needed.

it feels like we're building up some serious tech debt here that may make changes harder in the future and sets a precedent. what's your rationale?

TonyXiang8787:
> > @mgovers @Jerry-Jinfeng-Guo My proposal is to push this with minimal necessary refactoring. Do not change the logic of how validation function works unless it is really needed.
> 
> it feels like we're building up some serious tech debt here that may make changes harder in the future and sets a precedent. what's your rationale?

My rationale is that we might one day move the whole validation process in C++ core with semi-automatically generated code based on the pre-defined attributes constraints.

mgovers:
> > > @mgovers @Jerry-Jinfeng-Guo My proposal is to push this with minimal necessary refactoring. Do not change the logic of how validation function works unless it is really needed.
> > 
> > 
> > it feels like we're building up some serious tech debt here that may make changes harder in the future and sets a precedent. what's your rationale?
> 
> My rationale is that we might one day move the whole validation process in C++ core (but still separate optional function in C API) with semi-automatically generated code based on the pre-defined attributes constraints.

while i very much like that, that don't believe that that's anywhere near prioritized by @petersalemink95 now

petersalemink95:
> > > > @mgovers @Jerry-Jinfeng-Guo My proposal is to push this with minimal necessary refactoring. Do not change the logic of how validation function works unless it is really needed.
> > > 
> > > 
> > > it feels like we're building up some serious tech debt here that may make changes harder in the future and sets a precedent. what's your rationale?
> > 
> > 
> > My rationale is that we might one day move the whole validation process in C++ core (but still separate optional function in C API) with semi-automatically generated code based on the pre-defined attributes constraints.
> 
> while i very much like that, that don't believe that that's anywhere near prioritized by @petersalemink95 now

I think this wil happen in the long future. For now, I'd like to keep the code similar everywhere and not make new tech dept for different checks. We can do them all in a similar way as has been done before

src/power_grid_model/validation/validation.py
petersalemink95: That is because `power_sigma` was required previously. Now it is not, as long as `p_sigma` and `q_sigma` were provided

Jerry-Jinfeng-Guo:
> > > > > @mgovers @Jerry-Jinfeng-Guo My proposal is to push this with minimal necessary refactoring. Do not change the logic of how validation function works unless it is really needed.
> > > > 
> > > > 
> > > > it feels like we're building up some serious tech debt here that may make changes harder in the future and sets a precedent. what's your rationale?
> > > 
> > > 
> > > My rationale is that we might one day move the whole validation process in C++ core (but still separate optional function in C API) with semi-automatically generated code based on the pre-defined attributes constraints.
> > 
> > 
> > while i very much like that, that don't believe that that's anywhere near prioritized by @petersalemink95 now
> 
> I think this wil happen in the long future. For now, I'd like to keep the code similar everywhere and not make new tech dept for different checks. We can do them all in a similar way as has been done before

Well, I will ask for some clearance here then: my proposed solution (A) and Martijn's opinion (B) are both trying to do just this, but from different perspective. Mine kept the original checks and only made it more tailored; Marthijn's follows existing paradigm of incrementally adding new checks whenever there are new features needed special treatment. In a way, this is the result of keeping the same goal but with different interpretations. So here comes the question, A or B? @TonyXiang8787 @petersalemink95

TonyXiang8787:
@Jerry-Jinfeng-Guo @mgovers @petersalemink95 the situation about `power_sigma`, `p_sigma` and `q_sigma` is really exceptional. Making some tailored code for that would not build much technical debt, I think.

So I am in favor of @Jerry-Jinfeng-Guo's current solution.

src/power_grid_model/validation/validation.py
TonyXiang8787, line 397: @mgovers @Jerry-Jinfeng-Guo decision after discussion:

Clean this part up to make it more readable and less intrusive.

Jerry-Jinfeng-Guo: Code cleaned in the latest commit.

src/power_grid_model/validation/validation.py
mgovers: this function should be module-internal. no user should ever want to do this unless they really know what they are doing (and even then, they are most likely better off writing their own function because we don't guarantee a stable API for functions prefixed `_` by PEP convention)

src/power_grid_model/validation/validation.py
```diff
@@ -313,8 +351,19 @@ def validate_required_values(
...
+            required["power_sensor"].copy() for _ in range(data["sym_power_sensor"].shape[0])  # type: ignore
+        ]
+    except KeyError:
+        pass
+    try:
+        required["asym_power_sensor"] = [
+            required["power_sensor"].copy() for _ in range(data["asym_power_sensor"].shape[0])  # type: ignore
+        ]
+    except KeyError:
+        pass
```
mgovers, line 366:
```suggestion
    # Different requirements for individual sensors. Avoid shallow copy.
    for sensor_type in ("sym_power_sensor", "asym_power_sensor"):
        try:
            required[sensor_type] = [
                required["power_sensor"].copy() for _ in range(data[sensor_type].shape[0])  # type: ignore
            ]
        except KeyError:
            pass
```

mgovers (commented):
love it. just some minor remarks and hopefully we can get this out the door today

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: Included in the latest commit.

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: Included in the latest commit.

PR #505, opened 2024-02-21
[BUG] Netwon-Raphson power flow sometimes selects wrong initial value

TonyXiang8787:
# Introduction

The Newton-Raphson power flow (NRPF) chooses a flat start (all nodes start with the same voltage as source voltage) as initial value. This works in most of the time. However, if the actual solution is too far away from the initial value, the NRPF fails to converge or converge to unrealistic results.

# Reproduce error

In this PR a test case has been added to reproduce the error. In the test case there are two transformers which boost the voltage to almost 1.5 pu. The deviation between the actual solution and the initial value is too large. NRPF converges to a solution with almost zero voltages which are not realistic. Other methods give the correct solution.

# Solution

We need to re-think the way NRPF set the initial values. Maybe we run first a linear power flow (const impedance) with no loads/generators, we can use that value as initial value for NRPF. This is a PR working in progress to solve the issue.

# Code approach

We propose to extract the common part (the loop to add sources) of `prepare_matrix_and_rhs` of linear and iterative current solver into a common function. Also consider to re-organize the loop to loop source first, because sources are very rare.

## Code section in linear solver

https://github.com/PowerGridModel/power-grid-model/blob/037a0e43cef1e36f49df73f177a601967ecf553b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp#L97-L104

## Code section in iterative current solver

https://github.com/PowerGridModel/power-grid-model/blob/037a0e43cef1e36f49df73f177a601967ecf553b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_current_pf_solver.hpp#L115-L123

## Adjustment in NRPF

In the initialization of NRPF, we also construct a complex Matrix and RHS with the new common function. We solve the linear system, use the resulted voltage for the initial value for further NRPF.


mgovers:
i've actually been brainstorming about this in the context of state estimation. sometimes, the initial conditions are quite well-known. e.g. for a load, `p_specified` may actually be known. if that is the case for all of the grid, voltages provided by doing a single linear powerflow calculation run may provide a better initial value than just plainly `u_rated`.

I do think that a single linear powerflow run is the best solution. how we design that may actually be related to any refactoring we think we should undergo for the automatic tap math workout

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -211,7 +211,18 @@ template <symmetry_tag sym> class NewtonRaphsonPFSolver : public IterativePFSolv
           perm_(y_bus.size()) {}
 
     // Initilize the unknown variable in polar form
-    void initialize_derived_solver(YBus<sym> const& /* y_bus */, MathOutput<sym> const& output) {
+    void initialize_derived_solver(YBus<sym> const& y_bus, PowerFlowInput<sym> const& input, MathOutput<sym>& output) {
+        ComplexTensorVector<sym> linear_mat_data(y_bus.nnz_lu());
+        SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>> linear_sparse_solver{
+            y_bus.shared_indptr_lu(), y_bus.shared_indices_lu(), y_bus.shared_diag_lu()};
+        typename SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>>::BlockPermArray linear_perm(
```
mgovers, line 218: please use a `using` declaration at the start of the body of this function for readability

mgovers (commented):
very nice refactor so far

nitbharambe:
The source results would have slight difference. `sk` is set to 1e20. If the solution gets converged at a different place than original but within tolerance, the effect on source results would move it out of rtol / atol.


```
/home/runner/work/power-grid-model/power-grid-model/tests/cpp_validation_tests/test_validation.cpp:272: ERROR: CHECK( match ) is NOT correct!
  values: CHECK( false )
  logged: dataset scenario: #0, Component: source #0, attribute: p: actual = 2402572.687939726 vs. expected = 2401665

/home/runner/work/power-grid-model/power-grid-model/tests/cpp_validation_tests/test_validation.cpp:272: ERROR: CHECK( match ) is NOT correct!
  values: CHECK( false )
  logged: dataset scenario: #0, Component: source #0, attribute: q: actual = -2946135.465605061 vs. expected = -2946045

/home/runner/work/power-grid-model/power-grid-model/tests/cpp_validation_tests/test_validation.cpp:272: ERROR: CHECK( match ) is NOT correct!
  values: CHECK( false )
  logged: dataset scenario: #0, Component: source #0, attribute: i: actual = 206.86595025[375](https://github.com/PowerGridModel/power-grid-model/actions/runs/8348788311/job/22851503528?pr=505#step:7:376)17 vs. expected = 206.830905

/home/runner/work/power-grid-model/power-grid-model/tests/cpp_validation_tests/test_validation.cpp:272: ERROR: CHECK( match ) is NOT correct!
  values: CHECK( false )
  logged: dataset scenario: #0, Component: source #0, attribute: s: actual = 3801587.7870345 vs. expected = 3800944

/home/runner/work/power-grid-model/power-grid-model/tests/cpp_validation_tests/test_validation.cpp:272: ERROR: CHECK( match ) is NOT correct!
  values: CHECK( false )
  logged: dataset scenario: #0, Component: source #0, attribute: pf: actual = 0.6319918998408551 vs. expected = 0.63186
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -211,7 +211,19 @@ template <symmetry_tag sym> class NewtonRaphsonPFSolver : public IterativePFSolv
           perm_(y_bus.size()) {}
 
     // Initilize the unknown variable in polar form
-    void initialize_derived_solver(YBus<sym> const& /* y_bus */, MathOutput<sym> const& output) {
+    void initialize_derived_solver(YBus<sym> const& y_bus, PowerFlowInput<sym> const& input, MathOutput<sym>& output) {
```
TonyXiang8787, line 214: Maybe we want to also store this LU-factorization in the class variable to save time. The factorization does not need to recompute as long as the math model parameters do not change. Have a look at the iterative current solver.

TonyXiang8787 (commented):
@nitbharambe see my suggestion.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
```diff
@@ -87,29 +87,12 @@ template <symmetry_tag sym> class LinearPFSolver {
     // sparse linear equation
     ComplexTensorVector<sym> mat_data_;
     // sparse solver
-    SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>> sparse_solver_;
-    typename SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>>::BlockPermArray perm_;
+    using SparseSolverType = SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>>;
+    SparseSolverType sparse_solver_;
+    typename SparseSolverType::BlockPermArray perm_;
```
mgovers, line 92:
```suggestion
    using SparseSolverType = SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>>;
    using BlockPermArray = typename SparseSolverType::BlockPermArray;

    SparseSolverType sparse_solver_;
    BlockPermArray perm_;
```

or move the `using` declarations to top of class

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -269,9 +281,11 @@ template <symmetry_tag sym> class NewtonRaphsonPFSolver : public IterativePFSolv
     // 2. power unbalance: p/q_specified - p/q_calculated
     // 3. unknown iterative
     std::vector<ComplexPower<sym>> del_x_pq_;
-    SparseLUSolver<PFJacBlock<sym>, ComplexPower<sym>, PolarPhasor<sym>> sparse_solver_;
+
+    using SparseSolverType = SparseLUSolver<PFJacBlock<sym>, ComplexPower<sym>, PolarPhasor<sym>>;
+    SparseSolverType sparse_solver_;
     // permutation array
-    typename SparseLUSolver<PFJacBlock<sym>, ComplexPower<sym>, PolarPhasor<sym>>::BlockPermArray perm_;
+    typename SparseSolverType::BlockPermArray perm_;
```
mgovers, line 288: like before, please also make a `using` declaration for the block perm array

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_pf_solver.hpp
```diff
@@ -104,18 +91,72 @@ template <symmetry_tag sym, typename DerivedSolver> class IterativePFSolver {
...
+        std::vector<double> const& phase_shift = *phase_shift_;
+        // average u_ref of all sources
+        DoubleComplex const u_ref = [&]() {
+            DoubleComplex sum_u_ref = 0.0;
+            for (auto const& [bus, sources] : enumerated_zip_sequence(*sources_per_bus_)) {
+                for (Idx const source : sources) {
+                    sum_u_ref += input.source[source] * std::exp(1.0i * -phase_shift[bus]); // offset phase shift
+                }
+            }
+            return sum_u_ref / (double)input.source.size();
```
Jerry-Jinfeng-Guo, line 132: I was kinda waiting for Martijn's comment on this cool casting ;)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_pf_solver.hpp
```diff
@@ -104,18 +91,72 @@ template <symmetry_tag sym, typename DerivedSolver> class IterativePFSolver {
...
+                }
+            }
+            // prefactorize
+            LinearBlockPermArray perm(n_bus_);
+            linear_sparse_solver_.prefactorize(mat_data, perm);
+            // move pre-factorized version into shared ptr
+            linear_mat_data_ = std::make_shared<ComplexTensorVector<sym> const>(std::move(mat_data));
+            linear_perm_ = std::make_shared<LinearBlockPermArray const>(std::move(perm));
+        }
+        parameters_changed_ = false;
```
Jerry-Jinfeng-Guo, line 119: This line is redundant?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/linear_pf_solver.hpp
```diff
@@ -87,25 +87,27 @@ template <symmetry_tag sym> class LinearPFSolver {
...
         IdxVector const& bus_entry = y_bus.lu_diag();
         for (auto const& [bus_number, load_gens, sources] :
              enumerated_zip_sequence(*load_gens_per_bus_, *sources_per_bus_)) {
             Idx const diagonal_position = bus_entry[bus_number];
             auto& diagonal_element = mat_data_[diagonal_position];
             auto& u_bus = output.u[bus_number];
-            add_loads(load_gens, bus_number, input, diagonal_element);
-            add_sources(sources, bus_number, y_bus, input.source, diagonal_element, u_bus);
+            add_loads_to_lhs(load_gens, bus_number, input, diagonal_element);
+            add_sources_to_lhs_rhs(sources, bus_number, y_bus, input.source, diagonal_element, u_bus);
```
Jerry-Jinfeng-Guo, line 105: I think here it makes more sense to just do:
```
detail::add_sources_to_lhs_rhs(sources, bus_number, y_bus, input.source, diagonal_element, u_bus);
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_pf_solver.hpp
nitbharambe: It was vague. Better now.

mgovers:
@nitbharambe can we delete this branch?

nitbharambe:
@mgovers Lets keep this one as it has commits for different initializations refactored

nitbharambe:
A new branch is created: `feature/different-nr-initializations` to retain commits for the refactoring of different initialization methods.

Options tried:
1) LHS built with Ybus, ysource and const impedance loadgens. RHS with source injections. 
2) LHS built with Ybus, ysource. RHS with source injection.
3) LHS built with Ybus, ysouce. RHS built with source injection and load injections as const impedance. 
3.1) using flat start + phase shift for injections
3.2) using flat start for injections 
4) LHS built with Ybus, ysouce. RHS built with source injection and load injections as respective types. (ie. Linear current method)
4.1) using flat start + phase shift for injections
4.2)  using flat start for injections

Finally option 1 was selected and added in https://github.com/PowerGridModel/power-grid-model/pull/549
This PR is then closed as this bug was addressed.

PR #506, opened 2024-02-23
Feature/nrse example

zhen0427:
Update the example notebook of state estimation for Newton Raphson method.


docs/examples/Power Flow Example.ipynb
```diff
@@ -352,7 +352,7 @@
    "source": [
     "### Calculation Method\n",
     "\n",
-    "There are four calculation methods available it is defined as enumrations:\n",
+    "There are four calculation methods available and they are defined as the following enumrations:\n",
```
mgovers, line 355:
```suggestion
    "There are four calculation methods available. They are listed in the `CalculationMethod` enumeration as follows:\n",
```

docs/examples/State Estimation Example.ipynb
```diff
@@ -36,10 +36,22 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 4,
```
mgovers, line 39: please clear output and re-run all cells from scratch. this also provides a double-check that whatever you implemented is indeed correct

docs/examples/State Estimation Example.ipynb
```diff
@@ -36,10 +36,22 @@
...
    "id": "ae11dc9a",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "ModuleNotFoundError",
```
mgovers, line 44: you get errors... that shouldn't be, right? did you install all optional dependencies required to run the examples? you may need to create a new optional dependency group `[examples]` in the `pyproject.toml` file (just like how `[dev]` is defined)

docs/examples/State Estimation Example.ipynb
```diff
@@ -504,6 +528,87 @@
...
+   "id": "b2c7c914",
+   "metadata": {},
+   "source": [
+    "### Calculation Method\n",
+    "\n",
+    "There are two calculation methods available and they are defined as the following enumrations:\n",
+    "* **Iterative linear method**: \n",
+    "    * `CalculationMethod.iterative_linear`: All measurements are linearized before running the method.\n",
+    "* **Newton Raphson method**: The Newton-Raphson method considers the problem as a system of real, non-linear equations. It does not make any assumptions on linearization. It could be more accurate and converge in less iterations than the iterative linear method.\n",
+    "    * `CalculationMethod.newton_raphson`: traditional Newton-Raphson method."
```
mgovers, line 542: this is repetition of what you said before. please remove and refer to the earlier statements, or make the explanation such that there's no duplication anymore

docs/examples/State Estimation Example.ipynb
```diff
@@ -504,6 +528,87 @@
...
+    "print(\"Node result\")\n",
+    "display(pd.DataFrame(output_data_NR[\"node\"]))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a90ab8fb",
+   "metadata": {},
+   "source": [
+    "In **Newton Raphson method**, standard deviations of active and reactive power mesurement errors can also be provided seperately. We refer a detailed explanation about providing standard deviations for different methods to the [Documentation](../user_manual/calculations.md). Below is a example of a state estimation with seperate standard deviations of p and q measurement.  "
```
mgovers, line 571: maybe rephrase. this statement is true in `iterative_linear` as well, but the behaviour is different.

in particular, the degeneracy of the pair `sigma_p, sigma_q` breaks whenever there are 2 sensors with different ratios between `sigma_p` and `sigma_q` and they are aggregated into a single measurement (which happens only at the node level at most)

for `newton_raphson`, it doesn't matter whether they are on the same terminal or not and it is therefore more general

docs/examples/State Estimation Example.ipynb
```diff
@@ -36,25 +36,10 @@
...
+   "execution_count": null,
    "id": "ae11dc9a",
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "C:\\Users\\AL29374\\AppData\\Local\\Temp\\ipykernel_16184\\2477656666.py:3: DeprecationWarning: \n",
-      "Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),\n",
```
mgovers: something like this should do the trick

```py
import warnings

with warnings.catch_warnings(action="ignore", category=DeprecationWarning):
    # suppress warning about pyarrow as future required dependency
    import pandas
```
or otherwise
```py
import warnings

with warnings.catch_warnings():
    # suppress warning about pyarrow as future required dependency
    warnings.filterwarnings("ignore", category=DeprecationWarning)
    import pandas
```

mgovers (commented):
sorry i had completely forgotten about the fact that i would help you out with the deprecation warning

docs/examples/State Estimation Example.ipynb
```diff
@@ -36,14 +36,18 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 14,
```
mgovers, line 39: please restart kernel, clear output and rerun again to set the execution count back to `1`

docs/examples/Serialization Example.ipynb
```diff
@@ -18,13 +18,17 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 3,
```
mgovers, line 21: please restart kernel, clear output and rerun again to set the execution count back to `1`

docs/examples/Power Flow Example.ipynb
```diff
@@ -38,14 +38,18 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
```
mgovers, line 41: please restart kernel, clear output and rerun again to set the execution count back to `1`

docs/examples/Make Test Dataset.ipynb
```diff
@@ -153,15 +153,19 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 3,
```
mgovers, line 156: please restart kernel, clear output and rerun again to set the execution count back to `1`

mgovers (commented):
Everything looks good but i would still recommend resetting the execution count to `1` by restarting kernel, clear output and rerun

The main reason behind this is that it explicitly shows that there's no hidden dependencies and that the order of cells is correct.

As an example of where it could go wrong:
```python
# cell 0
a = np.array([])

# cell 1
import numpy as np
```

In that case, if you run the first cell, it will fail. If you then manually run the second cell, you import `numpy`. Then, if you re-run the first cell, it will actually run successfully, therefore making the example actually incorrect.

mgovers:
I will try to do the final changes myself. In addition, I found that a PR to the power-grid-model-workshop is needed to bump `power-grid-model[doc] >= 1.7`. I will do that as well

mgovers (approved):

mgovers:
> I will try to do the final changes myself.
Done

> In addition, I found that a PR to the power-grid-model-workshop is needed to bump `power-grid-model[doc] >= 1.7`. I will do that as well

done in https://github.com/PowerGridModel/power-grid-model-workshop/pull/24

mgovers:
FYI: i manually ran the workflow to sync the examples into the workshop in https://github.com/PowerGridModel/power-grid-model-workshop/actions/runs/8109082586

PR #507, opened 2024-02-28
extra trafo getters

petersalemink95:
Add extra getters to (3w) trafo that are needed for the automatic tap changer


mgovers (approved):
@Jerry-Jinfeng-Guo and I are considering the possibility of letting the tap position regulator dictate that the transformer should calculate the calculation parameters with a different tap position instead of actively modifying the component. this enables that as well

PR #512, opened 2024-03-04
Feature/tags as template args

mgovers:
replace bool template args with a named tag for comprehensibility

This is something that I've been looking into for a long time now. Dev work on #486 kept getting slower and slower because I was struggling to forward the `true` and `false` types. I ended up getting frustrating myself so much that I decided to go ahead with this change.

## Example for illustration

```cpp
template <bool sym>
struct A{
  RealValue<sym> a;
};

using SymA = A<true>;
using AsymA = A<false>;
```

becomes

```cpp
template <symmetry_tag sym>
struct A{
  RealValue<sym> a;
};

using SymA = A<symmetric_t>;
using AsymA = A<asymmetric_t>;
```

This makes types like `RealValue<symmetric_t>` and functions `get_output<asymmetric_t>` much more comprehensible than existing `RealValue<false>` and `get_output<false>`, resp.

## Example on datasets to really show why we need this

* `DataPointer` uses `is_const`
* but `DataHandler` uses `data_mutable`

this is confusing, because in one case `true` means `const` and in the other `true` means `mutable`

```cpp
// dataset.hpp
template <bool is_const> class DataPointer {
  // ...
};

// dataset_handler.hpp
template <bool data_mutable, bool indptr_mutable>
    requires(data_mutable || !indptr_mutable)
class DataHandler {
  // ...
};
```

## Introduced constructs

* For every changed construct, 2 tag types are created, one representing the old `true` value, and one representing the old `false` value
* A `tag` concept is created that matches on only those two tag types and nothing else. This is equivalent to the `bool` part of the template parameter declaration
* A `constexpr bool` is created to access the `true` resp `false` value of the tag types

```cpp
// tag types
struct symmetric_t{};
struct asymmetric_t{};

// tag concept
template <typename T> symmetry_tag = std::same_as<symmetric_t> || std::same_as<asymmetric_t>;

// tag value
template <symmetry_tag sym> constexpr bool is_symmetric = std::same_as<T, symmetric_t>;

// static asserts
static_assert(symmetry_tag<symmetric_t>);    // it is indeed a tag
static_assert(symmetry_tag<asymmetric_t>);   // it is indeed a tag

static_assert(is_symmetric<symmetric_t>);    // represents current 'true'
static_assert(!is_symmetric<asymmetric_t>);  // represents current 'false'
```

## Changed constructs

* `bool sym` for calculation symmetry and object symmetry
* `bool is_const` for the data set
* `bool data_mutable` for the data set handler
* `bool indptr_mutable` for the data set handler
* `bool is_gen` for load/gen


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset_handler.hpp
```diff
@@ -31,12 +33,15 @@ struct DatasetInfo {
     std::vector<ComponentInfo> component_info;
 };
 
-template <bool data_mutable, bool indptr_mutable>
-    requires(data_mutable || !indptr_mutable)
+template <data_mutable_tag data_mutable_type, indptr_mutable_tag indptr_mutable_type>
```
mgovers, line 36: please LMK if you have a better name for `indptr_mutable_type`

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -30,6 +30,16 @@ struct Idx2D {
...
+struct symmetric_t {};
+struct asymmetric_t {};
+
+template <typename T>
+concept symmetry_tag = std::derived_from<T, symmetric_t> || std::derived_from<T, asymmetric_t>;
+
+template <symmetry_tag T> constexpr bool is_symmetric = std::derived_from<T, symmetric_t>;
+
+template <symmetry_tag T> using other_symmetry_t = std::conditional_t<is_symmetric<T>, asymmetric_t, symmetric_t>;
+
```
mgovers, line 42: this is actually only used a couple of times and i don't particularly like it. If you prefer, I can move this outside `common`

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -1226,10 +1226,10 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         return sc_input;
     }
 
-    template <bool sym> void prepare_y_bus() {
+    template <symmetry_tag sym> void prepare_y_bus() {
         std::vector<YBus<sym>>& y_bus_vec = get_y_bus<sym>();
         // also get the vector of other Y_bus (sym -> asym, or asym -> sym)
-        std::vector<YBus<!sym>>& other_y_bus_vec = get_y_bus<!sym>();
+        std::vector<YBus<other_symmetry_t<sym>>>& other_y_bus_vec = get_y_bus<other_symmetry_t<sym>>();
```
mgovers, line 1232: we may want to move the declaration of `other_symmetry_t` here

tests/cpp_unit_tests/test_common.cpp
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+#include <power_grid_model/common/common.hpp>
+
+namespace power_grid_model {
+namespace {
+static_assert(symmetry_tag<symmetric_t>);
+static_assert(symmetry_tag<symmetric_t>);
+static_assert(is_symmetric<symmetric_t>);
+static_assert(!is_symmetric<asymmetric_t>);
```
mgovers, line 12: should i also add a test here for `other_symmetry_t`?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset_handler_fwd.hpp
```diff
@@ -0,0 +1,41 @@
...
+template <indptr_mutable_tag T> constexpr bool is_indptr_mutable = std::same_as<T, indptr_mutable_t>;
+
+static_assert(indptr_mutable_tag<indptr_mutable_t>);
+static_assert(indptr_mutable_tag<indptr_immutable_t>);
+static_assert(is_indptr_mutable<indptr_mutable_t>);
+static_assert(!is_indptr_mutable<indptr_immutable_t>);
+
+template <data_mutable_tag data_mutable, indptr_mutable_tag indptr_mutable>
+    requires(is_data_mutable<data_mutable> || !is_indptr_mutable<indptr_mutable>)
+class DatasetHandler;
```
mgovers, line 39: alternatively, I could create 3 tags:

```suggestion
struct const_dataset_t {};   // may be merged with the definition from dataset.hpp
struct mutable_dataset_t {}; // may be merged with the definition from dataset.hpp
struct writable_dataset_t{};

// to make the distinction between dataset handler tags and dataset tags
template <typename T>
concept dataset_handler_tag = std::same_as<T, const_dataset_t> || std::same_as<T, mutable_dataset_t> || std::same_as<T, writable_dataset_t>;

template <dataset_handler_tag T> constexpr bool is_data_mutable = std::same_as<T, mutable_dataset_t> || std::same_as<T, writable_dataset_t>;
template <dataset_handler_tag T> constexpr bool is_indptr_mutable = std::same_as<T, writable_dataset_t>;

static_assert(dataset_handler_tag<const_dataset_t>);
static_assert(dataset_handler_tag<mutable_dataset_t>);
static_assert(dataset_handler_tag<writable_dataset_t>);
static_assert(!is_data_mutable<const_dataset_t>);
static_assert(is_data_mutable<mutable_dataset_t>);
static_assert(is_data_mutable<writable_dataset_t>);
static_assert(!is_indptr_mutable<const_dataset_t>);
static_assert(!is_indptr_mutable<mutable_dataset_t>);
static_assert(is_indptr_mutable<writable_dataset_t>);

template <dataset_handler_tag dataset_handler_type>
class DatasetHandler;
```

The three types defined in `dataset_handler.hpp` would then become

```cpp
using ConstDatasetHandler = DatasetHandler<const_dataset_t>;
using MutableDatasetHandler = DatasetHandler<mutable_dataset_t>;
using WritableDatasetHandler = DatasetHandler<writable_dataset_t>;
```

power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
```diff
@@ -17,6 +17,18 @@
 
 namespace power_grid_model {
 
+struct load_appliance_t {};
+struct gen_appliance_t {};
```
mgovers, line 21: There is ambiguity here between (a)symmetric calculation type and (a)symmetric component type. Idem for power sensor

To work around that, we may actually want to create different tags for component type and calculation type symmetry. My proposal is to do that in `base.hpp`

Alternatively, we may actually want to create 4 tags here:

```suggestion
struct sym_load_appliance_t {};
struct asym_load_appliance_t {};
struct sym_gen_appliance_t {};
struct asym_gen_appliance_t {};
```

that would fix the ambiguity between (a)symmetric calculation type and (a)symmetric component type

power_grid_model_c/power_grid_model/include/power_grid_model/component/power_sensor.hpp
```diff
@@ -59,19 +59,19 @@ class GenericPowerSensor : public Sensor {
...
+    virtual PowerSensorCalcParam<asymmetric_t> asym_calc_param() const = 0;
 
-    virtual PowerSensorOutput<true> get_sym_output(ComplexValue<true> const& s) const = 0;
-    virtual PowerSensorOutput<false> get_asym_output(ComplexValue<false> const& s) const = 0;
+    virtual PowerSensorOutput<symmetric_t> get_sym_output(ComplexValue<symmetric_t> const& s) const = 0;
+    virtual PowerSensorOutput<asymmetric_t> get_asym_output(ComplexValue<asymmetric_t> const& s) const = 0;
 };
 
-template <bool sym> class PowerSensor : public GenericPowerSensor {
+template <symmetry_tag sym> class PowerSensor : public GenericPowerSensor {
```
mgovers, line 69: idem as for load_gen.hpp: we may want to add separate tags here. that would fix the ambiguity between (a)symmetric calculation type and (a)symmetric component type. I would propose to do that change in `base.hpp`

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -30,6 +30,16 @@ struct Idx2D {
     friend constexpr bool operator==(Idx2D x, Idx2D y) = default;
 };
 
+struct symmetric_t {};
+struct asymmetric_t {};
+
+template <typename T>
+concept symmetry_tag = std::derived_from<T, symmetric_t> || std::derived_from<T, asymmetric_t>;
+
+template <symmetry_tag T> constexpr bool is_symmetric = std::derived_from<T, symmetric_t>;
```
TonyXiang8787, line 39: Follow convension of `type_traits`, use `is_symmetric_v` for meta function of value.

TonyXiang8787: Define first a struct `is_symmetric` which derived from `std::true_type` or `std::false_type` depending on the template argument. Then use variable template.

```c++
template <symmetry_tag T>
struct is_symmetric;

template<>
struct is_symmetric<symmetric_t> : std::true_type {};
template<>
struct is_symmetric<asymmetric_t> : std::false_type {};

template <symmetry_tag T> constexpr bool is_symmetric_v = is_symmetric<T>::value;
```

tests/cpp_unit_tests/test_common.cpp
TonyXiang8787: Just do it.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
TonyXiang8787: That's okay common for such a small struct.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset_handler.hpp
TonyXiang8787: Please follow the conversion, a type ending with `_t`, a value ending with `_v`, a type as meta-function without any suffix.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset_handler_fwd.hpp
TonyXiang8787: I prefer the alternative way. Using three tags. The original implementation has two template arguments has to do that `bool` only has two states.

power_grid_model_c/power_grid_model/include/power_grid_model/component/load_gen.hpp
TonyXiang8787: The ambiguity should be solved by the name of template argument, not creating another tag type. E.g.

```c++
template<symmetry_tag loadgen_symmetry>
class LoadGen {

template<symmetry_tag calculation_symmetry>
LoadGenMathInput<calculation_symmetry> calc_params();
};

```

power_grid_model_c/power_grid_model/include/power_grid_model/component/power_sensor.hpp
TonyXiang8787: See the comment above. Do not create additional tag types, but use proper template argument name.

TonyXiang8787 (changes requested):
Please read the suggestions.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset_handler.hpp
mgovers: how should i work around the name conflict between:

* the concept
* the tag type itself
* the template argument
* the `using` declaration later on

proposal 1:

* the concept                       => suffixed with `_tag`
* the tag type itself               => suffixed with `_t`
* the template argument             => proposed: suffixed with `_`
* the `using` declaration later on  => proposed: without suffix

proposal 2:

* the concept                       => suffixed with `_c`
* the tag type itself               => suffixed with `_t`
* the template argument             => proposed: suffixed with `_tag`
* the `using` declaration later on  => proposed: without suffix

TonyXiang8787: Prefer option 1.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
mgovers: what is the reason you prefer type traits over concept syntax?

mgovers: ok, will use the following syntax then 😄 

```cpp
template <symmetry_tag T> constexpr bool is_symmetric_v = std::derived_from<T, symmetric_t>;
```

Jerry-Jinfeng-Guo (commented):
Tons of repeated changes, I can see why you were frustrated ;) One minor suggestion: why do we not already declare 
`
using SymComplexValue = ComplexValue<symmetric_t>;
using AsymComplexValue = ComplexValue<asymmetric_t>;
`
 throughout the code? This is just one instance.

PR #515, opened 2024-03-07
Fixes incorrect SonarCloud warning on template class deduction

Jerry-Jinfeng-Guo:
SonarCloud incorrectly generates warnings about a series of intended change of template. This branch fixes (supresses) that and removes the code smell.

ToDo:
The rest of the SonarCloud warnings on template class deduction will be addressed when a decision is made how to approach this.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/appliance.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/appliance.hpp
@@ -93,7 +93,7 @@ class Appliance : public Base {
         return output;
     }
     ApplianceShortCircuitOutput get_sc_output(ComplexValue<symmetric_t> const& i) const {
-        ComplexValue<asymmetric_t> const iabc{i};
+        ComplexValue<asymmetric_t> const iabc{i}; // NOSONAR
         return get_sc_output(iabc);
     }
     template <symmetry_tag sym> ApplianceOutput<sym> get_output(ComplexValue<sym> const& u) const {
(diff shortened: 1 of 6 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/component/appliance.hpp
```diff
@@ -93,7 +93,7 @@ class Appliance : public Base {
         return output;
     }
     ApplianceShortCircuitOutput get_sc_output(ComplexValue<symmetric_t> const& i) const {
-        ComplexValue<asymmetric_t> const iabc{i};
+        ComplexValue<asymmetric_t> const iabc{i}; // NOSONAR
```
mgovers, line 96: i don't particularly like this approach. Instead, please try either of the following (please discuss with the rest of the team which approach is preferred, I don't have any particular preference):

* `using AsymComplexValue = ComplexValue<asymmetric_t>;` so that here it can be `AsymComplexValue const iabc{i};`
* `auto const iabc{static_cast<ComplexValue<asymmetric_t>>(i)};`

The former has the advantage that everything is shorter and it requires less repetition, but refactoring may be every so slightly harder.

The latter explicitly shows that an explicit cast is being done here, but is more verbose

Jerry-Jinfeng-Guo: Please relax. This is a tentative commit that I will be using during the knowledge sharing today, not the final change. The goal is indeed to get a decision from the team today and finalize on the fix.

Jerry-Jinfeng-Guo: > i will convert this to `Draft` for that reason

Your input is always appreciated, but please do enjoy the training and trust that we will handle this well. ;)

mgovers: haha i am enjoying the training and i trust you but i'm waiting until we start today :joy: i was in the office early and whatcha gonna do other than just working

Jerry-Jinfeng-Guo:
This PR is closed after the discussion today (7-Mar). The decision made is to update corresponding warning checks on the SonarCloud, not on the code/project level.

PR #520, opened 2024-03-11
bump gcc in CI to 13 to test faster compilation

mgovers:
Relates to #486 

Reorganizing headers did not seem to help. Instead, there appear to have been some significant changes in gcc since `gcc-10` (C++ CI uses `gcc-11`, python build uses `gcc-12`).

This PR checks `gcc-13`, which is also likely the version that will be used on `ubuntu-24.04` that will come out next month.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
@@ -15,16 +15,6 @@
 
 namespace power_grid_model {
 
-// calculation info
-#ifdef POWER_GRID_MODEL_CPP_BENCHMARK
-#if defined(__cpp_lib_hardware_interference_size)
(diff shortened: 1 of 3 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -15,16 +15,6 @@
...
 
-// calculation info
-#ifdef POWER_GRID_MODEL_CPP_BENCHMARK
-#if defined(__cpp_lib_hardware_interference_size)
-constexpr size_t cache_line_size = std::hardware_destructive_interference_size;
-#else
-constexpr size_t cache_line_size = 64;
-#endif
-class alignas(cache_line_size) CalculationInfo : public std::map<std::string, double, std::less<>> {};
-#else
```
TonyXiang8787, line 26: @mgovers also remove the `<new>` include above. That's not needed.

TonyXiang8787 (approved):

PR #514, opened 2024-03-06
Enhanced documentation on batch datasets 

zhen0427:
In the [Dataset Terminology](https://power-grid-model.readthedocs.io/en/stable/user_manual/dataset-terminology.html) in the documentation, only a very brief mention is made on what a batch dataset is. There's even less documentation on sparse datasets - not even in the advanced documentation.

This PR is related with issue #434

PR #521, opened 2024-03-12
Quick patch on remaining issues on the docs.

Jerry-Jinfeng-Guo:
ToDo:

The hierarchy of batch datatypes is rather arbitrary and non-consistent, should we consider change that in the future?

Relates to https://github.com/PowerGridModel/power-grid-model/pull/518 and https://github.com/PowerGridModel/power-grid-model/pull/514

Preview in https://power-grid-model--521.org.readthedocs.build/en/521/api_reference/python-api-reference.html#power_grid_model.data_types.Dataset

### DIFF
--- a/docs/api_reference/python-api-reference.md
+++ b/docs/api_reference/python-api-reference.md
@@ -40,21 +40,14 @@ SPDX-License-Identifier: MPL-2.0
 ## data types
 
 ```{eval-rst}
-.. autoclass:: power_grid_model.data_types.SparseBatchArray
-.. autoclass:: power_grid_model.data_types.BatchArray
+.. autoclass:: power_grid_model.data_types.Dataset
 .. autoclass:: power_grid_model.data_types.SingleDataset
 .. autoclass:: power_grid_model.data_types.BatchDataset
-.. autoclass:: power_grid_model.data_types.Dataset
-.. autoclass:: power_grid_model.data_types.BatchList
-.. autoclass:: power_grid_model.data_types.NominalValue
-.. autoclass:: power_grid_model.data_types.RealValue
-.. autoclass:: power_grid_model.data_types.AsymValue
-.. autoclass:: power_grid_model.data_types.AttributeValue
-.. autoclass:: power_grid_model.data_types.Component
-.. autoclass:: power_grid_model.data_types.ComponentList
-.. autoclass:: power_grid_model.data_types.SinglePythonDataset
-.. autoclass:: power_grid_model.data_types.BatchPythonDataset
-.. autoclass:: power_grid_model.data_types.PythonDataset
+.. autoclass:: power_grid_model.data_types.DataArray
+.. autoclass:: power_grid_model.data_types.SingleArray
+.. autoclass:: power_grid_model.data_types.BatchArray
+.. autoclass:: power_grid_model.data_types.SparseBatchArray
+.. autoclass:: power_grid_model.data_types.DenseBatchArray
 ```
 
 ## utils
--- a/docs/api_reference/python-api-reference.md
+++ b/docs/api_reference/python-api-reference.md
@@ -31,6 +31,12 @@ SPDX-License-Identifier: MPL-2.0
 .. autofunction:: power_grid_model.validation.errors_to_string
 ```
 
+### errors
+
+```{eval-rst}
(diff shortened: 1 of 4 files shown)


docs/api_reference/python-api-reference.md
```diff
@@ -31,30 +31,23 @@ SPDX-License-Identifier: MPL-2.0
...
-.. autoclass:: power_grid_model.data_types.BatchList
-.. autoclass:: power_grid_model.data_types.NominalValue
-.. autoclass:: power_grid_model.data_types.RealValue
-.. autoclass:: power_grid_model.data_types.AsymValue
-.. autoclass:: power_grid_model.data_types.AttributeValue
-.. autoclass:: power_grid_model.data_types.Component
-.. autoclass:: power_grid_model.data_types.ComponentList
-.. autoclass:: power_grid_model.data_types.SinglePythonDataset
-.. autoclass:: power_grid_model.data_types.BatchPythonDataset
-.. autoclass:: power_grid_model.data_types.PythonDataset
```
mgovers, line 51: https://github.com/PowerGridModel/power-grid-model/pull/521

docs/api_reference/python-api-reference.md
```diff
@@ -31,30 +31,23 @@ SPDX-License-Identifier: MPL-2.0
...
+.. autoclass:: power_grid_model.validation.errors.ValidationError
```
mgovers, line 37: this should be part of the validation error logic, not datatypes

docs/user_manual/dataset-terminology.md
```diff
@@ -12,9 +12,12 @@ Some terms regarding the data structures are explained here, including the defin
...
-        - **SparseBatchDataset:**  Dictionaries with a one-dimensional numpy int32 array and a one-dimensional structured numpy arrays. 
-        - **DenseBatchDataset:** A dictionary where the keys are the component types and the values are two-dimensional structured numpy arrays.
+    - **BatchDataset:** A data type storing update and or output data for one or more scenarios. A batch dataset can contain sparse or dense data, depending on the component.
+- **DataArray** A data array can be a single or a batch array. It is a numpy structured array.     
+    - **SingleArray** A dictionary where the keys are the component types and the values are one-dimensional structured numpy arrays.
+    - **BatchArray:** An array of dictionaries where the keys are the component types and the values are two-dimensional structured numpy arrays.
+        - **DenseBatchArray:** A two-dimensional structured numpy array containing a list of components of the same type for each scenario.
+        - **SparseBatchArray:** A dictionary with a one-dimensional numpy int64 array and a one-dimensional structured numpy arrays.
```
mgovers, line 20: should i cross-reference with the python API reference here?

mgovers (commented):
if we want to move the contents of this file to `core` and then import the ones we need, we should use `__all__` as mentioned in https://stackoverflow.com/questions/50585305/include-specific-imported-functions-in-sphinx-documentation

src/power_grid_model/data_types.py
```diff
@@ -13,60 +13,84 @@
...
+"""
 
-sparse: {"indptr": <1d-array>, "data": <1d-array>}
+DataArray = Union[SingleArray, BatchArray]
+"""
+A data array can be a :class:`SingleArray` or a :class:`BatchArray`.
 """
 
-SingleDataset = Dict[str, np.ndarray]
+SingleDataset = Union[SingleArray]
```
nitbharambe, line 63: Is this equivalent to Dict[str, ndarray]? This data type is being used externally too.

mgovers: whoops, was a booboo

src/power_grid_model/data_types.py
```diff
@@ -13,60 +13,84 @@
 
 # When we're dropping python 3.8, we should introduce proper NumPy type hinting
 
-SparseBatchArray = Dict[str, np.ndarray]
+SingleArray = Union[np.ndarray]
```
nitbharambe, line 16: Is there a specific reason to use Union on a single type, Union[np.ndarray]? Does it make a difference?

mgovers: yes the problem is that np.ndarray not subclassable is. there's a way around that using `np.NDArray` but it does not completely solve our problem. Otherwise, we would need python3.10

nitbharambe (approved):

PR #543, opened 2024-03-20
Feature/auto tap changer skeleton

mgovers:
This creates the skeleton for #74 

NOTE: This is still in draft. I still need to create stub functions for the automatic tap optimizer

NOTE: Unfortunately, I had to make a few shortcuts to make sure I can unblock other people:

* For powerflow and state estimation calculations, it was easy to make `prepare_*_input` `static`. For short circuit calculations, that is where the topology is constructed for fault components, which means a little bit more refactoring is necessary to move that part out of `MainModel` as well. I considered that out of scope of this PR and added a `TODO` for follow-up due to the blocking nature of this PR.
* `MainModel` is the object keeping track of the changes in state, which means I had to provide an update function to the optimizer.
  * The alternative would be to introduce a signaling-like functionality (in which main model registers to the optimizer a callback for state changes). With the current state of the main model (many responsibilities) that would cause future refactorings to be come more difficult.
  * Instead, I consider introducing signaling logic to be part of another future refactoring in which a state object owns the update logic, and in which `MainModel` registers an `update_state(UpdateChange const&)` callback. That is out of scope of this PR

Other design considerations:

* I wanted to use a factory pattern to enable us to refactor at a later stage to at least keep the amount of `calculate_*` functions somewhat maintainable. If we would solve everything statically (like with `MathSolver`) the amount of overloads for `calculate_*` would grow indefinitely, or we would not be able to refactor at a later stage.
* I had to work with inheritance because otherwise, the factory model would not work.
* I had to use `shared_ptr` because otherwise, ownership would be lost unless I would implement rule-of-five. `shared_ptr` is actually able to use rule-of-zero without much overhead.


mgovers:
newly introduced sonar cloud issues are TODO's for follow-up tickets in this sprint

power_grid_model_c/power_grid_model/include/power_grid_model/common/enum.hpp
```diff
@@ -98,4 +98,16 @@ enum class CType : IntS { c_int32 = 0, c_int8 = 1, c_double = 2, c_double3 = 3 }
 
 enum class SerializationFormat : IntS { json = 0, msgpack = 1 };
 
+enum class OptimizerType : IntS {
+    noop = 0,                     // do nothing
```
mgovers, line 102:
```suggestion
    no_operation = 0,                     // do nothing
```

mgovers:
> @TonyXiang8787 can you also review please?

In a call, @TonyXiang8787 has approved the overall design choices. This PR can be merged after someone else does a regular review

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/base_optimizer.hpp
```diff
@@ -0,0 +1,87 @@
...
+    using typename Base::ResultType;
+    using typename Base::State;
+
+    NoopOptimizer(Calculator func) : func_{std::move(func)} {}
+
+    auto optimize(State const& state) -> ResultType final { return func_(state); }
+
+  private:
+    Calculator func_;
+};
```
Jerry-Jinfeng-Guo, line 85: `NoOpOptimizer`?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/base_optimizer.hpp
```diff
@@ -0,0 +1,87 @@
...
+    using ResultType = state_calculator_result_t<Calculator, State>;
+
+    BaseOptimizer() = default;
+    BaseOptimizer(BaseOptimizer const&) = delete;
+    BaseOptimizer(BaseOptimizer&&) noexcept = default;
+    BaseOptimizer& operator=(BaseOptimizer const&) = delete;
+    BaseOptimizer& operator=(BaseOptimizer&&) noexcept = default;
+    virtual ~BaseOptimizer() = default;
+
+    virtual auto optimize(State const& state) -> ResultType = 0;
```
Jerry-Jinfeng-Guo, line 51: How would this interface facilitate the fact that it modifies states in the middle, i.e., `math_model_state` ?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -0,0 +1,86 @@
...
+namespace power_grid_model::optimizer {
+namespace tap_position_optimizer {
+
+namespace detail = power_grid_model::optimizer::detail;
+
+using GraphIdx = size_t;
+struct Edge {
+    Idx2D pos{};
+    Idx weight{};
+};
```
Jerry-Jinfeng-Guo, line 23: `TrafoGraphIdx` and `TrafoGraphEdge`? Or if a very general index and `edge` concept in graph sense, it should reside somewhere in `core`.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -0,0 +1,86 @@
...
+using TransformerGraph = boost::compressed_sparse_row_graph<boost::directedS, boost::no_property, Edge,
+                                                            boost::no_property, GraphIdx, GraphIdx>;
+
+template <main_core::main_model_state_c State>
+inline auto build_transformer_graph(State const& /*state*/) -> TransformerGraph {
+    // TODO(mgovers): implement
+    return {};
+}
+
+inline auto determine_distance_from_source(TransformerGraph const& /*graph*/) -> std::vector<std::pair<Idx2D, Idx>> {
```
Jerry-Jinfeng-Guo, line 35:
```suggestion
inline auto get_distances_from_source(TransformerGraph const& /*graph*/) -> std::vector<std::pair<Idx2D, EdgeWeight >> {
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -0,0 +1,86 @@
...
+        auto const order = rank_transformers(state);
+        return optimize(state, order);
+    }
+
+    constexpr auto strategy() { return strategy_; }
+
+  private:
+    auto optimize(State const& /*state*/, std::vector<Idx2D> const& /*order*/) -> ResultType {
+        // TODO(mgovers): rank Idx2D of transformers as listed in the container
+        throw PowerGridError{};
```
Jerry-Jinfeng-Guo, line 73: Would we want to include optimizer specific error types here in the next stages?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -0,0 +1,86 @@
...
+    using typename Base::State;
+    using StateUpdater = StateUpdater_;
+
+    TapPositionOptimizer(Calculator calculate, StateUpdater update, OptimizerStrategy strategy)
+        : calculate_{std::move(calculate)}, update_{std::move(update)}, strategy_{strategy} {}
+
+    auto optimize(State const& state) -> ResultType final {
+        auto const order = rank_transformers(state);
+        return optimize(state, order);
+    }
```
Jerry-Jinfeng-Guo, line 66: I think we might need more tolerating constructors and `set_*` later?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    constexpr auto strategy() { return strategy_; }
```
Jerry-Jinfeng-Guo, line 68: Maybe a traditional `getter` and `setter` combo? Unless we eliminate the possibility of alternating optimization goals and opt for const exp able `adjust_tap` methods

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -0,0 +1,86 @@
...
+             std::invocable<std::remove_cvref_t<StateUpdater_>, ConstDataset const&>
+class TapPositionOptimizer : public detail::BaseOptimizer<StateCalculator, State_> {
+  public:
+    using Base = detail::BaseOptimizer<StateCalculator, State_>;
+    using typename Base::Calculator;
+    using typename Base::ResultType;
+    using typename Base::State;
+    using StateUpdater = StateUpdater_;
+
+    TapPositionOptimizer(Calculator calculate, StateUpdater update, OptimizerStrategy strategy)
```
Jerry-Jinfeng-Guo, line 60: Perhaps more semantically inline:
```suggestion
    TapPositionOptimizer(Calculator calculator, StateUpdater updater, OptimizerStrategy strategy)
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace power_grid_model::optimizer {
+namespace tap_position_optimizer {
+
+namespace detail = power_grid_model::optimizer::detail;
+
+using GraphIdx = size_t;
+struct Edge {
+    Idx2D pos{};
+    Idx weight{};
```
Jerry-Jinfeng-Guo, line 22: Since we are doing this already in the project:
```
using EdgeWeight = int64_t;
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: The more I read the OneNote, the more I feel that it's not `distances`, but `edge_weights`?

power_grid_model_c/power_grid_model/include/power_grid_model/common/enum.hpp
```diff
@@ -98,4 +98,16 @@ enum class CType : IntS { c_int32 = 0, c_int8 = 1, c_double = 2, c_double3 = 3 }
...
+    no_optimization = 0,          // do nothing
+    automatic_tap_adjustment = 1, // power flow with automatic tap adjustment
+};
+
+enum class OptimizerStrategy : IntS {
+    any = 0,
+    lowest = 1,
+    highest = 2,
+    nan = na_IntS,
+};
```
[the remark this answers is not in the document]
mgovers: the problem here is that this will become part of the public API. IMO, this is similar to how `newton_raphson` calclation method can be used for both power flow and state estimation, but not for short circuit (as of yet). They are indeed specific, but not specific enough to restrict to only this optimizer type

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/base_optimizer.hpp
mgovers: that goes via the `update` function that is provided

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: you can do that in your ticket. i deem that out of scope of this PR

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: yes if applicable. if there are no errors that can be thrown, then no 😬

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: in what sense?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: that would only apply for the tests as of now. every scenario has its own optimizer. lots of refactoring needed to change that (this PR prepares for that)

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: out of scope of this PR. that's IMO the responsibility of the follow-up tickets

power_grid_model_c/power_grid_model/include/power_grid_model/common/enum.hpp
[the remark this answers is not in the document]
mgovers: `minimum`, `maximum`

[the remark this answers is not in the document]
mgovers: any -> `any_optimum`

Jerry-Jinfeng-Guo (approved):

Issue #509, opened 2024-02-29
[FEATURE] Expose derived power grid error types


mgovers:
## Describe the feature request

Expose all different error types thrown by the calculation core as separate error classes that derive from `PowerGridError`.

## Rationale
Currently, all power grid errors are derived from `PowerGridError`. The only difference between e.g. `SparseMatrixError` and `IterationDivergeError` is the string message of the error. As a user, parsing strings (that may change in the future) is suboptimal compared to actual different errors derived from the `PowerGridError`.

## Design

* [ ] Python side
  * [ ] Create separate error classes, e.g.
        ```py
        class SparseMatrixError(PowerGridError):
            """Sparse matrix error class."""
            pass
        ```
  * [ ] Obtain the correct error
    * [ ] Short term solution: parse from the error message string
    * [ ] Sustainable solution: raise the correct error from the C API (see below)
* [ ] C API
  * [ ] Define other error codes for all error types defined in `exceptions.hpp`
  * [ ] Update the `call_with_catch` to produce the correct error code based on the raised exception.
    * [ ] Use a Lippincott pattern (see e.g. https://github.com/PowerGridModel/power-grid-model/blob/037a0e43cef1e36f49df73f177a601967ecf553b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp#L574 on usage)

PR #546, opened 2024-03-26
Feature/separate python error types

mgovers:
Part 1 of #509 : public interface

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
@@ -57,7 +57,7 @@ class InvalidBranch3 : public PowerGridError {
 class InvalidTransformerClock : public PowerGridError {
   public:
     InvalidTransformerClock(ID id, IntS clock) {
-        append_msg("Invalid clock for transformer " + std::to_string(id) + ", clock  " + std::to_string(clock) + '\n');
+        append_msg("Invalid clock for transformer " + std::to_string(id) + ", clock " + std::to_string(clock) + '\n');
     }
 };
 
@@ -140,7 +140,7 @@ class InvalidCalculationMethod : public CalculationError {
 class UnknownAttributeName : public PowerGridError {
   public:
     explicit UnknownAttributeName(std::string const& attr_name) {
-        append_msg("Unknown attribute name!" + attr_name + "\n");
+        append_msg("Unknown attribute name! " + attr_name + "\n");
     }
 };
 
@@ -168,7 +168,7 @@ class InvalidShortCircuitPhases : public PowerGridError {
 class InvalidShortCircuitPhaseOrType : public PowerGridError {
   public:
     InvalidShortCircuitPhaseOrType() {
-        append_msg("During one calculation the short circuit types phases should be similar for all faults \n");
+        append_msg("During one calculation the short circuit types phases should be similar for all faults\n");
     }
 };
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
@@ -137,13 +137,6 @@ class InvalidCalculationMethod : public CalculationError {
     InvalidCalculationMethod() : CalculationError("The calculation method is invalid for this calculation!") {}
 };
 
-class UnknownAttributeName : public PowerGridError {
-  public:
-    explicit UnknownAttributeName(std::string const& attr_name) {
-        append_msg("Unknown attribute name! " + attr_name + "\n");
-    }
-};
-
 class InvalidShortCircuitType : public PowerGridError {
   public:
     explicit InvalidShortCircuitType(FaultType short_circuit_type) {
--- a/src/power_grid_model/core/power_grid_model.py
+++ b/src/power_grid_model/core/power_grid_model.py
@@ -200,8 +200,10 @@ class PowerGridModel:
             setattr(opt, key, value.value if isinstance(value, IntEnum) else value)
         return opt
 
-    def _handle_errors(self, continue_on_batch_error: bool, batch_size: int):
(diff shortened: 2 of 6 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -137,13 +137,6 @@ class InvalidCalculationMethod : public CalculationError {
     InvalidCalculationMethod() : CalculationError("The calculation method is invalid for this calculation!") {}
 };
 
-class UnknownAttributeName : public PowerGridError {
```
mgovers, line 140: this error was never thrown (most likely outdated since switching to new metadata structure)

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -57,7 +57,7 @@ class InvalidBranch3 : public PowerGridError {
 class InvalidTransformerClock : public PowerGridError {
   public:
     InvalidTransformerClock(ID id, IntS clock) {
-        append_msg("Invalid clock for transformer " + std::to_string(id) + ", clock  " + std::to_string(clock) + '\n');
+        append_msg("Invalid clock for transformer " + std::to_string(id) + ", clock " + std::to_string(clock) + '\n');
```
mgovers, line 60: could this be breaking for our users?

src/power_grid_model/core/power_grid_model.py
```diff
@@ -334,6 +344,7 @@ def _calculate_short_circuit(
         threading: int = -1,
         output_component_types: Optional[Union[Set[str], List[str]]] = None,
         continue_on_batch_error: bool = False,
+        decode_error: bool = True,
         short_circuit_voltage_scaling: Union[ShortCircuitVoltageScaling, str] = ShortCircuitVoltageScaling.maximum,
```
mgovers, line 348: changing the order here is OK because of the `*` in line `341`. all arguments are kwargs

tests/unit/test_error_handling.py
```diff
@@ -23,5 +36,165 @@ def test_empty_model():
...
+    line_input["from_node"] = [0]
+    line_input["to_node"] = [1]
+
+    with pytest.raises(ConflictVoltage):
+        PowerGridModel(input_data={"node": node_input, "line": line_input})
+
+
+@pytest.mark.skip(reason="TODO")
+def test_handle_missing_case_for_enum_error():
+    pass
```
mgovers, line 59: it is not easy to find an example that will work here - now and in the future. what should we do with these tests?

tests/unit/test_error_handling.py
```diff
@@ -23,5 +36,165 @@ def test_empty_model():
...
+    transformer_input["to_node"] = [1]
+    transformer_input["clock"] = [-1]
+
+    with pytest.raises(InvalidTransformerClock):
+        PowerGridModel(input_data={"node": node_input, "transformer": transformer_input})
+
+
+@pytest.mark.skip(reason="TODO")
+def test_handle_sparse_matrix_error():
+    pass
```
mgovers, line 108: it is not easy to find an example that will work here - now and in the future. what should we do with these tests?

tests/unit/test_error_handling.py
```diff
@@ -23,5 +36,165 @@ def test_empty_model():
...
+    sym_load_input["type"] = [LoadGenType.const_power]
+
+    model = PowerGridModel(input_data={"node": node_input, "source": source_input, "sym_load": sym_load_input})
+    with pytest.raises(NotObservableError):
+        model.calculate_state_estimation(calculation_method=CalculationMethod.iterative_linear)
+
+
+@pytest.mark.skip(reason="TODO")
+def test_handle_iteration_diverge_error():
+    pass
```
mgovers, line 134: it is not easy to find an example that will work here - now and in the future. what should we do with these tests?

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
TonyXiang8787: We do not guarantee the error messages. So any change in error message as far as I concern no breaking changes.

mgovers: ok great. in that case, do you think it's a good idea to have a look at the other error messages as well? e.g. we use `\n ` with the whitespace after the newline for all errors. Is that intentional?

tests/unit/test_error_handling.py
Jerry-Jinfeng-Guo: I'd say "Create corresponding issue/tickets and put it on higher priority than before"

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: let's revisit later after we've given our users enough time to migrate to the new errors

tests/unit/test_error_handling.py
mgovers: The thing is that i don't know whether we can actually find these cases.. best we can do to test the regex is hack such that it is raised and then mock the PGM response i guess? Or do you have a better idea?

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
Jerry-Jinfeng-Guo: Would this line be further processed in this PR?

mgovers: Nope, removed entirely. My comment was just an FYI

Jerry-Jinfeng-Guo (approved):
Hereby approve. Let's leave the skipped tests up to the team to decide whether we remove them all together or spend some effort crafting a couple cases.

PR #550, opened 2024-03-26
Preparing steps 2-3 so that step 1 can start

Jerry-Jinfeng-Guo:
In this PR, the tentative implementation for steps 2-3 of the transformers ranking is cleaned up. The code part related to the Dijkstra algorithm could use some further optimization but it will be for next steps. The goal is to merge steps 2-3 a.s.a.p.


Jerry-Jinfeng-Guo:
Please do not delete this `step-4` branch. It will have its-own / another PR

mgovers:
It's usually better practice to branch from this branch, but fine either way

PR #479, opened 2024-01-26
Test macOS arm64 Python wheel using Fly CI

TonyXiang8787:
Closes #476 

# Approach

- Only test if we want to publish wheel, before the publish step. 
- We download the built `arm64` wheel, install it with dependencies and run tests.

PR #553, opened 2024-03-27
also run macos arm64 in merge group

mgovers:
Relates to #479 .

We cannot enable MacOS arm64 testing on PRs, because PRs have multiple (even up til quite a lot of) reruns/commits/... with a successful CI run. That would eventually result in our total amount of free minutes being used up for FlyCI.

Since merge queue CI runs typically only run once for each PR, we may "abuse" it to do this additional check before final release to `main` without prohibitively increasing the risk of going over the limit for free minutes on FlyCI runners.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -263,7 +263,7 @@ jobs:
   test-python-macos-arm64:
     runs-on: flyci-macos-large-latest-m1
     needs: [build-cpp-test-macos, build-and-test-python]
-    if: (github.event_name == 'push') || (github.event_name == 'workflow_dispatch')
+    if: (github.event_name == 'push') || (github.event_name == 'workflow_dispatch') || (github.event_name == 'merge_group')
 
     steps:
       - uses: actions/checkout@v4


nitbharambe (approved):
Nice idea

Jerry-Jinfeng-Guo:
What is the difference of doing it like this comparing to the previous way? Does it replace the previous way?

mgovers:
> What is the difference of doing it like this comparing to the previous way? Does it replace the previous way?

it's an additional check. before, the test would only happen on `main` after the rest of CI and before releasing. `main` would be broken if the test would fail.

Now, it's checked on the merge queue in addition to `main`, which means that we can catch issues earlier in the process

Jerry-Jinfeng-Guo:
> > What is the difference of doing it like this comparing to the previous way? Does it replace the previous way?
> 
> it's an additional check. before, the test would only happen on `main` after the rest of CI and before releasing. `main` would be broken if the test would fail.
> 
> Now, it's checked on the merge queue in addition to `main`, which means that we can catch issues earlier in the process

I am very much not sure if that is a good idea though... it we are far from consuming the free runs, sure; but otherwise, we are most of the time doing the same check twice. Perhaps wait until Tony's input to merge?

mgovers:
> > > What is the difference of doing it like this comparing to the previous way? Does it replace the previous way?
> > 
> > 
> > it's an additional check. before, the test would only happen on `main` after the rest of CI and before releasing. `main` would be broken if the test would fail.
> > Now, it's checked on the merge queue in addition to `main`, which means that we can catch issues earlier in the process
> 
> I am very much not sure if that is a good idea though... if we are far from consuming the free runs, sure; but otherwise, we are most of the time doing the same check twice. Perhaps wait until Tony's input to merge?

we indeed are far from consuming the free runs. @TonyXiang8787 did an investigation before we introduced this CI job in which he concluded that we are barely in the danger zone, even if we would enable it for PRs. A single run of the job costs about 45s, while we have >500min available per month for free (taken from https://www.flyci.net/pricing#features ).

That means that we would have to have like 600 CI runs, or 300 PR merges per month, before we go over our limits. I feel like we're safe for the time being

PR #557, opened 2024-03-28
Tap position optimizer outer loop

mgovers:
relates to #74 

Depends on #578 (cherry-picked from this branch)
Depends on #577 (blocking fix)


mgovers:
Rebasing to https://github.com/PowerGridModel/power-grid-model/pull/565 because those changes were cherry-picked from this branch but needed in this branch

mgovers:
CFR. discussed during standup: current state is good enough for first iteration

* This PR contains tests to verify working of different strategies with equivalent output state for a single transformer
* Out of scope tests for:
  * multiple transformers
    * same ranking
    * different ranking
  * different voltage levels
  * line drop compensation

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +366,550 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+                                         CalculationMethod method) const -> ResultType {
+        auto result = calculate_with_retry(state, method);
+
+        bool tap_changed = true;
+        while (tap_changed) {
+            tap_changed = false;
+            UpdateBuffer update_data;
+
+            for (auto const& same_rank_regulators : regulator_order) {
+                for (auto const& regulator : same_rank_regulators) {
```
nitbharambe, line 688: Status check here would be redundant i suppose. The regulator_order would contain only fully connected transformers.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +366,550 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+
+            auto const v_measured = mean_val(cabs(u_measured)); // TODO(mgovers): handle asym correctly
+
+            auto new_tap_pos = [&transformer, &v_measured, &param] {
+                if (v_measured > param.u_set + 0.5 * param.u_band) {
+                    return tap_one_step_up(transformer);
+                }
+                if (v_measured < param.u_set - 0.5 * param.u_band) {
+                    return tap_one_step_down(transformer);
+                }
```
Jerry-Jinfeng-Guo, line 740: This relates to the comment above. And, since the default condition considered in speak is that `tap_max > tap_min`, shouldn't it be:
```suggestion
                if (v_measured > param.u_set + 0.5 * param.u_band) {
                    return tap_one_step_down(transformer);
                }
                if (v_measured < param.u_set - 0.5 * param.u_band) {
                    return tap_one_step_up(transformer);
                }
```

Jerry-Jinfeng-Guo: also, let's make a `Idx2D .u_set_range` already ;)

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +366,550 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+        initialize(regulator_order);
+
+        if (auto result = try_calculation_with_regulation(state, regulator_order, method);
+            strategy_ == OptimizerStrategy::any) {
+            return result;
+        }
+
+        // refine solution
+        step_all(regulator_order);
+        return try_calculation_with_regulation(state, regulator_order, method);
```
Jerry-Jinfeng-Guo, line 674: How could this work? 
(Let's first not discuss the fact that `::any` strategy is isolated and that if the `try_calc_` call fails to return anything meaningful, the subsequent runs should also be skipped because something is not right.)

Shouldn't there be an outer loop over like the following sketch?

```suggestion
        res{};
        while (...) {
                step_all(regulator_order); // the ::any check should happen here?
                res = try_calculation_with_regulation(state, regulator_order, method);
        }
        return res;
```

Jerry-Jinfeng-Guo (commented):
Comments over the design and detail

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: indeed. if we do want to add it in this class, it should be where the regulators + transformer wrappers are appended but not here (that's why it's a class and not a struct: not all possible input combinations are valid and it is assumed that all invalid combinations have already been resolved before the calculation enters this function. This is usually called a class invariant)

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -8,26 +8,120 @@
...
+    requires main_core::component_container_c<typename State::ComponentContainer, ComponentType>
+constexpr auto get_math_id(State const& state, Idx topology_sequence_idx) {
+    return state.topo_comp_coup->branch3[topology_sequence_idx];
+}
+
+template <std::derived_from<Regulator> ComponentType, typename State>
+    requires main_core::component_container_c<typename State::ComponentContainer, ComponentType>
+constexpr auto get_math_id(State const& state, Idx topology_sequence_idx) {
+    return state.topo_comp_coup->regulator[topology_sequence_idx];
+}
```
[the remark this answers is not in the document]
mgovers: should be in `main_core` because it involves statefullness, but i 100% agree with your comment. initially kept this here to localize the changes compared to other things that are invasive to main model.

how do we want to name this? also part of `main_core/state_queries.hpp`?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +366,550 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+    auto const tap_pos = transformer.tap_pos();
+    auto const tap_max = transformer.tap_max();
+    return tap_pos < tap_max ? tap_pos + 1 : tap_max;
+}
+
+constexpr IntS tap_one_step_down(transformer_c auto const& transformer) {
+    auto const tap_pos = transformer.tap_pos();
+    auto const tap_min = transformer.tap_min();
+    return tap_pos > tap_min ? tap_pos - 1 : tap_min;
+}
```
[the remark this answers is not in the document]
mgovers: good question. i did want to add this to the tests (hence here the checkpoint) but tbh i don't know. there are 2 possible resolutions with opposite effect:

* [ ] `local_maximum` strategy results in the `tap_pos` that is closest to `127`.
  * this can be ensured by setting `tap_max = std::max(t.tap_min(), t.tap_max())` (idem for `tap_min`)
  * in that way, `tap_one_step_up` always amounts to `tap_pos = t.tap_pos() + 1` but clamped to the valid range
* [ ] `local_maximum` strategy returns the closest value to `t.tap_max()` (and vice versa for `local_minimum` strategy)
  * this can be achieved by setting `tap_pos -> t.tap_pos() + sgn(t.tap_max() - t.tap_min())`
  * in that way, the behaviour of `tap_one_step_up` depends on whether `t.tap_max() > t.tap_min()`

This requires expert input. @TonyXiang8787 do you know which of the two is desired?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +366,550 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+template <typename... Ts> using transformer_types_t = typename transformer_types_s<std::tuple<Ts...>>::type;
+
+template <typename State>
+    requires main_core::component_container_c<typename State::ComponentContainer, TransformerTapRegulator>
+TransformerTapRegulator const& find_regulator(State const& state, ID regulated_object) {
+    auto const regulators = get_component_citer<TransformerTapRegulator>(state);
+
+    auto result_it = std::ranges::find_if(regulators, [regulated_object](auto const& regulator) {
+        return regulator.regulated_object() == regulated_object;
+    });
```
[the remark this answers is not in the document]
mgovers: your comment does not explicitly provide a mapping from one to the other but if i assume correctly, then yes: there can be zero or one regulator of the same type mapped to a regulated object. otherwise, there will be conflicting inputs and you might end up in an infinite loop

NOTE: the fact that there cannot be multiple regulators mapped to the same transformer should be checked by `main_model` (i will add that to the Jira ticket). ~in the optimizer, my guess is that it can be assumed to be undefined behavior~ EDIT: due to the severity of inifinite loops, we cannot make it UB and we indeed need to check here.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +366,550 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+        if (!update_dataset.empty()) {
+            update_(update_dataset);
+        }
+    }
+
+    auto initialize(std::vector<std::vector<RegulatedTransformer>> const& regulator_order) const {
+        using namespace std::string_literals;
+
+        constexpr auto get_max = [](transformer_c auto const& transformer) -> IntS { return transformer.tap_max(); };
+        constexpr auto get_min = [](transformer_c auto const& transformer) -> IntS { return transformer.tap_min(); };
```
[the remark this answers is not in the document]
mgovers: TBD; needs expert knowledge

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: i don't really understand what you mean.

the `try_calculation_with_regulation` contains the actual loop to saturate in a valid set of tap positions. However, that solution might not be optimal yet. To achieve an optimal solution, one more run is necessary where you re-try with one tap position lower for all transformers and re-run the optimization.

This logic was written by Tony. I'm fine with changing it, but in that case, also the outer loop document should be reviewed and updated accordingly.

Jerry-Jinfeng-Guo: Yes, you're correct that I missed one line last Friday (reading code on github is not 100% optimal for me ;) ). It is actually the function name that misled me into understanding the `try_calculation_with_regulation` function call being a unit calculation call. Whereas in reality, it is the optimization iteration core. In that sense, I correct my second part of the comment to suggesting a function renaming.

The first part, however, is still up for discussion in my view.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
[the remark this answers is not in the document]
mgovers: i changed my mind: this is 100% the transformer ranker's responsibility

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: 1. shouldn't be an `Idx2D` (cause it's not an index and not even discrete)
2. we could make it a struct but this directly involves the calculation params and that already is basically that same struct 😬 

what is it you're looking fore here?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: about your first part:

* `any` is isolated because **any** valid output of `try_calculation_with_regulation` is valid; no finetuning is necessary to achieve a more optimal state
* if `try_calculation_with_regulation` fails, then it will throw an error (e.g. `IterationDiverge` or `SparseMatrixError`). no additional check is necessary

about your second part:

* i get that `try_calculation_with_regulation` is a little bit cryptic but don't really have a better name. do you?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: > what is it you're looking fore here?

First comment


> 1. shouldn't be an `Idx2D` (cause it's not an index and not even discrete)
> 2. we could make it a struct but this directly involves the calculation params and that already is basically that same struct 😬
> 

Minor details. Whatever type that can make a `.u_set_range` to save this:
```C++
param.u_set +/- 0.5 * param.u_band
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: > about your first part:
> 
> * `any` is isolated because **any** valid output of `try_calculation_with_regulation` is valid; no finetuning is necessary to achieve a more optimal state

This is due to the design that requires a refinement step, which is totally avoidable. I will attach a short pseudo code at the end of this comment/OneNote page as well as bring that up during the stand-up for discussion

> * if `try_calculation_with_regulation` fails, then it will throw an error (e.g. `IterationDiverge` or `SparseMatrixError`). no additional check is necessary
> 

In this case, why not simply put the actual logic in a try block instead of doing this abstraction?

> about your second part:
> 
> * i get that `try_calculation_with_regulation` is a little bit cryptic but don't really have a better name. do you?

I would name this `optimize` or `iterate` and instead call the existing `opt...` simply `run` or `run_opt`.


```python
# NB. extremely inaccurate use of names and types, which are not the point here
update_data = [{}*len(regulator_order)]
res = {}
last_iter_check = [False*len(regulator_order)]
while():
    tap_changed = False
    propose_data = {}
    while(rank < len(regulator_order)):
        same_rank_regulators = regulator_order[rank]:
        tap_changed = control_transformers(opt_strategy, same_rank_regulators, ..., res, propose_data)
        
        if not tap_changed and (opt_strategy == 'any' or last_iter_check[rank]):
            update_state(..., update_data)
            rank += 1
            continue # go to next rank

        if tap_changed or !last_iter_check[rank]:
            update_state(..., propose_data)
            res = calc_(...)

        last_run_check[rank] = !tap_changed
        update_data = update_data if last_iter_check[rank] else propose_data
        continue # still do current rank

return res
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +284,551 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+
+        bool tap_changed = true;
+        while (tap_changed) {
+            tap_changed = false;
+            UpdateBuffer update_data;
+
+            for (auto const& same_rank_regulators : regulator_order) {
+                for (auto const& regulator : same_rank_regulators) {
+                    if (regulator.transformer.connected_at_tap_side() &&
+                        regulator.transformer.connected_at_control_side(regulator.regulator)) {
```
Jerry-Jinfeng-Guo, line 609: I believe this check is also redundant

mgovers: you're right. i'll make an assert for this

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: done. hope you agree with the new implementation

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: > This is due to the design that requires a refinement step, which is totally avoidable. I will attach a short pseudo code at the end of this comment/OneNote page as well as bring that up during the stand-up for discussion

100% agree but IMHO out of scope of this PR. it's already long enough as is and this may be a future optimization

> In this case, why not simply put the actual logic in a try block instead of doing this abstraction?

actual errors are real errors. the optimizer should throw - not catch. no `try` block necessary. the `try_*` was just an unfortunate choice of words because of its ambiguity. see below.

> I would name this optimize or iterate and instead call the existing opt... simply run or run_opt.

i prefer `iterate` and keep `optimize` as is. if not, then maybe `find_local_optimum`. the `optimize` function would then specifically request a global optimum. that way, we could even go as far as to make the `local_minimum` and `local_maximum` strategies distinct from the `global_*` ones.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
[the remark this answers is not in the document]
mgovers: after discussion with tony, the answer is the 2nd option, which is the 3rd option + the assumption that the user correctly identifies `tap_max` as the tap position with the highest `u_rated` on the tap side, which is the lowest `u_rated` at the control side. This works because of the check that throws `AutomaticTapCalculationError` with the error message equivalent to `f"Automatic tap changing regulator with tap_side at LV side is not supported. Found at id {id}"`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +284,597 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+            using sym = typename ResultType::value_type::sym;
+
+            auto const param = regulator.regulator.get().template calc_param<sym>();
+            auto const node_state =
+                NodeState<sym>{.u = u_pu_controlled_node<TransformerType>(regulator, state, math_output),
+                               .i = i_pu_controlled_node<TransformerType>(regulator, state, math_output)};
+
+            auto const cmp = node_state <=> param;
+            auto new_tap_pos = [&transformer, &cmp] {
+                if (cmp > 0) { // NOLINT(modernize-use-nullptr)
```
mgovers, line 693: will be fixed in clang-tidy 18.1.0 https://releases.llvm.org/18.1.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize/use-nullptr.html#cmdoption-arg-IgnoredTypes

hopefully it will be included in Ubuntu 24.04 quite soon, but it's unlikely to be included in 22.04

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +284,598 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+            [[fallthrough]];
+        case OptimizerStrategy::local_minimum:
+            adjust_voltage(min_voltage_pos, regulator_order);
+            break;
+        default:
+            throw MissingCaseForEnumError{"TapPositionOptimizer::initialize"s, strategy_};
+        }
+    }
+
+    void step_all_back(std::vector<std::vector<RegulatedTransformer>> const& regulator_order) const {
```
Jerry-Jinfeng-Guo, line 758: I do not quite understand the new function name? 
And honestly I do not see why this function call is necessary in the first place.

(if this call is indeed necessary) this function could be re-configured for the `initialize` function?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +284,598 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+        case OptimizerStrategy::local_maximum:
+            adjust_voltage(one_step_up, regulator_order);
+            break;
+        case OptimizerStrategy::global_minimum:
+            [[fallthrough]];
+        case OptimizerStrategy::local_minimum:
+            adjust_voltage(one_step_down, regulator_order);
+            break;
+        default:
+            throw MissingCaseForEnumError{"TapPositionOptimizer::initialize"s, strategy_};
```
Jerry-Jinfeng-Guo, line 782:
```suggestion
            throw MissingCaseForEnumError{"TapPositionOptimizer::step_all_back"s, strategy_};
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +284,598 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+                for (auto const& regulator : same_rank_regulators) {
+                    tap_changed = tap_changed || control_transformer(regulator, state, result, update_data);
+                }
+                if (tap_changed) {
+                    break;
+                }
+            }
+            if (tap_changed) {
+                update_state(update_data);
+                result = calculate_(state, method);
```
Jerry-Jinfeng-Guo, line 659: I have one doubt here: is it guaranteed that `calculate_(state, method)` is good? No need to specify other `::linear` alternative?

```c++
// code changed somethere else
auto calculate_with_retry(State const& state, CalculationMethod method_in, CalculationMethod & method_used) const {
    try {
        method_used = method_in;
        return calculate_(state, method);
    } catch (const SparseMatrixError& e, const IterationDiverge& e) {
        method_used = CalculationMethod::linear;
        return calculate_(state, CalculationMethod::linear);
    }
}
```
and then:
```c++
    auto iterate(State const& state, std::vector<std::vector<RegulatedTransformer>> const& regulator_order,
                 CalculationMethod method) const -> ResultType {
        CalculationMethod method_used;
        auto result = calculate_with_retry(state, method, method_used);

        bool tap_changed = true;
        while (tap_changed) {
            tap_changed = false;
            UpdateBuffer update_data;

            for (auto const& same_rank_regulators : regulator_order) {
                for (auto const& regulator : same_rank_regulators) {
                    tap_changed = tap_changed || control_transformer(regulator, state, result, update_data);
                }
                if (tap_changed) {
                    break;
                }
            }
            if (tap_changed) {
                update_state(update_data);
                result = calculate_(state, method_used);
            }
        }

        add_tap_regulator_output(state, regulator_order, result);

        return result;
    }
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +284,598 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+        result.id = transformer.id();
+        result.tap_pos = new_tap_pos;
+        get<std::remove_cvref_t<decltype(transformer)>>(update_data).push_back(result);
+    }
+
+    template <typename Func>
+        requires((std::invocable<Func, TransformerTypes const&> &&
+                  std::same_as<std::invoke_result_t<Func, TransformerTypes const&>, IntS>) &&
+                 ...)
+    auto adjust_voltage(Func new_tap_pos, std::vector<std::vector<RegulatedTransformer>> const& regulator_order) const {
```
Jerry-Jinfeng-Guo, line 797: I believe this function should be named `regulate_transformers`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -273,41 +284,598 @@ inline auto rank_transformers(State const& state) -> RankedTransformerGroups {
...
+            return calculate_(state, method);
+        } catch (SparseMatrixError const&) {
+            return calculate_(state, CalculationMethod::linear);
+        } catch (IterationDiverge const&) {
+            return calculate_(state, CalculationMethod::linear);
+        }
+    }
+
+    bool control_transformer(RegulatedTransformer const& regulator, State const& state, ResultType const& math_output,
+                             UpdateBuffer& update_data) const {
```
Jerry-Jinfeng-Guo, line 679: I believe this function should be renamed `adjust_transformer`

Jerry-Jinfeng-Guo (commented):
A few comments. If merging is high priority, I am fine with going ahead with minor tweaks. The one thing that bothers me the most for now is the fact that there are multiple channels to adjust transformer tap positions.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: Let's park this discussion for later.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: maybe better: `reinitialize` or `initialize_refinement`?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: > I have one doubt here: is it guaranteed that calculate_(state, method) is good? No need to specify other ::linear alternative?

indeed. in fact, I believe that with #505 in place, this might not even be necessary anymore. @TonyXiang8787 do you know?

mgovers:
```cpp
CalculationMethod method_used;
auto result = calculate_with_retry(state, method, method_used);
// ...
result = calculate_(state, method_used);
```

this would mean it will always use `linear` as a fallback and keep using it, while in practice, it should only be an improved initial guess to ensure convergence on the first run that is later refined with the actual provided method

Jerry-Jinfeng-Guo: `calculate_with_retry` says otherwise

mgovers: no, the current `calculate_with_retry` is only used during the initial result. The fallback `linear` method is used there. During the actual calculation, the `method` provided by the user is used.

In your suggestion, you set `method_used` based on which method you provided and you only use that, completely ignoring whatever the user provided.

mgovers: i have a proposal: let's remove the `calculate_with_retry` and only use `calculate_`. The retry logic is something that can be worked out in a follow-up step. This suits the Continuous Improvement WoW. Since @TonyXiang8787 's input is needed here anyways, I would deem this a next iteration

Jerry-Jinfeng-Guo: > no, the current `calculate_with_retry` is only used during the initial result. The fallback `linear` method is used there. During the actual calculation, the `method` provided by the user is used.

`calculate_with_retry` is called every time `iterate` is called.

> In your suggestion, you set `method_used` based on which method you provided and you only use that, completely ignoring whatever the user provided.

That's what the code in its original state suggests: if the users method doesn't work, use linear. Twice.

mgovers: OK, i do see a mistake in my implementation:
```
try 
  math_result = math_solver.power_flow(calculation_method=calculation_method) 
  if (abs(u_pu) < 0.2).any(): 
    Throw InvalidPFResult 
catch (InvalidPFResult, SparseMatrixError, IterationDiverge) 
  Run run_pf_with_regulation_per_solver(math_solver=math_solver, calculation_method=linear),   discard return 
  math_result = math_solver.power_flow(calculation_method=calculation_method) // i forgot this line
```

I added a Jira ticket for that because I have reason to believe that #505 should solve this because of the discard step: the initial state is actually set to the solution of the `linear` calculation.

That leaves me with even more reason that all the other fallback logic is also unnecessary and we can just remove `calculate_with_retry` and, instead, directly call `calculate_`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: `exploit_neighborhood`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: removing from this PR.

mgovers: Moved to https://github.com/PowerGridModel/power-grid-model/issues/580

Jerry-Jinfeng-Guo (approved):
I hereby agree with moving parts of review comments to later follow-ups and approve the PR. (e.g., https://github.com/PowerGridModel/power-grid-model/issues/580)

PR #558, opened 2024-03-28
Feature/transformer ranking step 4

Jerry-Jinfeng-Guo:
Tentative implementation of step 4. Notice that the return type of `rank_transformer` is now a vector of vector, meaning that each group contains one or more trafo with equal trafo edge weight / rank. This is the result of the discussion of not having an order within same rank. Code might need further change once step 1 is finalized, but now it is clear what is expected.


tests/cpp_unit_tests/test_transformer_tap_regulator.cpp
```diff
@@ -31,8 +31,6 @@ TEST_CASE("Test transformer tap regulator") {
 
     TransformerTapRegulator transformer_tap_regulator{input, u_rated};
 
-    // TODO: test inverse update
-
```
Jerry-Jinfeng-Guo, line 35: Please don't comment here.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -23,6 +23,7 @@ using TrafoGraphIdx = Idx;
 using EdgeWeight = int64_t;
 using WeightedTrafo = std::pair<Idx2D, EdgeWeight>;
 using WeightedTrafoList = std::vector<WeightedTrafo>;
+using RankedTransformerGroups = std::vector<std::vector<Idx2D>>;
```
mgovers, line 26: the more i think about it, the more i'm starting to think that this might not be the correct approach after all. we shouldn't need the trafo's `Idx2D` but the regulator's `Idx` (which is equivalent to `Idx2D{.group = Container::template get_type_idx<AutomaticTapRegulator>, .idx = regulator_idx}` because of how the container works)

in addition, i'm also thinking more and more about how we actually want to know not only the rank but also which connected component the regulator belongs to...

I am quite hesitant to make a vector of vectors of vectors but i am not sure if we can come up with a better way. any thoughts?

Jerry-Jinfeng-Guo: The other option I had in mind is to simply have an ordered list, with additional edge weight / rank associated with each entry. That makes basically everything else more complicated.

`{ [id=*, rank=100], ...,[id=*, rank=10], [id=*, rank=10], [id=*, rank=10], ...}`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -23,6 +23,7 @@ using TrafoGraphIdx = Idx;
...
 const int infty = INT_MAX;
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: This change is not strictly necessary in our case here but indeed nice to have.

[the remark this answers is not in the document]
mgovers: well, the code would break if we ever decide to make `Idx` a `size_t` or a `long long` or whatever

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: that's OK too. i FINALLY found a way to handle different transformer types nicely in my branch. will push now so you can have a look

mgovers: pushed to https://github.com/PowerGridModel/power-grid-model/pull/557

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: "break" in what sense? It is only used to initialize to a large than necessary value, which is unlikely to be exceeded in most systems. I could comfortably retire if I have `2147483647` even in JPY.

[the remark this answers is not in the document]
mgovers: wraparounds are easy to create when people write scripts. since we support things like negative IDs etc. it's better to be safe than sorry. you never know who copy-pastes your code ;D

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: out of scope of this PR

mgovers (approved):

PR #562, opened 2024-04-03
Transformer ranking part 1 

nitbharambe:


Jerry-Jinfeng-Guo (commented):
I like the spirit ;)

TonyXiang8787:
@nitbharambe @Jerry-Jinfeng-Guo  why is #567 merged but this not? How can you test the ranking without the full implementation?

Jerry-Jinfeng-Guo:
> @nitbharambe @Jerry-Jinfeng-Guo why is #567 merged but this not? How can you test the ranking without the full implementation?

The merged tests are for consequtive steps in the ranking. Since steps 2-3-4 and step 1 have inter-dependency, so are the tests. Therefore we figure it to be a good idea to merge stuff from 2-3-4 once ready and continue working on 1, while cross-referencing.

Jerry-Jinfeng-Guo:
CI is stuck

mgovers:
> CI is stuck

that's because @nitbharambe prefixed the last commit [847723d](https://github.com/PowerGridModel/power-grid-model/pull/562/commits/847723dda45b351488ccea0a4d48614543192d1a) with `[skip ci]` in

Jerry-Jinfeng-Guo:
> > CI is stuck
> 
> that's because @nitbharambe prefixed the last commit [847723d](https://github.com/PowerGridModel/power-grid-model/pull/562/commits/847723dda45b351488ccea0a4d48614543192d1a) with `[skip ci]`
> 


That should `skip` it, shouldn't it?

mgovers:
> > > CI is stuck
> > 
> > 
> > that's because @nitbharambe prefixed the last commit [847723d](https://github.com/PowerGridModel/power-grid-model/pull/562/commits/847723dda45b351488ccea0a4d48614543192d1a) with `[skip ci]`
> 
> That should `skip` it, shouldn't it?

yes but the way github implemented it is that if there are `Required` checks (configurable in the repo admin settings) then it will always mark them as `Expected`.

This also is the reason why the `Publish` job is awaited in the merge queue but it doesn't link to the pipeline that will set it (because the job is the last one in the pipeline and won't report its status until it has started). I've been thinking about actually adding a report step that tells the merge queue that the pipeline that will set run those checks, but haven't come up with a sustainable solution yet


image

TonyXiang8787:
> > > > CI is stuck
> > > 
> > > 
> > > that's because @nitbharambe prefixed the last commit [847723d](https://github.com/PowerGridModel/power-grid-model/pull/562/commits/847723dda45b351488ccea0a4d48614543192d1a) with `[skip ci]`
> > 
> > 
> > That should `skip` it, shouldn't it?
> 
> yes but the way github implemented it is that if there are `Required` checks (configurable in the repo admin settings) then it will always mark them as `Expected`.
> 
> This also is the reason why the `Publish` job is awaited in the merge queue but it doesn't link to the pipeline that will set it (because the job is the last one in the pipeline and won't report its status until it has started). I've been thinking about actually adding a report step that tells the merge queue that the pipeline that will set run those checks, but haven't come up with a sustainable solution yet
> 
> image

We can just push a new commit with `skip ci` in the message right?

mgovers:
> > > > > CI is stuck
> > > > 
> > > > 
> > > > that's because @nitbharambe prefixed the last commit [847723d](https://github.com/PowerGridModel/power-grid-model/pull/562/commits/847723dda45b351488ccea0a4d48614543192d1a) with `[skip ci]`
> > > 
> > > 
> > > That should `skip` it, shouldn't it?
> > 
> > 
> > yes but the way github implemented it is that if there are `Required` checks (configurable in the repo admin settings) then it will always mark them as `Expected`.
> > This also is the reason why the `Publish` job is awaited in the merge queue but it doesn't link to the pipeline that will set it (because the job is the last one in the pipeline and won't report its status until it has started). I've been thinking about actually adding a report step that tells the merge queue that the pipeline that will set run those checks, but haven't come up with a sustainable solution yet
> > image
> 
> We can just push a new commit with `skip ci` in the message right?

yes but that is not the issue 😉 my second comment was only related to the merge queue

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -2,42 +2,127 @@
...
+                                                 edge_array.cend(), edge_prop.cbegin(), 10};
+        pgm_tap::TransformerGraph actual_graph = pgm_tap::build_transformer_graph(state);
+
+        std::vector<pgm_tap::TrafoGraphVertex> expected_vertex_props{{true},  {false}, {false}, {false}, {false},
+                                                                     {false}, {false}, {false}, {false}, {false}};
+
+        boost::graph_traits<pgm_tap::TransformerGraph>::vertex_iterator vi, vi_end;
+        for (boost::tie(vi, vi_end) = vertices(actual_graph); vi != vi_end; ++vi) {
+            actual_graph[*vi].is_source = expected_vertex_props[*vi].is_source;
+        }
```
Jerry-Jinfeng-Guo, line 103: Why is this necessary? the `// Mark sources` section inside the `build_transformer_graph` function is already doing that, no?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -35,14 +38,165 @@ struct TrafoGraphEdge {
     EdgeWeight weight{};
```
nitbharambe, line 38: @Jerry-Jinfeng-Guo Let us use either WeightedTrafo or TrafoGraphEdge. Is there a reason to use WeightedTrafo?

Jerry-Jinfeng-Guo: Indeed redundant, let's do:
```
using WeightedTrafoList = std::vector<TrafoGraphEdge>;
```

Jerry-Jinfeng-Guo: A few places where `WeightedTrafo` were also used could use an update as well.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -69,11 +69,10 @@ constexpr void add_edges(main_core::MainModelState<ComponentContainer> const& st
             auto const& to_node = transformer3w.node(to_side);
             const TrafoGraphEdge edge_prop{main_core::get_component_idx_by_id(state, transformer3w.id()), 1};
 
-            const bool single_direction_condition =
-                regulated_objects.transformers3w.contains(transformer3w.id()) &&
-                (transformer3w.tap_side() == from_side || transformer3w.tap_side() == to_side);
+            auto const tap_at_from_side = transformer3w.tap_side() == from_side;
+            auto const single_direction_condition = regulated_objects.transformers3w.contains(transformer3w.id()) &&
+                                                    (tap_at_from_side || transformer3w.tap_side() == to_side);
```
mgovers, line 86: three winding transformer has `side_1`, `side_2` and `side_3`, which are also overloaded in the `ControlSide`. what's the reason for using `from` and `to` here for `3w`?

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -118,6 +118,14 @@ class InvalidRegulatedObject : public PowerGridError {
     }
 };
 
+class AutomaticTapCalculationError : public PowerGridError {
+  public:
+    AutomaticTapCalculationError(ID id) {
+        append_msg("Automatic tap changing regulator is at LV side for id" + std::to_string(id) +
+                   "which is not supported");
+    }
+};
```
mgovers, line 127: please also add to the python error handling

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -23,7 +23,7 @@ class Transformer : public Branch {
     using UpdateType = TransformerUpdate;
     static constexpr char const* name = "transformer";
 
-    explicit Transformer(TransformerInput const& transformer_input, double u1_rated, double u2_rated)
+    explicit Transformer(TransformerInput const& transformer_input, double u1_rated = 1.0, double u2_rated = 1.0)
```
mgovers, line 26: why? this is different from what we currently do in the rest of the PGM so we should consider this carefully. if you really need this, please add an explicit overload

```suggestion
    explicit Transformer(TransformerInput const& transformer_input): Transformer{.transformer_input = transformer_input, .u1_rated = 1.0, .u2_rated = 1.0} {};
    // or just explicit Transformer(TransformerInput const& transformer_input): Transformer{.transformer_input, 1.0, 1.0} {};
    explicit Transformer(TransformerInput const& transformer_input, double u1_rated, double u2_rated)
```

or if that doesn't compile just
```suggestion
    explicit Transformer(TransformerInput const& transformer_input): Transformer{.transformer_input, 1.0, 1.0} {};
    explicit Transformer(TransformerInput const& transformer_input, double u1_rated, double u2_rated)
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,112 +25,249 @@ namespace detail = power_grid_model::optimizer::detail;
...
 constexpr auto infty = std::numeric_limits<Idx>::max();
 
 struct TrafoGraphVertex {
-    bool is_source{}; // is_source = true if the vertex is a source
+    bool is_source{};
 };
 
 struct TrafoGraphEdge {
-    Idx2D pos{};
+    bool is_trafo{false};
```
mgovers, line 35: implied
```suggestion
    bool is_trafo{};
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,112 +25,249 @@ namespace detail = power_grid_model::optimizer::detail;
...
+
+    bool operator<(const TrafoGraphEdge& other) const {
+        if (weight != other.weight) {
+            return weight < other.weight;
+        }
+        if (from_to.group != other.from_to.group) {
+            return from_to.group < other.from_to.group;
+        }
+        return from_to.pos < other.from_to.pos;
+    }
```
mgovers, line 51: please instead, define the spaceship operator `<=>`. this was added in C++20 precisely because it is otherwise so verbose to write. this is a really good writeup on that: https://indii.org/blog/overloading-the-spaceship-operator/

i believe that in this case the syntax is something like

```suggestion
    // weak ordering because we don't explicitly check for is_trafo
    std::weak_ordering operator<=>(const TrafoGraphEdge& other) const {
        if (auto cmp = weight <=> other.weight; cmp != 0) {
            return cmp;
        }
        if (auto cmp = from_to.group <=> other.from_to.group; cmp != 0) {
            return cmp;
        }
        return from_to.pos <=> other.from_to.pos;
    }
```

we can also decide to overload this operator<=> for `Idx2D` and then it becomes even shorter. i think that makes sense cause Idx2D is basically ordered due to the implied meaning of `group` and `pos`

```suggestion
    // weak ordering because we don't explicitly check for is_trafo
    std::weak_ordering operator<=>(const TrafoGraphEdge& other) const {
        if (auto cmp = weight <=> other.weight; cmp != 0) {
            return cmp;
        }
        return from_to <=> other.from_to;
    }
```

fun fact, if you declare `weight` before `from_to`, then it becomes even easier, because the first argument is compared first by default (implying strong ordering):

```suggestion
    friend auto operator<=>(const TrafoGraphEdge& a, const TrafoGraphEdge& b) = default;
```

i believe that that is also sufficient for `Idx2D`:
```cpp
    friend auto operator<=>(const Idx2D& a, const Idx2D& b) = default;
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,112 +25,249 @@ namespace detail = power_grid_model::optimizer::detail;
...
 
+template <std::same_as<ThreeWindingTransformer> Component, class ComponentContainer>
+    requires main_core::model_component_state_c<main_core::MainModelState, ComponentContainer, Component>
+constexpr void add_edges(main_core::MainModelState<ComponentContainer> const& state,
+                         RegulatedObjects const& regulated_objects, TrafoGraphEdges& edges,
+                         TrafoGraphEdgeProperties& edge_props) {
+    std::array<std::tuple<Branch3Side, Branch3Side>, 3> const branch3_combinations{
+        {{Branch3Side::side_1, Branch3Side::side_2},
+         {Branch3Side::side_2, Branch3Side::side_3},
+         {Branch3Side::side_1, Branch3Side::side_3}}};
```
mgovers, line 75:
```suggestion
    using enum Branch3Side;
    std::array<std::tuple<Branch3Side, Branch3Side>, 3> const branch3_combinations{
        {{side_1, side_2}, {side_2, side_3}, {side_1, side_3}}};
```

mgovers (commented):
part 1 of review. more to come

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,112 +25,249 @@ namespace detail = power_grid_model::optimizer::detail;
...
 
 struct TrafoGraphVertex {
-    bool is_source{}; // is_source = true if the vertex is a source
+    bool is_source{};
 };
 
 struct TrafoGraphEdge {
-    Idx2D pos{};
+    bool is_trafo{false};
+    Idx2D from_to{};
```
nitbharambe, line 36: Should this be group and pos? ie. idx2d from the container?
The from_to can be found in the edges right?

mgovers: you're right. the current `Idx2D` is specifically for hierarchical indices.

in this case, just `using FromTo = std::pair<Idx, Idx>` would do the trick, or you can create a struct in this namespace that contains `struct FromTo{Idx to{}; Idx from{}};`

Jerry-Jinfeng-Guo: My initial intent making this change is due to the fact that `Idx2D` is a pain * * * concept to work with outside its own territory. In actuality, the field `is_trafo` and `weight` are redundant. My proposal is to swap the `is_trafo` field to the old `Idx2D id`. Accordingly, the `struct FromTo` as suggested by @mgovers would be probably/eventually `struct FromTo {Idx2D from{}; Idx2D to{}};`. This will be up to discussion per later use in the tap changing optimizer.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: The comparison operators are actually necessary evil in test cases as in production a sorted list within group is not necessary. If shuffling the lines a bit makes things so much simpler, why nut?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: This field might be obsolete.

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/issues/574 
Revert is simple so long as it doesn't impose any issues.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -2,67 +2,303 @@
...
+                return a.group < b.group;
+            }
+            if (a.pos != b.pos) {
+                return a.pos < b.pos;
+            }
+            if (a.group != b.group) {
+                return a.group < b.group;
+            }
+            return a.pos < b.pos;
+        };
```
mgovers, line 287: ???????

1. there's duplication
2. since `Idx2D` represents a hierarchical index, it's probably good to implement this function where it's declared anyways. this is not the correct location.
3. even more so, it is probably possible to `default` it cause by default it is strong ordering in the order the fields are declared
4. if you really need it here, please make it like this
```suggestion
        // necessary evil, made possible by platform differences
        auto compareIdx2D = [](const Idx2D& a, const Idx2D& b) {
            if (a.group != b.group) {
                return a.group < b.group;
            }
            return a.pos < b.pos;
        };
```

i'm very interested to see what the platform differences are

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -2,67 +2,303 @@
...
+        }
+        CHECK(order == ref_order);
+    }
 }
 
 TEST_CASE("Test Tap position optimizer" * doctest::skip(true)) {
     // TODO: Implement unit tests for the tap position optimizer
 }
-
+//}
```
mgovers, line 303: ?
```suggestion
```

mgovers (changes requested):

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: > But having the RankedTransformerGroups by from/to node and weight would not help in finding the exact regulated object.

Indeed, hence the proposal of swapping the redundant is_trafo to Idx_2D

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,112 +25,249 @@ namespace detail = power_grid_model::optimizer::detail;
...
 
-inline auto process_edges_dijkstra(Idx v, std::vector<EdgeWeight>& edge_weight, std::vector<Idx2D>& edge_pos,
-                                   TransformerGraph const& graph) -> void {
+inline auto process_edges_dijkstra(Idx v, std::vector<EdgeWeight>& vtx_distances, TransformerGraph const& graph)
+    -> void {
     using TrafoGraphElement = std::pair<EdgeWeight, TrafoGraphIdx>;
     std::priority_queue<TrafoGraphElement, std::vector<TrafoGraphElement>, std::greater<>> pq;
-    edge_weight[v] = 0;
-    edge_pos[v] = {v, v};
+    vtx_distances[v] = 0;
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Either vertex or vtx.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: > ???????

?????????

> 
> 1. there's duplication

Don't work on a Sunday afternoon.

> 2. since `Idx2D` represents a hierarchical index, it's probably good to implement this function where it's declared anyways. this is not the correct location.
> 3. even more so, it is probably possible to `default` it cause by default it is strong ordering in the order the fields are declared
> 4. if you really need it here, please make it like this
> 

As mentioned in the code, only necessary for the test cases.

> i'm very interested to see what the platform differences are

{..., {5, 7}, {5, 4}, ...}
vs
{..., {5, 4}, {5, 7}, ...}

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: // TEST_SUIT

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: let's see if the same thing happens if you declare the strong ordering where `Idx2D` is implemented. IMO it makes sense to add fundamental behavior like this on a global level (reinforcing the notion of hierarchical indices)

mgovers: how, when and why do these differences arise? is it caused by boost?

as long as the weak ordering criterion is satisfied, that's OK and that's exactly what you check

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
nitbharambe: So this is to be added or removed?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+
+            auto const tap_at_first_side = transformer3w.tap_side() == first_side;
+            auto const single_direction_condition = regulated_objects.transformers3w.contains(transformer3w.id()) &&
+                                                    (tap_at_first_side || transformer3w.tap_side() == second_side);
+            if (single_direction_condition) {
+                auto const& tap_side_node = tap_at_first_side ? from_node : to_node;
+                auto const& non_tap_side_node = tap_at_first_side ? to_node : from_node;
+                Idx2D const regulated_idx =
+                    assigned_idx ? unregulated_idx : main_core::get_component_idx_by_id(state, transformer3w.id());
+                assigned_idx = true;
```
mgovers, line 92: wouldn't this logic be much easier if you would loop over `{(side_1, side_2), (side_2, side_3), (side_3, side_1)}` instead of `(side_1, side_3)` for the last element?

you can then check whether `first_side` is regulated. if so, assign `regulated_idx`. if not, assign `unregulated_idx`. The fact that the loop contains cyclic combinations ensures uniqueness. no need for complex logic like `assigned_idx`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+        auto const& to_node = transformer.to_node();
+
+        if (regulated_objects.transformers.contains(transformer.id())) {
+            auto const tap_at_from_side = transformer.tap_side() == BranchSide::from;
+            auto const& tap_side_node = tap_at_from_side ? from_node : to_node;
+            auto const& non_tap_side_node = tap_at_from_side ? to_node : from_node;
+            if (get_component<Node>(state, tap_side_node).u_rated() <
+                get_component<Node>(state, non_tap_side_node).u_rated()) {
+                throw AutomaticTapCalculationError(transformer.id());
+            }
```
mgovers, line 121: shouldn't a similar check happen for 3w transformer? if so, is this really the right location? can't we do that at construction time of the 3w transformer? feels like a typical use case of `assert_valid_input_data`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+    for (auto const& transformer3w : state.components.template citer<ThreeWindingTransformer>()) {
+        for (auto const& [first_side, second_side] : branch3_combinations) {
+            if (!transformer3w.status(first_side) || !transformer3w.status(second_side)) {
+                continue;
+            }
+            auto const& from_node = transformer3w.node(first_side);
+            auto const& to_node = transformer3w.node(second_side);
+
+            auto const tap_at_first_side = transformer3w.tap_side() == first_side;
+            auto const single_direction_condition = regulated_objects.transformers3w.contains(transformer3w.id()) &&
```
mgovers, line 85: the `map::contains` check is relatively expensive. please do this in the outer loop. in fact, you can make the distinction there between regulated and unregulated transformers in a function that calls different implementations for `add_unregulated_edges<TransformerType>` and `add_regulated_edges<TransformerType>`.

The reason is that that logic is common for both `Transformer` and `ThreeWindingTransformer`, so you can make a common function `template <transformer_c TransformerType> void add_edges(...)`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+            add_to_edge(edges, edge_props, to_node, from_node, {unregulated_idx, 1});
+        }
+    }
+}
+
+template <typename Component>
+concept non_regulating_branch_c = std::derived_from<Link, Component> || std::derived_from<Line, Component>;
+
+template <non_regulating_branch_c Component, class ComponentContainer>
+    requires main_core::model_component_state_c<main_core::MainModelState, ComponentContainer, Component>
```
mgovers, line 135: more general:

```suggestion
template <typename Component>
concept non_regulated_branch_c = std::derived_from<Component, Branch> && !transformer_c<Component>;

template <non_regulating_branch_c Component, class ComponentContainer>
    requires main_core::model_component_state_c<main_core::MainModelState, ComponentContainer, Component>
```

since you only use it once, you can also keep this logic but make it as follows:

```suggestion
template <std::derived_from<Component, Branch> Component, class ComponentContainer>
    requires main_core::model_component_state_c<main_core::MainModelState, ComponentContainer, Component> && !transformer_c<Component>
```

because then the compiler will raise an error if the edge type is not derived from `Branch` and also not a `Transformer` or `ThreeWindingTransformer`. The following 2 such cases may arise:

* For edge-type components that are not `Branch`, the user should specify overloads to create the unregulated edges.
* For transformer types (that can be regulated) that are not `Transformer` or `ThreeWindingTransformer`, the user should also specify overloads

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+    edges.reserve(std::distance(iter.begin(), iter.end()) * 2);
+    edge_props.reserve(std::distance(iter.begin(), iter.end()) * 2);
+    for (auto const& branch : iter) {
+        if (!branch.from_status() || !branch.to_status()) {
+            continue;
+        }
+        add_to_edge(edges, edge_props, branch.from_node(), branch.to_node(), {unregulated_idx, 0});
+        add_to_edge(edges, edge_props, branch.to_node(), branch.from_node(), {unregulated_idx, 0});
+    }
+}
```
mgovers, line 149: This logic is identical for unregulated transformers. it's probably good to do something like this:

```suggestion
template <std::derived_from<Branch> Component, class ComponentContainer>
    requires main_core::model_component_state_c<main_core::MainModelState, ComponentContainer, Component>
constexpr void add_unregulated_edge(Component const& branch,
                        TrafoGraphEdges& edges,
                        TrafoGraphEdgeProperties& edge_props) {
    if (!branch.from_status() || !branch.to_status()) {
        continue;
    }
    add_to_edge(edges, edge_props, branch.from_node(), branch.to_node(), {unregulated_idx, 0});
    add_to_edge(edges, edge_props, branch.to_node(), branch.from_node(), {unregulated_idx, 0});
}

template <non_regulating_branch_c Component, class ComponentContainer>
    requires main_core::model_component_state_c<main_core::MainModelState, ComponentContainer, Component>
constexpr void add_edge(main_core::MainModelState<ComponentContainer> const& state,
                        RegulatedObjects const& regulated_objects, TrafoGraphEdges& edges,
                        TrafoGraphEdgeProperties& edge_props) {
    auto const& iter = state.components.template citer<Component>();
    edges.reserve(std::distance(iter.begin(), iter.end()) * 2);
    edge_props.reserve(std::distance(iter.begin(), iter.end()) * 2);
    for (auto const& branch : iter) {
        add_unregulated_edge<Component>(branch, edges, edge_props);
    }
}
```

mgovers (commented):
part 1 of review

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+
+    if (state.components.template size<Source>() == 0) {
+        return trafo_graph;
+    }
+
+    // Mark sources
+    for (auto const& source : state.components.template citer<Source>()) {
+        // ignore disabled sources
+        trafo_graph[source.node()].is_source = source.status();
+    }
```
mgovers, line 203: overoptimization: if the amount of sources is 0 the `for` loop doesn't do anything anyway. (in fact, it introduces an unnecessary branch so it's actually even slower in almost every case)

```suggestion
    // Mark sources
    for (auto const& source : state.components.template citer<Source>()) {
        // ignore disabled sources
        trafo_graph[source.node()].is_source = source.status();
    }
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+
+    return trafo_graph;
 }
 
-inline auto process_edges_dijkstra(Idx v, std::vector<EdgeWeight>& edge_weight, std::vector<Idx2D>& edge_pos,
-                                   TransformerGraph const& graph) -> void {
+inline auto process_edges_dijkstra(Idx v, std::vector<EdgeWeight>& vertex_distances, TransformerGraph const& graph)
+    -> void {
```
mgovers, line 209: trailing return type `void` makes no sense. either keep it `auto` (no return type will be deduced) or just mark it `void` in the leading return type directly

```suggestion
inline void process_edges_dijkstra(Idx v, std::vector<EdgeWeight>& vertex_distances, TransformerGraph const& graph) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
-            groups.push_back(std::vector<Idx2D>{trafo.first});
-            last_weight = trafo.second;
-        } else {
-            groups.back().push_back(trafo.first);
-        }
-    }
+              [](const TrafoGraphEdge& a, const TrafoGraphEdge& b) { return a.weight < b.weight; });
+
+    RankedTransformerGroups groups(sorted_trafos.size());
+    std::ranges::transform(sorted_trafos.begin(), sorted_trafos.end(), groups.begin(),
```
mgovers, line 262: if you use `begin` and `end` but don't make use of the `|` functionality of `std::ranges` then you might as well use `std::ranges` directly

```suggestion
    std::transform(sorted_trafos.begin(), sorted_trafos.end(), groups.begin(),
```

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -2,67 +2,267 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
+#include <power_grid_model/main_core/input.hpp>
 #include <power_grid_model/optimizer/tap_position_optimizer.hpp>
 
+#include <algorithm>
+#include <ranges>
+
 #include <doctest/doctest.h>
```
mgovers, line 11: bad practice to include headers after `STL` because they might overload functionality, resulting in unexpected behaviour.

see also #236 

```suggestion
#include <doctest/doctest.h>

#include <algorithm>
#include <ranges>
```

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -2,67 +2,267 @@
...
-    }
+        // reference graph creation
+        // Inserted in order of transformer, transformer3w, line and link
+        std::vector<std::pair<Idx, Idx>> expected_edges;
+        expected_edges.insert(expected_edges.end(), {{0, 1}, {0, 1}, {5, 7}, {2, 3}, {8, 9}});
+        expected_edges.insert(expected_edges.end(), {{0, 4}, {4, 5}, {5, 4}, {0, 5}});
+        expected_edges.insert(expected_edges.end(), {{3, 6}, {6, 3}, {3, 9}, {9, 3}});
+        expected_edges.insert(expected_edges.end(), {{2, 1}, {1, 2}, {6, 4}, {4, 6}, {8, 7}, {7, 8}});
+
+        using pgm_tap::unregulated_idx;
```
mgovers, line 184: optionally: move to start of the scope

tests/unit/test_error_handling.py
```diff
@@ -205,6 +206,45 @@ def test_handle_invalid_calculation_method_error():
         model.calculate_power_flow(calculation_method=CalculationMethod.iec60909)
 
 
+@pytest.mark.skip(reason="TODO: Automatic tap changer")
```
mgovers, line 209: maybe instead of `skip` use `xfail` so that we are notified whenever this actually passes

```suggestion
@pytest.mark.xfail(reason="TODO: Automatic tap changer")
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+    using enum Branch3Side;
+    std::array<std::tuple<Branch3Side, Branch3Side>, 3> const branch3_combinations{
+        {{side_1, side_2}, {side_2, side_3}, {side_1, side_3}}};
+    for (auto const& transformer3w : state.components.template citer<ThreeWindingTransformer>()) {
+        for (auto const& [first_side, second_side] : branch3_combinations) {
+            if (!transformer3w.status(first_side) || !transformer3w.status(second_side)) {
+                continue;
+            }
+            auto const& from_node = transformer3w.node(first_side);
+            auto const& to_node = transformer3w.node(second_side);
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: No. These are concepts from two contexts. Nitish's original choice was correct.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
+
+            auto const tap_at_first_side = transformer3w.tap_side() == first_side;
+            auto const single_direction_condition = regulated_objects.transformers3w.contains(transformer3w.id()) &&
+                                                    (tap_at_first_side || transformer3w.tap_side() == second_side);
+            if (single_direction_condition) {
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: The suggestion makes it less descriptive.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
nitbharambe: Its not in algorithm for 3 winding transformer. Like i proposed at parallel 0 weighted edge case, adding this condition can also address that practically unlikely edge case.

Nice idea though, if implemented at `assert_valid_input_data`, it should not matter if automatic tap changer gives divergence or incorrect result by not satisfying this condition.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
nitbharambe: I thought of that, but its less desirable to refactor the few repeated lines with a new function that would need convoluted concept checks. Hardly any less duplication was seen with `add_regulated_edge` and `add_unregulated_edge`.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
nitbharambe: The `citer` not work if the component does not exist. Hence the size check is added first.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
nitbharambe: That change is thanks to sonar cloud nagging :D

mgovers: OHHHHH i understand why. this should be
```suggestion
    std::ranges::transform(sorted_trafos, groups.begin(),
```

cause `RankedTransformerGroups` is an `std::vector` and therefore supports direct range access

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: riiiiiight... i forgot that unregulated transformers have a different edge weight....

i was going to write the following. However, `Transformers` act like branches but then not entirely so that doesn't apply here

> well this also concerns testability:
> * the check whether something is regulated or not depends on whether it satisfies the `transformer_c` concept.
> * the adding of an edge depends on the `Branch` type
> 
> constraints should be as generic that only one function is needed to be tested (here: something can be automatic tap regulated if and only if it is a `transformer_c` with an updatable `tap_pos`) and so specific that overloads are not accidentally added to other things.
> 
> duplication is one thing but the current implementation may actually make the logic flow less obvious because

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,242 @@ namespace detail = power_grid_model::optimizer::detail;
...
 struct TrafoGraphEdge {
-    Idx2D pos{};
+    Idx2D regulated_idx{};
     EdgeWeight weight{};
+
+    bool operator==(const TrafoGraphEdge& other) const {
+        return regulated_idx == other.regulated_idx && weight == other.weight;
+    } // thanks boost
+
+    auto operator<=>(const TrafoGraphEdge& other) const {
```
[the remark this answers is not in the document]
mgovers: i'm starting to realize why so many different projects are looking for alternatives from boost... it's great but then not really....

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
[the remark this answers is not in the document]
mgovers: probably my confusion came from the fact that there's two logical steps in one statement

* it is regulated
* because of that, we model it as unidirectional

my guess is that one of my other comments on splitting up when the `contains` check happens should make this more clear

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: why doesn't it apply for 3w transformer? you still have the requirement that the u_rated at the tap side should be geq the u_rated at the regulated side, right? especially since it is modeled as pi-configuration and therefore effectively makes the following configuration
```
        /-oo- B
A -oo- D
        \-oo- C
```

if A is now regulated, then `u_rated(A) >= u_rated(D) >= u_rated(B), u_rated(C)`

otherwise, if the single-regulated-transformer configuration `A -oo- D` can give rise to divergences and errors, then so can the regulated 3w threewinding transformer

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: in my branch, i added this constexpr check: https://github.com/PowerGridModel/power-grid-model/pull/557/files#diff-1d8601478861bbe22e22c742aa5db8c51c0fe1fa9c5a225193dcb72cce714ef8R52-R53

i think you can combine that with adding to the container something like

```cpp
    // type checks
    template <typename... T>
    static constexpr has_gettable_type = supported_type_c<T, GettableTypes...>;
    template <typename... T>
    static constexpr has_storageable_type = supported_type_c<T, StorageableTypes...>;
```

mgovers: i can cherry-pick these changes if you like

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: The change is really not adding much, but since you typed so many letters already on this, I'll remove the 28 letters

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
nitbharambe: Then actually we dont need those size checks too ;(

mgovers: do you want me to or not? hahaha

nitbharambe: Yes please!

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
```diff
@@ -124,6 +127,31 @@ inline void add_component(MainModelState<ComponentContainer>& state, ForwardIter
                 }
             }();
 
+            if (regulated_object_idx.group == get_component_type_index<Transformer>(state)) {
```
mgovers, line 130: yeah this entire function needs refactoring. i guess it's ok for now (since it's already a wretch)

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -6,10 +6,14 @@
...
 #include "base_optimizer.hpp"
 
+#include "../all_components.hpp"
 #include "../auxiliary/dataset.hpp"
 #include "../common/enum.hpp"
+#include "../common/exception.hpp"
+#include "../main_core/state_queries.hpp"
 
 #include <boost/graph/compressed_sparse_row_graph.hpp>
+#include <boost/graph/dijkstra_shortest_paths.hpp>
```
mgovers, line 16: nitpick: whiteline between 3rd party libs and STL libs

```suggestion
#include <boost/graph/dijkstra_shortest_paths.hpp>

```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,240 @@ namespace detail = power_grid_model::optimizer::detail;
...
-using WeightedTrafo = std::pair<Idx2D, EdgeWeight>;
-using WeightedTrafoList = std::vector<WeightedTrafo>;
-using RankedTransformerGroups = std::vector<std::vector<Idx2D>>;
 constexpr auto infty = std::numeric_limits<Idx>::max();
+constexpr Idx2D unregulated_idx = {-1, -1};
+
+using enum Branch3Side;
+
+constexpr std::array<std::tuple<Branch3Side, Branch3Side>, 3> const branch3_combinations{
+    {{side_1, side_2}, {side_2, side_3}, {side_3, side_1}}};
```
mgovers, line 34: it's quite dangerous to do this `using enum` declaration here because the name for `side_1` etc. overlap between `Branch3Side` and `ControlSide`. In this case, I understand why you did it like this but it's probably better to scope it with a lambda. The main reason is that in the `tap_position_optimizer::detail` namespace, I would expect `ControlSide` to be the default used enum, not `Branch3Side`

```suggestion
using enum Branch3Side;

constexpr auto const branch3_combinations = []{
    using enum Branch3Side;
    return std::array<std::tuple<Branch3Side, Branch3Side>, 3>{
        {side_1, side_2}, {side_2, side_3}, {side_3, side_1}}};
```

in addition: is there a reason why you moved this outside the location where it is used? is it indeed used multiple times? if not, better to just move back to where it is used

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -21,114 +25,240 @@ namespace detail = power_grid_model::optimizer::detail;
...
+
+    bool operator==(const TrafoGraphEdge& other) const {
+        return regulated_idx == other.regulated_idx && weight == other.weight;
+    } // thanks boost
+
+    auto operator<=>(const TrafoGraphEdge& other) const {
+        if (auto cmp = weight <=> other.weight; cmp != 0) { // NOLINT(modernize-use-nullptr)
+            return cmp;
+        }
+        if (auto cmp = regulated_idx.group <=> other.regulated_idx.group; cmp != 0) { // NOLINT(modernize-use-nullptr)
```
mgovers, line 52: ~this is quite a severe clang-tidy error. might be good to investigate why.~ EDIT: bug in clang-tidy. See e.g. https://github.com/llvm/llvm-project/issues/64159#issuecomment-1653873572

probably good in the future to make a comment either inline or on the PR so the reviewer doesn't need to investigate

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -121,7 +121,7 @@ class InvalidRegulatedObject : public PowerGridError {
 class AutomaticTapCalculationError : public PowerGridError {
   public:
     AutomaticTapCalculationError(ID id) {
-        append_msg("Automatic tap changing regulator is at LV side is not supported. Found at id" +
+        append_msg("Automatic tap changing regulator with tap_side at LV side is not supported. Found at id" +
```
mgovers, line 124: if you change this exception, you also need to change the message in the python error

src/power_grid_model/core/error_handling.py
```diff
@@ -63,6 +64,9 @@
 _INVALID_REGULATED_OBJECT_RE = re.compile(
     r"(\w+) regulator is not supported for object "
 )  # potentially multiple different flavors
+_AUTOMATIC_TAP_CALCULATION_ERROR_RE = re.compile(
+    r"Automatic tap changing regulator is at LV side is not supported. Found at id (-?\d+)\n"
```
mgovers, line 68: this message is out of date. please update

mgovers (commented):
almost there 😬

mgovers (approved):
FINALLYYYYYYY 🎉

PR #566, opened 2024-04-08
Use GitHub hosted runner macos-14 for macOS build

TonyXiang8787:
Use GitHub hosted runner `macos-14`for macOS build. 

It is now out of beta:

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

The CPP test is built in `arm64`.

The Python wheels and tests are built and run in both `x64` (through `Rosetta 2` and `arm64`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -139,7 +139,7 @@ jobs:
           install\${{ env.PRESET }}\bin\power_grid_model_package_test; if(!$?) { Exit $LASTEXITCODE }
 
   build-cpp-test-macos:
-    runs-on: macos-13
+    runs-on: macos-latest
     strategy:
       matrix:
         build-option: [ debug, release ]
@@ -179,7 +179,7 @@ jobs:
             os: ubuntu-latest
             cibw_build: "*_aarch64"
           - platform: macos
-            os: macos-13
+            os: macos-latest
             cibw_build: "*"
           - platform: windows
             os: windows-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -139,7 +139,7 @@ jobs:
           install\${{ env.PRESET }}\bin\power_grid_model_package_test; if(!$?) { Exit $LASTEXITCODE }
 
   build-cpp-test-macos:
-    runs-on: macos-13
+    runs-on: macos-latest
```
TonyXiang8787, line 142: @mgovers, is the Python hack for 3.12 for brew still needed?

.github/workflows/main.yml
```diff
@@ -139,7 +139,7 @@ jobs:
           install\${{ env.PRESET }}\bin\power_grid_model_package_test; if(!$?) { Exit $LASTEXITCODE }
 
   build-cpp-test-macos:
-    runs-on: macos-13
+    runs-on: macos-14
```
TonyXiang8787, line 142: @mgovers, is the Python hack for 3.12 for brew still needed?

TonyXiang8787: I have removed it and it works!

mgovers (approved):

.github/workflows/main.yml
mgovers: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#language-and-runtime shows that it runs python 3.12.2, so that's great!

PR #572, opened 2024-04-16
[Feature] common state queries + transformer concept

mgovers:
Cherry-picked from https://github.com/PowerGridModel/power-grid-model/pull/557

* common state queries (e.g. shorthands for `state.common.template *`)
* `transformer_c` concept
* `using gettable_types` in `Container`
* `using SideType` in `Branch` and `Branch3`

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/branch.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/branch.hpp
@@ -22,6 +22,8 @@ class Branch : public Base {
     using UpdateType = BranchUpdate;
     template <symmetry_tag sym> using OutputType = BranchOutput<sym>;
     using ShortCircuitOutputType = BranchShortCircuitOutput;
+    using SideType = BranchSide;
+
     static constexpr char const* name = "branch";
     ComponentType math_model_type() const final { return ComponentType::branch; }
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/branch3.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/branch3.hpp
@@ -20,6 +20,8 @@ class Branch3 : public Base {
     using UpdateType = Branch3Update;
     template <symmetry_tag sym> using OutputType = Branch3Output<sym>;
     using ShortCircuitOutputType = Branch3ShortCircuitOutput;
+    using SideType = Branch3Side;
+
     static constexpr char const* name = "branch3";
     ComponentType math_model_type() const final { return ComponentType::branch3; }
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/three_winding_transformer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/three_winding_transformer.hpp
@@ -409,4 +409,6 @@ class ThreeWindingTransformer : public Branch3 {
     }
 };
 
+static_assert(transformer_c<ThreeWindingTransformer>);
+
 } // namespace power_grid_model
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
@@ -265,4 +265,6 @@ class Transformer : public Branch {
     }
 };
 
+static_assert(transformer_c<Transformer>);
+
 } // namespace power_grid_model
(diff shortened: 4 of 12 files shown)


Jerry-Jinfeng-Guo (commented):
A nice-to-have function is to verify if certain type of things exist in the container and if so, how many of it. 
I.e., a combination of the previous `size` call and a method to test of Type belongs to the list of types.

mgovers:
> > > A nice-to-have function is to verify if certain type of things exist in the container and if so, how many of it. I.e., a combination of the previous `size` call and a method to test of Type belongs to the list of types.
> > 
> > 
> > what specifically do you desire? is it something like `size_if_known<C>() = size<C>() if C in gettable else 0`? now that the `gettable_types` is exposed, I personally am of the opinion that that is a user thing. In fact, the compiler errors here help with not mistakenly call with the wrong type
> 
> Personally I think the existing `size` call should have a guarding step telling exactly `if C in gettable`. Instead, it now does an assert check on whether the container construction is set to done manually by the code.

the check you mean and the current assertion do two completely different things:

1. `size` returns the size of the container portion containing that `gettable` type.
  * the `gettable` type should really be `gettable`
  * if it is not, it should not compile. it's like calling `std::get<Foo>(std::make_tuple(Bar{}, Baz{}))` or `static_cast<std::vector<Foo>>(std::vector<Bar>())`;
2. the assertion checks the **status** of the container (not the content types)
  * we deliberately only support that the container is fully initialized and filled (but potentially empty).
  * if the container is not initialized, we do not want to call `size` because it doesn't make sense (it's like reading uninitialized memory)
    * in `Debug` mode, we raise an assertion error to make sure we catch issues in our code ASAP. however, this check is relatively expensive compared to the amount of times we call it. Also, it will never be the users' fault, but always our own fault as developers.
    * in production (`Release` mode), we ignore it to prevent user exceptions.
      * this at least hides the issue from the user in most cases
      * if we really mess up, this could result in potentially erroneous results, but we most likely will catch these edge cases in our tests

Jerry-Jinfeng-Guo:
> > > > A nice-to-have function is to verify if certain type of things exist in the container and if so, how many of it. I.e., a combination of the previous `size` call and a method to test of Type belongs to the list of types.
> > > 
> > > 
> > > what specifically do you desire? is it something like `size_if_known<C>() = size<C>() if C in gettable else 0`? now that the `gettable_types` is exposed, I personally am of the opinion that that is a user thing. In fact, the compiler errors here help with not mistakenly call with the wrong type
> > 
> > 
> > Personally I think the existing `size` call should have a guarding step telling exactly `if C in gettable`. Instead, it now does an assert check on whether the container construction is set to done manually by the code.
> 
> the check you mean and the current assertion do two completely different things:
> 
> 1. `size` returns the size of the container portion containing that `gettable` type.
> 
> * the `gettable` type should really be `gettable`
> * if it is not, it should not compile. it's like calling `std::get<Foo>(std::make_tuple(Bar{}, Baz{}))` or `static_cast<std::vector<Foo>>(std::vector<Bar>())`;


No. Code outside could call on whatever type to the `size` call. If the type passed in does not live in the container internal list, this should be checked instead of simply crash. The assumption that an actual `gettable` type is passed in is ungrounded. 

> 2. the assertion checks the **status** of the container (not the content types)
> 
> * we deliberately only support that the container is fully initialized and filled (but potentially empty).
> * if the container is not initialized, we do not want to call `size` because it doesn't make sense (it's like reading uninitialized memory)
>   
>   * in `Debug` mode, we raise an assertion error to make sure we catch issues in our code ASAP. however, this check is relatively expensive compared to the amount of times we call it. Also, it will never be the users' fault, but always our own fault as developers.
>   * in production (`Release` mode), we ignore it to prevent user exceptions.
>     
>     * normally if it goes wrong, it will only result in some useless code. by skipping the check, we at least hide that from the user
>     * if we really mess up, this could result in potentially erroneous results, but we most likely will catch these edge cases in our tests
>       
>       * this will only happen if we either deliberately add incorrect code (should be caught in review), or do stuff **during** initialization of the container (should be caught in tests), or some uncaught exception is thrown and the entire PGM becomes corrupt (should never happen due to [`call_with_catch` in the C API implementation](https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model_c/src/handle.hpp#L31-L45))

This is unfortunately not relevant to my question.

Jerry-Jinfeng-Guo (approved):
After discussion, the changes needed in the `size` call will be done in a separate PR.

mgovers:
conclusion of quick call:

1. a similar state query wrapper for `size` and `empty` is desirable
  * also related: `constexpr` check for `Gettable` type is desirable
  * these changes are out of scope of this PR and can be added when needed
2. `assert(construction_complete)` is indeed unrelated and instead comes from a different design choice

PR #577, opened 2024-04-25
[FEATURE] add support for same-ranked transformers

mgovers:
If you have a number of transformers like the following grid then `Trafo_1` and `Trafo_2` should be assigned equivalent rank.
```
      /-- Trafo_1 -- Trafo_3
Source
      \-- Trafo_2
```

The current `main` after #562 only has support for a linear ranking and therefore assigns order `[1, 2, 3]`

This PR introduces the transformer ranking output as a nested vector, resulting in `[[1, 2], 3]`, such that `1` and `2` are treated equivalently, while `3` is ranked differently because it has higher rank.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
@@ -53,7 +53,7 @@ struct TrafoGraphEdge {
 
 using TrafoGraphEdges = std::vector<std::pair<TrafoGraphIdx, TrafoGraphIdx>>;
 using TrafoGraphEdgeProperties = std::vector<TrafoGraphEdge>;
-using RankedTransformerGroups = std::vector<Idx2D>;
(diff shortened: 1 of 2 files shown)


nitbharambe (approved):

Jerry-Jinfeng-Guo:
> If you have a number of transformers like the following grid then `Trafo_1` and `Trafo_2` should be assigned equivalent rank.
> 
> ```
>       /-- Trafo_1 -- Trafo_3
> Source
>       \-- Trafo_2
> ```
> 
> The current `main` after #562 only has support for a linear ranking and therefore assigns order `[1, 2, 3]`
> 
> This PR introduces the transformer ranking output as a nested vector, resulting in `[[1, 2], 3]`, such that `1` and `2` are treated equivalently, while `3` is ranked differently because it has higher rank.

How was this true in the first place? The transformer ranking returned vector of vectors, which was originally `[[1,2],3]`

Jerry-Jinfeng-Guo:
OK, it became clear that the original design was accidentally changed to a simple list. The change in this PR is indeed a necessary revert.

PR #578, opened 2024-04-25
Feature/container + main core improvements

mgovers:
More cherry-picking from #557 

@nitbharambe ~either before or after merge of this branch, the size checks from #562 can be removed~ EDIT: [8bf6a0b](https://github.com/PowerGridModel/power-grid-model/pull/562/commits/8bf6a0bd71cdc6a037421e5aa0af3bcd65df100d) already removed the `size` checks for source. can you please have a look to verify that all of them have been removed?

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -131,7 +131,7 @@ template <symmetry_tag sym_type> struct TransformerTapRegulatorMathOutput {
     IntS tap_pos{na_IntS};
 };
 
-// from, to side
+// from side, to side
 // in case of indices for math model, -1 means the branch is not connected to that side
 using BranchIdx = std::array<Idx, 2>;
 // node 0, 1, 2 side
@@ -154,7 +154,6 @@ struct MathModelTopology {
     DenseGroupedIdxVector power_sensors_per_branch_from;
     DenseGroupedIdxVector power_sensors_per_branch_to;
     DenseGroupedIdxVector power_sensors_per_bus;
-    DenseGroupedIdxVector transformer_tap_regulators_per_branch;
 
     Idx n_bus() const { return static_cast<Idx>(phase_shift.size()); }
 
@@ -179,8 +178,6 @@ struct MathModelTopology {
     Idx n_branch_to_power_sensor() const { return power_sensors_per_branch_to.element_size(); }
 
     Idx n_bus_power_sensor() const { return power_sensors_per_bus.element_size(); }
-
-    Idx n_transformer_tap_regulator() const { return transformer_tap_regulators_per_branch.element_size(); }
 };
 
 template <symmetry_tag sym_type> struct MathModelParam {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -35,13 +35,6 @@ template <symmetry_tag sym_type> struct BranchCalcParam {
     ComplexTensor<sym> const& ytt() const { return value[3]; }
 };
 
-struct TransformerTapRegulatorCalcParam {
-    double u_set{};
-    double u_band{};
-    DoubleComplex z_compensation{};
-    IntS status{};
-};
-
 template <symmetry_tag sym_type> struct BranchMathOutput {
     using sym = sym_type;
 
@@ -117,13 +110,26 @@ template <symmetry_tag sym_type> struct PowerSensorCalcParam {
 
 template <typename T>
 concept sensor_calc_param_type =
-    std::same_as<T, VoltageSensorCalcParam<symmetric_t>> || std::same_as<T, VoltageSensorCalcParam<asymmetric_t>> ||
-    std::same_as<T, PowerSensorCalcParam<symmetric_t>> || std::same_as<T, PowerSensorCalcParam<asymmetric_t>>;
+    std::derived_from<T, VoltageSensorCalcParam<symmetric_t>> ||
+    std::derived_from<T, VoltageSensorCalcParam<asymmetric_t>> ||
+    std::derived_from<T, PowerSensorCalcParam<symmetric_t>> || std::derived_from<T, PowerSensorCalcParam<asymmetric_t>>;
 
 static_assert(sensor_calc_param_type<VoltageSensorCalcParam<symmetric_t>>);
 static_assert(sensor_calc_param_type<VoltageSensorCalcParam<asymmetric_t>>);
 static_assert(sensor_calc_param_type<PowerSensorCalcParam<symmetric_t>>);
 static_assert(sensor_calc_param_type<PowerSensorCalcParam<asymmetric_t>>);
+
+struct TransformerTapRegulatorCalcParam {
+    double u_set{};
+    double u_band{};
+    DoubleComplex z_compensation{};
+    IntS status{};
+};
+template <symmetry_tag sym_type> struct TransformerTapRegulatorMathOutput {
+    using sym = sym_type;
+
+    IntS tap_pos{na_IntS};
+};
 
 // from, to side
 // in case of indices for math model, -1 means the branch is not connected to that side
@@ -148,6 +154,7 @@ struct MathModelTopology {
     DenseGroupedIdxVector power_sensors_per_branch_from;
     DenseGroupedIdxVector power_sensors_per_branch_to;
     DenseGroupedIdxVector power_sensors_per_bus;
+    DenseGroupedIdxVector transformer_tap_regulators_per_branch;
 
     Idx n_bus() const { return static_cast<Idx>(phase_shift.size()); }
 
@@ -172,6 +179,8 @@ struct MathModelTopology {
     Idx n_branch_to_power_sensor() const { return power_sensors_per_branch_to.element_size(); }
 
     Idx n_bus_power_sensor() const { return power_sensors_per_bus.element_size(); }
+
+    Idx n_transformer_tap_regulator() const { return transformer_tap_regulators_per_branch.element_size(); }
 };
 
 template <symmetry_tag sym_type> struct MathModelParam {
@@ -221,7 +230,7 @@ struct ShortCircuitInput {
 
 template <typename T>
 concept symmetric_calculation_input_type =
-    std::same_as<T, PowerFlowInput<symmetric_t>> || std::same_as<T, StateEstimationInput<symmetric_t>>;
+    std::derived_from<T, PowerFlowInput<symmetric_t>> || std::derived_from<T, StateEstimationInput<symmetric_t>>;
 
 static_assert(symmetric_calculation_input_type<PowerFlowInput<symmetric_t>>);
 static_assert(symmetric_calculation_input_type<StateEstimationInput<symmetric_t>>);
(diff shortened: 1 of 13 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/component/component.hpp
```diff
@@ -0,0 +1,39 @@
...
+    }
+};
+
+template <typename T>
+concept component_c = requires(T t, T const& ct, typename T::UpdateType u, typename T::UpdateType const& cu) {
+                          typename T::InputType;
+                          typename T::UpdateType;
+
+                          { T::name } -> std::convertible_to<std::string_view>;
+                          { ct.math_model_type() } -> std::convertible_to<ComponentType>;
```
nitbharambe, line 31: In the state queries, I see
```
template <typename ComponentType, class ComponentContainer>
    requires model_component_state_c<MainModelState, ComponentContainer, ComponentType>
inline Idx get_component_type_index(MainModelState<ComponentContainer> const& state) {
    return state.components.template get_type_idx<ComponentType>();
}
```
and also in main core:
```
template <template <typename T> class StateType, typename ContainerType, typename ComponentType>
concept model_component_state_c =
    component_container_c<typename StateType<ContainerType>::ComponentContainer, ComponentType> &&
    std::same_as<StateType<ContainerType>, MainModelState<ContainerType>>;
```

Is it intentional for the type name and enum to be the same?

mgovers: not intentional but also not harmful because only one of the two is used within the template functions

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -117,14 +110,27 @@ template <symmetry_tag sym_type> struct PowerSensorCalcParam {
...
+    DoubleComplex z_compensation{};
+    IntS status{};
+};
+template <symmetry_tag sym_type> struct TransformerTapRegulatorMathOutput {
+    using sym = sym_type;
+
+    IntS tap_pos{na_IntS};
+};
+
 // from, to side
```
Jerry-Jinfeng-Guo, line 134:
```suggestion
// from side, to side
```

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -148,6 +154,7 @@ struct MathModelTopology {
     DenseGroupedIdxVector power_sensors_per_branch_from;
     DenseGroupedIdxVector power_sensors_per_branch_to;
     DenseGroupedIdxVector power_sensors_per_bus;
+    DenseGroupedIdxVector transformer_tap_regulators_per_branch;
```
Jerry-Jinfeng-Guo, line 157: `tap_regulator` is only present in `transformer` objects, hence redundant:
```suggestion
    DenseGroupedIdxVector tap_regulators_per_branch;
```

power_grid_model_c/power_grid_model/include/power_grid_model/component/component.hpp
```diff
@@ -0,0 +1,39 @@
+// SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+//
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
```
Jerry-Jinfeng-Guo, line 5: what, we are already doing this?

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/math_state.hpp
```diff
@@ -22,4 +22,43 @@ inline void clear(MathState& math_state) {
     math_state.y_bus_vec_asym.clear();
 }
 
+template <symmetry_tag sym>
+inline void update_y_bus(MathState& math_state, std::vector<MathModelParam<sym>> const& math_model_params) {
+    auto& y_bus_vec = [&math_state]() -> auto& {
```
Jerry-Jinfeng-Guo, line 27: Is this recommended? `auto` `-> auto`?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -94,44 +104,13 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
...
-            std::make_shared<MathModelParam<sym> const>(std::move(math_model_params[i])),
-            math_model_param_increments[i]);
-    }
+template <component_c Component, class ComponentContainer, std::forward_iterator ForwardIterator,
+          std::output_iterator<typename Component::UpdateType> OutputIterator>
+    requires model_component_state_c<MainModelState, ComponentContainer, Component>
+inline void update_inverse(MainModelState<ComponentContainer> const& state, ForwardIterator begin, ForwardIterator end,
+                           OutputIterator destination) {
+    return update_inverse<Component>(state, begin, end, destination,
+                                     get_component_sequence<Component>(state, begin, end));
```
Jerry-Jinfeng-Guo, line 113: Does the length justify making it a (-n inline) function?

Jerry-Jinfeng-Guo (commented):
Few comments

power_grid_model_c/power_grid_model/include/power_grid_model/component/component.hpp
mgovers: yes since #502 (last february)

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: this has been in there for 2 years. resolved now 😬

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: whoops should be out of scope of this PR

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/math_state.hpp
mgovers: the two `auto`s mean different things and both are needed.

* the second one declares that the lambda returns a reference
* the first one captures the result of the lambda by reference

The lambda allows for side-effects. The construct used here is:

```cpp
struct Foo;
struct Bar { Foo foo; };

Bar bar;
auto& foo_reference = [&bar]() -> auto& { return bar.foo; }();
  ^                      ^   ^      ^               ^       ^
  0                      1   2      3               4       5
```

which has the same result as first declaring only the function and then calling it, but with the additional advantage that the function is not stored, so the compiler will `inline` and maybe even resolve `constexpr`.

0. declares that `foo_reference` is reference (references cannot be forward-declared, so always needs to be obtained from a lambda in complex cases like this)
1. captures `bar` by reference
2. needed because of trailing return type (only way for lambdas to specify return types)
3. declare that return type is a reference (this is the only way to do that for lambdas without using `return *&result` construct, which may be illegal if the captured result is `const`)
4. return the result type
5. immediately call the function so that it doesn't need to be stored in a variable

hope that explains it

Jerry-Jinfeng-Guo: Clearly a good read but not what I was asking ;)
My question was since you already know `Foo` as return type, why still write `auto` when the type couldn't be inferred by mortal human brains with simple reading?

Jerry-Jinfeng-Guo: For the ease of discussion:

> ```c++
> struct Foo;
> struct Bar { Foo foo; };
> 
> Bar bar;
> auto& foo_reference = [&bar]() -> auto& { return bar.foo; }();
>   ^                      ^   ^      ^               ^       ^
>   0                      1   2      3               4       5
> ```

```c++
struct Foo;
struct Bar { Foo ofo; };

Bar bar;
auto& foo_reference = [&bar]() -> Foo& { return bar.ofo; }();
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: length is independent of `inline`. every function has to have one of the following to correctly resolve the linking

* `extern` (maybe forward-declared)
  * this tells the compiler that this function shall be exported by the binaries.
  * `int main(argc*, argv**)` xor `int main()` are implied to be `extern`
* declared `constexpr`
  * implies `inline`
  * has the additional directive that the compiler may use the function at compile-time whenever possible.
  * this may also help with compilation speed because the code may be partially compiled even before inlining
* inlined using `inline`
  * this means that the compiler can include the code where it is used and doesn't need to link it
* be unique inside a translation unit
  * this is a little bit hard to explain but basically it amounts to e.g. a cpp file inside a nameless namespace via `namespace { /* ... */ }`)
* inside a `class` or `struct`
  * this implies `inline` if you also create the body inside the class declaration itself
  * if you do forward declare the function instead, it is added to the namespace lookup, which the linker can use at a later stage. The most common use case for this is object oriented libraries that are not header-only
    * the `typedef` definition (implied by `struct` and `class` in C++) makes sure that the function body is uniquely defined in namespace lookup
* be unique and never ever included from another location
  * different compilers may have different linking behaviour if you don't specify functions inside a nameless namespace, so it is bad practice to depend on this.
  * `doctest` abuses this possibility to register the tests to the executable, which is why test cases cannot be declared in nameless namespaces

if you do not satisfy the above (e.g. by forgetting a `constexpr` or `inline` function) and effectively end up declaring the the same function name and signature multiple times (e.g. by including the same header in 2 or more other headers) then you will get compiler errors or - even worse and more difficult to debug - linking errors

Jerry-Jinfeng-Guo: Yet another nice read but not exactly what I was asking.
I think I get the answer to my question, which was the whole point of providing such overloads.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/math_state.hpp
mgovers: cause I don't know what the return type is in the case of complex operations. In fact, the return type depends on the template parameters (here: `std::vector<YBus<sym>>&`). In this case, I agree that we could still add it, but it also clutters the code with implementation details.

In the more general case, the return type might actually be more complex and not explicitly depend on the template parameters like `Transformer&` vs `ThreeWindingTransformer&`. In that case, the only alternative to provide a trailing return type other than `auto&` would be using `std::conditional_t`, which is already verbose in the case of 2 supported template parameters - let alone multiple

Jerry-Jinfeng-Guo: Pending approval upon resolution of this conversation.

mgovers: another advantage of using `auto&` is that the caller very often really doesn't care which type it is, as long as it compiles. the best you can get is slightly better type hinting, but only marginally because it involves - once again - templates

Jerry-Jinfeng-Guo: `auto` it is then

Jerry-Jinfeng-Guo (approved):

PR #579, opened 2024-04-29
Use NaN as default value in input/output/update struct

### LINKED ISSUE #568: [FEATURE] initialize input/update/output data to nan
Input, update and output data needs to initialize as nan.
To do this, update the code generation functions + rerun.

Some unit tests depend on the fact that the input/update/output data defaulted to zero - these have to be explicitly set to zero.

TonyXiang8787:
Closes #568 

@mgovers the number of places which we need explicit zeros are not as much as I expected.


code_generation/code_gen.py
```diff
@@ -16,6 +16,17 @@
...
 
+def _data_type_nan(data_type: str):
+    if data_type == "ID":
+        return "na_IntID"
+    elif data_type == "double" or "RealValue" in data_type:
+        return "nan"
+    elif data_type == "IntS":
+        return "na_IntS"
+    else:
+        return f"static_cast<{data_type}>(na_IntS)"
```
mgovers, line 27: right... enums...

power_grid_model_c/power_grid_model/include/power_grid_model/component/branch.hpp
```diff
@@ -144,13 +144,21 @@ class Branch : public Base {
     }
 
     template <symmetry_tag sym> BranchOutput<sym> get_null_output() const {
-        BranchOutput<sym> output{};
+        BranchOutput<sym> output{.loading = {},
```
mgovers, line 147: i think that some compilers will complain that there are missing initializers on this.

power_grid_model_c/power_grid_model/include/power_grid_model/component/node.hpp
```diff
@@ -57,13 +57,13 @@ class Node final : public Base {
         return get_sc_output(uabc_pu);
     }
     template <symmetry_tag sym> NodeOutput<sym> get_null_output() const {
-        NodeOutput<sym> output{};
+        NodeOutput<sym> output{.u_pu = {}, .u = {}, .u_angle = {}, .p = {}, .q = {}};
         static_cast<BaseOutput&>(output) = base_output(false);
         return output;
```
mgovers, line 62: it might be better to do something like
```cpp
auto output = base_output<NodeOutput<sym>>(false);
output.u_pu = {};
output.u = {};
output.u_angle = {};
output.p = {};
output.q = {};
```

mgovers (approved):
if the compilers are OK with it, so am I. great change!

power_grid_model_c/power_grid_model/include/power_grid_model/component/branch.hpp
TonyXiang8787: No. If the struct has all default values defined, compiler will not complain.

power_grid_model_c/power_grid_model/include/power_grid_model/component/node.hpp
TonyXiang8787: @mgovers can you make this change?

mgovers: the main reason for this comment was that i thought that the compiler might complain. since it doesn't, i'm OK with the current implementation.

whichever you prefer

TonyXiang8787: Please approve the PR.

TonyXiang8787: It's on merge queue

PR #583, opened 2024-05-01
Renaming `MathOutput` to `SolverOutput`

Jerry-Jinfeng-Guo:
Step 1 of the MathOutput transformation. Necessary for tap changer.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
@@ -329,15 +329,15 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         is_asym_parameter_up_to_date_ = is_asym_parameter_up_to_date_ && !changes.topo && !changes.param;
     }
 
-    template <math_output_type MathOutputType, typename MathSolverType, typename YBus, typename InputType,
+    template <solver_output_type SolverOutputType, typename MathSolverType, typename YBus, typename InputType,
               typename PrepareInputFn, typename SolveFn>
         requires std::invocable<std::remove_cvref_t<PrepareInputFn>, Idx /*n_math_solvers*/> &&
                  std::invocable<std::remove_cvref_t<SolveFn>, MathSolverType&, YBus const&, InputType const&> &&
                  std::same_as<std::invoke_result_t<PrepareInputFn, Idx /*n_math_solvers*/>, std::vector<InputType>> &&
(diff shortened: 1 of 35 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -329,15 +329,15 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         is_asym_parameter_up_to_date_ = is_asym_parameter_up_to_date_ && !changes.topo && !changes.param;
     }
 
-    template <math_output_type MathOutputType, typename MathSolverType, typename YBus, typename InputType,
+    template <solver_output_type SolverOutputType, typename MathSolverType, typename YBus, typename InputType,
```
mgovers, line 332: most of the objects in this file are actually `MathOutput` types (in the sense that it should be equal to the output after the optimizer is done with it).

in particular, anywhere where it doesn't matter whether or not it is actually a `std::vector` of `SolverOutput`s is a location where it should probably remain `MathOutput`

It might be a good idea to declare a `template <symmetry_tag sym> using MathOutput = std::vector<SolverOutput<sym>>;` in the `calculation_parameters.hpp`.

tests/cpp_integration_tests/test_main_model.cpp
```diff
@@ -238,10 +238,10 @@ TEST_CASE("Test main model - power flow") {
     }
 
     SUBCASE("Test calculate power flow") {
-        auto const math_output = main_model.calculate_power_flow<symmetric_t>(1e-8, 20, CalculationMethod::linear);
-        main_model.output_result<Node>(math_output, state.sym_node.begin());
-        main_model.output_result<Branch>(math_output, state.sym_branch.begin());
-        main_model.output_result<Appliance>(math_output, state.sym_appliance.begin());
+        auto const solver_output = main_model.calculate_power_flow<symmetric_t>(1e-8, 20, CalculationMethod::linear);
```
mgovers, line 241: this entire file: idem to my comment on the `main_model.hpp`

mgovers (commented):
it might be a good idea to squash all commits after the final review is done but before adding it to the merge queue. this will filter out the commits that do a partial revert (particularly on `main_model.hpp` and `test_main_model_*.cpp`

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
Jerry-Jinfeng-Guo: I believe this should be part of the next PR?

mgovers (approved):

PR #586, opened 2024-05-06
Feature/tap optimize voltage level tests

mgovers:


power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer_tap_regulator.hpp
```diff
@@ -35,10 +35,10 @@ class TransformerTapRegulator : public Regulator {
     UpdateChange update(TransformerTapRegulatorUpdate const& update_data) {
         assert(update_data.id == id());
         set_status(update_data.status);
-        u_set_ = update_data.u_set;
-        u_band_ = update_data.u_band;
-        line_drop_compensation_r_ = update_data.line_drop_compensation_r;
-        line_drop_compensation_x_ = update_data.line_drop_compensation_x;
+        update_real_value<symmetric_t>(update_data.u_set, u_set_, 1.0);
```
nitbharambe, line 38: Why was this needed? if update_data.u_set is nan, then it would have been replaced by nan anyway

mgovers: yes but if the original value is not `nan`, then the original code will make it `nan` and invalidate the input data. that is not according to our current contract with the users

mgovers:
remaining TODOs will be fixed in the near future

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -721,6 +748,72 @@ TEST_CASE("Test Tap position optimizer") {
...
+                state_b.tap_max = 5;
+                state_b.tap_pos = 3;
+
+                SUBCASE("unique value in band") {
+                    u_band = 0.01;
+                    check_b = check_exact(3);
+                }
+                SUBCASE("large compact band") {
+                    u_band = 1.01;
+                    check_b = check_exact_per_strategy(3, 5, 1);
```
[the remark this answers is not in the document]
mgovers: `tap_pos_min` is the expected tap position for `min` strategy. i unfortunately don't have a better name... feel free to help me out here

nitbharambe (approved):

PR #588, opened 2024-05-07
Feature/line drop compensation tests

mgovers:


tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -808,12 +807,57 @@ TEST_CASE("Test Tap position optimizer") {
...
+                check_b = test::check_exact_per_strategy(3, 5, 3);
+            }
+            SUBCASE("positive reactance") {
+                line_drop_compensation_x = 0.125;
+                check_b = test::check_exact_per_strategy(3, 5, 2);
+            }
+            SUBCASE("negative reactance") {
+                line_drop_compensation_x = -0.5;
+                check_b = test::check_exact_per_strategy(3, 5, 3);
+            }
```
mgovers, line 852: i could also add tests like what happens if i switch off `i_pu` or what happens if i make `i_pu` a purely real or purely imaginary function. thoughts?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -557,6 +557,11 @@ auto check_exact_per_strategy(IntS tap_pos_any, IntS tap_pos_min, IntS tap_pos_m
         }
     };
 }
+auto normalized_lerp(IntS value, IntS start, IntS stop) {
+    REQUIRE(start != stop);
+    return (static_cast<double>(value) - static_cast<double>(start)) /
+           (static_cast<double>(stop) - static_cast<double>(start));
```
Jerry-Jinfeng-Guo, line 563: `stop-start` sounds like cacheable ;)
Also would any check be necessary? `stop > start` is assumed? `value` falls in range assumed?

mgovers: if `stop < start`, the logic remains valid due to cancelling minus signs. they just cannot be equal to eachother

Jerry-Jinfeng-Guo: > if `stop < start`, the logic remains valid due to cancelling minus signs. they just cannot be equal to eachother.

Mathematically yes, but semantically? I was more asking about whether both could well be the case.

> if `value \nin [start, stop]` it's `linear extrapolation` instead of `linear interpolation` which is still the same logic

Interesting that this function does more that what it should ;)

Jerry-Jinfeng-Guo: > ... whether both could well be the case.
[should be read]
Whether `start < stop` and `start> stop` could be the case in reality and that they both mean valid configurations.

mgovers: > > ... whether both could well be the case.
> 
> [should be read] Whether `start < stop` and `start> stop` could both be the case in reality and that they both mean valid configurations.

yes, definitely. `tap_min < tap_max` and `tap_min > tap_max` are definitely both valid configurations.

> > if `value \nin [start, stop]` it's `linear extrapolation` instead of `linear interpolation` which is still the same logic
> 
> Interesting that this function does more that what it should ;)

well, `std::lerp` is a standard function for a reason 😬 the problem is that it doesn't serve our needs in this particular case because we need it normalized here (otherwise, the logic becomes explicitly dependent on `tap_min` and `tap_max` which is what we want to avoid in the first place)

mgovers:
failing build is a connection error of the build host. will restart when other jobs finish

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -809,12 +808,57 @@ TEST_CASE("Test Tap position optimizer") {
...
+                check_b = test::check_exact_per_strategy(3, 5, 2);
+            }
+            SUBCASE("negative reactance") {
+                line_drop_compensation_x = -0.5;
+                check_b = test::check_exact_per_strategy(3, 5, 3);
+            }
+
+            regulator_b.update({.id = 4,
+                                .u_set = u_set,
+                                .u_band = u_band,
```
nitbharambe, line 857: This is a bit confusing as we are dividing incorrect units. Also why should `update` be at bottom and not before SUBCASE?

mgovers: will change to one single `update_data` instance.

`update` has to come after the `SUBCASE` because it passes by value and not by reference. If something changes in the `SUBCASE`, it will not not end up in the actual component

nitbharambe: Also, my vscode app commented at wrong line. Dividing line_drop_compensation_r by base_power_3p seems odd. Why not have line_drop_compensation_r = 5e-8 directly?

mgovers: in my example, everything is `pu`, except for  `line_drop_compensation_*`. i want to make that explicit

nitbharambe (approved):

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: follow-up if necessary in #595

PR #589, opened 2024-05-07
add retry logic when tap optimizer diverges

### LINKED ISSUE #580: [Feature] Retry logic in automatic tap outer loop
Originally posted in #557 as https://github.com/PowerGridModel/power-grid-model/pull/557#discussion_r1584309983

```
try 
  math_result = math_solver.power_flow(calculation_method=calculation_method) 
catch (SparseMatrixError, IterationDiverge) 
  Run run_pf_with_regulation_per_solver(math_solver=math_solver, calculation_method=linear),   discard return 
  math_result = math_solver.power_flow(calculation_method=calculation_method)
```

I added a Jira ticket for that because I have reason to believe that #549 should solve this because of the discard step: the initial state is actually set to the solution of the `linear` calculation.

That leaves me with even more reason that all the other fallback logic is also unnecessary and we can just remove `calculate_with_retry` and, instead, directly call `calculate_`

FYI: I removed `InvalidPFResult` from this example due to below comments

mgovers:
fixes #580

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
@@ -632,12 +632,27 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
                   CalculationMethod method) const -> ResultType {
         initialize(regulator_order);
 
-        if (auto result = iterate(state, regulator_order, method); strategy_ == OptimizerStrategy::any) {
+        if (auto result = iterate_with_fallback(state, regulator_order, method); strategy_ == OptimizerStrategy::any) {
             return result;
         }
 
         // refine solution
         exploit_neighborhood(regulator_order);
+        return iterate_with_fallback(state, regulator_order, method);
+    }
+
+    auto iterate_with_fallback(State const& state,
+                               std::vector<std::vector<RegulatedTransformer>> const& regulator_order,
+                               CalculationMethod method) const -> ResultType {
+        try {
+            return iterate(state, regulator_order, method);
+        } catch (IterationDiverge const& /* ex */) {
+            // fallback
+        } catch (SparseMatrixError const& /* ex */) {
+            // fallback
+        }
+
+        std::ignore = iterate(state, regulator_order, CalculationMethod::linear);
         return iterate(state, regulator_order, method);
     }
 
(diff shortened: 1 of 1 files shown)


mgovers:
it's not really possible to black-box unit test this. it **is** possible to just create a mock state calculator but that's white-box testing, which we do not intend to do here. this is something for a validation test

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -632,13 +632,30 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
-        return iterate(state, regulator_order, method);
+        return iterate_with_fallback(state, regulator_order, method);
+    }
+
+    auto iterate_with_fallback(State const& state,
+                               std::vector<std::vector<RegulatedTransformer>> const& regulator_order,
+                               CalculationMethod method) const -> ResultType {
+        auto fallback = [this, &state, &regulator_order, &method] {
+            std::ignore = iterate(state, regulator_order, CalculationMethod::linear);
+            return iterate(state, regulator_order, method);
```
Jerry-Jinfeng-Guo, line 649: How does ignoring one single run from the `::linear` method help prevent `method` from throwing same errors?

Jerry-Jinfeng-Guo (commented):
Not yet 100% convinced that this would work

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: most often, the `SparseMatrixError` or `IterationDiverge` will come from unreasonable combinations of tap positions. the linear method will already make the tap positions more closely representing the actual positions. from there, the more precise iterative methods will refine the solution and hopefully end up in a valid final state.

that said, if the errors were thrown because of another reason (e.g. missing data) then the iterative methods will not be able to refine the solution and still throw the error. In that case, the user needs to investigate because it is most likely a problem on their end.

Jerry-Jinfeng-Guo: I understand that this might help narrow stuff down (a bit) (hopefully). 
I hate to break it but the whole logic is wishful thinking. I mean absolutely no offense.  
We can of course include this in the first implementation but I really do not like the concept given context. Let's totally make this a discussion point for later improvement.

nitbharambe (approved):

mgovers:
@Jerry-Jinfeng-Guo i propose merging this because the discussions on #594 and #595 may be delayed

mgovers:
> @Jerry-Jinfeng-Guo i propose merging this because the discussions on #594 and #595 may be delayed

merging cfr outcome of discussion offline

PR #590, opened 2024-05-07
automatic tap changer outer loop multiple transformer test

nitbharambe:
Multiple transformers whose control voltage is dependent on their ranking.

### DIFF
--- a/tests/cpp_unit_tests/test_tap_position_optimizer.cpp
+++ b/tests/cpp_unit_tests/test_tap_position_optimizer.cpp
@@ -826,7 +826,7 @@ TEST_CASE("Test Tap position optimizer") {
             state_a.u_pu = [&state_a, &regulator_a](ControlSide side) {
                 CHECK(side == regulator_a.control_side());
 
-                // u_2a = f(ta) for rank 0
+                // u_2a = f(tap_pos_a) for rank 0
                 // u_2a = (u_1a * n_1) / (1.0 + relative_tap_pos_a)
                 // consider u_1a = n_1 = 1.0
                 // For a tap_size of 0.1 and tap_nom of 0, tap_pos_relative_a = 0.1 * (tap_pos_a - 0)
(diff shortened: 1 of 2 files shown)


tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -814,6 +815,104 @@ TEST_CASE("Test Tap position optimizer") {
...
+                // For a tap_size of 0.1 and tap_nom of 0, tap_pos_relative_b = 0.1 * (tap_pos_b - 0)
+                auto const relative_tap_b = static_cast<double>(state_b.tap_pos) * 0.1;
+                return state_a.u_pu(regulator_a.control_side()) / (1.0 + relative_tap_b);
+            };
+
+            SUBCASE("Situation 1") {
+                regulator_a.update({.id = 3, .u_set = 1.25, .u_band = 0.01});
+                regulator_b.update({.id = 4, .u_set = 1.13636, .u_band = 0.01});
+                check_a = check_exact(-2);
+                check_b = check_exact(1);
```
Jerry-Jinfeng-Guo, line 852: I believe these two lines returned two functors, but they are not yet run/evaluated. Therefore nothing is actually checked. The same goes for basically all of the `check_a` and `check_b`?

Jerry-Jinfeng-Guo: I think the checks should be something like:
```suggestion
                
                check_a = check_exact(pos_expected);
                check_a(pos_actual);
```

nitbharambe: Yes, the evaluation happens below, after all subcases end. The double for loop below runs for each subcase

nitbharambe: Your 2nd comment didnt load for me previously. :D
`check_a = check_exact(pos_expected);` is  `check_a = check_exact(-2);`
and `check_a(pos_actual);` is  `check_a(get_state_tap_pos(result, state_a.id), strategy);`

Jerry-Jinfeng-Guo: I still do not understand after going through the code at the bottom. `check_exact` and the other one returned a `std::function<IntS, OptStrategy>` functor. If anything, inside the nested `for` loop it will only be equal to the last occurrance of assignment? I think I am missing some fundamental testing knowledge here ;)

Jerry-Jinfeng-Guo: OK, after a short call I got what doctest is messing here.

Jerry-Jinfeng-Guo (approved):

PR #592, opened 2024-05-08
fix performance regression

mgovers:
Fix performance regression after https://github.com/PowerGridModel/power-grid-model/pull/585

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
@@ -665,15 +665,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
 
     template <symmetry_tag sym>
     auto calculate_power_flow(double err_tol, Idx max_iter, CalculationMethod calculation_method) {
-        auto const result_pf =
+        auto result_pf =
             optimizer::get_optimizer<MainModelState, ConstDataset>(
                 OptimizerType::no_optimization, OptimizerStrategy::any, calculate_power_flow_<sym>(err_tol, max_iter),
                 [this](ConstDataset update_data) { this->update_component<permanent_update_t>(update_data); })
                 ->optimize(state_, calculation_method);
-        if constexpr (std::is_same<decltype(result_pf), OptimizerOutput>::value) {
-            return MathOutput<SolverOutput<sym>>{.solver_output = {}, .optimizer_output = result_pf};
-        }
-        return MathOutput<SolverOutput<sym>>{.solver_output = result_pf, .optimizer_output = {}};
+        return MathOutput<SolverOutput<sym>>{.solver_output = std::move(result_pf), .optimizer_output = {}};
     }
 
     // Single load flow calculation, propagating the results to result_data


power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -665,15 +665,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
 
     template <symmetry_tag sym>
     auto calculate_power_flow(double err_tol, Idx max_iter, CalculationMethod calculation_method) {
-        auto const result_pf =
+        auto result_pf =
```
mgovers, line 668: this `const` prevents moving and therefore causes a full copy of the entire powerflow output

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -665,15 +665,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
 
     template <symmetry_tag sym>
     auto calculate_power_flow(double err_tol, Idx max_iter, CalculationMethod calculation_method) {
-        auto const result_pf =
```
Jerry-Jinfeng-Guo, line 668: Why?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -665,15 +665,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+        auto result_pf =
             optimizer::get_optimizer<MainModelState, ConstDataset>(
                 OptimizerType::no_optimization, OptimizerStrategy::any, calculate_power_flow_<sym>(err_tol, max_iter),
                 [this](ConstDataset update_data) { this->update_component<permanent_update_t>(update_data); })
                 ->optimize(state_, calculation_method);
-        if constexpr (std::is_same<decltype(result_pf), OptimizerOutput>::value) {
-            return MathOutput<SolverOutput<sym>>{.solver_output = {}, .optimizer_output = result_pf};
-        }
-        return MathOutput<SolverOutput<sym>>{.solver_output = result_pf, .optimizer_output = {}};
+        return MathOutput<SolverOutput<sym>>{.solver_output = std::move(result_pf), .optimizer_output = {}};
```
Jerry-Jinfeng-Guo, line 673: Why 2.0?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
Jerry-Jinfeng-Guo: OK, found comment above.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: either the output should be provided by r-value or it needs to be explicitly moved.

The former means it's a "right-hand-side" value, e.g. the result returned by another function. That was the case before #585 .

The latter is what is added here because you preferred making a separate variable, which makes it an l-value, meaning it can be used on the left-hand side of an `=` sign. l-values always need to be moved explicitly.

Jerry-Jinfeng-Guo: And assign `result_pf` to `solver` leaving `opt` empty will be later adjusted then?

mgovers: yes. that's the responsibility of the one implementing auto tap optimizer in the main model

Jerry-Jinfeng-Guo (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: let's first of all make sure we don't have any regressions

PR #597, opened 2024-05-14
[Feature] Automatic Tap Changer C API

mgovers:
* Add automatic tap position as experimental feature to C API
* Add `PGM_TapChangingStrategy` enum
* Add `PGM_set_tap_changing_strategy` option setter

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
@@ -33,10 +33,12 @@ class InvalidArguments : public PowerGridError {
         append_msg(method + " is not implemented for " + arguments + "!\n");
     }
 
-    template <std::same_as<TypeValuePair>... Options>
+    template <class... Options>
+        requires(std::same_as<std::remove_cvref_t<Options>, TypeValuePair> && ...)
     InvalidArguments(std::string const& method, Options... options)
         : InvalidArguments{method, "the following combination of options"} {
-        (append_msg(" " + options.name + ": " + options.value + "\n"), ...);
+        (append_msg(" " + std::forward<Options>(options).name + ": " + std::forward<Options>(options).value + "\n"),
+         ...);
     }
 };
 
@@ -213,7 +215,6 @@ class DatasetError : public PowerGridError {
 };
 
 class ExperimentalFeature : public InvalidArguments {
-  public:
     using InvalidArguments::InvalidArguments;
 };
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
@@ -33,10 +33,12 @@ class InvalidArguments : public PowerGridError {
         append_msg(method + " is not implemented for " + arguments + "!\n");
     }
 
-    template <std::same_as<TypeValuePair>... Options>
-    InvalidArguments(std::string const& method, Options const&... options)
+    template <class... Options>
+        requires(std::same_as<std::remove_cvref_t<Options>, TypeValuePair> && ...)
+    InvalidArguments(std::string const& method, Options... options)
         : InvalidArguments{method, "the following combination of options"} {
-        (append_msg(" " + options.name + ": " + options.value + "\n"), ...);
+        (append_msg(" " + std::forward<Options>(options).name + ": " + std::forward<Options>(options).value + "\n"),
+         ...);
     }
 };
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
@@ -193,6 +193,22 @@ class DatasetError : public PowerGridError {
     explicit DatasetError(std::string const& msg) { append_msg("Dataset error: " + msg); }
 };
 
+class ExperimentalFeature : public PowerGridError {
+  public:
+    struct TypeValuePair {
+        std::string_view name;
+        std::string_view value;
+    };
+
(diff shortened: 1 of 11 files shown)


power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -79,6 +79,11 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
         auto const calculation_method = static_cast<CalculationMethod>(opt->calculation_method);
         switch (opt->calculation_type) {
         case PGM_power_flow:
+            if (opt->tap_changing_strategy != PGM_tap_changing_strategy_disabled &&
+                opt->experimental_features == PGM_experimental_features_disabled) {
+                // this option is experimental and should not be exposed to the user
+                throw MissingCaseForEnumError{"PGM_TapChangingStrategy", opt->tap_changing_strategy};
+            }
```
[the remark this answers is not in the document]
mgovers: one thought i have is that we can check even earlier in the process and instead of raising a `MissingCaseForEnum`, to raise an `ExperimentalFeature` if it is an experimental combination.

The danger of this, is that we first allow it (because it is ignored), then add it as a disallowed experimental feature and then allow it again. The 2nd step would be a breaking change.

To fix that, the `get_optimizer` function should get the `calculation_type` as input. Then, we can mark it as experimental in the C API and it will raise a `MissingCaseForEnum` in the `get_optimizer` function for state estimation/short circuit calculation

Alternatively, we also allow automatic tap optimization with state estimation. While the logic would allow this, it is not a correct approach, because it would be a variable in the math solver itself.

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -59,6 +59,92 @@ void PGM_get_indexer(PGM_Handle* handle, PGM_PowerGridModel const* model, char c
...
+    default:
+        throw MissingCaseForEnumError{"get_optimizer_strategy", opt.tap_changing_strategy};
+    }
+}
+
+constexpr auto get_short_circuit_voltage_scaling(PGM_Options const& opt) {
+    return static_cast<ShortCircuitVoltageScaling>(opt.short_circuit_voltage_scaling);
+}
+
+constexpr auto parse_options(PGM_Options const& opt) {
```
mgovers, line 137:
```suggestion
constexpr auto extract_calculation_options(PGM_Options const& opt) {
```

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -76,13 +162,9 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
-        auto const options =
-            MainModel::Options{.calculation_method = static_cast<CalculationMethod>(opt->calculation_method),
-                               .err_tol = opt->err_tol,
-                               .max_iter = opt->max_iter,
-                               .threading = opt->threading,
-                               .short_circuit_voltage_scaling =
-                                   static_cast<ShortCircuitVoltageScaling>(opt->short_circuit_voltage_scaling)};
+        check_calculate_valid_options(*opt);
+
+        auto const options = parse_options(*opt);
```
mgovers, line 167:
```suggestion
        auto const options = extract_calculation_options(*opt);
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -193,6 +212,11 @@ class DatasetError : public PowerGridError {
     explicit DatasetError(std::string const& msg) { append_msg("Dataset error: " + msg); }
 };
 
+class ExperimentalFeature : public InvalidArguments {
+  public:
+    using InvalidArguments::InvalidArguments;
```
mgovers, line 217:
```suggestion
```
@nitbharambe i think sonar cloud does not like this. Can you try to fix it? I think this suggestion should do

TonyXiang8787: I think this is a false alarm from SonarCloud. We do not need to do anything.

TonyXiang8787 (dismissed):

TonyXiang8787:
@mgovers I have reviewed and approved the PR. I am not sure if this should now be merged before `main_model` is ready.

nitbharambe (dismissed):

mgovers:
> @mgovers I have reviewed and approved the PR. I am not sure if this should now be merged before `main_model` is ready.

i don't think it matters. with the current implementation in this PR, you do get the optimal state - just not the optimal tap positions. I would consider it a `missing feature` rather than a `critical feature`.

After this PR is merged, the validation tests can already be created and even run - provided we do not check for the tap positions. We can also start the Python API creation.

Since we do not give any guarantees for experimental features, I think having `missing features` is OK.

If `experimental_feature` does not carry the load of `feature_under_construction`, then maybe we should add that flag as well. Maybe that's something up for discussion for our next big thing

TonyXiang8787 (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -95,9 +95,13 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         static constexpr Idx sequential = -1;
 
         CalculationMethod calculation_method{CalculationMethod::default_method};
+        OptimizerType optimizer_type{OptimizerType::no_optimization};
+        OptimizerStrategy optimizer_strategy{OptimizerStrategy::any};
+
```
Jerry-Jinfeng-Guo, line 100: Oh, I was under the impression this was intentionally left out of this PR?

mgovers: no that was #599 , which blocked work on the C API (this PR)

PR #598, opened 2024-05-15
feature/update main model for automatic tap changer

Jerry-Jinfeng-Guo:
- [x] Update main_core/output → function skeletons are already in place (main_core.hpp and output.hpp). 
- [x] User will need to be able to extract certain data (tap positions, etc.). 
- [x] Attribute input should be checked (should have been in place)
- [x] attr output (optimizer >all the way back> output structure) needs to be done
- [x] add check to ensure that there can at most be one regulator for each transformer
- [x] include unit tests for the input/update/output functions of automatic tap changer.


power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
```diff
@@ -158,6 +163,13 @@ inline void add_component(MainModelState<ComponentContainer>& state, ForwardIter
             emplace_component<Component>(state, id, input, regulated_object_type, u_rated);
         }
     }
+    // Make sure that each regulated object has at most one regulator
+    const std::unordered_set<Idx2D, Idx2DHash> unique_regulated_objects(regulated_objects.begin(),
+                                                                        regulated_objects.end());
+    if (unique_regulated_objects.size() != regulated_objects.size()) {
+        // There are duplicates
+        throw DuplicatedRegulatedObject();
+    }
```
mgovers, line 172: is this faster than just looping over the data?

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -118,6 +118,13 @@ class InvalidRegulatedObject : public PowerGridError {
     }
 };
 
+class DuplicatedRegulatedObject : public PowerGridError {
+  public:
+    DuplicatedRegulatedObject() {
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: `Duplicatively` then.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
Jerry-Jinfeng-Guo: 'just looping over the data' reads quadratic complexity ;)

mgovers: yep i realized after i posted this 😬

mgovers: maybe you can create a utility for this? like `has_duplicates(Range, Func)`

Jerry-Jinfeng-Guo: Possibility, yes, maybe later

mgovers:
> At the moment the `tap_position_optimizer` still writes everything to a `SolverOutput`. The addition of `MathOutput` to include both `SolverOutput` and `OptimizerOutput` seems ignored. @mgovers Could the `automatic tap changer` adapt before I continue?

Yes the optimizer needs to output a `MathSolver` that includes both the `SolverOutput` and the `OptimizerOutput` if that is waht you mean

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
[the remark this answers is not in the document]
mgovers: Maybe even better tonswitch it around: `DuplicateRegulation` because it would be ok if multiple regulators regulate the same object, as long as they dont regulate the same attribute.

Similarly, there could be one single regulator regulating multiple attributes of a single object.

We can maybe also come up with a regulator that regulates multiple components at the same time. Maybe a smart switch that switches on and off certain loads, some of which have const power, others const current or const impedance, maybe some can be regulated, ... . Could be depending on e.g. line load or energy price.

Regulated object does not fit the purpose well then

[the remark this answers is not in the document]
mgovers: Or `ConflictingRegulation`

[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: `Conflicting...` isn't all that descriptive. `DuplicativelyRegulated...` means something is regulated by more than one party, which is perfect for our purpose. **If** there was some error for the regulator side, that it regulates more than it should, it would be `DuplicativelyRagulating...`. I don't think we are at the granularity of attributes. If that would be an error, it should be in consult with expert input. Otherwise, I wouldn't create imaginary cases (yet).

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -30,6 +30,14 @@ struct Idx2D {
     friend constexpr bool operator==(Idx2D x, Idx2D y) = default;
 };
 
+struct Idx2DHash {
+    std::size_t operator()(const Idx2D& idx) const {
+        const std::size_t h1 = std::hash<Idx>{}(idx.group);
+        const std::size_t h2 = std::hash<Idx>{}(idx.pos);
```
mgovers, line 36:
```suggestion
        size_t consth1 = std::hash<Idx>{}(idx.group);
        size_t const h2 = std::hash<Idx>{}(idx.pos);
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -30,6 +30,14 @@ struct Idx2D {
     friend constexpr bool operator==(Idx2D x, Idx2D y) = default;
 };
 
+struct Idx2DHash {
+    std::size_t operator()(const Idx2D& idx) const {
```
mgovers, line 34: i think this can be `constexpr`

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
```diff
@@ -158,6 +163,13 @@ inline void add_component(MainModelState<ComponentContainer>& state, ForwardIter
             emplace_component<Component>(state, id, input, regulated_object_type, u_rated);
         }
     }
+    // Make sure that each regulated object has at most one regulator
+    const std::unordered_set<Idx2D, Idx2DHash> unique_regulated_objects(regulated_objects.begin(),
+                                                                        regulated_objects.end());
+    if (unique_regulated_objects.size() != regulated_objects.size()) {
+        // There are duplicates
+        throw DuplicativelyRegulatedObject();
```
mgovers, line 171:
```suggestion
        throw DuplicativelyRegulatedObject{};
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +584,41 @@ void add_tap_regulator_output(State const& state,
...
+    using ComponentContainerType = typename State::ComponentContainer;
+    (get_transformer_tap_positions<ComponentTypes, ComponentContainerType>(state, transformer_tap_positions), ...);
+}
+#ifdef _MSC_VER
+// MSVC compiler
+#pragma warning(pop)
+#elif defined(__GNUC__) || defined(__clang__)
+// GCC or Clang compilers
+#pragma GCC diagnostic pop
+#endif
```
mgovers, line 620: this feels like it's extremely dangerous. it is highly likely that the compiler better knows what happens under the hood than we are - especially if all three compilers mention the same thing.

Even more so: the severity level is `\W1` for MSVC (critical). If this really is a false positive, please reproduce a minimal case in https://godbolt.org/ and file a bug report

I believe that the issue comes from the fact that the compiler does not understand that it should use the function in line 591 instead of this line because the signatures are different. Making the signatures the same should fix this

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
Jerry-Jinfeng-Guo: If only `std::hash<Idx>` is `constexpr`

mgovers: welp, guess we'll have to wait until C++29 then... https://www.reddit.com/r/cpp_questions/comments/ld7yux/why_isnt_stdhash_constexpr/

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
TonyXiang8787: Maybe discuss this in the next session. This seems to be very far.

mgovers: yeah we've already discussed that that we'll go over this together in a cleanup round later today

petersalemink95: I also don't think this would be the right way to implement it :)

Jerry-Jinfeng-Guo: @petersalemink95 It most definitely isn't. We'll see what additional logic is needed for the compiler to know for sure.

Jerry-Jinfeng-Guo: Issue solved with a new way of doing variadic calls.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -126,9 +126,12 @@ struct TransformerTapRegulatorCalcParam {
     IntS status{};
 };
 
-struct TransformerTapRegulatorOptimizerOutput {
-    IntS tap_pos{na_IntS};
+struct TransformerTapPosition {
+    Idx2D index;
```
mgovers, line 130: maybe:
```suggestion
    Idx2D components_index;
```
?

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -126,9 +126,12 @@ struct TransformerTapRegulatorCalcParam {
...
+    IntS tap_position;
+    TransformerTapPosition(Idx2D _index, IntS _tap_position) : index(_index), tap_position(_tap_position) {}
```
mgovers, line 132:
```suggestion
    Idx2D index{};
    IntS tap_position{};
```

otherwise it does not satisfy `DefaultConstructible` which is a (soft) requirement for `std::vector`

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -126,9 +126,12 @@ struct TransformerTapRegulatorCalcParam {
...
 };
+using TransformerTapPositionResult = std::vector<TransformerTapPosition>;
```
mgovers, line 134: isn't `Output` better in this case since that's what we use across the repo

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
```diff
@@ -158,6 +163,13 @@ inline void add_component(MainModelState<ComponentContainer>& state, ForwardIter
             emplace_component<Component>(state, id, input, regulated_object_type, u_rated);
         }
     }
+    // Make sure that each regulated object has at most one regulator
+    const std::unordered_set<Idx2D, Idx2DHash> unique_regulated_objects(regulated_objects.begin(),
+                                                                        regulated_objects.end());
+    if (unique_regulated_objects.size() != regulated_objects.size()) {
+        // There are duplicates
+        throw DuplicativelyRegulatedObject{};
```
mgovers, line 171: i guess we'll need the input validation checks to get the actual duplicate ids for efficiency

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
mgovers, line 1: so far, it's only been compatibility. you don't actually output the transformer tap positions, right?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -560,26 +560,14 @@ inline void create_tap_regulator_output(State const& state, std::vector<SolverOu
...
+template <transformer_c Component, class ComponentContainer>
+    requires main_core::model_component_state_c<main_core::MainModelState, ComponentContainer, Component>
+constexpr void get_transformer_tap_positions(main_core::MainModelState<ComponentContainer> const& state,
+                                             TransformerTapPositionResult& transformer_tap_positions) {
+    constexpr auto group_index = ComponentContainer::template get_type_idx<Component>();
+    for (auto const& transformer : state.components.template citer<Component>()) {
+        transformer_tap_positions.emplace_back(
+            TransformerTapPosition{Idx2D{group_index, transformer.id()}, static_cast<IntS>(transformer.tap_pos())});
     }
 }
```
mgovers, line 572: as it is now, you produce tap positions for **all** components but the output for the user only concerns the tap positions of regulated transformers.

The way it is now, in the `main_core/output.hpp`, we'll need to loop over all transformers; find the regulator that regulates that transformer (if any); and then output the tap position to the transformer.

That's very expensive and actually not useful because set of regulated transformers is usually fairly small

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -615,14 +603,21 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
         auto const cache = this->cache_states(order);
-        auto optimal_output = optimize(state, order, method);
+        auto solver_output = optimize(state, order, method);
         update_state(cache);
 
-        return optimal_output;
+        TransformerTapPositionResult transformer_tap_positions;
+        (get_transformer_tap_positions<TransformerTypes, typename State::ComponentContainer>(state,
+                                                                                             transformer_tap_positions),
+         ...);
```
mgovers, line 617: `update_state(cache)` should happen after this (or, equivalently, this output needs to happen in the `optimize` implementation)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
Jerry-Jinfeng-Guo: I'd assume the same

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: All *transformers*, yes. The argument is that users wouldn't know which transformer is regulated, but they would be interested to know the tap of a transformer (given idx).

I don't see the need to go via the regulators.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: Yes, I think you are correct in this. Could you apply the change?

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
Jerry-Jinfeng-Guo: Hmm, if you mean to use `component_index` to distinguish it from a regular scalar id, perhaps? Otherwise, I am not so sure here...

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
Jerry-Jinfeng-Guo: You are right, this slipped my mind. Could you add this?

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: more like to distinguish explicitly index from math id, but i guess it's fine as is. let's close it

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: the user output of the transformer tap regulator is the actual tap position. The user chooses which transformer a regulator regulates, and from there, the regulator is able to tell the user which tap position it chose. This is defined in https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/output.hpp#L229

Since outputting this actually requires a regulator-to-transformer or transformer-to-regulator-lookup (expensive), my guess is that it would be easier to go directly via the regulator. Also: the user is not interested in all transformers' tap positions but only the regulated ones, which may be a very small subset of all transformers

Jerry-Jinfeng-Guo: > the user output of the transformer tap regulator is the actual tap position. 

A regulator is simply a middle man since the tap position is a physical property of transformers. This statement is not correct.

> The user chooses which transformer a regulator regulates, and from there, the regulator is able to tell the user which tap position it chose. 

In the ideal case, a list of transformers to regulate is made known by the users and the same list should be used to retrieve tap positions.

> Since outputting this actually requires a regulator-to-transformer or transformer-to-regulator-lookup (expensive), my guess is that it would be easier to go directly via the regulator. Also: the user is not interested in all transformers' tap positions but only the regulated ones, which may be a very small subset of all transformers

I don't want to argue about efficiency yet, without knowing whether it is actually a bottleneck at all. Saying something is expensive out of the box seems rather contextless. I do agree that there is a subset of what might be interesting for the users, as to which degree I have no knowledge and frankly I don't think there is a guarantee.

As I mentioned above, if a list is made known by the users, we should be using that, instead of going via the regulators.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
Jerry-Jinfeng-Guo: I have no strong objection ;)

mgovers (approved):

mgovers (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: > A regulator is simply a middle man since the tap position is a physical property of transformers. This statement is not correct.

in that case we will need to redesign the C API component output and we need to do it ASAP because if we release the C API with auto tap regulator with tap positions and we want to change that, it will be a breaking change.

mgovers: see e.g. https://power-grid-model--606.org.readthedocs.build/en/606/user_manual/components.html#id38

mgovers: in particular, see this change in the user output data attribute declaration https://github.com/PowerGridModel/power-grid-model/pull/534/files#diff-401bcdd1f75ea6c2e0f22a6b89435d2e0be22fec43a411e5f6f9776a5dd3f8c3R284-R288

mgovers (approved):

PR #602, opened 2024-05-16
pgm-automatic-tap validation tests

nitbharambe:
Comparison cases added from pgm-automatic-tap


tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -333,14 +342,17 @@ CalculationFunc calculation_func(CaseParam const& param) {
...
-            if (sym) {
-                return model.calculate_power_flow<symmetric_t>(get_default_options(calculation_method, threading),
-                                                               dataset, update_dataset);
+        return [param, get_default_options](MainModel& model, CalculationMethod calculation_method,
+                                            Dataset const& dataset, ConstDataset const& update_dataset, Idx threading) {
+            auto options = get_default_options(calculation_method, threading);
+            options.optimizer_type = param.tap_changing_strategy == "disabled"
+                                         ? OptimizerType::no_optimization
+                                         : OptimizerType::automatic_tap_adjustment;
+            options.optimizer_strategy = optimizer_strategy_mapping.at(param.tap_changing_strategy);
```
mgovers, line 351: there's probably a better way of doing things like this but since we are going to touch this when we pull calculation type + symmetry into the options as well, i don't think it's necessary to make this nicer than it is now.

tests/data/power_flow/pgm-automatic-tap/params.json
```diff
@@ -0,0 +1,15 @@
...
+  "calculation_method": "newton_raphson",
+  "tap_changing_strategy": "any_valid_tap",
+  "rtol": 1e-5,
+  "atol": 1e-5,
+  "extra_params": {
+    "newton_raphson": {
+      "experimental_features": "enabled",
+      "fail": {
+        "reason": "Experimental feature under development",
+        "raises": "PowerGridError"
```
mgovers, line 11: is this really the most generic error message we get? 😮 it's fine like this, just surprised 😬

mgovers (approved):

mgovers (changes requested):
@nitbharambe the tests fail. i think you broke the logic surrounding skipping tests that are in failing

tests/data/power_flow/pgm-automatic-tap/params.json
nitbharambe: Nope, we dont. I am going to add a `run` argument to xfail to avoid running this case in first place.

mgovers: what do you mean?

tests/data/power_flow/pgm-automatic-tap/params.json
```diff
@@ -0,0 +1,16 @@
...
+  "tap_changing_strategy": "any_valid_tap",
+  "rtol": 1e-5,
+  "atol": 1e-5,
+  "extra_params": {
+    "newton_raphson": {
+      "experimental_features": "enabled",
+      "fail": {
+        "reason": "Experimental feature under development",
+        "raises": "PowerGridError",
+        "run": false
```
mgovers, line 12: why not just
```suggestion
        "raises": "AssertionError"
```
?

nitbharambe: It raises OSError which is not helpful at all. Maybe later, it will not raise OSError, but some other error.

nitbharambe: I see it fails with this in sonar cloud. Maybe skipping is the way to go
```
........................................................................ [ 22%]
Fatal Python error: Segmentation fault
```

mgovers:
@nitbharambe if you temporarily disable this test, i'm OK with merging

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -300,13 +302,20 @@ std::map<std::string, ShortCircuitVoltageScaling> const sc_voltage_scaling_mappi
 using CalculationFunc =
     std::function<BatchParameter(MainModel&, CalculationMethod, Dataset const&, ConstDataset const&, Idx)>;
 
+std::map<std::string, OptimizerStrategy> const optimizer_strategy_mapping = {
```
mgovers, line 305: sonar cloud suggestion

```suggestion
std::map<std::string, OptimizerStrategy, std::less<>> const optimizer_strategy_mapping = {
```

tests/data/power_flow/pgm-automatic-tap/params.json
mgovers: cfr #608 : not a problem in this PR. let's discuss whether we disable this PR or do it another way

PR #603, opened 2024-05-21
try gcc 12 in the ci

TonyXiang8787:


.github/workflows/main.yml
```diff
@@ -84,9 +84,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install packages
         run: |
-          sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
           sudo apt-get update
-          sudo apt-get install -y ninja-build gcc-13 g++-13 clang-15
+          sudo apt-get install -y ninja-build gcc-12 g++-12 clang-15
```
mgovers, line 88: isn't it installed by default?
```suggestion
          sudo apt-get install -y ninja-build clang-15
```

TonyXiang8787: No the default is `gcc-11`.

TonyXiang8787: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

mgovers: interesting, they must've changed that

mgovers: no actually, because we set `CC=gcc-12 CXX=g++-12`, it should work just fine. They are installed by default, just not added as the default `CC` and `CXX` compilers

mgovers: fine with keeping it as is though. makes stuff explicit

mgovers (approved):

.github/workflows/main.yml
TonyXiang8787: In the readme of images `gcc-12` is not listed as installed.

mgovers: image

PR #604, opened 2024-05-21
[Feature] enable Python API for auto tap changer (experimental)

mgovers:
Python equivalent of #597

### DIFF
--- a/src/power_grid_model/enum.py
+++ b/src/power_grid_model/enum.py
@@ -67,6 +67,27 @@ class CalculationMethod(IntEnum):
     iec60909 = 5
 
 
+class TapChangingStrategy(IntEnum):
+    """Tap Changing Strategies"""
+
+    disabled = 0
+    """
+    Disable automatic tap adjustment
+    """
+    any_valid_tap = 1
+    """
+    Adjust tap position automatically; optimize for any value in the voltage band
+    """
+    min_voltage_tap = 2
+    """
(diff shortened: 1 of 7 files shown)


src/power_grid_model/enum.py
```diff
@@ -67,6 +67,27 @@ class CalculationMethod(IntEnum):
     iec60909 = 5
```
mgovers, line 69: TBD: should we add an enum `ControlSide` here?

```suggestion
class ControlSide(IntEnum):
    """Control Sides for Regulators on Branch Objects"""

    from_side = 0
    to_side = 1
    side_1 = 0
    side_2 = 1
    side_3 = 2


```

* [ ] If so: should we add a validation check to check that the from_side/to_side/side_1/side_2/side_3 is in range of the transformer type
* [ ] If not: should we add a validation check to check
  * [ ] that the correct enum type is being used and/or
  * [ ] that it is in range of the transformer type

mgovers:
I believe this is all that's needed for the time being

src/power_grid_model/enum.py
nitbharambe: I dont see an extra enum being useful. The intention is sufficiently conveyed by using BranchSide or Branch3Side itself.
A data validation check would be nice.

mgovers: Created a JIRA ticket for this

TonyXiang8787 (approved):

PR #606, opened 2024-05-22
[Feature] Automatic tap changer documentation

mgovers:


docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+| `u_set`                    | `double`                                                                                                                                                                                                                                                                                                           | volt (V) | the voltage setpoint (at the center of the band)                                                        |                         &#10004;                         | &#10004; |      `>= 0`       |
+| `u_band`                   | `double`                                                                                                                                                                                                                                                                                                           | volt (V) | the width of the voltage band                                                                           |                         &#10004;                         | &#10004; | `> 0` (see below) |
```
mgovers, line 744: TODO: elaborate more on lower bound/higher bound vs. per side

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,100 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+| ------------------------------------------------------------ | -------- | -------- | ----------------------------------------------------------------------------------------------------------- |
+| No automatic tap changing (regular power flow)               | &#10004; | &#10004; | {py:class}`TapChangingStrategy.disabled <power_grid_model.enum.TapChangingStrategy.disabled>`               |
+| Optimize tap position for any value in the voltage band      |          | &#10004; | {py:class}`TapChangingStrategy.any_valid_tap <power_grid_model.enum.TapChangingStrategy.any_valid_tap>`     |
+| Optimize tap position for lower voltage in the voltage band  |          |          | {py:class}`TapChangingStrategy.min_voltage_tap <power_grid_model.enum.TapChangingStrategy.min_voltage_tap>` |
+| Optimize tap position for higher voltage in the voltage band |          |          | {py:class}`TapChangingStrategy.max_voltage_tap <power_grid_model.enum.TapChangingStrategy.max_voltage_tap>` |
+
+The following control logic is used.
+
+- Regulated transformers are ranked according to the amount of transformers between itself and the nearest source {hoverxreftooltip}`user_manual/components:source`.
```
petersalemink95, line 608: Maybe good to mention that they are ranked based on voltage level.

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,100 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+        - If the $U_{\text{control}} < U_{\text{set}} - \frac{U_{\text{band}}}{2}$:
+          - The ratio between the voltage on the tap side and the voltage on the control side is too high.
+          - To decrease this ratio, the tap ratio must be decreased.
+          - Therefore, the tap position of the regulated transformer is decreased if it satisfies the bounds set by `tap_min` and `tap_max`.
+        - If, however, the $U_{\text{control}} > U_{\text{set}} + \frac{U_{\text{band}}}{2}$:
+          - The ratio between the voltage on the tap side and the voltage on the control side is too low.
+          - To increase this ratio, the tap ratio must be increase.
+          - Therefore, the tap position of the regulated transformer is increased if it satisfies the bounds set by `tap_min` and `tap_max`.
+        - If the tap position of this transformer did not change, the transformer is considered regulated.
+      - If not all transformers within this rank are regulated, step out of the loop and go to the next iteration step.
```
petersalemink95, line 626: Would be good to mention when this would happen

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,100 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+  - Exploit the neighbourhood of all transformers (see {hoverxreftooltip}`user_manual/calculations:Initialization and exploitation of regulated transformers`)
+    - Re-run the iteration in the above if any of the tap positions changed by the exploitation.
+
+The exploitation of the neighbourhood ensures that the actual optimum is not accidentally missed due to feedback mechanisms in the grid.
+
+```{note}
```
petersalemink95, line 632: This is a general note for all PF calculations right? Might be confusing that is is within a section about "Power flow with automatic tap changing"

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,100 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+The control logic assumes that the change in voltage level caused by changing a transformer is larger than the change caused by adjacent transformers.
+This is assumption is reflected in the requirements mentioned in {hoverxreftooltip}`user_manual/components:Transformer Tap Regulator`.
+```
+
+```{note}
+The control logic assumes that the (compensated) control voltage decreases when the tap position increases.
+If the line drop compensation impedance is high, and the control side has generator-like behavior, then this assumption does not hold, and the calculation may diverge.
+Hence, this is assumption is reflected in the requirements mentioned in {hoverxreftooltip}`user_manual/components:Line drop compensation`.
```
petersalemink95, line 647:
```suggestion
Hence, this assumption is reflected in the requirements mentioned in {hoverxreftooltip}`user_manual/components:Line drop compensation`.
```

docs/user_manual/components.md
```diff
@@ -114,17 +114,17 @@ If `i_n` is not provided, `loading` of line will be a `nan` value.
...
-| `c0`   | `double`  | farad (F)  | zero-sequence shunt capacitance            | &#10024; only for asymmetric calculations | &#10060; |                                   |
-| `tan0` | `double`  | -          | zero-sequence shunt loss factor (tan𝛿)     | &#10024; only for asymmetric calculations | &#10060; |                                   |
-| `i_n`  | `double`  | ampere (A) | rated current                              |                 &#10060;                  | &#10060; |               `> 0`               |
+| name   | data type | unit       | description                                |                 required                  |  update  |            valid values            |
+| ------ | --------- | ---------- | ------------------------------------------ | :---------------------------------------: | :------: | :--------------------------------: |
+| `r1`   | `double`  | ohm (Ω)    | positive-sequence serial resistance        |                 &#10004;                  | &#10060; | `r1` and `x1` cannot be both `0.0` |
```
petersalemink95, line 119: Why all the changes from "zero" to 0 and 0.0?

docs/user_manual/components.md
```diff
@@ -657,9 +657,9 @@ $$
 ## Fault
 
 * type name: `fault`
-* * base: {hoverxreftooltip}`user_manual/components:base`
+* base: {hoverxreftooltip}`user_manual/components:base`
 
-`fault` defines a short circuit location in the grid. At this moment a fault can only happen at a `node`.
+`fault` defines a short circuit location in the grid. At the time of writing, a fault can only happen at a `node`.
```
petersalemink95, line 662: Not just at the time of writing; For now, a fault can only happen at a node. Maybe just change it in "A fault can only happen at a `node`."

docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+| -------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------- | ------------------------------------------------------------------------------------------------------- | :------------------------------------------------------: | :------: | :---------------: |
+| `control_side`             | {py:class}`BranchSide <power_grid_model.enum.BranchSide>` if the regulated object is a {hoverxreftooltip}`user_manual/components:transformer` and {py:class}`Branch3Side <power_grid_model.enum.Branch3Side>` if it the regulated object is a {hoverxreftooltip}`user_manual/components:Three-Winding Transformer` | -        | the controlled side of the transformer                                                                  | &#10024; only for power flow with automatic tap changing | &#10004; |                   |
```
petersalemink95, line 742: Why are `u_set` and `u_band` required, but is `control_side` only needed when you want to perform a power flow with automatic tap?

docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+A `transformer_tap_regulator` has no short circuit output.
+
+#### Electric Model
+
+The transformer tap regulator itself does not have a direct contribution to the grid state.
+Instead, it regulates the tap position of the regulated object until the voltage at the control side is in the specified voltage band:
+
+$$
+   \begin{eqnarray}
+      U_{\text{control}}$ is in the range $\left[U_{\text{set}} - \frac{U_{\text{band}}}{2}, U_{\text{set}} + \frac{U_{\text{band}}}{2}\right]
```
petersalemink95, line 776: "is in the range" should be as text. Formatting is not good now, see: https://power-grid-model--606.org.readthedocs.build/en/606/user_manual/components.html#id40

docs/user_manual/components.md
mgovers: you're right

docs/user_manual/components.md
mgovers: because they are all `double` values. I found them confusing when going through the code. I can change it back if you prefer

docs/user_manual/calculations.md
mgovers: what do you mean? they are regulated based on the weighted shortest path to the source, right? Or do you mean that only paths from the higher voltage to the same or lower rated voltage are taken into account?

mgovers:
resolved the comments that i was able to resolve. please provide input to the others

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,102 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+
+```{note}
+For iterative [power flow calculation methods](#power-flow-algorithms), the initial state may not converge.
+If the iterative process failed to converge, the optimization procedure is executed twice.
+First, the {py:class}`linear <power_grid_model.enum.CalculationMethod.linear>` calculation method is used to find an approximate solution.
+The optimization procedure is then run a second time to find the actual optimum with the user-specified calculation method.
+```
+
+```{note}
+The control logic assumes that the change in voltage level caused by changing a transformer is larger than the change caused by adjacent transformers.
```
nitbharambe, line 642:
```suggestion
The control logic assumes that the change in voltage level caused by regulating a transformer is larger than the change caused by adjacent transformers.
```
Or maybe something else is meant

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,102 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+
+```{note}
+The control logic assumes that the (compensated) control voltage decreases when the tap position increases.
+If the line drop compensation impedance is high, and the control side has generator-like behavior, then this assumption does not hold, and the calculation may diverge.
+Hence, this assumption is reflected in the requirements mentioned in {hoverxreftooltip}`user_manual/components:Line drop compensation`.
+```
+
+##### Initialization and exploitation of regulated transformers
+
+Regulated transformers are initialized and exploited depending on the optimization strategy.
```
nitbharambe, line 654: In the context of all written text about `exploitation`, I think more information can be added here or below: 

Above the table,
Internally, the control algorithm sets initial tap position as following so we end up with a regulated tap_pos corresponding to the following conditions:

In the description section
`Find best possible tap position that gives highest control side voltage within the u_band.`
`Find best possible tap position that gives lowest control side voltage within the u_band.`

Tap side voltage seems less relevant i suppose.

docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+| `regulated_object` | `int32_t` | -    | ID of the regulated object                | &#10004; | &#10060; | a valid regulated object ID |
+| `status`           | `int8_t`  | -    | connection status to the regulated object | &#10004; | &#10004; |         `0` or `1`          |
+
+### Transformer tap regulator
+
+* type name: `transformer_tap_regulator`
+* base: {hoverxreftooltip}`user_manual/components:regulator`
 
+`transformer_tap_regulator` defines a regulator for transformers in the grid.
+At the time of writing, a transformer tap regulator regulates a component that is either a {hoverxreftooltip}`user_manual/components:transformer` or a {hoverxreftooltip}`user_manual/components:Three-Winding Transformer`.
```
nitbharambe, line 731:
```suggestion
A transformer tap regulator regulates a component that is either a {hoverxreftooltip}`user_manual/components:transformer` or a {hoverxreftooltip}`user_manual/components:Three-Winding Transformer`.
```
A lot of things are temporary to mention this everywhere.

docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+
+The transformer tap regulator overloads the `tap_pos` of the transformer it regulates in the range set by the user via `tap_min` and `tap_max` (i.e., `(tap_min <= tap_pos <= tap_max)` or `(tap_min >= tap_pos >= tap_max)`).
```
nitbharambe, line 733: While it is most accurate, `overloads` here is a non electrical term users can make do without.
(This is a nitpick, resolve without changes if you wish to)

docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+| `line_drop_compensation_x` | `double`                                                                                                                                                                                                                                                                                                           | ohm (Ω)  | compensation for voltage drop due to  reactance during transport (see [below](#line-drop-compensation)) |                  &#10060; default `0.0`                   | &#10004; |      `>= 0`       |
+
+The following additional requirements exist on the input parameters.
+
+- The `control_side` supports {py:class}`BranchSide <power_grid_model.enum.BranchSide>` if it regulates a {hoverxreftooltip}`user_manual/components:transformer` and {py:class}`Branch3Side <power_grid_model.enum.Branch3Side>` if it regulates a {hoverxreftooltip}`user_manual/components:Three-Winding Transformer`.
```
nitbharambe, line 750:
```suggestion
```
I think the message is clear from the table above.

docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+- The rated voltage at the `tap_side` must be greater than or equal to the rated voltage at the `control_side`.
```
nitbharambe, line 751:
```suggestion
- The automatic tap changer algorithm currently only supports tap changers connected at HV side of the transformer. Hence, the rated voltage of node at the `tap_side` must be greater than or equal to the rated voltage of node at the other side of transformer.
```

Control side can be both HV or LV. But only tap_side can be HV.

docs/user_manual/components.md
```diff
@@ -695,15 +695,116 @@ A `fault` has no steady state output.
...
+     regulator <=================================/
+```
+
+The control voltage is the voltage at the node, compensated with the voltage drop corresponding to the specified line drop compensation.
+
+$$
+   \begin{eqnarray}
+      & Z_{\text{compensation}} = r_{\text{compensation}} + \mathrm{j} x_{\text{compensation}} \\
+      & U_{\text{control}} = \left|\underline{U}_{\text{node}} - \underline{I}_{\text{node}} \cdot \underline{Z}_{\text{compensation}}\right|
+   \end{eqnarray}
```
nitbharambe, line 798: `u_control_side_node + i_out_transformer * z`
A node does not have a current. 

or maybe as is `u_control_side_node - i_transformer * z` with a repeat refence about the direction of current. The direction part always confuses the user at first.

docs/user_manual/components.md
mgovers: the reason is that documentation is very easily outdated (easy to miss something). changed it back to `A` but that was my initial rationale

docs/user_manual/components.md
mgovers: i resolved this, but semi-related: what's the difference between `u1`/`u2` of the transformers and the `u_rated` of the nodes they are connected to?

docs/user_manual/components.md
mgovers: whoops, you're right.. fixed

mgovers: fixed again cause of the sign conventions: it's canceling minus signs

`u_control_side_node - i_out_transformer * z = u_control_side_node + i_transformer * z`

docs/user_manual/components.md
nitbharambe: All these 4 can be different values. `u1`, `u2` are about the actual transformer parameters. They would usually determine the transformation ratio of voltage.

`u_rated` usually is almost equal to `u1` or `u2`. But it does not have to be equal. All the associated `%` or p.u. input grid parameters get affected by the `u_rated` voltages. You can attach any `u1`/`u2` transformer to any `u_rated_1` / `u_rated_2`

While manufacturers could indeed combine all these and consider only 1 value or 2  values, they remain separate as they are all linked to other parameters individually within & outside of power flow calculations.

Also, we could have done ranking against `u1` or `u2`, but that can potentially give more groups of ranks. Different brands of transformer might have different ratings: like maybe some have `10000` or some have `10050` (There might be better examples for this)

**TLDR**: And in relevance of this comment, the check we do is against u_rated of both nodes.

mgovers: great writeup. thanks!

nitbharambe (approved):
Great work. Quite detailed explainations!

nitbharambe (approved):

docs/user_manual/calculations.md
petersalemink95: I meant that it is not necessarily the amount of transformers: if you have a transformer that has similar rated voltage on both sides, it does not count in the ranking process

mgovers: It does. Taken from the mathematical workout of the ranking algorithm:

>    2. For each non-transformer branch (link, line) which are connected at both sides 
>       1. Create two edges for both direction, with the weight zero. 
>    3. **For each transformer which are connected at both sides** 
>       1. **If there is no automatic tap changer** 
>          1. **Create two edges for both direction, with the weight one.** 
>       2. **If there is automatic tap changer** 
>          1. **Create one edge from the node at the tap side to the node without tap, with the weight one.** 

Lines are not taken into account, but all transformers are. The transformers with similar `u_rated` on both sides are just taken into account bidirectionally, rather than unidirectionally. That means that in the following grid

docs/user_manual/components.md
petersalemink95: `0` and `0.0` is fine by me as well. For me personally `zero` was also sufficienct

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,104 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+| Optimize tap position for any value in the voltage band      |          | &#10004; | {py:class}`TapChangingStrategy.any_valid_tap <power_grid_model.enum.TapChangingStrategy.any_valid_tap>`     |
+| Optimize tap position for lower voltage in the voltage band  |          |          | {py:class}`TapChangingStrategy.min_voltage_tap <power_grid_model.enum.TapChangingStrategy.min_voltage_tap>` |
+| Optimize tap position for higher voltage in the voltage band |          |          | {py:class}`TapChangingStrategy.max_voltage_tap <power_grid_model.enum.TapChangingStrategy.max_voltage_tap>` |
+
+##### Control logic for power flow with automatic tap changing
+
+The following control logic is used.
+
+- Regulated transformers are ranked according to the amount of transformers between itself and the nearest source {hoverxreftooltip}`user_manual/components:source`.
+  - Transformers that are closer to a {hoverxreftooltip}`user_manual/components:source` are regulated first.
```
Jerry-Jinfeng-Guo, line 611:
```suggestion
  - Transformers of lower ranks are prioritized for regulation before those of higher ranks.
```

docs/user_manual/calculations.md
mgovers: cfr. discussion: indeed correct as is

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,104 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+| No automatic tap changing (regular power flow)               | &#10004; | &#10004; | {py:class}`TapChangingStrategy.disabled <power_grid_model.enum.TapChangingStrategy.disabled>`               |
+| Optimize tap position for any value in the voltage band      |          | &#10004; | {py:class}`TapChangingStrategy.any_valid_tap <power_grid_model.enum.TapChangingStrategy.any_valid_tap>`     |
+| Optimize tap position for lower voltage in the voltage band  |          |          | {py:class}`TapChangingStrategy.min_voltage_tap <power_grid_model.enum.TapChangingStrategy.min_voltage_tap>` |
+| Optimize tap position for higher voltage in the voltage band |          |          | {py:class}`TapChangingStrategy.max_voltage_tap <power_grid_model.enum.TapChangingStrategy.max_voltage_tap>` |
+
+##### Control logic for power flow with automatic tap changing
+
+The following control logic is used.
+
+- Regulated transformers are ranked according to the amount of transformers between itself and the nearest source {hoverxreftooltip}`user_manual/components:source`.
```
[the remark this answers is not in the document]
mgovers: the user doesn't care what is low and what is high, though. i agree the original was incomplete but i feel like your proposal also doesn't fully capture the info that is relevant to the user. i'll think about it for a second but if you have another proposal, please LMK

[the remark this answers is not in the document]
mgovers: maybe (including the next line):

```md
- Regulated transformers are prioritized according to the amount of transformers to a {hoverxreftooltip}`user_manual/components:source` in a transformer-source topological graph.
  - Transformers that are directly connected to a {hoverxreftooltip}`user_manual/components:source` are prioritized for regulation.
```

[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: In your own words: 
```
  - Transformers that are directly connected to a {hoverxreftooltip}`user_manual/components:source` are prioritized for regulation. 
```
... this captures even less ;)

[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: second line:
```
 - Transformers are regulated in order according to their ranks.
```

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,104 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+
+##### Control logic for power flow with automatic tap changing
+
+The following control logic is used:
+
+- Regulated transformers are ranked according to how close they are to {hoverxreftooltip}`sources <user_manual/components:source>` in terms of the amount of transformers inbetween.
+  - Transformers are regulated in order according to their ranks.
+- Initialize all transformers to their starting tap position (see {hoverxreftooltip}`user_manual/calculations:Initialization and exploitation of regulated transformers`)
+- Find the optimal state using the following procedure
+  - While not all transformers are regulated, iterate as follows:
```
Jerry-Jinfeng-Guo, line 614: This sounds like there is a state saying if a transformer is regulated already.
```suggestion
  - While not all transformers cannot be further regulated, iterate as follows:
```
or
```suggestion
  - While some transformers can still be further regulated, iterate as follows:
```

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,104 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+    - Run a power flow calculation with the current tap positions with the specified [calculation method](#power-flow-algorithms).
+    - Start with the lowest rank (with the transformers closest to a {hoverxreftooltip}`user_manual/components:source`, because the source provides a relatively stable voltage level).
```
Jerry-Jinfeng-Guo, line 616: I thought the users *don't care* about low or high? ;)

mgovers: indeed, will remove

docs/user_manual/calculations.md
```diff
@@ -554,12 +563,104 @@ The assumptions used for calculations in power-grid-model are aligned to the one
...
+The following control logic is used:
+
+- Regulated transformers are ranked according to how close they are to {hoverxreftooltip}`sources <user_manual/components:source>` in terms of the amount of transformers inbetween.
+  - Transformers are regulated in order according to their ranks.
+- Initialize all transformers to their starting tap position (see {hoverxreftooltip}`user_manual/calculations:Initialization and exploitation of regulated transformers`)
+- Find the optimal state using the following procedure
+  - While some transformers can still be further regulated, iterate as follows:
+    - Run a power flow calculation with the current tap positions with the specified [calculation method](#power-flow-algorithms).
+    - Start with the transformers ranked closest to a {hoverxreftooltip}`user_manual/components:source` (because the source provides a relatively stable voltage level).
+    - Loop over all ranks:
```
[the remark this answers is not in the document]
mgovers: the two are related. will update

nitbharambe (approved):
LGTM. Merge post other comment resolves.

TonyXiang8787:
@nitbharambe can we go ahead and merge this one?

Jerry-Jinfeng-Guo (approved):

PR #607, opened 2024-05-23
Feature/tap regulator output

mgovers:
output tap positions for the regulators with the original implementation of #598

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -127,7 +127,7 @@ struct TransformerTapRegulatorCalcParam {
 };
 
 struct TransformerTapPosition {
-    ID transformer{};
+    ID transformer_id{};
     IntS tap_position{};
 };
 using TransformerTapPositionResult = std::vector<TransformerTapPosition>;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -127,7 +127,7 @@ struct TransformerTapRegulatorCalcParam {
 };
 
 struct TransformerTapPosition {
-    Idx2D index{};
+    ID transformer{};
     IntS tap_position{};
 };
 using TransformerTapPositionResult = std::vector<TransformerTapPosition>;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
@@ -617,7 +617,8 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
         for (auto const& sub_order : regulator_order) {
             for (auto const& regulator : sub_order) {
                 auto const& transformer = regulator.transformer;
-                transformer_tap_positions.push_back({transformer.id(), transformer.tap_pos()});
+                transformer_tap_positions.push_back(
+                    {.transformer_id = transformer.id(), .tap_position = transformer.tap_pos()});
             }
         }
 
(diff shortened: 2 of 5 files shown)


mgovers:
@nitish let's discuss tomorrow whether to merge this into #598 and/or also to `main`. that would mean we reach the sprint goal and cleanup (making it efficient) would then be a follow-up

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -127,7 +127,7 @@ struct TransformerTapRegulatorCalcParam {
 };
 
 struct TransformerTapPosition {
-    Idx2D index{};
+    ID transformer{};
```
Jerry-Jinfeng-Guo, line 130:
```suggestion
    ID transformer_id{};
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -622,21 +610,24 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
         TransformerTapPositionResult transformer_tap_positions;
 
-        // TODO(mgovers): only output the transformers that are regulated
-        (get_transformer_tap_positions<TransformerTypes, typename State::ComponentContainer>(state,
-                                                                                             transformer_tap_positions),
-         ...);
+        for (auto const& sub_order : regulator_order) {
+            for (auto const& regulator : sub_order) {
+                auto const& transformer = regulator.transformer;
+                transformer_tap_positions.push_back({transformer.id(), transformer.tap_pos()});
```
Jerry-Jinfeng-Guo, line 620:
```suggestion
                transformer_tap_positions.emplace_back({transformer.id(), transformer.tap_pos()});
```
on behalf of sonar cloud

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -622,21 +610,24 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
+    auto produce_output(std::vector<std::vector<RegulatedTransformer>> const& regulator_order,
+                        ResultType solver_output) const -> MathOutput<ResultType> {
         TransformerTapPositionResult transformer_tap_positions;
 
-        // TODO(mgovers): only output the transformers that are regulated
-        (get_transformer_tap_positions<TransformerTypes, typename State::ComponentContainer>(state,
-                                                                                             transformer_tap_positions),
-         ...);
+        for (auto const& sub_order : regulator_order) {
+            for (auto const& regulator : sub_order) {
```
Jerry-Jinfeng-Guo, line 618: Sooo a nested `for` then

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -990,47 +1006,6 @@ TEST_CASE("Test tap position optmizer I/O") {
                                                                           bad_regulators.end(), 50.0),
                         DuplicativelyRegulatedObject);
     }
-
-    SUBCASE("transformer tap position retrieval") {
```
Jerry-Jinfeng-Guo, line 994: So the decision is to not unit test tap pos retrieval?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: i now know why you initially used `push_back`. see https://developercommunity.visualstudio.com/t/stdvectoremplace-back-does-not-work-for-a-class-wi/1691516 and in particular this minimal example: https://godbolt.org/z/xovjh9ffe . 

looks like a compiler bug. we can change it in the future when it's resolved

mgovers: FYI: this was fixed in clang `16.0.0` https://godbolt.org/z/vYohvbv7q

mgovers: another reason to migrate to Ubuntu 24.04 runner images

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: it is what it is. it's fine, i guess. it is definitely more readable than more variadic stuff (that is handled internally in the transformer regulator reference class)

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: it should be unit tested in `main_core` but here, the class' unit tests itself should take care of it. see https://github.com/PowerGridModel/power-grid-model/pull/607/files#diff-e3cb2a7b8db22693560826087f2c675630f9cd51a2337191aadc986dd50c4249R951-R958

PR #610, opened 2024-05-27
automatic tap changer regulation tactic validation cases

nitbharambe:
Add new validation cases from pgm-automatic-tap


mgovers:
@nitbharambe you can still keep the results but update them with the existing results

TonyXiang8787 (approved):

mgovers:
@nitbharambe can these be merged?

PR #614, opened 2024-05-29
feature/maximum iteration per rank to prevent infinite loop

Jerry-Jinfeng-Guo:
- [x] maximum tap ranges per rank
- [x] 2x max tap range as max iter for rank
- [x] unit test
- [x] new error


power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -665,8 +672,13 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
                 if (tap_changed) {
                     break;
                 }
+                iterations_per_rank[++rank_index] = 0;
             }
             if (tap_changed) {
+                if (++iterations_per_rank[rank_index] > 2 * regulator_order[rank_index][0].transformer.tap_range()) {
```
mgovers, line 678: should be (pseudocode):
```
max_iters[rank_k] = max( tap_range() for transformer in regulator_order[rank_idx] )
```
because trafo 1 in the same rank may have a larger tap range than trafo 0.

probably a good idea to pre-calculate (maybe even in a separate function)

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -112,6 +112,11 @@ class IterationDiverge : public PowerGridError {
     }
 };
 
+class MaxIterationReached : public PowerGridError {
```
mgovers, line 115: shouldn't this derive from `IterationDiverge`?

```suggestion
class MaxIterationReached : public IterationDiverge {
```

mgovers, line 115: please also add to the python exceptions incl parsing

mgovers, line 115: do we also want to update the `MathSolvers`?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -665,8 +672,13 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
+                    throw MaxIterationReached{
```
mgovers, line 679: i believe that if the optimizer throws, that the cache is not restored. this is a bug in the current implementation because it will already run into this edge case when a `SparseMatrixError` or `IterationDiverge` happens in the solver and it is not able to recover by re-running with the fallback method. Will create a Jira ticket

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
Jerry-Jinfeng-Guo: After discussion: original `IterationDiverge` is probably not terribly accurate, but for the sake of keeping minimum change to the user side, `MaxIteration` error should derive from it and only mention the maximum iteration reached situation. This has nothing to do with convergence per se.

Jerry-Jinfeng-Guo: See discussion above.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: Conclusion: maximum tap range should be taken

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -106,12 +106,18 @@ class NotObservableError : public PowerGridError {
 
 class IterationDiverge : public PowerGridError {
   public:
+    IterationDiverge() {}
```
mgovers, line 109:
```suggestion
    IterationDiverge() = default;
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: did you forget to push this by mistake?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -354,6 +354,9 @@ template <transformer_c... TransformerTypes> class TransformerWrapper {
     IntS tap_max() const {
         return apply([](auto const& t) { return t.tap_max(); });
     }
+    IntS tap_range() const {
+        return apply([](auto const& t) { return std::abs(t.tap_max() - t.tap_min()); });
```
mgovers, line 358: this will wrap around if `|tap_min - tap_max| >= 128`... not sure what would be a good solution though

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -665,8 +673,14 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
                 if (tap_changed) {
                     break;
                 }
+                iterations_per_rank[++rank_index] = 0;
             }
             if (tap_changed) {
+                if (static_cast<unsigned long>(++iterations_per_rank[rank_index]) >
+                    2 * max_tap_ranges_per_rank[rank_index]) {
+                    throw MaxIterationReached{
+                        "TapPositionOptimizer::iterate, maximum iterations reached, no solution."};
```
mgovers, line 682: it may be useful to add the `maximum iterations reached` to the constructor of the exception. that will also help when making the python exception parser

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -736,6 +750,9 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
             // min voltage at control side => max voltage at tap side => max tap pos
             return transformer.tap_max();
         };
+        constexpr auto tap_pos_range_cmp = [](RegulatedTransformer const& a, RegulatedTransformer const& b) {
+            return a.transformer.tap_range() < b.transformer.tap_range();
```
mgovers, line 754: probably this is a good location to take care of the wraparound

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -997,6 +987,85 @@ TEST_CASE("Test Tap position optimizer") {
...
+        state_b.u_pu = [&state_b, &regulator_b](ControlSide side) {
+            CHECK(side == regulator_b.control_side());
+
+            // tap pos closer to tap_max at tap side <=> lower voltage at control side
+            return static_cast<DoubleComplex>(test::normalized_lerp(state_b.tap_pos, state_b.tap_max, state_b.tap_min));
+        };
+
+        auto update_data = TransformerTapRegulatorUpdate{.id = 4, .u_set = 0.5, .u_band = 0.0};
+
+        // TODO(jguo): final sub test fail -> move to stand alone test to catch error
```
mgovers, line 1042: what do you mean?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -997,6 +987,85 @@ TEST_CASE("Test Tap position optimizer") {
...
+        SUBCASE("no valid value in band") {
+            state_b.tap_min = 1;
+            state_b.tap_max = 5;
+            state_b.tap_pos = 3;
```
mgovers, line 1046: due to the wraparound case: may be good to add an edge case with really large `tap_max - tap_min`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: could use some upgrade ;)

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: the upgrade renders this test unnecessary.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: meaning this was originally left for today to patch up but I found this to be minimum work yesterday and did it after all (forgetting a couple things of course...)

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -1004,6 +994,83 @@ TEST_CASE("Test Tap position optimizer") {
...
+
+        auto update_data = TransformerTapRegulatorUpdate{.id = 4, .u_set = 0.5, .u_band = 0.0};
+
+        SUBCASE("no valid value in band") {
+            state_b.tap_min = 1;
+            state_b.tap_max = 5;
+            state_b.tap_pos = 3;
+
+            update_data.u_set = 0.4;
+            update_data.u_set = 0.01;
```
nitbharambe, line 1054: Seems like you missed this

Jerry-Jinfeng-Guo: Oops. Either is valid

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -578,6 +583,7 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
     using TransformerRanker = TransformerRanker_;
 
   private:
+    mutable std::vector<uint64_t> max_tap_ranges_per_rank{};
```
nitbharambe, line 586: tap_min and tap_max are not going to change for a input_data.
I wonder if we can just have tap_range at `RegulatedObjects` instead. They can be calculated in `retrieve_regulator_info()`.
And then the max iteration of each rank in `RankedTransformerGroups`. They can be calculated at `rank_transformers()`

Jerry-Jinfeng-Guo: > tap_min and tap_max are not going to change for a input_data.
I wonder if we can just have tap_range at RegulatedObjects instead. They can be calculated in retrieve_regulator_info().

`tap_range` is already available in transformers as a function call. In terms of number of calls, they are the same whether they live as a member or as a function call, because only once will they be required throughout the code (here). If we put everything inside `RegulatedObjects`, which is container of indices, I am not sure if this is making things less scattered or more so XD

> And then the max iteration of each rank in RankedTransformerGroups. They can be calculated at rank_transformers()

Having this per rank information inside `RankedTransformerGroups` indeed makes sense, and that was my original code as well. After bumping into some template mismatch, I figure it might also make sense that this is a per optimizer requirement, and not necessarily order group property (which is again groups of indices).

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -1004,6 +994,81 @@ TEST_CASE("Test Tap position optimizer") {
...
+            state, 4,
+            TransformerTapRegulatorInput{.id = 4,
+                                         .regulated_object = 2,
+                                         .status = 1,
+                                         .control_side = ControlSide::side_2,
+                                         .u_set = 0.0,
+                                         .u_band = 0.0,
+                                         .line_drop_compensation_r = 0.0,
+                                         .line_drop_compensation_x = 0.0},
+            transformer_b.math_model_type(), 1.0);
```
nitbharambe, line 1028: Would be nice to avoid repetition here

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -666,8 +676,13 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
                 if (tap_changed) {
                     break;
                 }
+                iterations_per_rank[++rank_index] = 0; // NOSONAR
```
nitbharambe, line 679: out of curiosity, what does sonar say?

Jerry-Jinfeng-Guo: something like "omg, more than one thing at a line"

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers, line 586: i know it's already been merged and it doesn't necessarily need active changing, but what's the reasons you need `mutable`?

using `mutable` is usually dangerous because it overrides **any** `const` statement. that means that a statement like

```cpp
#include <iostream>

class A{
 private:
  int regular_a{};
  mutable int mutable_a{};

 public:
  int get_a() const {
    // return a++;  // compile error: instance is const
    return regular_a;       // OK
  }

  int get_mutable_a() const {
    return mutable_a++;  // OK: mutable overrides const
  }
};

void print_a(const A& const_a) {
    std::cout << "a: ";
    for (auto _ : {0, 1, 2}) {
        std::cout << const_a.get_a() << ", ";
    };
    std::cout << "\n";
}
void print_mutable_a(const A& const_a) {
    std::cout << "mutable a: ";
    for (auto _ : {0, 1, 2}) {
        std::cout << const_a.get_mutable_a() << ", ";
    };
    std::cout << "\n";
}

int main() {
    A const const_a{};
    print_a(const_a);
    print_mutable_a(const_a);
    return 0;
}
```
with output:
```txt
a: 0, 0, 0
mutable_a: 0, 1, 2
```

using `mutable` is usually only useful when dealing with things like loggers or other boilerplate code with intentional side effects on otherwise constant objects. in most other cases, changing the `const`ness in the function declaration or changing the order of function calls is usually a better way

mgovers (commented):
just want to understand one specific part of the code change. not overruling the approval in any way cause i do like the general solution

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: The `mutable` was used due to the fact that the only place that it seemingly fit is in the `initialize` function. Which is not actually the case because the function name is misleading. Since for now it's a private member and only exposed to this misleading function, active changing is indeed not needed. In the possible following-up refactoring, we could work a bit on the `initialize` and `exploit_neighbourhood` and move the max iteration mechanism somewhere proper.

Jerry-Jinfeng-Guo: Hahaha, that's the option I consider even more frightening XD

Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/pull/617

PR #615, opened 2024-05-31
[Feature] add automatic tap adjustment to the transformer notebook

mgovers:


docs/examples/Transformer Examples.ipynb
```diff
@@ -533,16 +527,1066 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "01a473bd",
+   "metadata": {},
+   "source": [
+    "### Automatic tap changing with line drop compensation\n",
+    "\n",
+    "In the above examples, the voltage at the low-voltage side of the transformer is controlled. However, the load is still connected to a node with a voltage lower than 390 V.\n",
+    "\n",
+    "To regulate the node that the lode is connected to, the voltage needs to be compensated for the voltage drop over the line.\n",
```
nitbharambe, line 1422:
```suggestion
    "To regulate the node that the load is connected to, the voltage needs to be compensated for the voltage drop over the line.\n",
```

docs/examples/Transformer Examples.ipynb
```diff
@@ -533,16 +527,1066 @@
...
+    "\n",
+    "```txt\n",
+    "                   (tap_side)  (control side)\n",
+    "source_1 --- node_2 --- transformer_3 --- node_4 --- line_5 --- node_6 --- load_7\n",
+    "                        |                 |                         :\n",
+    "                        regulator_8 <-----/ <......................./ (line drop compensation)\n",
+    "                                     (control voltage) \n",
+    "```"
```
nitbharambe, line 1430: Resolve without changes:

I wish there was a draw.io like tool for power grid. What do you think about changing these to hand drawn image svgs from whiteboard instead?

docs/examples/Transformer Examples.ipynb
```diff
@@ -533,16 +527,1066 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "id": "0c89a335",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# node\n",
+    "node = initialize_array(\"input\", \"node\", 3)\n",
```
nitbharambe, line 596: We probably can do a copy here as well from `input_data`. And in description above this cell, refer to `#input-dataset` section

docs/examples/Transformer Examples.ipynb
mgovers: we have a couple options: we can use GraphViz or we can get some input from the team that's working on the data science toolkit

mgovers: let's talk about it when discussing the improvmenets this summer

docs/examples/Transformer Examples.ipynb
```diff
@@ -533,16 +527,1066 @@
...
+    "print(\"------node result------\")\n",
+    "display(pd.DataFrame(output_data5[\"node\"])[[\"id\", \"u\"]])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "07af8e93",
+   "metadata": {},
+   "source": [
+    "### Power flow with any voltage tap position strategy\n",
```
nitbharambe, line 771:
```suggestion
    "#### Any voltage tap position strategy\n",
```
And for sections below too

docs/examples/Transformer Examples.ipynb
mgovers: unfortunately not because we need 3 nodes here instead of 2. alternatively, we can copy the ones from `input_data` here but then we would need to create a full new one with 3 nodes for the line drop compensation. I think it's better to only do one input data set for the auto tap changing entirely. I would propose to keep it as is

nitbharambe (approved):

PR #617, opened 2024-06-02
Feature / State restoration after exception in optimizer

Jerry-Jinfeng-Guo:
- [x] state restoration after exception
- [x] unit test
- [x] unit test refactor
- [x] polish max iteration exception mechanism

### DIFF
--- a/tests/cpp_unit_tests/test_tap_position_optimizer.cpp
+++ b/tests/cpp_unit_tests/test_tap_position_optimizer.cpp
@@ -564,6 +564,45 @@ auto normalized_lerp(IntS value, IntS start, IntS stop) {
     return (static_cast<double>(value) - static_cast<double>(start)) /
            (static_cast<double>(stop) - static_cast<double>(start));
 }
+void checkNormalTapRange(MockTransformerState& state_b, TransformerTapRegulator& regulator_b,
+                         TransformerTapRegulatorUpdate& update_data, auto& check_b) {
+    state_b.tap_min = 1;
+    state_b.tap_max = 5;
+    state_b.tap_pos = 3;
+
+    SUBCASE("unique value in band") {
+        update_data.u_band = 0.01;
+        check_b = test::check_exact(3);
+    }
+    SUBCASE("large compact band") {
+        update_data.u_band = 1.01;
+        check_b = test::check_exact_per_strategy(3, 5, 1);
+    }
+    SUBCASE("small open band") {
+        update_data.u_band = 0.76;
+        check_b = test::check_exact_per_strategy(3, 4, 2);
+    }
+}
+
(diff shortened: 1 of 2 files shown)


tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -1025,7 +1047,9 @@ TEST_CASE("Test Tap position optimizer") {
                     state_a.tap_side = tap_side;
 
                     auto optimizer = get_optimizer(strategy);
+                    auto const cached_state = state;
                     CHECK_THROWS_AS(optimizer.optimize(state, CalculationMethod::default_method), MaxIterationReached);
+                    CHECK(twoStatesEqual(cached_state, state));
```
mgovers, line 1052: with the current initial tap position being `3` it actually tests nothing.

To repro: you can do some TDD-style backtracing

* checkout this branch for the tests and `origin/main` for the tap position optimizer implementation
  ```bash
  git checkout origin/main -- power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```
* build + run tests
  * see that they pass
  * this is not expected, because it means the original code was already compliant with this test, which should not be the case
* To debug:
  * set breakpoints before it enters the `optimizer.optimize` and in the mock transformer update function
  * step into this subcase
  * initial `tap_pos == 3`
  * it will repeatedly update between `tap_pos == 4` and `tap_pos == 3`
  * due to the even number of updates, the final `tap_pos == 3`
  * `twoStatesEqual` will therefore return `true`
* To fix:
  * set the initial to `state_b.tap_pos = 2`
    * rebuild + run the tests
    * Now, the `origin/main` version of this fails
  * restore the current branch implementation:
    * `git reset power_grid_model_c && git checkout power_grid_model_c` to checkout the current implementation on this branch
  * rebuild + run the tests
    * now, the test passes again, meaning the new implementation really is the fix for the problem

File: tests/cpp_unit_tests/test_tap_position_optimizer.cpp [hunk 2]
```diff
@@ -597,6 +634,25 @@ TEST_CASE("Test Tap position optimizer") {
                                                      std::back_inserter(changed_components));
     };
 
+    auto twoStatesEqual = [](const MockState& state1, const MockState& state2) {
+        if (state1.components.template size<MockTransformer>() != state2.components.template size<MockTransformer>()) {
+            return false;
+        }
+        std::vector<std::pair<IntS, IntS>> trafo_state_1, trafo_state_2;
```
mgovers, line 641: sonar cloud
```suggestion
        std::vector<std::pair<IntS, IntS>> trafo_state_1
        std::vector<std::pair<IntS, IntS>> trafo_state_2;
```

mgovers (approved):
nice

PR #621, opened 2024-06-04
Make tap_pos optional

### LINKED ISSUE #576: [FEATURE] make tap_pos optional
For users that want to use the automatic tap changer power flow, it does not really matter what the original `tap_pos` is. They are just interested in the final `tap_pos`, found by the automatic tap algorithm. At this moment `tap_pos` is required, since it is necessary for the other algorithms.

We can make `tap_pos` optional, if we assign it the value of `tap_nom` when it is not given

Update:
- [ ] transformer
- [ ] three winding transformer
- [ ] documentation
- [ ] validation checks
- [ ] unit tests should be updated

NOTE: `tap_nom` is Optional. If `tap_nom` is not given, it will be set to 0 as a default. In this case `tap_pos` will still be set to the value of `tap_nom = 0`.

petersalemink95:
closes #576 

- [x] transformer
- [x] three winding transformer
- [x] documentation
- [x] validation checks
- [x] validation unit tests 
- [x] rules unit tests
- [x] C++ unit tests should be updated


power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -37,7 +37,9 @@ class Transformer : public Branch {
           winding_to_{transformer_input.winding_to},
           clock_{transformer_input.clock},
           tap_side_{transformer_input.tap_side},
-          tap_pos_{transformer_input.tap_pos},
+          tap_pos_{transformer_input.tap_pos == na_IntS
+                       ? (transformer_input.tap_nom == na_IntS ? (IntS)0 : transformer_input.tap_nom)
```
mgovers, line 41:
```suggestion
                       ? (transformer_input.tap_nom == na_IntS ? IntS{0} : transformer_input.tap_nom)
```

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -71,14 +73,14 @@ class Transformer : public Branch {
...
-    constexpr IntS tap_pos() const { return tap_pos_; }
-    constexpr BranchSide tap_side() const { return tap_side_; }
-    constexpr IntS tap_min() const { return tap_min_; }
-    constexpr IntS tap_max() const { return tap_max_; }
-    constexpr IntS tap_nom() const { return tap_nom_; }
+    IntS tap_pos() const { return tap_pos_; }
+    BranchSide tap_side() const { return tap_side_; }
+    IntS tap_min() const { return tap_min_; }
+    IntS tap_max() const { return tap_max_; }
+    IntS tap_nom() const { return tap_nom_; }
```
mgovers, line 80: why are these not `constexpr` anymore? there's no complicated logic in the function that requires it to be non-`constexpr` and it may help the compiler

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -265,6 +267,4 @@ class Transformer : public Branch {
     }
 };
 
-static_assert(transformer_c<Transformer>);
-
```
mgovers, line 269: ?

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
petersalemink95: This is weird, I didn't do that; guess the formatter updated this -> I'll dive into it

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
petersalemink95: Same here

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
petersalemink95: Why this comment on this specific branch? It's written this way everywhere in the C++ code (see `tap_nom_` and `tap_direction_` for instance)

nitbharambe, line 41: How about we set it at `tap_nom` by default instead of `0`? That would disable the taps and make it safer.

petersalemink95: @nitbharambe it is set to `tap_nom`, but `tap_nom` is optional, so if `tap_nom` it is not specified default will be 0, just as for `tap_nom`

nitbharambe: Oh my bad. I guess I got confused by the double `?`

mgovers: it's a separate issue then 😬 . In general, logic like `(IntS)0` will create a 64-bit (or otherwise depending on the OS + hardware) and then reinterpret that as an 8-bit value. For `0`, that's not really an issue, but it may cause issues for other values.

Instead, `IntS{0}` or `IntS{}` will immediately create the value as 8-bit integer and initialize as `0`.

There are 2 ways to resolve this: either using incremental improvements by changing it whenever we touch the code, or by going over the entire code base in one go and change all occurrences. I'm fine either way but would like to resolve this sometime

mgovers: we can also initialize `tap_nom` before `tap_pos` and then obtain `tap_pos` from `tap_nom`, e.g.

```cpp
tap_nom_{transformer_input.tap_nom == na_IntS ? IntS{0} : transformer_input.tap_nom},
tap_pos_{transformer_input.tap_pos == na_IntS ? tap_nom_ : transformer_input.tap_pos}
```

note that to stop (especially Unix) compilers from complaining, the order of the attributes at the declaration needs to be changed as well (`-Wreorder`)

petersalemink95: I think it should be fine like this. The documentation mentions how this initialization is done; if things are unclear

petersalemink95: Thanks for the response. For now I updated all `(IntS)0` to `IntS{0}` in transformer and 3w transformer .hpp

src/power_grid_model/validation/rules.py
```diff
@@ -316,7 +319,7 @@ def outside(val: np.ndarray, *ref: np.ndarray) -> np.ndarray:
         return np.logical_or(np.less(val, np.minimum(*ref)), np.greater(val, np.maximum(*ref)))
 
     return none_match_comparison(
-        data, component, field, outside, (ref_value_1, ref_value_2), NotBetweenOrAtError, default_value
+        data, component, field, outside, (ref_value_1, ref_value_2), NotBetweenOrAtError, default_value, default_value_2
```
Jerry-Jinfeng-Guo, line 322: Is there a deeper reason the parameters aren't renamed accordingly? Here and everywhere else necessary ofc.
```suggestion
        data, component, field, outside, (ref_value_1, ref_value_2), NotBetweenOrAtError, default_value_1, default_value_2
```

tests/unit/validation/test_rules.py
```diff
@@ -143,6 +143,19 @@ def test_all_between_or_at():
     assert len(errors) == 1
     assert NotBetweenOrAtError("test", "value", [1, 4, 5, 6], (0.2, -0.2)) in errors
 
+    int8_nan = power_grid_meta_data["input"]["transformer"].nans["tap_pos"]
```
mgovers, line 146: maybe rename this to just

```suggestion
    nan_value = power_grid_meta_data["input"]["transformer"].nans["tap_pos"]
```

src/power_grid_model/validation/rules.py
petersalemink95: @Jerry-Jinfeng-Guo no particular reason. made it default 1/2

docs/user_manual/components.md
```diff
@@ -301,7 +301,7 @@ voltage levels. An example of usage of three-winding transformer is given in [Tr
...
 | `tap_side`      | {py:class}`Branch3Side <power_grid_model.enum.Branch3Side>` | -                | side of tap changer                                                                                       |             &#10004;             | &#10060; |                    `side_1` or `side_2` or `side_3`                    |
-| `tap_pos`       | `int8_t`                                                    | -                | current position of tap changer                                                                           |             &#10004;             | &#10004; | `(tap_min <= tap_pos <= tap_max)` or `(tap_min >= tap_pos >= tap_max)` |
+| `tap_pos`       | `int8_t`                                                    | -                | current position of tap changer                                                                           |             &#10060; default `tap_nom`, if no `tap_nom` default `0`             | &#10004; | `(tap_min <= tap_pos <= tap_max)` or `(tap_min >= tap_pos >= tap_max)` |
```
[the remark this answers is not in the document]
petersalemink95: That would already raise an error. Both for `tap_pos` and `tap_nom` we do a check that it's between `tap_min` and `tap_max`. If `tap_min` and `tap_max` are both >0 then you should just specify the value as a user. We don't do data corrections -> garbage in = garbage out

src/power_grid_model/validation/rules.py
```diff
@@ -347,13 +351,18 @@ def none_match_comparison(
         default_value: Some values are not required, but will receive a default value in the C++ core. To do a proper
         input validation, these default values should be included in the validation. It can be a fixed value for the
         entire column (int/float) or be different for each element (np.ndarray).
+        default_value_2: Some values can have a double default: the default will be set to another attribute of the
+        component, but if that attribute is missing, the default will be set to a fixed value.
```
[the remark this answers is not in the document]
petersalemink95: I think it's nice to mention the waterfall effect here

docs/user_manual/components.md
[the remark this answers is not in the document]
mgovers: perfect

mgovers:
@petersalemink95 is this ready for final review/merge?

mgovers (commented):
overall LGTM. can you please resolve the sonar cloud warnings?

Jerry-Jinfeng-Guo (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -53,6 +52,13 @@ class Transformer : public Branch {
               calculate_z_pu(transformer_input.r_grounding_from, transformer_input.x_grounding_from, u1_rated)},
           z_grounding_to_{
               calculate_z_pu(transformer_input.r_grounding_to, transformer_input.x_grounding_to, u2_rated)} {
+        // init tap_pos_ linter smell
+        if (transformer_input.tap_pos == na_IntS) {
+            tap_pos_ = transformer_input.tap_nom == na_IntS ? IntS{0} : transformer_input.tap_nom;
+        } else {
+            tap_pos_ = transformer_input.tap_pos;
+        }
+
```
nitbharambe, line 61: I think this can go in the transformer_utils.hpp since  its repeated

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
nitbharambe: As I tried to do this, i saw there is a whole lot of other duplicated code too. A good amount of refactoring would be needed here. Hence lets do that later

Jerry-Jinfeng-Guo: Could use an immediate jira ticket ;)

PR #625, opened 2024-06-05
temporarily add DCO check

mgovers:


Jerry-Jinfeng-Guo (dismissed):

Jerry-Jinfeng-Guo:
For other approved PRs, do we need to push empty commit to trigger the ci?

mgovers:
> For other approved PRs, do we need to push empty commit to trigger the ci?

that wouldn't be sufficient because the workflow wouldn't be there. merge with main would do the trick

mgovers:
merge queue build failed due to https://github.com/actions/runner-images/issues/10001

PR #635, opened 2024-06-07
[Fix] handle nullptr on serialization error

### LINKED ISSUE #582: [BUG] Serialization should fail with a readable error when required memory cannot be allocated
**Describe the bug**

Serialization fails with an difficult to interpret error message. Likely cause as per the msgpack error meassage is that required memory cannot be allocated.

**To Reproduce**

A big output dataset was attempted to serialize to file using json and msgpack serializer.

**Expected behavior**

Some understandable error message like `PowerGridSerializationError` would make it clear.

**Screenshots**

json_serialize_to_file gives out following:
image

same data with msgpack export it gives out `bad allocation` message.

mgovers:
Closes #582 

Fix unintelligible error message when serializing to JSON fails.

The behavior to msgpack has not changed because "bad allocation" really is the only valid error message: we don't know what or where it goes wrong, but there was a memory allocation that was bad.

## Current `main` (before bugfix)

Behavior depends on serialization format

### serialize to msgpack

Throw `PowerGridSerializationError` with the message `bad allocation`

image

### serialize to JSON

Throw `AttributeError` with the message `'NoneType' object has no attribute 'decode'`

image

## After

Throw `PowerGridSerializationError` with the message `bad allocation`

### serialize to msgpack

image

### serialize to JSON

image


tests/c_api_tests/test_serialization.cpp
```diff
@@ -51,33 +51,47 @@ TEST_CASE("Serialization") {
             auto* const serializer = json_serializer.get();
             CHECK(PGM_error_code(hl) == PGM_no_error);
 
-            // to string
-            std::string json_result = PGM_serializer_get_to_zero_terminated_string(hl, serializer, 0, -1);
-            CHECK(PGM_error_code(hl) == PGM_no_error);
-            CHECK(json_result == json_data);
+            SUBCASE("to zero-terminated string") {
```
Jerry-Jinfeng-Guo, line 54:
```suggestion
            SUBCASE("To zero-terminated string") {
```

tests/c_api_tests/test_serialization.cpp
```diff
@@ -51,33 +51,47 @@ TEST_CASE("Serialization") {
...
+            }
 
-            // to buffer
-            char const* buffer_data{};
-            Idx buffer_size{};
-            PGM_serializer_get_to_binary_buffer(hl, serializer, 0, &buffer_data, &buffer_size);
-            CHECK(PGM_error_code(hl) == PGM_no_error);
-            std::string const json_string{buffer_data, static_cast<size_t>(buffer_size)};
-            CHECK(json_result == json_string);
+            SUBCASE("to binary buffer") {
```
Jerry-Jinfeng-Guo, line 60:
```suggestion
            SUBCASE("To binary buffer") {
```

src/power_grid_model/core/power_grid_core.py
```diff
@@ -191,7 +191,7 @@ def cbind_func(self, *args, **kwargs):
         res = getattr(_CDLL, f"PGM_{name}")(*c_inputs)
         # convert to string for CStr
         if c_restype == CStr:
-            res = res.decode()
+            res = res.decode() if res is not None else ""
```
Jerry-Jinfeng-Guo, line 194: so this is the bug

mgovers: Yep, the C API was behaving as intended but the Python wrapper wasn't

mgovers:
You can apply the comments yourself @Jerry-Jinfeng-Guo , i agree with them but i can't sign off with my alliander mail on this device

Jerry-Jinfeng-Guo (approved):

PR #638, opened 2024-06-11
Drop python 3.9

Jerry-Jinfeng-Guo:


mgovers:
are you sure it's a good idea to do this right after the release of `v1.8.0`? what if there are important bugfixes? I'd propose to do this change at least a week from now

TonyXiang8787 (approved):

mgovers:
@TonyXiang8787 @Jerry-Jinfeng-Guo  I think it's good to wait with dropping Python 3.9 support until #672 is merged so that users that are pinned to Python 3.9 still have the benefits of that

mgovers:
Scheduled for Monday morning (to prevent Friday afternoon releases)

PR #648, opened 2024-06-19
Feature/metadata datatypes

### LINKED ISSUE #613: [FEATURE] Power Grid Meta Data components enum
## The issue with the current implementation

Currently, all power grid model components are listed only by their strings on the python side. That is bugprone, because typos might go undetected until they end up in production.

Additional problems include:

* string data types do not support type hinting.
* pylint and mypy don't understand the supported string data types
* reviewing may be difficult, both for the user and for the developers

## Proposed solution

* [ ] Runtime creating two `Enum`s from the `power_grid_meta_data` makes sure the enums are always up-to-date. The following code snippet illustrates how that would be possible.
  * [ ] one for the data types
  * [ ] one for the component types
* [ ] Expose in the `__init__.py`
* [ ] Update input data types in all functions where e.g. `data_type` and `component_type` are used
* [ ] Update tests
* [ ] Update docs
* [ ] Update examples
* [ ] Update workshop

```py
>>> from enum import Enum
>>> from power_grid_model import power_grid_meta_data
>>>
>>> power_grid_meta_data.keys()
dict_keys(['input', 'update', 'sym_output', 'asym_output', 'sc_output'])
>>>
>>> PowerGridDataTypes = Enum("PowerGridDataTypes", {data_type: data_type for data_type in power_grid_meta_data}) 
>>>
>>> PowerGridDataTypes
<enum 'PowerGridDataTypes'>
>>> list(PowerGridDataTypes)
[<PowerGridDataTypes.input: 'input'>, <PowerGridDataTypes.update: 'update'>, <PowerGridDataTypes.sym_output: 'sym_output'>, <PowerGridDataTypes.asym_output: 'asym_output'>, <PowerGridDataTypes.sc_output: 'sc_output'>]
>>>
>>> PowerGridComponents = Enum("PowerGridComponents", {component: component for component in power_grid_meta_data[PowerGridDataTypes.input]})
>>> PowerGridComponents
<enum 'PowerGridComponents'>
>>> list(PowerGridComponents)  
[<PowerGridComponents.node: 'node'>, <PowerGridComponents.line: 'line'>, <PowerGridComponents.link: 'link'>, <PowerGridComponents.transformer: 'transformer'>, <PowerGridComponents.three_winding_transformer: 'three_winding_transformer'>, <PowerGridComponents.shunt: 'shunt'>, <PowerGridComponents.source: 'source'>, <PowerGridComponents.sym_gen: 'sym_gen'>, <PowerGridComponents.asym_gen: 'asym_gen'>, <PowerGridComponents.sym_load: 'sym_load'>, <PowerGridComponents.asym_load: 'asym_load'>, <PowerGridComponents.sym_power_sensor: 'sym_power_sensor'>, <PowerGridComponents.asym_power_sensor: 'asym_power_sensor'>, <PowerGridComponents.sym_voltage_sensor: 'sym_voltage_sensor'>, <PowerGridComponents.asym_voltage_sensor: 'asym_voltage_sensor'>, <PowerGridComponents.fault: 'fault'>, <PowerGridComponents.transformer_tap_regulator: 'transformer_tap_regulator'>]
```

### LINKED ISSUE #657: [Release] v1.9.0
Decision during standup: minor version bump due to potential pylint/mypy stuff for our users from #648 

**Before release**

* [x] Update [`VERSION`](#VERSION) file if applicable
* [x] Make PR
* [ ] Merge PR

**After release**

* [ ] Update the release notes
  * [ ] Add `Highlights` section if needed
  * [ ] Any breaking changes: explain + provide migration approach
  * [ ] Any deprecations: explain + provide migration approach
* [ ] Share with stakeholders
  * [ ] Notify maintainers
  * [ ] Post to PGM LF Energy mailing list
  * [ ] Post to PGM LF Energy Slack channel
  * [ ] Post to LinkedIn
  * [ ] Share in sprint review

figueroa1395:
Closes #613
Closes #657 

Implement data types for the power grid dataset and component types:

1. Add a code generator template (Jinja) to create both datatypes as the `Union` between an `Enum` (for new datatype support, e.g. `DataType`.input, `Component`.node) and `Literal` (to keep old support, e.g. “input“, “node“).
2. Run `code_gen.py`.
3. Expose new datatypes in `__init__.py`.
4. Update input data types in all functions where e.g. data_type and component_type are used.
5. Add unit test.
6. Update docs.


figueroa1395:
I found an issue and I can't take back to draft review. Don't merge.

mgovers (commented):
Posting comments to keep unresolved as blocker

src/power_grid_model/__init__.py
```diff
@@ -6,6 +6,7 @@
 
 from power_grid_model.core.power_grid_meta import initialize_array, power_grid_meta_data
 from power_grid_model.core.power_grid_model import PowerGridModel
+from power_grid_model.dataset_definitions import ComponentType, DataType
```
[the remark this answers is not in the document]
nitbharambe: These names are too similar. 
Lets have the type hints (ie. `ComponentType`) in the data_types.py along with `SingleDataset` etc. 
Or maybe a separate file `typing.py` or something for the distinction.

It feels out of place in `__init__.py`. Here we can only keep the enum, `ComponentTypes`.

mgovers:
> I found an issue and I can't take back to draft review. Don't merge.

I converted it back to `draft`. You can always add the `do not merge` label as well

code_generation/code_gen.py
```diff
@@ -117,17 +117,45 @@ def render_dataset_class_maps(self, template_path: Path, data_path: Path, output
...
+                prefixes = ["sym_", "asym_"]
+            else:
+                prefixes = [""]
+            for prefix in prefixes:
+                dataset_types.append(f"{prefix}{dataset.name}")
+
+                if dataset.name == "input":
+                    for component in dataset.components:
+                        components.append(component.names)
+
```
nitbharambe, line 138: It feels like complexity here is more than necessary.

Correct me if im wrong, but wouldnt all components simply be:
`components = [name for components in dataset_meta_data["all_datasets"]["input"]["components"] for name in components["names"]]`

And 
`data_types = [dataset["name"] for dataset in dataset_meta_data["all_datasets"]]`

code_generation/code_gen.py
```diff
@@ -117,17 +117,45 @@ def render_dataset_class_maps(self, template_path: Path, data_path: Path, output
...
         }
 
         # render attribute classes
         for template_name, render_func in render_funcs.items():
             for template_path in TEMPLATE_DIR.rglob(f"{template_name}.*.jinja"):
                 output_suffix = template_path.with_suffix("").suffix
                 output_dir = template_path.parent.relative_to(TEMPLATE_DIR)
+                if template_name == "metadata_enums":
+                    template_name = "dataset_class_maps"  # To use existing data.
                 for data_path in DATA_DIR.glob(f"{template_name}/*.json"):
```
nitbharambe, line 159:
```suggestion
                data_name = "dataset_class_maps" if template_name == "metadata_enums" else template_name
                for data_path in DATA_DIR.glob(f"{data_name}/*.json"):
```
Sort of a nitpick here for a script. But creating new variable is always better than modifying existing one.

scripts/quick_example.py
```diff
@@ -41,7 +45,12 @@
 source["u_ref"] = [1.0]
 
 # input_data
-input_data = {"node": node, "line": line, "sym_load": sym_load, "source": source}
+input_data: Dict[ComponentType, np.ndarray] = {
+    "node": node,
+    "line": line,
+    "sym_load": sym_load,
+    "source": source,
+}
```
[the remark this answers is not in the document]
nitbharambe: Why not both? :P enum with hint for LHS.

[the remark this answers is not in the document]
nitbharambe: Oh, better is to name this as `SingleDataset`. And update the SingleDataset with the new type hint too

src/power_grid_model/core/power_grid_dataset.py
```diff
@@ -375,7 +389,7 @@ def __init__(self, dataset_ptr: WritableDatasetPtr):
         self._schema = power_grid_meta_data[self._dataset_type]
 
         self._component_buffer_properties = self._get_buffer_properties(info)
-        self._data: Dict[str, Union[np.ndarray, Dict[str, np.ndarray]]] = {}
+        self._data: Dict[ComponentType, Union[np.ndarray, Dict[str, np.ndarray]]] = {}
```
nitbharambe, line 392:
```suggestion
        self._data: Dataset = {}
```
Or maybe check if it is some other dataset type in the data_types.py

src/power_grid_model/validation/errors.py
```diff
@@ -317,19 +320,19 @@ class InvalidIdError(SingleFieldValidationError):
...
         field: str,
         ids: List[int],
-        ref_components: Union[str, List[str]],
+        ref_components: Union[ComponentType, List[ComponentType]],
         filters: Optional[Dict[str, Any]] = None,
     ):
         # pylint: disable=too-many-arguments
         super().__init__(component=component, field=field, ids=ids)
-        self.ref_components = [ref_components] if isinstance(ref_components, str) else ref_components
+        self.ref_components = [ref_components] if isinstance(ref_components, str) else ref_components  # type: ignore
```
nitbharambe, line 335: Will this make the warning disappear?

```suggestion
        self.ref_components = [ref_components] if isinstance(ref_components, (str, ComponentTypes)) else ref_components  # type: ignore
```

src/power_grid_model/validation/utils.py
```diff
@@ -161,7 +162,7 @@ def errors_to_string(
     return msg
 
 
-def nan_type(component: str, field: str, data_type="input"):
+def nan_type(component: ComponentType, field: str, data_type: DataType = "input"):
```
nitbharambe, line 165:
```suggestion
def nan_type(component: ComponentType, field: str, data_type: DataType = DataTypes.input):
```

src/power_grid_model/validation/utils.py
```diff
@@ -243,7 +246,7 @@ def get_valid_ids(data: SingleDataset, ref_components: Union[str, List[str]]) ->
     """
     # For convenience, ref_component may be a string and we'll convert it to a 'list' containing that string as it's
     # single element.
-    if isinstance(ref_components, str):
+    if not isinstance(ref_components, list):
```
nitbharambe, line 249: While it is correct, there is no need to modify logic :D
```suggestion
    if isinstance(ref_components, (str, ComponentTypes)):
```

src/power_grid_model/utils.py
```diff
@@ -190,7 +191,7 @@ def import_json_data(json_file: Path, data_type: str, *args, **kwargs) -> Datase
     if kwargs:
         warnings.warn(f"Provided keyword arguments {list(kwargs.keys())} are deprecated.", DeprecationWarning)
 
-    return _compatibility_deprecated_import_json_data(json_file=json_file, data_type=data_type)
+    return _compatibility_deprecated_import_json_data(json_file=json_file, data_type=data_type)  # type: ignore
```
nitbharambe, line 194: The data_type argument can be of `DataType` and this suppresion might not be needed

src/power_grid_model/utils.py
```diff
@@ -274,10 +275,13 @@ def import_update_data(json_file: Path) -> BatchDataset:
     """
     warnings.warn(_DEPRECATED_JSON_DESERIALIZATION_MSG, DeprecationWarning)
 
-    return cast_type(BatchDataset, _compatibility_deprecated_import_json_data(json_file=json_file, data_type="update"))
+    return cast_type(
+        BatchDataset,
+        _compatibility_deprecated_import_json_data(json_file=json_file, data_type="update"),
```
nitbharambe, line 280:
```suggestion
        _compatibility_deprecated_import_json_data(json_file=json_file, data_type=DataTypes.update),
```

tests/unit/validation/test_assertions.py
```diff
@@ -28,10 +28,10 @@ def test_validation_exception(errors_to_string_mock: MagicMock):
 def test_assert_valid_input_data(validate_mock: MagicMock):
     validate_mock.return_value = None
     assert_valid_input_data(
-        input_data={"foo": np.array([1])}, calculation_type=CalculationType.state_estimation, symmetric=False
+        input_data={"foo": np.array([1])}, calculation_type=CalculationType.state_estimation, symmetric=False  # type: ignore
```
nitbharambe, line 31: I feel these suppresions might not be needed. And maybe "foo" can be modified to one of the supported type since now we only expect `ComponentTypes` instead of a a generic string.

mgovers: Well you can do 2 things: 1 is test what happens when it actuallily really is forbidden (needs suppression), the other is to remove it entirely because we now actually have the literals, but that means new UB and demanding the user uses a linter/type checker. I'd keep it as is

code_generation/code_gen.py
```diff
@@ -117,17 +117,45 @@ def render_dataset_class_maps(self, template_path: Path, data_path: Path, output
 
         self.render_template(template_path=template_path, output_path=output_path, all_map=all_map)
 
+    def render_metadata_types(self, template_path: Path, data_path: Path, output_path: Path):
```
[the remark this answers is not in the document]
figueroa1395: Mainly I wanted to separate responsibilities a bit (C++ from python generated code). Also, it kind of provided it but within `render_dataset_class_maps`, so I just made its own implementation.

code_generation/code_gen.py
figueroa1395: `data_types = [dataset["name"] for dataset in dataset_meta_data["all_datasets"]]` is incorrect as `dataset_meta_data["all_datasets"]]` doesn't include all datatypes. `sym_output` and `asym_output` come attached to `dataset.is_template` (this is how the json config. files are structured).

Making it a comprehension can also be done, but since the first loop is going to be used for components as well and the comprehension would be 'large', I think it is reasonable to keep it this way.

code_generation/templates/src/power_grid_model/metadata_enums.py.jinja
```diff
@@ -0,0 +1,77 @@
...
+
+# To avoid conflicts with src/power_grid_model/enum.py
+# pylint: disable=wrong-import-position
+
+sys_path = sys.path.pop(0)
+from enum import Enum
+
+# pylint: enable=wrong-import-position
+
+sys.path.append(sys_path)
```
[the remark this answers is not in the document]
figueroa1395: When moved to `src/power_grid_model/core` I will remove this as there is no name conflict there.

code_generation/templates/src/power_grid_model/metadata_enums.py.jinja
```diff
@@ -0,0 +1,77 @@
...
+
+
+DataType = Union[DataTypes, DataTypesLiteral]
+"""
+A DataType is the type of a :class:`BatchDataset`.
+
+- Examples: 
+
+    - DataType.input = "input"
+    - DataType.update = "update"
```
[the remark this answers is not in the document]
figueroa1395: Awaiting discussion with the rest of the team, but in principle I think we should only place examples of the new datatypes as it may lead to confusion because mixing both cannot be done if we want any sort of type annotations.

scripts/quick_example.py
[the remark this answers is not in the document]
figueroa1395: @nitbharambe I think being explicit with the component types as in @mgovers suggestion is more readable for the users. Also, I don't think we should show examples with both the old and new supported way of writing component types.

src/power_grid_model/__init__.py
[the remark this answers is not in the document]
figueroa1395: @nitbharambe Per my recent investigation, `Literal` does not really provide advantages in this scenario since dict[Literal, Any] is just interpreted as dict[str, Any] by Python, hence my new solution to this.

@mgovers In the end we will probably end up With `DataType` containing the `Enum` definition.

src/power_grid_model/core/power_grid_model.py
```diff
@@ -158,9 +159,9 @@ def _get_output_component_count(self, calculation_type: CalculationType):
             CalculationType.short_circuit: ["sensor"],
         }.get(calculation_type, [])
 
-        def include_type(component_type: str):
+        def include_type(component_type: ComponentType):
             for exclude_type in exclude_types:
-                if exclude_type in component_type:
+                if exclude_type == component_type:
```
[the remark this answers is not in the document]
figueroa1395: ?

src/power_grid_model/utils.py
figueroa1395: The suppression comes because the function is deprecated, hence I am not updating it with the new feature, but just ignoring the possible type error. Nevertheless, it is still backwards compatible.

src/power_grid_model/utils.py
figueroa1395: Deprecated function.

src/power_grid_model/validation/errors.py
figueroa1395: isinstance() doesn't work directly with a datatype such as ComponentTypes, hence why I added it.

src/power_grid_model/validation/utils.py
figueroa1395: I modified it because `isinstance()` gives me trouble with `ComponentTypes` :/. Any other suggestion to keep the logic is welcomed.

tests/unit/validation/test_assertions.py
figueroa1395: Since the literals don't really enforce type annotations, I believe it should be kept as it is as well.

src/power_grid_model/validation/utils.py
mgovers: what kind of trouble, specifically?

mgovers:
@figueroa1395 i recommend merging with `main` soon because there's a conflict

src/power_grid_model/core/power_grid_model.py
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: How did the old code not fail in the first place?

Jerry-Jinfeng-Guo (commented):
Just a general question: I see quite some linter disables in this PR, much more that average PRs. While this is for sure no average PR, but were these disables double checked?

src/power_grid_model/dataset_definitions.py
```diff
@@ -0,0 +1,129 @@
...
+
+    input = "input"
+
+    sym_output = "sym_output"
+
+    asym_output = "asym_output"
+
+    update = "update"
+
+    sc_output = "sc_output"
```
Jerry-Jinfeng-Guo, line 37: I don't think it is necessary to have empty lines here (and below), there is no logic purpose behind

src/power_grid_model/core/power_grid_model.py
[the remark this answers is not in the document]
mgovers: my guess (and nothing more than that) is that the string `"output"` was checked against things like `"sym_output"`, `"asym_output"` and `"sc_output"`, which for all cases is `True` in the original code.

which does mean that we now have to verify that that still works correctly.

code_generation/code_gen.py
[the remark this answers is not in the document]
mgovers: i believe most of the things are the same right? it's ok to have multiple things defined that you don't actually use in the jinja template, but since we don't do language-specific case-sensitive stuff, i don't really see a reason to duplicate this function.

the separation of concerns should be restricted to the actual template. only if we start generating different conventions for Python, R, Java, C#, ... whatever other languages we might support in the future, then we might want to start making separate generators for each of those languages.

but for the time being, it's almost only data declarations, so it's fine to combine them

code_generation/code_gen.py
mgovers: see my other component: this may be useful for multiple languages we might ever want to generate an API for

code_generation/templates/src/power_grid_model/metadata_enums.py.jinja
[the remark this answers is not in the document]
mgovers: alright 👍

code_generation/templates/src/power_grid_model/metadata_enums.py.jinja
```diff
@@ -0,0 +1,77 @@
...
+
+
+ComponentType = Union[ComponentTypes, ComponentTypesLiteral]
+"""
+A ComponentType is the type of a grid component.
+
+- Examples: 
+
+    - ComponentType.node = "node"
+    - ComponentType.line = "line"
```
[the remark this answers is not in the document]
mgovers: cfr. discussion above: i agree with keeping as is

src/power_grid_model/dataset_definitions.py
```diff
@@ -0,0 +1,129 @@
...
+
+# pylint: enable=wrong-import-position
+
+sys.path.append(sys_path)
+
+# Value names are defined in lower case instead of upper case
+# pylint: disable=invalid-name
+
+
+class DataTypes(Enum):
```
mgovers, line 26: the more i think about it, the more i'm thinking we should make this
```suggestion
class DatasetTypes(Enum):
```

cause `DataTypes` is kind of generic in programming context and therefore may be confusing to the reader/developer

src/power_grid_model/dataset_definitions.py
```diff
@@ -0,0 +1,129 @@
...
+    sc_output = "sc_output"
+
+
+DataTypesLiteral = Literal[
+    "input",
+    "sym_output",
+    "asym_output",
+    "update",
+    "sc_output",
+]
```
mgovers, line 46: do these types conflict with the `str` data type? if so, we can keep exposing them for the users that don't want to change too much in their existing code base

src/power_grid_model/validation/errors.py
```diff
@@ -165,18 +167,18 @@ class MultiComponentValidationError(ValidationError):
...
 
-    def __init__(self, fields: List[Tuple[str, str]], ids: List[Tuple[str, int]]):
+    def __init__(self, fields: List[Tuple[ComponentType, str]], ids: List[Tuple[ComponentType, int]]):
         """
         Args:
             fields: List of field names, formatted as tuples (component, field)
             ids: List of component IDs (not row indices), formatted as tuples (component, id)
         """
-        self.component = sorted(set(component for component, _ in fields))
+        self.component = sorted(set(component for component, _ in fields), key=str)
```
mgovers, line 181: wouldn't `component.value` be more correct?

scripts/quick_example_batch.py
```diff
@@ -54,7 +57,12 @@
 source["u_ref"] = [1.0]
 
 # input_data
-input_data = {"node": node, "line": line, "asym_load": asym_load, "source": source}
+input_data: Dict[ComponentType, np.ndarray] = {
+    "node": node,
+    "line": line,
+    "asym_load": asym_load,
+    "source": source,
+}
```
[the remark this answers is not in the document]
nitbharambe:
```suggestion
input_data: Dict[ComponentType, np.ndarray] = {
    ComponentTypes.node: node,
    ComponentTypes.line: line,
    ComponentTypes.sym_load: sym_load,
    ComponentTypes.source: source,
}
```

I  meant this

code_generation/code_gen.py
nitbharambe: > `data_types = [dataset["name"] for dataset in dataset_meta_data["all_datasets"]]` is incorrect as `dataset_meta_data["all_datasets"]]` doesn't include all datatypes. `sym_output` and `asym_output` come attached to `dataset.is_template` (this is how the json config. files are structured).
> 
> Making it a comprehension can also be done, but since the first loop is going to be used for components as well and the comprehension would be 'large', I think it is reasonable to keep it this way.

I meant to point that if the goal is to get all components, only  reading the "input" must be enough if im not mistaken. And not sym_output and asm_output. But its fine either way!

src/power_grid_model/core/power_grid_model.py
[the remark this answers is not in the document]
figueroa1395: This is how the new implementation looks like.
```
def _get_output_component_count(self, calculation_type: CalculationType):
        exclude_types = {
            CalculationType.power_flow: [
                ComponentType.sym_voltage_sensor,
                ComponentType.asym_voltage_sensor,
                ComponentType.sym_power_sensor,
                ComponentType.asym_power_sensor,
                ComponentType.fault,
            ],
            CalculationType.state_estimation: [ComponentType.fault],
            CalculationType.short_circuit: [
                ComponentType.sym_voltage_sensor,
                ComponentType.asym_voltage_sensor,
                ComponentType.sym_power_sensor,
                ComponentType.asym_power_sensor,
            ],
        }.get(calculation_type, [])

        def include_type(component_type: ComponentType):
            for exclude_type in exclude_types:
                if exclude_type.value in component_type.value:
                    return False
            return True

        return {k: v for k, v in self.all_component_count.items() if include_type(k)}
```

src/power_grid_model/dataset_definitions.py
figueroa1395: I will bring it up in the standup, that way people know about the change when navigating through the Python API, but I agree.

src/power_grid_model/dataset_definitions.py
figueroa1395: These were removed, now `class DataType (str, Enum)` is the implementation.

src/power_grid_model/validation/errors.py
figueroa1395: @mgovers It really depends. If we add the helper function _str_to_componenttype() to all the error classes' constructors, then yes. As it is now, it can process components both as `str` or `ComponentType`.  Maybe we can discuss a bit about this.

src/power_grid_model/validation/errors.py
figueroa1395: @nitbharambe Never mind, you were correct. It's been fixed.

src/power_grid_model/validation/utils.py
figueroa1395: @mgovers @nitbharambe Neither mypy nor `isinstance()`  like the fact that `ComponentTypes = Union[ComponentType, ComponentTypeLiterals]`. However, with the new definition of `class ComponentType(str, Enum)` this is fine. Nitish suggestion has been implemented.

figueroa1395: @mgovers  @nitbharambe  Neither mypy nor isinstance() like the fact that ComponentTypes = Union[ComponentType, ComponentTypeLiterals]. However, with the new definition of `class ComponentType(str, Enum)` this is fine. Nitish suggestion has been implemented.

code_generation/code_gen.py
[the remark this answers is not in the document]
figueroa1395: Changes made. You were right, I misunderstood the your comment initially.

code_generation/templates/src/power_grid_model/core/metadata_enums.py.jinja
```diff
@@ -0,0 +1,72 @@
...
+
+# This file is automatically generated. DO NOT modify it manually!
+
+{% set dataset_types = all_map.keys() %}
+{% set components = all_map['input'].keys() %}
+
+from enum import Enum
+from typing import Any, Dict, Mapping
+
+# pylint: disable=invalid-name
```
mgovers, line 15: Why?

docs/advanced_documentation/native-data-interface.md
```diff
@@ -98,10 +98,11 @@ For example, the following update dataset will set `from_status` of the line #5
 and keep the `to_status` as unchanged (change not available).
 
 ```python
-from power_grid_model import power_grid_meta_data
+from power_grid_model import ComponentType, DataType, power_grid_meta_data
 import numpy as np
 
-line_update = np.empty(shape=1, dtype=power_grid_meta_data['update']['line']['dtype'])
+line_update = np.empty(shape=1, dtype=power_grid_meta_data[DataType.update][ComponentType.line]['dtype'])
+# line_update = np.empty(shape=1, dtype=power_grid_meta_data['update']['line']['dtype']) is supported as well.
```
Jerry-Jinfeng-Guo, line 105: Remove commented out lines.

docs/advanced_documentation/native-data-interface.md
```diff
@@ -137,9 +138,10 @@ One can import the `power_grid_meta_data` to get all the predefined `numpy.dtype
 The code below creates an array which is compatible with transformer input dataset.
 
 ```python
-from power_grid_model import power_grid_meta_data
+from power_grid_model import ComponentType, DataType, power_grid_meta_data
 
-transformer = np.empty(shape=5, dtype=power_grid_meta_data['input']['transformer']['dtype'])
+transformer = np.empty(shape=5, dtype=power_grid_meta_data[DataType.input][ComponentType.transformer]['dtype'])
+# transformer = np.empty(shape=5, dtype=power_grid_meta_data['input']['transformer']['dtype']) is supported as well.
```
Jerry-Jinfeng-Guo, line 144: Remove commented out lines.

docs/examples/Asymmetric Calculation Example.ipynb
```diff
@@ -389,7 +389,7 @@
     }
    ],
    "source": [
-    "display(pd.DataFrame(output_data[\"line\"][\"p_from\"][0]))"
+    "display(pd.DataFrame(output_data[ComponentType.line][\"p_from\"][0]))"
```
Jerry-Jinfeng-Guo, line 392: So we are now only doing this at component level, not attribute level ;)

docs/examples/Asymmetric Calculation Example.ipynb
```diff
@@ -519,7 +519,7 @@
     }
    ],
    "source": [
-    "display(pd.DataFrame(output_data[\"line\"][\"i_from\"][:,0]))"
+    "display(pd.DataFrame(output_data[ComponentType.line][\"i_from\"][:,0]))"
```
Jerry-Jinfeng-Guo, line 522: Worth discussion, probably as a follow up: this `data.line['i_from']` mixed way is a bit awkward imo

docs/examples/Make Test Dataset.ipynb
Jerry-Jinfeng-Guo, line 1: Hmm? The use of C..Type.component is not consistent in this file, why?

src/power_grid_model/_utils.py
```diff
@@ -143,7 +145,7 @@ def get_batch_size(batch_data: BatchArray) -> int:
     return n_batches
 
 
-def split_numpy_array_in_batches(data: np.ndarray, component: str) -> List[np.ndarray]:
+def split_numpy_array_in_batches(data: np.ndarray, component: ComponentType) -> List[np.ndarray]:
```
Jerry-Jinfeng-Guo, line 148: I believe there is a data type to go for the first parameter here, instead of plainly `np.ndarray`

src/power_grid_model/core/dataset_definitions.py
Jerry-Jinfeng-Guo, line 1: I am curious, what is the complain from the linter? Which ones are `invalid-name`s?

src/power_grid_model/core/power_grid_dataset.py
```diff
@@ -183,14 +188,17 @@ class CMutableDataset:
...
+            Mapping[ComponentType, Union[np.ndarray, Mapping[str, np.ndarray]]],
+        ],
+        dataset_type: Any = None,
     ):
         instance = super().__new__(cls)
         instance._mutable_dataset = MutableDatasetPtr()
         instance._buffer_views = []
 
-        instance._dataset_type = dataset_type if isinstance(dataset_type, str) else get_dataset_type(data)
+        instance._dataset_type = dataset_type if dataset_type in list(DataType) else get_dataset_type(data)
```
Jerry-Jinfeng-Guo, line 201: This would mean we are doing `list` every single time if we want to check something. Instead, define a method in `DataType`, something like:
```
    @classmethod
    def contains(cls, a):
        return a in cls._member_map_
```
so that we could do:
```suggestion
        instance._dataset_type = dataset_type if DataType.contains(dataset_type) else get_dataset_type(data)
```

Jerry-Jinfeng-Guo (commented):
Half way (train is arriving)

code_generation/templates/src/power_grid_model/core/metadata_enums.py.jinja
figueroa1395: I assume you are asking about `# pylint: disable=invalid-name`. Pylint doesn't like non-capitalized `Enum` keys, but I am following the pgm convention as in `src/power_grid_model/enum.py`.

docs/advanced_documentation/native-data-interface.md
figueroa1395: @Jerry-Jinfeng-Guo I left the commented line on purpose to explain that the old format is still supported. It was a suggestion by @mgovers

docs/advanced_documentation/native-data-interface.md
figueroa1395: @Jerry-Jinfeng-Guo I left the commented line on purpose to explain that the old format is still supported. It was a suggestion by @mgovers

src/power_grid_model/core/dataset_definitions.py
figueroa1395: @Jerry-Jinfeng-Guo Pylint doesn't like non-capitalized `Enum` keys, but I am following the pgm convention as in src/power_grid_model/enum.py. Hence why the `# pylint: disable=invalid-name`.

docs/examples/Make Test Dataset.ipynb
figueroa1395: @Jerry-Jinfeng-Guo I must have missed it for some reason. Thanks!

docs/examples/Asymmetric Calculation Example.ipynb
figueroa1395: @Jerry-Jinfeng-Guo Do you mean that having component types with their own data type whereas their attributes being treated as regular strings is strange?

docs/examples/Asymmetric Calculation Example.ipynb
figueroa1395: @Jerry-Jinfeng-Guo Yes. Only custom datatypes have been introduced for `DataType` and `ComponentType`.

src/power_grid_model/_utils.py
figueroa1395: @Jerry-Jinfeng-Guo It would be of the type `Union[DenseBatchArray, SingleArray]`. Although the former is shorter, it probably makes sense to use the existing custom datatypes as you are suggesting.

src/power_grid_model/core/power_grid_model.py
```diff
@@ -459,7 +482,7 @@ def calculate_power_flow(
             error_tolerance=error_tolerance,
             max_iterations=max_iterations,
             calculation_method=calculation_method,
-            update_data=update_data,
+            update_data=(_map_to_componenttypes(update_data) if update_data is not None else None),
```
mgovers, line 485: i'm actually a little bit suprised that the linter doesn't complain about the import of functions prefixed `_`

tests/unit/test_dataset.py
```diff
@@ -92,7 +93,7 @@ def test_const_dataset__conflicting_data():
 
 
 def test_const_dataset__single_data(dataset_type):
-    components = {"node": 3, "sym_load": 2, "asym_load": 4}
+    components = {ComponentType.node: 3, ComponentType.sym_load: 2, ComponentType.asym_load: 4}
```
Jerry-Jinfeng-Guo, line 96: A bit awkward if I have to be brutally honest about the new usage... I wouldn't be surprised if users would 1, stick to the old way, or 2, come up with ways like:
 ```suggestion
    def map_components(enum_class, simple_mapping):
        return {getattr(enum_class, k): v for k, v in simple_mapping.items()}
    
    components = map_components(ComponentType, {
        'node': 3,
        'sym_load': 2,
        'asym_load': 4
    })
```
or plainly:
```suggestion
    temp_mapping = {
        'node': 3,
        'sym_load': 2,
        'asym_load': 4
    }
    
    components = {getattr(ComponentType, k): v for k, v in temp_mapping.items()}
```

src/power_grid_model/core/power_grid_dataset.py
figueroa1395: @Jerry-Jinfeng-Guo You are correct. Thanks for the suggestion.

src/power_grid_model/validation/utils.py
```diff
@@ -96,15 +97,15 @@ def update_input_data(input_data: SingleDataset, update_data: SingleDataset):
...
+def update_component_data(component: ComponentType, input_data: np.ndarray, update_data: np.ndarray) -> None:
     """
     Update the data in a numpy array, with another numpy array,
     indexed on the "id" field and only non-NaN values are overwritten.
     """
     for field in update_data.dtype.names:
         if field == "id":
             continue
-        nan = nan_type(component, field, "update")
+        nan = nan_type(component, field, DataType.input)
```
mgovers, line 108: i think this is incorrect?
```suggestion
       nan = nan_type(component, field, DataType.update)
```

tests/unit/test_dataset.py
```diff
@@ -136,21 +137,21 @@ def test_const_dataset__uniform_batch_data(dataset_type, batch_size):
...
             "indptr": np.array([0, 0, 1, 2]),
         },
-        "asym_load": {
+        ComponentType.asym_load: {
             "data": np.zeros(shape=4, dtype=power_grid_meta_data[dataset_type]["asym_load"]),
             "indptr": np.array([0, 2, 3, 4]),
         },
-        "link": np.zeros(shape=(batch_size, 4), dtype=power_grid_meta_data[dataset_type]["link"]),
+        ComponentType.link: np.zeros(shape=(batch_size, 4), dtype=power_grid_meta_data[dataset_type]["link"]),
     }
```
Jerry-Jinfeng-Guo, line 155: This relates to one of my previous comment: this mixed way of doing at component level but not on attribute level could be confusing.

tests/unit/test_serialization.py
```diff
@@ -32,7 +33,7 @@ def from_msgpack(data):
     return msgpack.unpackb(data)
 
 
-def empty_dataset(dataset_type: str = "input"):
+def empty_dataset(dataset_type: str = DataType.input):
```
Jerry-Jinfeng-Guo, line 36: Hmm?

tests/unit/test_power_grid_model.py
```diff
@@ -153,7 +154,7 @@ def test_batch_calculation_error_continue(model: PowerGridModel, case_data):
     assert "The id cannot be found:" in error.error_messages[0]
     # assert value result for scenario 0
     result = {"node": result["node"][error.succeeded_scenarios, :]}
-    expected_result = {"node": case_data["output_batch"]["node"][error.succeeded_scenarios, :]}
+    expected_result = {ComponentType.node: case_data["output_batch"]["node"][error.succeeded_scenarios, :]}
```
mgovers, line 157: shouldn't `case_data["output_batch"]["node"]` be `case_data["output_batch"][ComponentType.node]` as well? or is that unrelated?

mgovers (commented):
one minor note: in the future it's probably better to do the replacing of everything in the tests in a separate PR. basically, you then split concerns:

1. PR to introduce the new functionality. Existing tests will prove that it is backwards compatible, while new tests prove that the new functionality works
2. PR to apply the new functionality to the internal codebase to modernize it

tests/unit/validation/test_validation_functions.py
```diff
@@ -33,7 +34,7 @@
     validate_values,
 )
 
-NaN = power_grid_meta_data["input"]["node"].nans["id"]
+NaN = power_grid_meta_data[DataType.input][ComponentType.node].nans["id"]
```
Jerry-Jinfeng-Guo, line 37: This line really beats the purpose of this PR...

src/power_grid_model/core/dataset_definitions.py
Jerry-Jinfeng-Guo: If this we are globally doing non-captalized `Enum` keys, we should change the rule in the pylint config file

Jerry-Jinfeng-Guo (commented):
Second half. 
If we can not config Pylint globally, please at least do the `# pylint: disable=invalid-name` - `# pylint: enable=invalid-name` combo

docs/examples/Asymmetric Calculation Example.ipynb
figueroa1395: @mgovers @Jerry-Jinfeng-Guo Since this design discussion is probably outside the scope of this ticket, I will resolve the conversation and open a Jira ticket in the backlog for future possible discussion and implementation.

docs/examples/Asymmetric Calculation Example.ipynb
Jerry-Jinfeng-Guo: Same as above, this mixed way of doing it: component and attribute level.

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -74,7 +75,7 @@ def _generate_meta_data() -> PowerGridMetaData:
     n_datasets = pgc.meta_n_datasets()
     for i in range(n_datasets):
         dataset = pgc.meta_get_dataset_by_idx(i)
-        py_meta_data[pgc.meta_dataset_name(dataset)] = _generate_meta_dataset(dataset)
+        py_meta_data[_str_to_datatype(pgc.meta_dataset_name(dataset))] = _generate_meta_dataset(dataset)
```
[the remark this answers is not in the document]
figueroa1395: @Jerry-Jinfeng-Guo Since `py_meta_data` is defined as a plain `dict` (check line 74), my initial solution should be the correct one.

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -91,7 +92,7 @@ def _generate_meta_dataset(dataset: DatasetPtr) -> DatasetMetaData:
     n_components = pgc.meta_n_components(dataset)
     for i in range(n_components):
         component = pgc.meta_get_component_by_idx(dataset, i)
-        py_meta_dataset[pgc.meta_component_name(component)] = _generate_meta_component(component)
+        py_meta_dataset[_str_to_componenttype(pgc.meta_component_name(component))] = _generate_meta_component(component)
```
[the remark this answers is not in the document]
figueroa1395: @Jerry-Jinfeng-Guo See above.

src/power_grid_model/core/power_grid_model.py
```diff
@@ -127,11 +132,12 @@ def update(self, *, update_data: Dict[str, np.ndarray]):
         Returns:
             None
         """
+        update_data = _map_to_componenttypes(update_data)
         prepared_update = prepare_update_view(update_data)
         pgc.update_model(self._model, prepared_update.get_dataset_ptr())
         assert_no_error()
 
-    def get_indexer(self, component_type: str, ids: np.ndarray):
+    def get_indexer(self, component_type: Any, ids: np.ndarray):
```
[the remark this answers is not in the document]
figueroa1395: @Jerry-Jinfeng-Guo Oops. Fixed it.

src/power_grid_model/core/power_grid_meta.py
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: It's not wrong in its current form, it's just still a mixture

src/power_grid_model/core/power_grid_model.py
```diff
@@ -385,13 +403,18 @@ def calculate_power_flow(
         error_tolerance: float = 1e-8,
         max_iterations: int = 20,
         calculation_method: Union[CalculationMethod, str] = CalculationMethod.newton_raphson,
-        update_data: Optional[Dict[str, Union[np.ndarray, Dict[str, np.ndarray]]]] = None,
+        update_data: Optional[
+            Union[
+                Dict[str, Union[np.ndarray, Dict[str, np.ndarray]]],
+                Dataset,
+            ]
+        ] = None,
```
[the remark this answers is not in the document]
figueroa1395: @mgovers I implemented it this way as this is a user available function: Basically it can take `Dataset` as defined with `ComponentType` but also as defined with `Literal`. I had to do it this way for mypy to like it when users decide to use `Literal`s (backwards support) instead of the new datatypes.

src/power_grid_model/core/power_grid_model.py
figueroa1395: @mgovers I am not surprised, I am happy that it doesn't!

tests/unit/test_dataset.py
figueroa1395: @Jerry-Jinfeng-Guo But the good thing is that support is there for the new and old implementations. Users can do `ComponentType.node: 3` or `"node": 3`, and it is going to be treated equally within pgm. However, now they will have proper warnings from linters if they make typos and use the new implementation. It is a matter of perspective.

tests/unit/test_dataset.py
figueroa1395: @Jerry-Jinfeng-Guo I understand your point of view. I guess, as pointed out by @mgovers, we should make some design decisions if we want to also include some sort of `AttributeType`. Since it is out of scope for now, I will resolve and open a new ticket in the backlog.

tests/unit/test_power_grid_model.py
figueroa1395: @mgovers it could, yes. However, to be truthful, I only updated the tests that dealt directly with internal (non-user accessible) functions of pgm, since now we expect to don't use Literals as component or data types anymore. For the tests that take core of user friendly functions, I mostly left them as is, as they should provide support for both ways and did not think it was really useful to change. If it is desired for me to change all the tests, please let me know.

tests/unit/test_serialization.py
figueroa1395: @Jerry-Jinfeng-Guo Oops.

tests/unit/validation/test_validation_functions.py
figueroa1395: @Jerry-Jinfeng-Guo why? `power_grid_meta_data` takes both `Literal`s and `DataType`s / `ComponenType`s

src/power_grid_model/core/dataset_definitions.py
figueroa1395: @Jerry-Jinfeng-Guo I agree that if we decide to keep lower-case members for the `Enum`s globally, we should just change the config. file. Will bring that up in the standup.

mgovers: we're not doing this globally - only in this file and it is enabled afterwards. Not sure but maybe it works to disable it on the class instead of the member variables

tests/unit/test_power_grid_model.py
mgovers: you can leave it, i guess

src/power_grid_model/dataset_definitions.py
figueroa1395: DatasetType it is.

code_generation/code_gen.py
```diff
@@ -114,21 +114,22 @@ def render_dataset_class_maps(self, template_path: Path, data_path: Path, output
...
         }
 
         # render attribute classes
         for template_name, render_func in render_funcs.items():
             for template_path in TEMPLATE_DIR.rglob(f"{template_name}.*.jinja"):
                 output_suffix = template_path.with_suffix("").suffix
                 output_dir = template_path.parent.relative_to(TEMPLATE_DIR)
-                for data_path in DATA_DIR.glob(f"{template_name}/*.json"):
+                data_name = template_name if template_name != "metadata_enums" else "dataset_class_maps"
+                for data_path in DATA_DIR.glob(f"{data_name}/*.json"):
```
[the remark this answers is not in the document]
figueroa1395: `metadata_enums.py` renamed to `dataset_class_maps.py` as suggested.

tests/unit/test_dataset_definitions.py
```diff
@@ -0,0 +1,23 @@
...
+    gen_power_grid_data_types.sort()
+    assert power_grid_data_types == gen_power_grid_data_types
+
+
+def test_power_grid_components():
+    power_grid_components = [component for component in power_grid_meta_data[DatasetType.input]]
+    gen_power_grid_components = [member.value for member in ComponentType]
+    power_grid_components.sort()
+    gen_power_grid_components.sort()
+    assert power_grid_components == gen_power_grid_components
```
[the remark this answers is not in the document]
figueroa1395: You are correct. Done.

src/power_grid_model/core/data_handling.py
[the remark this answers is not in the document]
figueroa1395: @Jerry-Jinfeng-Guo I think you might be right. However, it may be related with the C binding. We can raise some discussion and make a simple PR if needed.

scripts/quick_example_batch.py
```diff
@@ -15,12 +15,12 @@
 """
 
 # node
-node = initialize_array("input", "node", 3)
+node = initialize_array(DatasetType.input, ComponentType.node, 3)  # initialize_array("input", "node", 3) is also OK
```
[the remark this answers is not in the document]
figueroa1395: Corrected.

src/power_grid_model/core/data_handling.py
```diff
@@ -24,9 +25,9 @@ class OutputType(Enum):
         - asym_output
```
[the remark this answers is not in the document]
figueroa1395: Done

docs/examples/Make Test Dataset.ipynb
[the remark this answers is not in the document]
figueroa1395: I will bring back the original metadata then.

src/power_grid_model/core/data_handling.py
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Yup, perhaps let's discuss this next week with @mgovers

Jerry-Jinfeng-Guo (commented):
I have no more complains. If @nitbharambe could give a final look, I think we can merge this PR. (hence not yet an 'Approve' approving review)

nitbharambe (dismissed):

src/power_grid_model/core/power_grid_model.py
```diff
@@ -555,13 +583,18 @@ def calculate_short_circuit(
...
         threading: int = -1,
-        output_component_types: Optional[Union[Set[str], List[str]]] = None,
+        output_component_types: Optional[Union[Set[ComponentType], List[ComponentType]]] = None,
         continue_on_batch_error: bool = False,
         decode_error: bool = True,
         short_circuit_voltage_scaling: Union[ShortCircuitVoltageScaling, str] = ShortCircuitVoltageScaling.maximum,
-    ) -> Dict[str, np.ndarray]:
+    ) -> Dict[Any, np.ndarray]:
```
mgovers, line 597: idem

src/power_grid_model/core/dataset_definitions.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        Check if member is part of the Enum.
+
+        Args:
+            member: Member to check.
+
+        Returns:
+            bool: True if the member is part of the Enum, False otherwise.
+        """
+        return member in cls.__members__
```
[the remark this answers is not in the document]
figueroa1395: @mgovers You are correct. I did not think about `__contains__`. I have overloaded it.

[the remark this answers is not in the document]
mgovers: did the default one not work?

src/power_grid_model/core/power_grid_model.py
```diff
@@ -385,13 +403,18 @@ def calculate_power_flow(
...
+            ]
+        ] = None,
         threading: int = -1,
-        output_component_types: Optional[Union[Set[str], List[str]]] = None,
+        output_component_types: Optional[Union[Set[ComponentType], List[ComponentType]]] = None,
         continue_on_batch_error: bool = False,
         decode_error: bool = True,
         tap_changing_strategy: Union[TapChangingStrategy, str] = TapChangingStrategy.disabled,
-    ) -> Dict[str, np.ndarray]:
+    ) -> Dict[Any, np.ndarray]:
```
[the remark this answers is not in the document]
figueroa1395: All fixed.

mgovers (approved):

docs/examples/Asymmetric Calculation Example.ipynb
mgovers: relevant input for the discussion later today:

```py
>>> from enum import StrEnum
>>> class A(StrEnum):
...     a = "a"
... 
>>> class B(StrEnum):
...     a = "a"
... 
>>> A.a == B.a
True
>>> A.a is B.a
False
```

PR #666, opened 2024-07-08
[FEATURE] Numerically stable source injection in power flow.

### LINKED ISSUE #458: [FEATURE] Use more numerically stable method to calculate source injection in power flow
# Background

The current implementation of power flow treats each source as a [Thévenin's equivalence](https://en.wikipedia.org/wiki/Th%C3%A9venin%27s_theorem) with an ideal voltage source and an internal impedance. The size of the impedance depends on the short circuit power of the source. In this way, we can cope with multiple sources within one interconnected grid. The contribution of each source is determined by their reference voltage and internal impedance.

# Problem

You do not get the same voltage at the node as the reference voltage of the source, due to the voltage drop in the internal impedance. Users who wish to match the reference voltage exactly need to specify a large short circuit power (`sk`). However, this might cause numerical stability issue.

# Experiment

We have conducted an [experiment](https://github.com/PowerGridModel/power-grid-model-workshop/blob/experiment/source-calculation/source_calculation/source_calculation.ipynb) to see the effect of large `sk` for source. We have drawn the following conclusions.

* A high short circuit power `sk` in sources will force the node voltage to approach to the source `u_ref`. The calculation itself will not have numerical stability issue.
* The current way of calculating source injection power does suffer from numerical instability. The votlage difference between source node and source `u_ref` is too close.
* We need to use the node admittance matrix to help with calculation of source injection.

# Proposal

As the conclusion of the experiment, we need to use node admittance matrix to calculation source injection, as described below in the conclusion chapter of the [experiment](https://github.com/PowerGridModel/power-grid-model-workshop/blob/experiment/source-calculation/source_calculation/source_calculation.ipynb).

figueroa1395:
Closes #458


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,11 +59,24 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+        i_norton_total += dot(y_ref, u_ref);
+    }
+
+    for (Idx const source : sources) {
+        ComplexValue<sym> const u_ref{input.source[source]};
+        ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
+        ComplexValue<sym> const i_norton = dot(y_ref, u_ref);
+        ComplexTensor<sym> const y_ref_normalized = y_ref / y_ref_total;
+        output.source[source].i = (i_norton - dot(y_ref_normalized, i_norton_total)) + dot(y_ref_normalized, i_source_total);
         output.source[source].s = output.u[bus_number] * conj(output.source[source].i);
```
Jerry-Jinfeng-Guo, line 80: The whole function:
```suggestion
template <symmetry_tag sym>
inline void calculate_source_result(IdxRange const& sources, Idx bus_number, YBus<sym> const& y_bus,
                                    PowerFlowInput<sym> const& input, SolverOutput<sym>& output) {
                                    PowerFlowInput<sym> const& input, SolverOutput<sym>& output,
                                    ComplexValue<sym> const& i_load_gen_bus) {
    ComplexValue<sym> const i_source_total = conj(output.bus_injection[bus_number] / output.u[bus_number]) - i_load_gen_bus;
    std::vector<ComplexTensor<sys>> y_ref_acc(sources.size());
    std::vector<ComplexValue<sym>> i_norton_acc(sources.size());

    std::transform(sources.begin(), sources.end(), y_ref_acc.begin(),
                [&](Idx const source) -> ComplexTensor<sym> {
                    ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
                    return y_ref; 
                });
    std::transform(sources.begin(), sources.end(), i_norton_acc.begin(),
                [&](Idx const source) -> ComplexTensor<sym> {
                    ComplexValue<sym> const u_ref{input.source[source]};
                    ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
                    return dot(y_ref, u_ref); 
                });
    ComplexTensor<sym> const y_ref_total_inv = 1.0 / std::accumulate(y_ref_acc.begin(), y_ref_acc.end() - 1, ComplexTensor<sym>{});
    ComplexValue<sym> const i_norton_total = std::accumulate(i_norton_acc.begin(), i_norton_acc.end() - 1, ComplexValue<sym>{});

    for (size_t i = 0; i < sources.size(); ++i) {
        Idx const source = sources[i];
        ComplexTensor<sym> const y_ref_normalized = y_ref_acc[i] * y_ref_total_inv;
        output.source[source].i = (i_norton_acc[i] - dot(y_ref_normalized, i_norton_total)) + dot(y_ref_normalized, i_source_total);
        output.source[source].s = output.u[bus_number] * conj(output.source[source].i);
    }
}
```

Jerry-Jinfeng-Guo (commented):
I know it's still in draft, but I just could not un-see it :)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
mgovers: if possible, please use `std::ranges::transform(vector, target.begin())` instead of `std::transform(vector.begin(), vector.end(), target.begin())` if possible

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,11 +59,26 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+    ComplexValue<sym> const i_source_total =
+        conj(output.bus_injection[bus_number] / output.u[bus_number]) - i_load_gen_bus;
+    std::vector<ComplexTensor<sym>> y_ref_acc(sources.size());
+    std::vector<ComplexValue<sym>> i_norton_acc(sources.size());
     for (Idx const source : sources) {
         ComplexValue<sym> const u_ref{input.source[source]};
         ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
-        output.source[source].i = dot(y_ref, u_ref - output.u[bus_number]);
+        y_ref_acc[source] = y_ref;
+        i_norton_acc[source] = dot(y_ref, u_ref);
```
Jerry-Jinfeng-Guo, line 72: This may not give you the correct indices

figueroa1395: @Jerry-Jinfeng-Guo You were right. It's been corrected.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
figueroa1395: @mgovers @Jerry-Jinfeng-Guo Implemented something along those lines. Thanks for the suggestion.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,20 +60,63 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
-        ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
-        output.source[source].i = dot(y_ref, u_ref - output.u[bus_number]);
-        output.source[source].s = output.u[bus_number] * conj(output.source[source].i);
+                                    PowerFlowInput<sym> const& input, SolverOutput<sym>& output,
+                                    ComplexValue<sym> const& i_load_gen_bus) {
+    ComplexValue<sym> const i_source_total =
+        conj(output.bus_injection[bus_number] / output.u[bus_number]) - i_load_gen_bus;
+    std::vector<Idx> sources_acc(sources.size());
+    std::ranges::transform(sources, sources_acc.begin(), [&](Idx const source) { return source; });
+    if (sources_acc.size() == 1) {
```
nitbharambe, line 69: The size=0 goes via other condition.
```suggestion
    if (sources_acc.size() < 2) {
```
Or better reverse the condition for early return.

nitbharambe: size=0 is possible? :O

figueroa1395: Addressed by explicitly taking care of the size=0 case. Thanks!

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -105,7 +151,7 @@ inline void calculate_pf_result(YBus<sym> const& y_bus, PowerFlowInput<sym> cons
                                 LoadGenFunc&& load_gen_func) {
     assert(sources_per_bus.size() == load_gens_per_bus.size());
 
-    // call y bus
+    // call y bus //
```
Jerry-Jinfeng-Guo, line 154: hmm?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,20 +60,64 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+            z_ref_t = ComplexTensor<sym>{s_z, m_z};
+        }
+        ComplexValue<sym> const i_ref_t = std::accumulate(i_ref_acc.begin(), i_ref_acc.end(), ComplexValue<sym>{});
+        for (size_t i = 0; i < sources.size(); ++i) {
+            Idx const source = sources_acc[i];
+            ComplexValue<sym> const u_ref_t = dot(z_ref_t, i_ref_t);
+            ComplexValue<sym> const u_ref_i{input.source[i]};
+            ComplexValue<sym> delta_u = (u_ref_i - u_ref_t);
+            // Arbitrary precision threshold (20.0) needs further discussion, as well as this whole cutoff.
+            constexpr auto precision = 20.0 * std::numeric_limits<double>::epsilon();
```
Jerry-Jinfeng-Guo, line 104: Make a variable for it:
```suggestion
            // Arbitrary precision threshold (20.0) needs further discussion, as well as this whole cutoff.
            constexpr double theta = 20.0;
```
and then
```suggestion
            constexpr auto precision = theta * std::numeric_limits<double>::epsilon();
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,20 +60,64 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+                                    ComplexValue<sym> const& i_load_gen_bus) {
+    ComplexValue<sym> const i_inj_t = conj(output.bus_injection[bus_number] / output.u[bus_number]) - i_load_gen_bus;
+    std::vector<Idx> sources_acc(sources.size());
+    std::ranges::transform(sources, sources_acc.begin(), [&](Idx const source) { return source; });
+    if (sources_acc.empty()) {
+        ; // Do nothing
+    } else if (sources_acc.size() == 1) {
+        output.source[sources_acc[0]].i = i_inj_t;
+        output.source[sources_acc[0]].s = output.u[bus_number] * conj(output.source[sources_acc[0]].i);
+    } else {
```
Jerry-Jinfeng-Guo, line 73:
```suggestion
    if (sources_acc.size() == 1) {
        output.source[sources_acc[0]].i = i_inj_t;
        output.source[sources_acc[0]].s = output.u[bus_number] * conj(output.source[sources_acc[0]].i);
    } else if (!sources_acc.empty()) {
```

figueroa1395: I was suggested to make it explicit that nothing must be done when there are no sources, but I feel that your way accomplishes this more elegantly. Thanks!

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
figueroa1395: I wanted to leave it as an explicit number such that it would be impossible to miss, at least after the discussion is had. But you are totally right.

tests/data/power_flow/source-big-sk/input.json
```diff
@@ -39,15 +39,17 @@
         "status": 0,
         "u_ref": 1.0,
         "sk": 2e50,
-        "rx_ratio": 0.1
+        "rx_ratio": 0.1,
+        "z01_ratio": 2.0
```
petersalemink95, line 43: To double check, since you're changing values in the test in parallel with updating the code: is this correct? Default `z01_ratio` is `1.0`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,20 +60,66 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
-                                    PowerFlowInput<sym> const& input, SolverOutput<sym>& output) {
-    for (Idx const source : sources) {
-        ComplexValue<sym> const u_ref{input.source[source]};
-        ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
-        output.source[source].i = dot(y_ref, u_ref - output.u[bus_number]);
-        output.source[source].s = output.u[bus_number] * conj(output.source[source].i);
+                                    PowerFlowInput<sym> const& input, SolverOutput<sym>& output,
+                                    ComplexValue<sym> const& i_load_gen_bus) {
+    ComplexValue<sym> const i_inj_t = conj(output.bus_injection[bus_number] / output.u[bus_number]) - i_load_gen_bus;
+    std::vector<Idx> sources_acc(sources.size());
```
petersalemink95, line 66: What does "acc" stand for in `sources_acc`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,20 +60,66 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
 
 template <symmetry_tag sym>
 inline void calculate_source_result(IdxRange const& sources, Idx bus_number, YBus<sym> const& y_bus,
```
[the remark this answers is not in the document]
figueroa1395: Did a bit of cleaning and tried to left there what is directly relevant to the function, please let me know.

tests/data/power_flow/source-big-sk/input.json
figueroa1395: This was done on purpose. This way, as pointed out by @nitbharambe, the output will be the same, but now the off-diagonal (`m`) elements of the admittance matrix will be and non-zero. The intention was to make sure that no other instability errors would propagate, and if so, that the cut-off can mitigate it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
figueroa1395: "acc" stands for accumulated. Maybe better naming is needed?

tests/data/power_flow/source-big-sk/input.json
mgovers: maybe good to make 2 separate test cases of this

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -284,6 +284,34 @@ inline std::pair<DoubleComplex, DoubleComplex> inv_sym_param(DoubleComplex const
...
+    if constexpr (is_symmetric_v<sym>) {
+        return 1.0 / x;
+    } else {
+        auto [s, m] = inv_sym_param(x(0), x(1));
+        return ComplexTensor<sym>(s, m);
+    }
+}
+
+// inverse of Y via s, m
+template <symmetry_tag sym> inline ComplexTensor<sym> inv_y(ComplexTensor<sym> const& y) {
```
mgovers, line 298: `y` is very specific EE terminology. This file is about the generic mathematical concept of a 3x3 tensor together with the conventions about the basis vectors (`1,a,a²`) that stems from EE, but is actually more generic.

The explicit reference to `y` does not belong here, or requires different terminology (e.g. referencing a more generic symmetric tensor `t`).

See e.g. how we do not explicitly access `(0)` and `(1)` when calling `inv_sym_param` above this. Instead, the input is diagonal part `s` and off-diagonal part `m`. As such, usage is e.g. `inv_sym_param(y(0), y(1))`.

There are 2 options to solve it here:
1. change the input type from `ComplexTensor<sym>` to a `ComplexSymmetricTensor<sym>`, which needs to be created because it doesn't exist yet. NOTE: this also requires defining the related operators as we did for `DiagonalTensor<sym>`. `Y` would have to be redefined in multiple locations. While I think that this is worth the effort (less, more stable matrix calculations), it's also very much a scope-creep and I don't think we should go down this route in this PR.
2. use the same formalism as for `inv_sym_param` here, with `DoubleComplex const& s, DoubleComplex const& m` as input parameters

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -8,7 +8,9 @@
 #include "y_bus.hpp"
 
 #include "../calculation_parameters.hpp"
+#include "../common/common.hpp"
 #include "../common/exception.hpp"
+#include <numeric>
```
mgovers, line 13: convention: space before 3rd party, system and STL libraries

```suggestion

#include <numeric>
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,60 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
     });
 }
 
+template <symmetry_tag sym>
+inline void truncate_to_zero(double const& theta, ComplexValue<sym> const& val1, ComplexValue<sym> const& val2,
+                             ComplexValue<sym>& result) {
```
mgovers, line 64: please provide a docstring. while it's not difficult to understand, the rationale behind the need for this function may not be straightforward. Someone may encounter this and be like "why would I need this? wouldn't the calculation be more accurate without it?" and remove it. Or, worse, they may not even want to touch this code in the first place.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -59,20 +60,66 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+        }
+        ComplexValue<sym> const i_ref_t = std::accumulate(i_ref_acc.begin(), i_ref_acc.end(), ComplexValue<sym>{});
+        for (size_t i = 0; i < sources.size(); ++i) {
+            Idx const source = sources_acc[i];
+            // The resulting error (from the matrix multiplication) for every u_ref_t element is 54E
+            ComplexValue<sym> const u_ref_t = dot(z_ref_t, i_ref_t);
+            ComplexValue<sym> const u_ref_i{input.source[i]};
+            ComplexValue<sym> delta_u = (u_ref_i - u_ref_t);
+            // This truncation needs further discussion.
+            constexpr double theta = 54.0; // From counting floating point operations involving E.
```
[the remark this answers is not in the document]
mgovers: one thing I just thought of is that we can also do something like calculate the same function multiple times (3?) but with a small `std::numeric_limits<double>::epsilon()` applied to all input arguments. The standard deviation would then be an estimation of the precision and the mean would be an estimation of the actual value.

The 2 major downsides are:
1. there may be quite a number of input parameters (e.g. an arbitrary symmetric 3-dimensional tensor has 6 inputs. fortunately, the symmetric one with all diagonal elements the same and all off-diagonal elements the same only has 2, so that could help, but only if there's no risk that they differ slightly due to other calculations). We can work around this by doing some arbitrary stuff but it should still be deterministic.
2. depending on the number of input parameters and sources, this can introduce a combinatorically expensive calculation. Fortunately, the expectation is that the amount of sources is small (typically 1 to a couple in the entire grid, let alone on the same node), so this may be premature optimization.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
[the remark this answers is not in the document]
mgovers: it's still quite long. can we split it up by "relevant part"? e.g.: the first part is about accumulating, the second part about calculating the source contribution. They may be separate functions (if possible). If not, it's fine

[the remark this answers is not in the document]
figueroa1395: It should be a lot better now. Will split and reorganize the asymmetric part after the still failing tests are fixed.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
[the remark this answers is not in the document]
figueroa1395: I am currently working on how to not use truncation. Since the symmetric tests pass with the current implementation, that seems to indicate that working in the 012 domain may remove the instability. Work in progress.

tests/data/power_flow/source-big-sk/input.json
figueroa1395: @mgovers Is it needed though? In the end the results are the same, this just makes the admittance matrix not a perfect diagonal one.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
figueroa1395: Due to the rework being done to try to remove instabilities in a different way, this shall be removed as it is no longer needed at the moment. However, I believe your suggestion in 1 is very valid and may be worth investigating if it is of value outside of this PR, and if it is, we should create an issue with it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
figueroa1395: The code may be straightforward, but I totally agree that the reason why it is being done in such a way is not. Well noted, and will add it.

tests/data/power_flow/source-big-sk/input.json
mgovers: Perfect and non-perfect are different edge cases (location where rounding error occurs)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -58,21 +60,119 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+
+    for (size_t i = 0; i < sources.size(); ++i) {
+        Idx const source = sources_acc[i];
+        ComplexValue<asymmetric_t> const i_ref_i_012 = dot(get_sym_matrix_inv(), i_ref_acc[i]);
+        ComplexValue<asymmetric_t> const y_ref_i_012_over_y_ref_t_012 = y_ref_012_acc[i] / y_ref_012_t;
+        ComplexValue<asymmetric_t> const i_inj_012{
+            y_ref_i_012_over_y_ref_t_012(0) * i_ref_t_012(0),
+            (i_ref_i_012(1) - y_ref_i_012_over_y_ref_t_012(1) * i_ref_t_012(1)) +
+                (y_ref_i_012_over_y_ref_t_012(1) * i_ref_t_012(1)),
+            y_ref_i_012_over_y_ref_t_012(2) * i_ref_t_012(2),
```
nitbharambe, line 148:
```suggestion
        ComplexValue<asymmetric_t> const i_inj_012{
            y_ref_i_012_over_y_ref_t_012(0) * i_inj_t_012(0),
            (i_ref_i_012(1) - y_ref_i_012_over_y_ref_t_012(1) * i_inj_t_012(1)) +
                (y_ref_i_012_over_y_ref_t_012(1) * i_inj_t_012(1)),
            y_ref_i_012_over_y_ref_t_012(2) * i_inj_t_012(2),
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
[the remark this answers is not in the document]
figueroa1395: There is no need for truncation anymore after reorganizing the equations cleverly and going to 012 domain for the asymmetric case.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
figueroa1395: This has been removed in the current implementation.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
figueroa1395: After going to 012 domain in the asymmetric case, there is no longer need for such an inverse. This has been removed.

tests/data/power_flow/source-big-sk/input.json
figueroa1395: This has been left untouched, i.e. `z01_ratio = 1` as it is not really relevant to the instability issue.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,94 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+        DoubleComplex const i_inj_i_0 = (y0_y1[source].first / y_ref_t_012[0]) * i_inj_t_012(0);
+        DoubleComplex const i_inj_i_2 = (y0_y1[source].second / y_ref_t_012[2]) * i_inj_t_012(2);
+        ComplexValue<asymmetric_t> const i_inj_012{i_inj_i_0, i_inj_i_1, i_inj_i_2};
+        output.source[source].i = dot(get_sym_matrix(), i_inj_012);
         output.source[source].s = output.u[bus_number] * conj(output.source[source].i);
     }
 }
 
+// Current implementation avoids numerical instability. For details refer to:
+// https://github.com/PowerGridModel/power-grid-model-workshop/blob/experiment/source-calculation/source_calculation/source_calculation.ipynb
```
Jerry-Jinfeng-Guo, line 130: Should we make this notebook official and provide a more proper linkage here?

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -284,6 +284,16 @@ inline std::pair<DoubleComplex, DoubleComplex> inv_sym_param(DoubleComplex const
...
+// inverse of symmetric tensor
+template <symmetry_tag sym> inline ComplexTensor<sym> inv_sym_tensor(ComplexTensor<sym> const& x) {
+    if constexpr (is_symmetric_v<sym>) {
+        return 1.0 / x;
+    } else {
+        auto [s, m] = inv_sym_param(x(0), x(1));
+        return ComplexTensor<sym>(s, m);
+    }
+}
+
```
TonyXiang8787, line 296: This function is never used. Maybe delete it?

figueroa1395: You are right. It was being used in a previous implementation, but not anymore. I will remove it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,94 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+    ComplexVector const y_ref = y_bus.math_model_param().source_param;
+    DoubleComplex const y_ref_t =
+        std::accumulate(sources.begin(), sources.end(), DoubleComplex{},
+                        [&](DoubleComplex sum, Idx const source) { return y_ref[source] + sum; });
+    DoubleComplex const i_ref_t =
+        std::accumulate(sources.begin(), sources.end(), DoubleComplex{}, [&](DoubleComplex sum, Idx const source) {
+            return input.source[source] * y_ref[source] + sum;
+        });
+    for (Idx const source : sources) {
+        DoubleComplex const y_ref_i_over_y_ref_t = y_ref[source] / y_ref_t;
```
Jerry-Jinfeng-Guo, line 81: Please cache the inverse `1.0 / y_ref_t` as this is inside a `for` loop

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -190,6 +190,7 @@ template <symmetry_tag sym_type> struct MathModelParam {
     std::vector<BranchCalcParam<sym>> branch_param;
     ComplexTensorVector<sym> shunt_param;
     ComplexTensorVector<sym> source_param;
+    std::vector<std::pair<DoubleComplex, DoubleComplex>> source_param_y0_y1;
```
TonyXiang8787, line 193: I am against the idea that we have two `source` stuff in `MathModelParam`. I think we need to change the way of math parameters of source in general. Concretely, I propose:

1. Define a new struct `SourceCalcParam` which is not a template.

```c++
struct SourceCalcParam {
    DoubleComplex y1;
    DoubleComplex y0;
};
```

2. Change the old `source_param` definition in `MathModelParam` to the new.

from 
```c++
    ComplexTensorVector<sym> source_param;
```

to
```c++
    std::vector<SourceCalcParam> source_param;
```

3. Rewrite the `template <symmetry_tag sym> ComplexTensor<sym> math_param()` in source to the new way of just exporting `y1` and `y0` instead of a scalar or tensor.

4. Rewrite the power flow solver functions where the source scalar/tensor is expected. Either directly use `y1` for symmetric calculation, or build source y tensor on the fly for asymmetric calculation. You can create an inline function in the `common_solver_functions.hpp` to build the y tensor matrix.

5. In the calculate source results, you can then directly use the `y1` and `y0`.

power_grid_model_c/power_grid_model/include/power_grid_model/component/source.hpp
```diff
@@ -42,7 +42,9 @@ class Source : public Appliance {
         y0_ref_ = y1_ref_ / z01_ratio;
     }
 
-    // getter for calculation param, y_ref
+    // getter for y0 y1 ref
+    std::pair<DoubleComplex, DoubleComplex> get_y0_y1() const { return {y0_ref_, y1_ref_}; }
```
TonyXiang8787, line 46: As I said before, put this logic directly into `template <symmetry_tag sym> ComplexTensor<sym> math_param()`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,94 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
-        ComplexValue<sym> const u_ref{input.source[source]};
-        ComplexTensor<sym> const y_ref = y_bus.math_model_param().source_param[source];
-        output.source[source].i = dot(y_ref, u_ref - output.u[bus_number]);
+        DoubleComplex const y_ref_i_1_over_y_ref_t_1 = y0_y1[source].second / y_ref_t_012[1];
+        DoubleComplex const i_inj_i_1_lhs =
+            y_ref_i_1_over_y_ref_t_1 * (input.source[source] * y_ref_t_012[1] - i_ref_1_t);
+        DoubleComplex const i_inj_i_1 = i_inj_i_1_lhs + (y_ref_i_1_over_y_ref_t_1 * i_inj_t_012(1));
+
+        DoubleComplex const i_inj_i_0 = (y0_y1[source].first / y_ref_t_012[0]) * i_inj_t_012(0);
+        DoubleComplex const i_inj_i_2 = (y0_y1[source].second / y_ref_t_012[2]) * i_inj_t_012(2);
```
Jerry-Jinfeng-Guo, line 122:
```suggestion
        DoubleComplex const i_inj_i_0 = (y0_y1[source].first / y_ref_t_012[0]) * i_inj_t_012(0);
        DoubleComplex const i_inj_i_1 = i_inj_i_1_lhs + (y_ref_i_1_over_y_ref_t_1 * i_inj_t_012(1));
        DoubleComplex const i_inj_i_2 = (y0_y1[source].second / y_ref_t_012[2]) * i_inj_t_012(2);
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -874,6 +874,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
             math_param[i].branch_param.resize(state_.math_topology[i]->n_branch());
             math_param[i].shunt_param.resize(state_.math_topology[i]->n_shunt());
             math_param[i].source_param.resize(state_.math_topology[i]->n_source());
+            math_param[i].source_param_y0_y1.resize(state_.math_topology[i]->n_source());
```
TonyXiang8787, line 877: We don't need this anymore if we re-define the `source_param`.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -917,6 +918,8 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
             // assign parameters
             math_param[math_idx.group].source_param[math_idx.pos] =
                 state_.components.template get_item_by_seq<Source>(i).template math_param<sym>();
+            math_param[math_idx.group].source_param_y0_y1[math_idx.pos] =
+                state_.components.template get_item_by_seq<Source>(i).get_y0_y1();
```
TonyXiang8787, line 922: We don't need this anymore if we re-define the `source_param`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,94 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+// Note: u_ref_i_0 = u_ref_i_2 = 0
+// 012 domain is used to solve the numerical instability as for the symmetric case.
+inline void calculate_multiple_source_result(IdxRange const& sources, YBus<asymmetric_t> const& y_bus,
+                                             PowerFlowInput<asymmetric_t> const& input,
+                                             ComplexValue<asymmetric_t> const& i_inj_t,
+                                             SolverOutput<asymmetric_t>& output, Idx const& bus_number) {
+    std::vector<std::pair<power_grid_model::DoubleComplex, power_grid_model::DoubleComplex>> const y0_y1 =
+        y_bus.math_model_param().source_param_y0_y1;
+    ComplexValue<asymmetric_t> const y_ref_t_012 =
+        std::accumulate(sources.begin(), sources.end(), ComplexValue<asymmetric_t>{},
```
TonyXiang8787, line 102: `std::accumulate` is discouraged now. Use `std::transform_reduce`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,94 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+                        [&](ComplexValue<asymmetric_t> sum, Idx const source) {
+                            sum(0) += y0_y1[source].first;
+                            sum(1) += y0_y1[source].second;
+                            sum(2) = sum(1); // y1 = y2
+                            return sum;
+                        });
+    DoubleComplex const i_ref_1_t =
+        std::accumulate(sources.begin(), sources.end(), DoubleComplex{}, [&](DoubleComplex sum, Idx const source) {
```
TonyXiang8787, line 110: `std::accumulate` is discouraged now. Use `std::transform_reduce`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,94 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+// Phase space is used because y_ref is a DoubleComplex in the symmetric case.
+// The equation is organized in such a way to avoid the numerical instability induced in the substraction if y_ref_i
+// is large.
+inline void calculate_multiple_source_result(IdxRange const& sources, YBus<symmetric_t> const& y_bus,
+                                             PowerFlowInput<symmetric_t> const& input,
+                                             ComplexValue<symmetric_t> const& i_inj_t,
+                                             SolverOutput<symmetric_t>& output, Idx const& bus_number) {
+    ComplexVector const y_ref = y_bus.math_model_param().source_param;
+    DoubleComplex const y_ref_t =
+        std::accumulate(sources.begin(), sources.end(), DoubleComplex{},
```
TonyXiang8787, line 74: `std::accumulate` is discouraged now. Use `std::transform_reduce`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,22 +59,94 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+                        [&](DoubleComplex sum, Idx const source) { return y_ref[source] + sum; });
+    DoubleComplex const i_ref_t =
+        std::accumulate(sources.begin(), sources.end(), DoubleComplex{}, [&](DoubleComplex sum, Idx const source) {
```
TonyXiang8787, line 77: `std::accumulate` is discouraged now. Use `std::transform_reduce`.

Jerry-Jinfeng-Guo (commented):
I do not know the full progress of this PR as it has changed in shape a couple times since I last checked. I am only here leaving comments regarding the code itself.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -114,11 +189,12 @@ inline void calculate_pf_result(YBus<sym> const& y_bus, PowerFlowInput<sym> cons
     output.load_gen.resize(load_gens_per_bus.element_size());
     output.bus_injection.resize(sources_per_bus.size());
 
+    output.bus_injection = y_bus.calculate_injection(output.u);
     for (auto const& [bus_number, sources, load_gens] : enumerated_zip_sequence(sources_per_bus, load_gens_per_bus)) {
-        calculate_source_result<sym>(sources, bus_number, y_bus, input, output);
-        calculate_load_gen_result<sym>(load_gens, bus_number, input, output, load_gen_func);
+        ComplexValue<sym> i_load_gen_bus{};
```
TonyXiang8787, line 194: So don't define a variable like this.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -114,11 +189,12 @@ inline void calculate_pf_result(YBus<sym> const& y_bus, PowerFlowInput<sym> cons
...
+        calculate_load_gen_result<sym>(load_gens, bus_number, input, output, load_gen_func, i_load_gen_bus);
+        calculate_source_result<sym>(sources, bus_number, y_bus, input, output, i_load_gen_bus);
```
TonyXiang8787, line 196: Instead, first check if there are sources at the bus at all.

If yes, then calculate the `i_load_gen_bus`. If not just continue to the next bus.

TonyXiang8787 (changes requested):
Nice job on the mathematics. Some comments on the code review.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
figueroa1395: @TonyXiang8787 What do you think?

TonyXiang8787: The notebook is already tagged so the everyone can read it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
figueroa1395: Done.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,17 +57,91 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+//     i_inj_i = (y_ref_i * z_ref_t) [ (u_ref_i * y_ref_t - i_ref_t) + i_inj_t]
+// Phase space is used because y_ref is a DoubleComplex in the symmetric case.
+// The equation is organized in such a way to avoid the numerical instability induced in the substraction if y_ref_i
+// is large.
+inline void calculate_multiple_source_result(IdxRange const& sources, YBus<symmetric_t> const& y_bus,
+                                             PowerFlowInput<symmetric_t> const& input,
+                                             ComplexValue<symmetric_t> const& i_inj_t,
+                                             SolverOutput<symmetric_t>& output, Idx const& bus_number) {
+    std::vector<SourceCalcParam> const y_ref = y_bus.math_model_param().source_param;
+    DoubleComplex const y_ref_t = std::transform_reduce(sources.begin(), sources.end(), DoubleComplex{}, std::plus<>{},
```
TonyXiang8787, line 71: angle bracket is not needed. just `std::pluc{}` is enough. Please change in all places

tests/cpp_integration_tests/test_math_solver.cpp
```diff
@@ -196,7 +196,7 @@ TEST_CASE("Test math solver") {
     DoubleComplex const yref = 10.0 - 50.0i;
     pf_input.source = {vref};
     // source param and result
-    param.source_param = {yref};
+    param.source_param = {SourceCalcParam{yref, yref}};
```
Jerry-Jinfeng-Guo, line 199: Here and below: my blind wonder without knowing anything, wouldn't it suffice to simply do
```suggestion
    param.source_param = SourceCalcParam{yref, yref};
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,17 +57,91 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+    ComplexValue<sym> const i_load_gen_bus =
+        std::transform_reduce(load_gens.begin(), load_gens.end(), ComplexValue<sym>{}, std::plus{},
+                              [&](Idx const load_gen) { return output.load_gen[load_gen].i; });
+    ComplexValue<sym> const i_inj_t = conj(output.bus_injection[bus_number] / output.u[bus_number]) - i_load_gen_bus;
+    if (sources.size() == 1) {
+        output.source[*sources.begin()].i = i_inj_t;
+        output.source[*sources.begin()].s = output.u[bus_number] * conj(output.source[*sources.begin()].i);
+    } else {
+        calculate_multiple_source_result(sources, y_bus, input, i_inj_t, output, bus_number);
+    }
```
Jerry-Jinfeng-Guo, line 142: Out of curiosity: isn't single source just one case of multiple sources?

figueroa1395: Refer to this [notebook](https://github.com/PowerGridModel/power-grid-model-workshop/blob/experiment/source-calculation/source_calculation/source_calculation.ipynb) to see the mathematical difference in the calculation of one versus multiple sources. The detailed discussion we can have elsewhere, together with someone more experienced in power systems.

tests/cpp_integration_tests/test_math_solver.cpp
figueroa1395: `{SourceCalcParam{yref, yref}}` is needed because `param.source_param` is of type `std::vector<power_grid_model::SourceCalcParam>`

Jerry-Jinfeng-Guo: Then a bad naming detected ;)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,17 +57,91 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
 
-template <symmetry_tag sym>
-inline void calculate_source_result(IdxRange const& sources, Idx bus_number, YBus<sym> const& y_bus,
-                                    PowerFlowInput<sym> const& input, SolverOutput<sym>& output) {
+// Calculates current and power injection of source i for multiple symmetric sources at a node.
+// The current injection of source i to the bus in phase space is:
+//     i_inj_i = (y_ref_i * z_ref_t) [ (u_ref_i * y_ref_t - i_ref_t) + i_inj_t]
+// Phase space is used because y_ref is a DoubleComplex in the symmetric case.
+// The equation is organized in such a way to avoid the numerical instability induced in the substraction if y_ref_i
+// is large.
```
nitbharambe, line 65:
```suggestion
/// @brief Calculates current and power injection of source i for multiple symmetric sources at a node.
/// The current injection of source i to the bus in phase space is:
///     i_inj_i = (y_ref_i * z_ref_t) [ (u_ref_i * y_ref_t - i_ref_t) + i_inj_t]
/// The equation is organized in such a way to avoid the numerical instability induced in the substraction if y_ref_i
/// is large.
```
- docstring would be recommended instead of normal comments.
- Also, there are no phases in symmetrical case in a mathematical sense.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -57,17 +57,91 @@ template <symmetry_tag sym> inline void copy_y_bus(YBus<sym> const& y_bus, Compl
...
+    }
+}
+
+// Calculates current and power injection for source i for multiple asymmetric sources at a node.
+// The current injection of source i to the bus in 012 domain is:
+// i_inj_0 = y_ref_i_0 / y_ref_t_0 * i_inj_t_0
+// i_inj_1 = y_ref_i_1 * [(u_ref_i_1 - i_ref_t_1 / y_ref_t_1) + i_inj_t_1 / y_ref_t_1]
+// i_inj_2 = y_ref_i_2 / y_ref_t_2 * i_inj_t_2
+// Note: u_ref_i_0 = u_ref_i_2 = 0
+// 012 domain is used to solve the numerical instability as for the symmetric case.
```
nitbharambe, line 92:
```suggestion
/// @brief Calculates current and power injection for source i for multiple asymmetric sources at a node.
/// The current injection of source i to the bus in 012 domain is:
/// i_inj_0 = y_ref_i_0 / y_ref_t_0 * i_inj_t_0
/// i_inj_1 = y_ref_i_1 * [(u_ref_i_1 - i_ref_t_1 / y_ref_t_1) + i_inj_t_1 / y_ref_t_1]
/// i_inj_2 = y_ref_i_2 / y_ref_t_2 * i_inj_t_2
/// Note: u_ref_i_0 = u_ref_i_2 = 0
/// 012 domain is used to solve the numerical instability as for the symmetric case.
```

nitbharambe (approved):

Issue #548, opened 2024-03-26
[FEATURE] Add support columnar data buffer to save memory usage


TonyXiang8787:
# Background

Power Grid Model currently uses the row-based buffer to share the data across the C-API. For example, the memory layout of a node input buffer looks like:

```
| XXXX OOOO  XXXXXXXX  | XXXX OOOO  XXXXXXXX  | ... |
| id_0       u_rated_0 | id_1       u_rated_1 | ... |
```

Where `X` is a meaningful byte and `O` is an empty byte to align the memory.

In this way we can match the input/update/output data structs exactly as we do in the calculation core. This can deliver the performance benefits as we avoid any copies and the memory layout is exactly matching.

When we need to leave some attributes unspecified in input/update, we set the pre-defined [`null`](https://power-grid-model.readthedocs.io/en/stable/advanced_documentation/native-data-interface.html#basic-data-types) value defined in the place so the core knows the value should be ignored.

# Problem

While this design is CPU-wise very efficient, it could be memory-wise inefficient due to several reasons:

1. If you have many unspecified attributes, you waste the memory to have many `null`.
2. To align the data structure, the compiler will add paddings, as shown above. Those memory will not be used at all.
3. If we are only interested in certain attributes in the output, we still need to create the memory buffer for all the attributes. See the related issue here: #542.

There is a strong case to support columnar data buffers. We give two real-world examples of this issue.

## Example of update buffer

If we have an update buffer of 1000 scenarios of 1000 `sym_load`, the buffer size is `24 * 1000 * 1000 = 24,000,000` bytes. However, we might only need to specify `id` and `p_specified`. If we could provide these two array separately, the buffer size in total is `(8 + 4) * 1000 * 1000 = 12,000,000` bytes. The reduction on memory footprint is 50%!

## Example of output buffer

If we get a result buffer of 1000 scenarios of 1000 `line`, the buffer size is `80 * 1000 * 1000 = 80,000,000` bytes. However, we might only need to know the `loading` output, not even `id`, since we already know the `id` order in the input. The buffer size is `8 * 1000 * 1000 = 8,000,000` bytes. We can save 90% of memory footprint!

# Proposal

We propose to support columnar data buffers across the C-API (and further in the Python API). Both the PGM core and serialization need to support that.

## C-API

[...]

PR #675, opened 2024-07-15
Feature/test row based dataset

mgovers:
Prep-work for #548 : freeze + test current dataset


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -163,6 +166,9 @@ template <dataset_type_tag dataset_type_> class Dataset {
         if (!is_batch() && scenario > 0) {
             throw DatasetError{"Cannot export a single dataset with specified scenario\n"};
         }
+        if (scenario >= batch_size()) {
```
TonyXiang8787, line 169: I proposed to put an assert instead of exception. It is an UB if the user has an overflow, just as many places in the code.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -180,7 +186,12 @@ template <dataset_type_tag dataset_type_> class Dataset {
     }
 
     // get individual dataset from batch
-    Dataset get_individual_scenario(Idx scenario) {
+    Dataset get_individual_scenario(Idx scenario)
+        requires(!is_indptr_mutable_v<dataset_type>)
+    {
+        if (scenario < 0 || scenario >= batch_size()) {
+            throw DatasetError{"Scenario cannot be less than 0 or greater or equal to batch size!\n"};
```
TonyXiang8787, line 193: Same here, an UB tested by assert.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -216,9 +231,14 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
             if constexpr (std::same_as<check_indptr_content, immutable_t>) {
-                if (indptr[0] != 0 || indptr[batch_size()] != total_elements) {
-                    throw DatasetError{
-                        "For a non-uniform buffer, indptr should begin with 0 and end with total_elements!\n"};
+                auto indptr_span = get_indptr_span(indptr);
+                if (auto decreasing_found = std::ranges::adjacent_find(indptr_span, std::greater<>());
+                    decreasing_found != indptr_span.end()) {
+                    throw DatasetError{"For a non-uniform buffer, indptr must be increasing!\n"};
+                }
+                if (indptr_span.front() < 0 || indptr_span.back() > total_elements) {
```
TonyXiang8787, line 239: I am against this. Just keep the original requirement: `indptr` begins with `zero` and ends with `total_elements`. That's the standard for all the `indptr` places.

mgovers: then we need to modify the documentation in https://power-grid-model.readthedocs.io/en/stable/api_reference/python-api-reference.html#power_grid_model.data_types.SparseBatchArray , which says

> The elements are the indices in the data that point to the first element of that scenario.
> The last element is one after the data index of the last element of the last scenario.
> Usually, the last element will therefore be the size of the data.

mgovers: in this particular case, I would actually even go as far as to make this UB (which just so happens to work as expected) if we feel like explicitly adding this support is going too far

TonyXiang8787: Modify the documentation is fine.



> it also would prevent us from making subranges

The user of `Dataset` can provide a shifted pointer to the function. I don't think `Dataset` itself needs to support subranges.

TonyXiang8787: > in this particular case, I would actually even go as far as to make this UB (which just so happens to work as expected) if we feel like explicitly adding this support is going too far

Note that we can get all elements in the get range which is from `zero` to `total_elements`. If we allow `indptr` to have a sub-range, this is confusing.

TonyXiang8787 (approved):

figueroa1395 (approved):

TonyXiang8787 (approved):

PR #663, opened 2024-07-05
add columnar data to data types (but don't use yet)

mgovers:
* This is preparation to #548 (relates to Python API step 1)
* This relates to #465

Issue #465, opened 2024-01-11
Add type hinting to Python API


mgovers:
See the TODO in https://github.com/PowerGridModel/power-grid-model/blob/81b1f8c900152fd32ce64e86cb3fb0edb4f2b7d7/src/power_grid_model/data_types.py#L13

Support for Python 3.8 was dropped in #464

Cbit11:
I'll do it

mgovers:
Awesome! LMK if you need any help

mgovers:
@Cbit11 how is it going with the development?

Jerry-Jinfeng-Guo:
I don't understand this issue yet, what's wrong with this [data_types.py](https://github.com/PowerGridModel/power-grid-model/blob/81b1f8c900152fd32ce64e86cb3fb0edb4f2b7d7/src/power_grid_model/data_types.py#L13) document?

mgovers:
> this may already be done in #613 and #648

this is actually not entirely true.

* Starting python3.9, we have `list[Foo]` and `set[Foo]` instead of `List[Foo]` and `Set[foo]`, as well as `collections.abc.Iterable` and `collections.abc.Mapping` as opposed to `typing.Iterable` and `typing.Mapping`.

* Starting python3.10 ( #638 ) , we will have support for `TypeVar` including `bounds` https://docs.python.org/3.10/library/typing.html#typing.TypeVar . This will enable us to remove syntax like `Union[SingleDataset, Dict[str, SingleComponentData]]` and replace it with `SingleDataset`, where `SingleDataset = Dict[ComponentTypeVar, SingleComponentData]` and `ComponentTypeVar = TypeVar['ComponentTypeVar', bounds=[ComponentType, str]]`
* Starting python3.10, `Union` is deprecated and replaced with the much simpler syntax `|`

PR #676, opened 2024-07-16
Columnar data Python API step 2 

nitbharambe:
Change behavior of output_component type to the decided logic.

### DIFF
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -294,7 +294,7 @@ def copy_output_to_columnar_dataset(
     output_component_types: ComponentAttributeMapping,
     output_type: OutputType,
     available_components: list[ComponentType],
-) -> dict[ComponentType, np.ndarray] | dict[ComponentType, dict[str, np.ndarray]]:
+) -> Dataset:
     """Temporary function to copy row based dataset to a column based dataset as per output_component_types.
     The purpose of this function is to mimic columnar data without any memory overhead benefits.
 
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -294,7 +294,7 @@ def copy_output_to_columnar_dataset(
     output_component_types: ComponentAttributeMapping,
     output_type: OutputType,
     available_components: list[ComponentType],
-) -> Dataset:
+) -> dict[ComponentType, np.ndarray] | dict[ComponentType, dict[str, np.ndarray]]:
     """Temporary function to copy row based dataset to a column based dataset as per output_component_types.
     The purpose of this function is to mimic columnar data without any memory overhead benefits.
 
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -296,6 +296,7 @@ def copy_output_to_columnar_dataset(
     available_components: list[ComponentType],
 ) -> Dataset:
     """Temporary function to copy row based dataset to a column based dataset as per output_component_types.
+    The purpose of this function is to mimic columnar data without any memory overhead benefits.
 
     Args:
         data (Dataset):
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -296,7 +296,7 @@ def copy_output_to_columnar_dataset(
     available_components: list[ComponentType],
 ) -> Dataset:
     """Temporary function to copy row based dataset to a column based dataset as per output_component_types.
-    The purpose of this function is to mimic columnar data without any memory overhead benefits.
+    The purpose of this function is to mimic columnar data without any memory footprint benefits.
 
     Args:
         data (Dataset):
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -10,10 +10,12 @@ Disclaimer!
 We do not officially support this functionality and may remove features in this library at any given time!
 """
 
+from copy import deepcopy
 from typing import List, Optional, Union, cast
 
 import numpy as np
 
+from power_grid_model.core.data_handling import OutputType, process_output_component_types
 from power_grid_model.core.dataset_definitions import ComponentType
 from power_grid_model.data_types import (
     BatchArray,
@@ -27,6 +29,7 @@ from power_grid_model.data_types import (
     SinglePythonDataset,
     SparseBatchArray,
 )
+from power_grid_model.typing import ComponentAttributeMapping
 
 
 def is_nan(data) -> bool:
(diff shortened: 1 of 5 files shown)


tests/unit/test_data_handling.py
```diff
@@ -0,0 +1,93 @@
...
+)
+def test_create_output_data(output_component_types, is_batch, expected):
+    # TODO use is_batch and shorten parameterization after columnar data implementation
+    all_component_count = {CT.node: 4, CT.sym_load: 3, CT.source: 1}
+    batch_size = 15 if is_batch else 1
+    actual = create_output_data(
+        output_component_types=output_component_types,
+        output_type=OutputType.SYM_OUTPUT,
+        all_component_count=all_component_count,
+        is_batch=is_batch,
```
nitbharambe, line 55: The is_batch will be tested after columnar data testing is possible. The whole parameterization can be shortened based on that.

src/power_grid_model/core/data_handling.py
```diff
@@ -151,6 +156,44 @@ def create_output_data(
             shape: Union[Tuple[int], Tuple[int, int]] = (batch_size, count)
         else:
             shape = (count,)
-        result_dict[name] = initialize_array(output_type.value, name, shape=shape, empty=True)
+        attributes = output_component_types[name]
+        if attributes is None:
+            result_dict[name] = initialize_array(output_type.value, name, shape=shape, empty=True)
+        elif attributes in [[], set()]:
```
mgovers, line 162: can't you check whether it's empty in a different way?

src/power_grid_model/core/data_handling.py
```diff
@@ -101,7 +102,7 @@ def prepare_output_view(output_data: Mapping[ComponentType, np.ndarray], output_
 
 
 def create_output_data(
-    output_component_types: Union[Set[ComponentType], List[ComponentType]],
+    output_component_types: OutputComponentNamesType,
```
mgovers, line 105: sparse batch data is not captured by this ticket, correct?

nitbharambe: output would only be dense data.

src/power_grid_model/core/data_handling.py
figueroa1395: @mgovers why is this not sufficient?

src/power_grid_model/core/data_handling.py
```diff
@@ -151,6 +156,44 @@ def create_output_data(
...
 
     return result_dict
+
+
+def validate_output_component_types(output_type: OutputType, dict_output_types: _OutputComponentTypeDict):
+    """Checks dict_output_types for any invalid component names and attribute names
+
+    Args:
+        output_type (OutputType): the type of output that the user will see (as per the calculation options)
+        dict_output_types (Dict[ComponentType, Optional[str]]): output_component_types converted to dictionary
```
figueroa1395, line 178: I think the correct typing for `dict_output_types` in the docstring should be `Dict[ComponentType, Optional[Union[Set[str], List[str]]]]`

src/power_grid_model/typing.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+Type hints for PGM. This includes all miscellaneous type hints not under dataset or dataset_definitions categories
+"""
+from typing import Dict, List, Optional, Set, Union
+
+from power_grid_model.core.dataset_definitions import ComponentType
+
+_OutputComponentTypeDict = Dict[ComponentType, Optional[Union[Set[str], List[str]]]]
+
+OutputComponentNamesType = Optional[Union[Set[ComponentType], List[ComponentType], _OutputComponentTypeDict]]
```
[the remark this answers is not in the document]
figueroa1395: Shouldn't we leave this for #465 entirely? If we start implementing this changes right now, maybe is hard to draw the line on when to stop and one would end up doing half of the other issue? Maybe starting to incorporate small changes like the use of `|` or `list[foo]`, etc, but only to the explicit code in the scope of this issue and leaving `TypeVar` out of this?

src/power_grid_model/core/data_handling.py
figueroa1395: Does it make sense for the user to provide only components for the output but ask them to be in `ColumnarData` type? If yes, the output would be of sparse data, right?

src/power_grid_model/typing.py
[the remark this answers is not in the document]
nitbharambe: Either would do. But since this is new we can do it in this PR itself

src/power_grid_model/core/data_handling.py
mgovers: Check for `isinstance` and then for contents?

src/power_grid_model/core/data_handling.py
mgovers: Sparse/dense is only the case for now and may change in the future if we start applying filtering by e.g. id as well

mgovers: Use case is when a user is only interested in a specific id or only in energized components

src/power_grid_model/core/data_handling.py
nitbharambe: So `isinstance(attributes, (set, list) and len(attributes) == 0`. Both would do, but is there a reason any of one should be preferred?

mgovers: Prevents not creating new objects

mgovers: I also believe (please verify) that since python 3.10 also types like `Mapping` can be used in `isinstance`, which ensures you don't have to do 2 checks

nitbharambe: Yes, but the earlier type hint was specially list or set. Should we extend it to sequence?

> I also believe (please verify) that since python 3.10 also types like `Sequence` can be used in `isinstance`, which ensures you don't have to do 2 checks

mgovers: hmm good call. i believe so but maybe only after #465

figueroa1395 (approved):
Everything looks good to me.

figueroa1395 (dismissed):

src/power_grid_model/_utils.py
```diff
@@ -284,3 +287,40 @@ def convert_single_dataset_to_python_single_dataset(data: SingleDataset) -> Sing
...
     }
+
+
+def copy_output_to_columnar_dataset(
+    output_data: Dataset,
+    output_component_types: ComponentAttributeMapping,
+    output_type: OutputType,
+    available_components: list[ComponentType],
+) -> Dataset:
+    """Temporary function to copy row based dataset to a column based dataset as per output_component_types.
```
Jerry-Jinfeng-Guo, line 298: Since the new return type in `Dataset` is now somewhat like a `BatchArray` but not really, I think we could augment the definition of `Dataset` in data_types.py to differentiate the `ColumnarArray` from `BatchArray`, which in reality are concepts of two categories. In the end, I think there is little difference to the result recipient and the changes I proposed is merely for semantics.

src/power_grid_model/_utils.py
```diff
@@ -296,6 +296,7 @@ def copy_output_to_columnar_dataset(
     available_components: list[ComponentType],
 ) -> Dataset:
     """Temporary function to copy row based dataset to a column based dataset as per output_component_types.
+    The purpose of this function is to mimic columnar data without any memory overhead benefits.
```
Jerry-Jinfeng-Guo, line 299:
```
The purpose of this function is to mimic columnar data without any memory footprint benefits.
```

Jerry-Jinfeng-Guo (changes requested):
I think for now the temporary type hint of `dict[ComponentType, np.ndarray] | dict[ComponentType, dict[str, np.ndarray]]` is ok. We can make subsequent changes when merging the other PR where data types are touched. Only request a minor change in the docstring. i.e., "memory overhead" -> "memory footprint"

figueroa1395:
@Jerry-Jinfeng-Guo I believe you are referring to #465. I will leave it here for reference.

Jerry-Jinfeng-Guo:
> @Jerry-Jinfeng-Guo I believe you are referring to #465. I will leave it here for reference.

https://github.com/PowerGridModel/power-grid-model/pull/663, but yes

Jerry-Jinfeng-Guo (approved):

PR #678, opened 2024-07-18
Feature/columnar dataset

mgovers:
Relates to #548 .

This preps the C++ core dataset for columnar data support.

TODO:

* [x] create row-like element view on columnar data using a `Proxy` pattern
  * [x] `ConstDataset` should only have a getter
  * [x] `MutableDataset` and `WritableDataset` should have both a getter and a setter
* [x] add the view of columnar object to the actual `Dataset` structure

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -38,18 +38,15 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
 struct ComponentInfo {
     MetaComponent const* component;
     // for non-uniform component, this is -1, we use indptr to describe the elements per scenario
-    Idx elements_per_scenario;
-    Idx total_elements;
+    Idx elements_per_scenario{};
+    Idx total_elements{};
 };
 
 struct DatasetInfo {
-    bool is_batch;
-    Idx batch_size; // for single dataset, the batch size is one
-    MetaDataset const* dataset;
+    bool is_batch{false};
+    Idx batch_size{0}; // for single dataset, the batch size is one
+    MetaDataset const* dataset{nullptr};
     std::vector<ComponentInfo> component_info;
-
-    DatasetInfo(bool is_batch, Idx batch_size, MetaDataset const* dataset)
-        : is_batch(is_batch), batch_size(batch_size), dataset(dataset) {}
 };
 
 template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRange {
@@ -75,7 +72,10 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
             for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
                 auto const& meta_attribute = get_meta_attribute(attribute_idx);
                 std::byte* buffer_ptr = reinterpret_cast<std::byte*>(data_[attribute_idx]) + meta_attribute.size * idx_;
-                meta_attribute.get_value(&value, buffer_ptr, 0);
+                
+                ctype_func_selector(meta_attribute.ctype, [&value, &meta_attribute, &buffer_ptr]<class T> {
+                    *reinterpret_cast<T*>(buffer_ptr) = meta_attribute.get_attribute<T>(reinterpret_cast<RawDataConstPtr>(&value));
+                });
             }
             return *this;
         }
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -56,9 +56,6 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
     class iterator;
 
     class Proxy {
-      private:
-        using ConstProxy = ColumnarAttributeRange<T, const_dataset_t>::Proxy;
-
       public:
         using value_type = T;
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -47,6 +47,9 @@ struct DatasetInfo {
     Idx batch_size; // for single dataset, the batch size is one
     MetaDataset const* dataset;
     std::vector<ComponentInfo> component_info;
+
+    DatasetInfo(bool is_batch, Idx batch_size, MetaDataset const* dataset)
+        : is_batch(is_batch), batch_size(batch_size), dataset(dataset), component_info() {}
 };
 
 template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRange {
@@ -183,6 +186,8 @@ template <dataset_type_tag dataset_type_> class Dataset {
         Data* data;
         AttributeBuffers attributes;
         std::span<Indptr> indptr;
+
+        Buffer(Data* data = nullptr, std::span<Indptr> indptr = {}) : data(data), indptr(indptr) {}
     };
 
     template <class StructType>
@@ -192,11 +197,7 @@ template <dataset_type_tag dataset_type_> class Dataset {
     static constexpr Idx invalid_index{-1};
 
     Dataset(bool is_batch, Idx batch_size, std::string_view dataset_name, MetaData const& meta_data)
-        : meta_data_{&meta_data},
-          dataset_info_{.is_batch = is_batch,
-                        .batch_size = batch_size,
-                        .dataset = &meta_data.get_dataset(dataset_name),
-                        .component_info = {}} {
+        : meta_data_{&meta_data}, dataset_info_{is_batch, batch_size, &meta_data.get_dataset(dataset_name)} {
         if (dataset_info_.batch_size < 0) {
             throw DatasetError{"Batch size cannot be negative!\n"};
         }
@@ -213,7 +214,7 @@ template <dataset_type_tag dataset_type_> class Dataset {
         : meta_data_{&other.meta_data()}, dataset_info_{other.get_description()} {
         for (Idx i{}; i != other.n_components(); ++i) {
             auto const& buffer = other.get_buffer(i);
-            buffers_.push_back(Buffer{.data = buffer.data, .indptr = buffer.indptr});
+            buffers_.push_back(Buffer{buffer.data, buffer.indptr});
         }
     }
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -119,7 +119,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         iterator(Idx idx, std::span<AttributeBuffer<Data> const> attribute_buffers)
             : current_{idx, attribute_buffers}, idx_{idx} {}
 
-        Idx const get_idx() const { return idx_; }
+        Idx get_idx() const { return idx_; }
 
       private:
         friend class boost::iterator_core_access;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -140,7 +140,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         : size_{size}, attribute_buffers_{std::move(attribute_buffers)} {}
     ColumnarAttributeRange(ColumnarAttributeRange::iterator begin, ColumnarAttributeRange::iterator end)
         : size_{std::distance(begin, end)},
-          start_{std::distance(get(0), begin)},
+          start_{begin->idx_},
           attribute_buffers_{begin->attribute_buffers_.begin(), begin->attribute_buffers_.end()} {
         assert(begin + std::distance(begin, end) == end);
     }
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -60,7 +60,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         using value_type = T;
 
         Proxy() = default;
-        Proxy(Idx idx, std::span<Data*> data,
+        Proxy(Idx idx, std::span<Data* const> data,
               std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes)
             : idx_{idx}, data_{data}, meta_attributes_{meta_attributes} {
             assert(data.size() == meta_attributes.size());
@@ -78,7 +78,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         }
         operator value_type() const { return get(); }
         value_type get() const {
-            value_type result{};
+            std::remove_const_t<value_type> result{};
             for (Idx attribute_idx = 0; attribute_idx < meta_attributes_.size(); ++attribute_idx) {
                 auto const& meta_attribute = get_meta_attribute(attribute_idx);
                 char const* buffer_ptr =
@@ -96,7 +96,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         }
 
         Idx idx_{};
-        std::span<Data*> data_{};
+        std::span<Data* const> data_{};
         std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes_;
     };
 
@@ -105,7 +105,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         using value_type = Proxy;
 
         iterator() = default;
-        iterator(Idx idx, std::span<Data*> data,
+        iterator(Idx idx, std::span<Data* const> data,
                  std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes)
             : current_{idx, data, meta_attributes} {}
 
@@ -128,11 +128,12 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
     };
 
     ColumnarAttributeRange() = default;
-    ColumnarAttributeRange(Idx size, std::span<Data*> data,
+    ColumnarAttributeRange(Idx size, std::span<Data* const> data,
                            std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes)
         : size_{size}, data_{data}, meta_attributes_{meta_attributes} {}
 
-    Idx size() const { return size_; }
+    constexpr Idx size() const { return size_; }
+    constexpr bool empty() const { return size_ == 0; }
     iterator begin() const { return get(0); }
     iterator end() const { return get(size_); }
     auto iter() const { return boost::make_iterator_range(begin(), end()); }
@@ -142,7 +143,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
     iterator get(Idx idx) const { return iterator{idx, data_, meta_attributes_}; }
 
     Idx size_{};
-    std::span<Data*> data_{};
+    std::span<Data* const> data_{};
     std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes_;
 };
 
@@ -161,9 +162,16 @@ template <dataset_type_tag dataset_type_> class Dataset {
     template <class StructType>
     using DataStruct = std::conditional_t<is_data_mutable_v<dataset_type>, StructType, StructType const>;
 
+    struct AttributeBuffers {
+        std::vector<Data*> data;
+        std::vector<std::reference_wrapper<MetaAttribute const>> meta_attributes;
+    };
+
+    // for columnar buffers, Data* data is empty and attributes is filled
+    // for uniform buffers, indptr is empty
     struct Buffer {
         Data* data;
-        // for uniform buffer, indptr is empty
+        AttributeBuffers attributes;
         std::span<Indptr> indptr;
     };
 
@@ -263,6 +271,22 @@ template <dataset_type_tag dataset_type_> class Dataset {
         } else {
             buffers_[idx].indptr = {};
         }
+    }
+
+    void add_attribute_buffer(std::string_view component, std::string_view attribute, Data* data) {
+        Idx const idx = find_component(component, true);
+        Buffer& buffer = buffers_[idx];
+        if (buffer.data != nullptr) {
+            throw DatasetError{"Cannot add attribute buffers to row-based dataset!\n"};
+        }
+        if (std::ranges::find_if(buffer.attributes.meta_attributes, [&attribute](MetaAttribute const& meta_attribute) {
+                return meta_attribute.name == attribute;
+            }) != buffer.attributes.meta_attributes.end()) {
+            throw DatasetError{"Cannot have duplicated attribute buffers!\n"};
+        }
+        buffer.attributes.data.emplace_back(data);
+        buffer.attributes.meta_attributes.emplace_back(
+            dataset_info_.component_info[idx].component->get_attribute(attribute));
     }
 
     // get buffer by component type
@@ -399,14 +423,16 @@ template <dataset_type_tag dataset_type_> class Dataset {
 
     template <class StructType>
     RangeObject<StructType> get_columnar_buffer_span_impl(Idx scenario, Idx component_idx) const {
-        return {};
-        // // return empty span if the component does not exist
-        // if (component_idx < 0) {
-        //     return {};
-        // }
-        // // return span based on uniform or non-uniform buffer
-        // ComponentInfo const& info = dataset_info_.component_info[component_idx];
-        // Buffer const& buffer = buffers_[component_idx];
+        // return empty span if the component does not exist
+        if (component_idx < 0) {
+            return {};
+        }
+        // return span based on uniform or non-uniform buffer
+        ComponentInfo const& info = dataset_info_.component_info[component_idx];
+        Buffer const& buffer = buffers_[component_idx];
+        assert(buffer.data == nullptr);
+
+        return RangeObject<StructType>{info.total_elements, buffer.attributes.data, buffer.attributes.meta_attributes};
         // auto const ptr = reinterpret_cast<StructType*>(buffer.data);
         // if (scenario < 0) {
         //     return RangeObject<StructType>{ptr, ptr + info.total_elements};
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -69,7 +69,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         Proxy& operator=(value_type const& value)
             requires is_data_mutable_v<dataset_type>
         {
-            for (Idx attribute_idx = 0; attribute_idx < meta_attributes_.size(); ++attribute_idx) {
+            for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
                 auto const& meta_attribute = get_meta_attribute(attribute_idx);
                 char* buffer_ptr = reinterpret_cast<char*>(data_[attribute_idx]) + meta_attribute.size * idx_;
                 meta_attribute.get_value(&value, buffer_ptr, 0);
@@ -79,7 +79,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         operator value_type() const { return get(); }
         value_type get() const {
             std::remove_const_t<value_type> result{};
-            for (Idx attribute_idx = 0; attribute_idx < meta_attributes_.size(); ++attribute_idx) {
+            for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
                 auto const& meta_attribute = get_meta_attribute(attribute_idx);
                 char const* buffer_ptr =
                     reinterpret_cast<char const*>(data_[attribute_idx]) + meta_attribute.size * idx_;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -74,7 +74,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         {
             for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
                 auto const& meta_attribute = get_meta_attribute(attribute_idx);
-                char* buffer_ptr = reinterpret_cast<char*>(data_[attribute_idx]) + meta_attribute.size * idx_;
+                std::byte* buffer_ptr = reinterpret_cast<std::byte*>(data_[attribute_idx]) + meta_attribute.size * idx_;
                 meta_attribute.get_value(&value, buffer_ptr, 0);
             }
             return *this;
@@ -84,8 +84,8 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
             std::remove_const_t<value_type> result{};
             for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
                 auto const& meta_attribute = get_meta_attribute(attribute_idx);
-                char const* buffer_ptr =
-                    reinterpret_cast<char const*>(data_[attribute_idx]) + meta_attribute.size * idx_;
+                std::byte const* buffer_ptr =
+                    reinterpret_cast<std::byte const*>(data_[attribute_idx]) + meta_attribute.size * idx_;
                 meta_attribute.set_value(&result, buffer_ptr, 0);
             }
             return result;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -422,7 +422,6 @@ template <dataset_type_tag dataset_type_> class Dataset {
         ComponentInfo const& info = dataset_info_.component_info[component_idx];
         Buffer const& buffer = buffers_[component_idx];
         auto const ptr = reinterpret_cast<StructType*>(buffer.data);
-        // std::span<StructType> const total_range{ptr, ptr + info.total_elements};
         if (scenario < 0) {
             return std::span<StructType>{ptr, ptr + info.total_elements};
         }
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
@@ -16,7 +16,6 @@
 
 #include <span>
 #include <string_view>
-#include <variant>
 
 namespace power_grid_model {
 
@@ -181,7 +180,6 @@ template <dataset_type_tag dataset_type_> class Dataset {
     template <class StructType>
     using RangeObject = std::conditional_t<is_data_mutable_v<dataset_type>, mutable_range_object<StructType>,
                                            const_range_object<StructType>>;
-    template <typename StructType> using BufferSpanType = std::variant<std::span<StructType>, RangeObject<StructType>>;
 
     static constexpr Idx invalid_index{-1};
 
@@ -291,7 +289,7 @@ template <dataset_type_tag dataset_type_> class Dataset {
             throw DatasetError{"Cannot have duplicated attribute buffers!\n"};
         }
         AttributeBuffer<Data> attribute_buffer{
-            .data = static_cast<void*>(data),
+            .data = data,
             .meta_attribute = &dataset_info_.component_info[idx].component->get_attribute(attribute)};
         buffer.attributes.emplace_back(attribute_buffer);
     }
@@ -299,7 +297,7 @@ template <dataset_type_tag dataset_type_> class Dataset {
     // get buffer by component type
     template <template <class> class type_getter, class ComponentType,
               class StructType = DataStruct<typename type_getter<ComponentType>::type>>
-    BufferSpanType<StructType> get_buffer_span(Idx scenario = invalid_index) const {
+    std::span<StructType> get_buffer_span(Idx scenario = invalid_index) const {
         assert(scenario < batch_size());
 
         if (!is_batch() && scenario > 0) {
@@ -310,14 +308,38 @@ template <dataset_type_tag dataset_type_> class Dataset {
         return get_buffer_span_impl<StructType>(scenario, idx);
     }
 
+    template <template <class> class type_getter, class ComponentType,
+              class StructType = DataStruct<typename type_getter<ComponentType>::type>>
+    RangeObject<StructType> get_columnar_buffer_span(Idx scenario = invalid_index) const {
+        assert(scenario < batch_size());
+
+        if (!is_batch() && scenario > 0) {
+            throw DatasetError{"Cannot export a single dataset with specified scenario\n"};
+        }
+
+        Idx const idx = find_component(ComponentType::name, false);
+        return get_columnar_buffer_span_impl<StructType>(scenario, idx);
+    }
+
     // get buffer by component type for all scenarios in vector span
     template <template <class> class type_getter, class ComponentType,
               class StructType = DataStruct<typename type_getter<ComponentType>::type>>
-    std::vector<BufferSpanType<StructType>> get_buffer_span_all_scenarios() const {
+    std::vector<std::span<StructType>> get_buffer_span_all_scenarios() const {
         Idx const idx = find_component(ComponentType::name, false);
-        std::vector<BufferSpanType<StructType>> result(batch_size());
+        std::vector<std::span<StructType>> result(batch_size());
         for (Idx scenario{}; scenario != batch_size(); scenario++) {
             result[scenario] = get_buffer_span_impl<StructType>(scenario, idx);
+        }
+        return result;
+    }
+
+    template <template <class> class type_getter, class ComponentType,
+              class StructType = DataStruct<typename type_getter<ComponentType>::type>>
+    std::vector<RangeObject<StructType>> get_columnar_buffer_span_all_scenarios() const {
+        Idx const idx = find_component(ComponentType::name, false);
+        std::vector<RangeObject<StructType>> result(batch_size());
+        for (Idx scenario{}; scenario != batch_size(); scenario++) {
+            result[scenario] = get_columnar_buffer_span_impl<StructType>(scenario, idx);
         }
         return result;
     }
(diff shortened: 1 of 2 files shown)


tests/cpp_unit_tests/test_dataset.cpp
```diff
@@ -443,7 +629,336 @@ TEST_CASE_TEMPLATE("Test dataset (common)", DatasetType, ConstDataset, MutableDa
...
+                                    static_assert(std::same_as<decltype(buffer_span[idx]),
+                                                               mutable_range_object<A::InputType>::Proxy>);
+                                }
+                                return static_cast<A::InputType>(buffer_span[idx]);
+                            }();
+                            CHECK(element.id == id_buffer[idx]);
+                            CHECK(element.a1 == a1_buffer[idx]);
+                            CHECK(is_nan(element.a0));
+                        }
+                    };
```
Jerry-Jinfeng-Guo, line 662: Sonar cloud
```suggestion
template <typename DatasetType, typename BufferSpan, typename Idx>
auto get_element(BufferSpan const& buffer_span, Idx idx) {
    if constexpr (std::same_as<DatasetType, ConstDataset>) {
        static_assert(std::same_as<decltype(buffer_span[idx]), const_range_object<A::InputType const>::Proxy>);
    } else {
        static_assert(std::same_as<decltype(buffer_span[idx]), mutable_range_object<A::InputType>::Proxy>);
    }
    return static_cast<A::InputType>(buffer_span[idx]);
}

auto const check_span = [&](auto const& buffer_span) {
    CHECK(buffer_span.size() == total_elements);
    for (Idx idx = 0; idx < buffer_span.size(); ++idx) {
        auto const element = get_element<DatasetType>(buffer_span, idx);
        CHECK(element.id == id_buffer[idx]);
        CHECK(element.a1 == a1_buffer[idx]);
        CHECK(is_nan(element.a0));
    }
};
```

figueroa1395: Thanks. Will test and resolve if passes SonarCloud.

figueroa1395:
Sonarcloud is still complaining about something that doesn't really make sense. Anyhow, I will try something different to fix it.

tests/cpp_unit_tests/test_dataset.cpp
figueroa1395: I had to remove the `static_assert` statements and refactor that, but now it passes.

Jerry-Jinfeng-Guo:
The number of sonar cloud issues is a bit on the higher side. Could you address them? If there is no solution, you can leave them and we can eventually disable them if we couldn't find solutions.

figueroa1395:
@Jerry-Jinfeng-Guo I addressed most of them now. I didn't notice before. Thanks for the heads up.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
     Idx batch_size; // for single dataset, the batch size is one
     MetaDataset const* dataset;
     std::vector<ComponentInfo> component_info;
+
+    DatasetInfo(bool is_batch, Idx batch_size, MetaDataset const* dataset)
```
TonyXiang8787, line 51: Why creating a constructor? This type is not aggregate anymore and we cannot use the named attribute initialization.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -59,20 +175,29 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+
+    // for columnar buffers, Data* data is empty and attributes is filled
+    // for uniform buffers, indptr is empty
     struct Buffer {
         Data* data;
-        // for uniform buffer, indptr is empty
+        AttributeBuffers attributes;
         std::span<Indptr> indptr;
+
+        Buffer(Data* data = nullptr, std::span<Indptr> indptr = {}) : data(data), indptr(indptr) {}
```
TonyXiang8787, line 190: same here. Don't create a constructor.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -59,20 +175,29 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+
     static constexpr Idx invalid_index{-1};
 
     Dataset(bool is_batch, Idx batch_size, std::string_view dataset_name, MetaData const& meta_data)
-        : meta_data_{&meta_data},
-          dataset_info_{.is_batch = is_batch,
-                        .batch_size = batch_size,
-                        .dataset = &meta_data.get_dataset(dataset_name),
-                        .component_info = {}} {
+        : meta_data_{&meta_data}, dataset_info_{is_batch, batch_size, &meta_data.get_dataset(dataset_name)} {
```
TonyXiang8787, line 200: because you have now a constructor, we cannot use the nice named attributes anymore.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -89,7 +214,7 @@ template <dataset_type_tag dataset_type_> class Dataset {
         : meta_data_{&other.meta_data()}, dataset_info_{other.get_description()} {
         for (Idx i{}; i != other.n_components(); ++i) {
             auto const& buffer = other.get_buffer(i);
-            buffers_.push_back(Buffer{.data = buffer.data, .indptr = buffer.indptr});
+            buffers_.push_back(Buffer{buffer.data, buffer.indptr});
```
TonyXiang8787, line 217: because you have now a constructor, we cannot use the nice named attributes anymore.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -59,20 +175,29 @@ template <dataset_type_tag dataset_type_> class Dataset {
     template <class StructType>
     using DataStruct = std::conditional_t<is_data_mutable_v<dataset_type>, StructType, StructType const>;
 
+    struct AttributeBuffers {
```
TonyXiang8787, line 178: I propose to use one struct per attribute.

```c++
struct AttributeBuffer {
    Data* data;
    MetaAttribute const* attribute;
    bool is_c_order{true};
    Idx stride{};
};
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -59,20 +175,29 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+        std::vector<Data*> data;
+        std::vector<std::reference_wrapper<MetaAttribute const>> meta_attributes;
+    };
+
+    // for columnar buffers, Data* data is empty and attributes is filled
+    // for uniform buffers, indptr is empty
     struct Buffer {
         Data* data;
-        // for uniform buffer, indptr is empty
+        AttributeBuffers attributes;
```
TonyXiang8787, line 187: this becomes

```c++
std::vector<AttributeBuffer> attributes;
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+        friend class ColumnarAttributeRange;
+        friend class iterator;
+
+        MetaAttribute const& get_meta_attribute(Idx attribute_idx) const {
+            return meta_attributes_[attribute_idx].get();
+        }
+
+        Idx idx_{};
+        std::span<Data* const> data_{};
+        std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes_;
```
TonyXiang8787, line 104: Similar as below, define something like `std::vector<AttributeBuffer>`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+    iterator end() const { return get(size_); }
+    auto iter() const { return boost::make_iterator_range(begin(), end()); }
+    auto operator[](Idx idx) const { return *get(idx); }
+
+  private:
+    iterator get(Idx idx) const { return iterator{idx, data_, meta_attributes_}; }
+
+    Idx size_{};
+    std::vector<Data*> data_{};
+    std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes_;
```
TonyXiang8787, line 160: Similar as below, define something like `std::vector<AttributeBuffer>`.

TonyXiang8787, line 160: Similar as below, define something like `std::vector<AttributeBuffer>`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
 
+template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRange {
+  public:
+    using Data = std::conditional_t<is_data_mutable_v<dataset_type>, void, void const>;
+
+    class iterator;
+
+    class Proxy {
+      public:
+        using value_type = T;
```
TonyXiang8787, line 63: The value type should also be a non-const. Return a const value is very tricky. So static assert here that `T` is not const.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+            for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
+                auto const& meta_attribute = get_meta_attribute(attribute_idx);
+                std::byte* buffer_ptr = reinterpret_cast<std::byte*>(data_[attribute_idx]) + meta_attribute.size * idx_;
+                meta_attribute.get_value(&value, buffer_ptr, 0);
+            }
+            return *this;
+        }
+        operator value_type() const { return get(); }
+        value_type get() const {
+            std::remove_const_t<value_type> result{};
```
TonyXiang8787, line 84: This is not needed if we enforce `T` to be non-const.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+            requires is_data_mutable_v<dataset_type>
+        {
+            for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
+                auto const& meta_attribute = get_meta_attribute(attribute_idx);
+                std::byte* buffer_ptr = reinterpret_cast<std::byte*>(data_[attribute_idx]) + meta_attribute.size * idx_;
+                meta_attribute.get_value(&value, buffer_ptr, 0);
+            }
+            return *this;
+        }
+        operator value_type() const { return get(); }
```
TonyXiang8787, line 82: Nice design using implicit conversion operator!

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+
+      private:
+        friend class boost::iterator_core_access;
+
+        constexpr auto dereference() const -> value_type { return *&current_; }
+        constexpr auto dereference() -> value_type { return *&current_; }
+        constexpr auto equal(iterator const& other) const {
+            return current_.idx_ == other.current_.idx_ &&
+                   current_.meta_attributes_.size() == other.current_.meta_attributes_.size() &&
+                   current_.meta_attributes_.begin() == other.current_.meta_attributes_.begin();
```
TonyXiang8787, line 124: These two comparison is confusing. Especially the comparing the `begin`. If the attributes are empty, this is actually UB.

We don't need to do this comparison. Just compare the index. (internal) Users are not supposed to compare two iterators from different range.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+        iterator() = default;
+        iterator(Idx idx, std::span<Data* const> data,
+                 std::span<std::reference_wrapper<MetaAttribute const> const> meta_attributes)
+            : current_{idx, data, meta_attributes} {}
+
+      private:
+        friend class boost::iterator_core_access;
+
+        constexpr auto dereference() const -> value_type { return *&current_; }
+        constexpr auto dereference() -> value_type { return *&current_; }
```
TonyXiang8787, line 120: You don't need this one right?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -172,6 +313,18 @@ template <dataset_type_tag dataset_type_> class Dataset {
         Idx const idx = find_component(ComponentType::name, false);
         return get_buffer_span_impl<StructType>(scenario, idx);
     }
+    template <template <class> class type_getter, class ComponentType,
+              class StructType = DataStruct<typename type_getter<ComponentType>::type>>
+    RangeObject<StructType> get_columnar_buffer_span(Idx scenario = invalid_index) const {
+        assert(scenario < batch_size());
```
TonyXiang8787, line 319: assert here if the buffer if the row buffer is `nullptr`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -172,6 +313,18 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+
+        if (!is_batch() && scenario > 0) {
+            throw DatasetError{"Cannot export a single dataset with specified scenario\n"};
+        }
+
+        Idx const idx = find_component(ComponentType::name, false);
+        return get_columnar_buffer_span_impl<StructType>(scenario, idx);
+    }
```
TonyXiang8787, line 327: we need another public function to tell if a component is row or column based.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+            assert(data.size() == meta_attributes.size());
+        }
+
+        Proxy& operator=(value_type const& value)
+            requires is_data_mutable_v<dataset_type>
+        {
+            for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
+                auto const& meta_attribute = get_meta_attribute(attribute_idx);
+                std::byte* buffer_ptr = reinterpret_cast<std::byte*>(data_[attribute_idx]) + meta_attribute.size * idx_;
+                meta_attribute.get_value(&value, buffer_ptr, 0);
```
TonyXiang8787, line 78: This will go through a function pointer call every time. As this is a very tight loop, we have another function which does not go to function pointer:

https://github.com/PowerGridModel/power-grid-model/blob/e22b528cc8084d45ef86c7140fe97e8aa0fb5e45/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp#L102-L107

TonyXiang8787: You need a template lamda to do that.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -45,8 +47,122 @@ struct DatasetInfo {
...
+            return *this;
+        }
+        operator value_type() const { return get(); }
+        value_type get() const {
+            std::remove_const_t<value_type> result{};
+            for (Idx attribute_idx = 0; attribute_idx < static_cast<Idx>(meta_attributes_.size()); ++attribute_idx) {
+                auto const& meta_attribute = get_meta_attribute(attribute_idx);
+                std::byte const* buffer_ptr =
+                    reinterpret_cast<std::byte const*>(data_[attribute_idx]) + meta_attribute.size * idx_;
+                meta_attribute.set_value(&result, buffer_ptr, 0);
```
TonyXiang8787, line 89: Same here
This will go through a function pointer call every time. As this is a very tight loop, we have another function which does not go to function pointer:

https://github.com/PowerGridModel/power-grid-model/blob/e22b528cc8084d45ef86c7140fe97e8aa0fb5e45/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp#L102-L107

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: See example here

https://github.com/PowerGridModel/power-grid-model/blob/e22b528cc8084d45ef86c7140fe97e8aa0fb5e45/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp#L781-L787

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: See example here

https://github.com/PowerGridModel/power-grid-model/blob/e22b528cc8084d45ef86c7140fe97e8aa0fb5e45/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp#L504-L509

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Addressed: Made ComponentInfo and DatasetInfo aggregate with default values.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers: moved the `remove_const_t` to the `value_type` using declaration

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers: added `remove_const_t`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers: done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: This check is done in `get_columnar_buffer_span_impl`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Reverted to old after removing constructors.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Reverted to old after removing constructors.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Added `is_columnar` public member function to address this.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done

figueroa1395:
With the use of `is_columnar`, some cleanup is to be implemented. Then, the tests will be refactored to use the new implementations.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -12,8 +12,11 @@
 #include "dataset_fwd.hpp"
 #include "meta_data.hpp"
 
+#include <boost/range.hpp>
+
 #include <span>
 #include <string_view>
+#include <variant>
```
mgovers, line 19: currently not used

figueroa1395: It is used at `template <typename StructType> using BufferSpanType = std::variant<std::span<StructType>, RangeObject<StructType>>;`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -36,16 +39,125 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
...
+    Idx elements_per_scenario{};
+    Idx total_elements{};
 };
 
 struct DatasetInfo {
-    bool is_batch;
-    Idx batch_size; // for single dataset, the batch size is one
-    MetaDataset const* dataset;
+    bool is_batch{false};
+    Idx batch_size{0}; // for single dataset, the batch size is one
```
TonyXiang8787, line 48: I think the default `batch_size` should be one because you set `is_batch` default to `false`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: Why use a variant?

figueroa1395: I was trying to combine together `get_columnar_buffer_span` and `get_buffer_span` into one, as much of the code between both was repeated, hence I used `variant` to hold both possible return types. However, now I am having many thoughts about this.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Correct. Done.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: I don't think we need `variant` here. Just define two different functions with different return type. In both function you put a `assert` to check if the user is calling the correct function.

The user is supposed to call `is_coloumnar` first and then choose to what function to call.

figueroa1395: Done.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Removed the overload after passing just a `std::span` of `AttributeBuffer` instead of a vector to `Proxy`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -12,6 +12,8 @@
 #include "dataset_fwd.hpp"
 #include "meta_data.hpp"
 
+#include <boost/range.hpp>
```
mgovers, line 15: do you still use this header? if so, you can keep it

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -36,16 +38,125 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
...
-    Idx batch_size; // for single dataset, the batch size is one
-    MetaDataset const* dataset;
+    bool is_batch{false};
+    Idx batch_size{1}; // for single dataset, the batch size is one
+    MetaDataset const* dataset{nullptr};
     std::vector<ComponentInfo> component_info;
 };
+template <typename Data> struct AttributeBuffer {
+    Data* data;
+    MetaAttribute const* meta_attribute;
```
mgovers, line 53:
```suggestion
    Data* data{nullptr};
    MetaAttribute const* meta_attribute{nullptr};
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -104,6 +220,8 @@ template <dataset_type_tag dataset_type_> class Dataset {
     Buffer const& get_buffer(std::string_view component) const { return get_buffer(find_component(component, true)); }
     Buffer const& get_buffer(Idx i) const { return buffers_[i]; }
 
+    bool is_columnar(Data* data) const { return data == nullptr; }
```
mgovers, line 223: this usage looks weird... maybe it's better to make this the private implementation and only `public`ly expose `bool is_columnar(std::string_view component)` and `bool is_columnar(Idx i)` and optionally `bool is_columnar(Buffer const& buffer)`

or (unlikely) you can add this to the `ComponentInfo` of a certain component

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -36,16 +38,125 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
...
+    iterator begin() const { return get(0); }
+    iterator end() const { return get(size_); }
+    auto iter() const { return boost::make_iterator_range(begin(), end()); }
+    auto operator[](Idx idx) const { return *get(idx); }
+
+  private:
+    iterator get(Idx idx) const { return iterator{idx, attribute_buffers_}; }
+
+    Idx size_{};
+    std::vector<AttributeBuffer<Data>> attribute_buffers_;
```
mgovers, line 155: ~i wonder whether this should be a `span` as well... really unsure about that tho, but it does make the iterator nicer, i guess~ EDIT: see my other comment on the bug with the current implementation

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -36,16 +38,125 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
...
+    };
+
+    ColumnarAttributeRange() = default;
+    ColumnarAttributeRange(Idx size, std::vector<AttributeBuffer<Data>> attribute_buffers)
+        : size_{size}, attribute_buffers_{std::move(attribute_buffers)} {}
+    ColumnarAttributeRange(ColumnarAttributeRange::iterator begin, ColumnarAttributeRange::iterator end)
+        : size_{std::distance(begin, end)},
+          attribute_buffers_{begin->attribute_buffers_.begin(), begin->attribute_buffers_.end()} {
+        assert(begin + std::distance(begin, end) == end);
+    }
```
mgovers, line 142: i have a feeling there's still a bug in this: let's say you have a range `[0, 1, 2, 3, 4]` and you take sub-range `[1, 2, 3]` (like we do in `get_columnar_buffer_span_impl`), then you take:

* `auto const start = total_range.begin() + 1` ==> `iterator{Proxy{.idx_ = 1, .attribute_buffers_ = total_range.attribute_buffers}}`
* `auto const stop = total_range.begin() + 4` ==> `iterator{Proxy{.idx_ = 4, .attribute_buffers_ = total_range.attribute_buffers}}`
* create new `result = ColumnarAttributeRange(start, stop)` ==> `ColumnarAttributeRange{.size = 4 - 1 = 3, .attribute_buffers_ = {begin->attribute_buffers_.begin(), begin->attribute_buffers_.end()} = std::vector(total_range.attribute_buffers_.begin(), total_range.attribute_buffers_.end()) = copy of total_range.attribute_buffers}`
* meaning `result.attribute_buffers_ == total_range.attribute_buffers_`, so that
* `result[0].get() == total_range[0].get()`
* this is incorrect, because we need `result[1].get() == total_range[0].get()`

(please add a test to verify that this is indeed true)

we need to add an `Idx start_` to this `ColumnarAttributeRange` and below set `iterator get(Idx idx) const { return iterator{start_ + idx, attribute_buffers_}; }`

NOTE: this wasn't an issue before because we could offset the data buffers for all attributes individually before when doing the range copy. Now, we can't anymore. That this also means that now, the entire `ColumnarAttributeRange` itself can also just directly contain an `std::span<AttributeBuffer const>` instead of an owned `std::vector<AttributeBuffer>`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -268,6 +428,29 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+        RangeObject<StructType> const total_range{info.total_elements, buffer.attributes};
+        if (scenario < 0) {
+            return total_range;
+        }
+        if (info.elements_per_scenario < 0) {
+            return RangeObject<StructType>{total_range.begin() + buffer.indptr[scenario],
+                                           total_range.begin() + buffer.indptr[scenario + 1]};
+        }
+        return RangeObject<StructType>{total_range.begin() + info.elements_per_scenario * scenario,
+                                       total_range.begin() + info.elements_per_scenario * (scenario + 1)};
```
mgovers, line 452: this still contains a bug. see my other comment. Maybe that's also the reason you were struggling with combining the implementation with the row-based buffer

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -55,16 +167,22 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
-        std::span<Indptr> indptr;
+        Data* data{nullptr};
+        std::vector<AttributeBuffer<Data>> attributes;
+        std::span<Indptr> indptr{};
     };
 
+    template <class StructType>
+    using RangeObject = std::conditional_t<is_data_mutable_v<dataset_type>, mutable_range_object<StructType>,
+                                           const_range_object<StructType>>;
+    template <typename StructType> using BufferSpanType = std::variant<std::span<StructType>, RangeObject<StructType>>;
```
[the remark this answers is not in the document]
figueroa1395: Maybe you reviewed an outdated version? I removed `std::variant` entirely, hence `BufferSpanType` was removed as well.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: I need it for `boost::make_iterator_range`. However, it may be removed after the possible bug you found. I'll update this here.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: I realize how the original implementation of `is_columnar(Data* data)` is very weird. I think that leaving the three implementations that you suggest and remove the original should be better.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: I get your point, but I think you made a typo: What we should be getting is `result[0].get() == total_range[1].get()`, shouldn't we? I will add a test to first check this and then implement the fix.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: The bug you have encountered does not affect the use of `boost::make_iterator_range`. So `<boost/range.hpp>` will stay.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers: Whoops yes you're right. Covariance vs contravariance traumas all over again

figueroa1395: The bug is indeed present. A test has been added to confirm it.

figueroa1395: Test passes now with new implementation after introducing `start_` for the `ColumnarAttributeRange` constructor with `iterator`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: This implementation should be fine now. The tests are a little bit simplified as well, since taking sub-ranges now works as intended.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: You were right, only `span` is needed now.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -36,16 +38,130 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
...
+        friend class boost::iterator_core_access;
+
+        constexpr auto dereference() const { return current_; }
+        constexpr auto equal(iterator const& other) const { return current_.idx_ == other.current_.idx_; }
+        constexpr auto distance_to(iterator const& other) const { return other.current_.idx_ - current_.idx_; }
+        constexpr void increment() { ++current_.idx_; }
+        constexpr void decrement() { --current_.idx_; }
+        constexpr void advance(Idx n) { current_.idx_ += n; }
+
+        Idx idx_{};
```
mgovers, line 134: obsolete because it's not touched by `increment`, `decrement`, `advance`, etc.
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -36,16 +38,130 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
...
+
+    class iterator : public boost::iterator_facade<iterator, T, boost::random_access_traversal_tag, Proxy, Idx> {
+      public:
+        using value_type = Proxy;
+
+        iterator() = default;
+        iterator(Idx idx, std::span<AttributeBuffer<Data> const> attribute_buffers)
+            : idx_{idx}, current_{idx, attribute_buffers} {}
+
+        Idx get_idx() const { return idx_; }
```
mgovers, line 122:
```suggestion
        Idx get_idx() const { return current.idx_; }
```

mgovers, line 122: why do you need this function?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -104,6 +225,12 @@ template <dataset_type_tag dataset_type_> class Dataset {
     Buffer const& get_buffer(std::string_view component) const { return get_buffer(find_component(component, true)); }
     Buffer const& get_buffer(Idx i) const { return buffers_[i]; }
 
+    bool is_columnar(std::string_view component) const {
+        return buffers_[find_component(component, true)].data == nullptr;
```
mgovers, line 229: try to keep the logic in one location
```suggestion
        return is_columnar(find_component(component, true));
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -104,6 +225,12 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+    }
+    bool is_columnar(Idx const i) const { return buffers_[i].data == nullptr; }
```
mgovers, line 231: try to keep the logic in one location
```suggestion
    constexpr bool is_columnar(Idx const i) const { return is_columnar(buffers_[i]); }
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -104,6 +225,12 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+    bool is_columnar(Buffer const& buffer) const { return buffer.data == nullptr; }
```
mgovers, line 232: `constexpr` where possible
```suggestion
    constexpr bool is_columnar(Buffer const& buffer) const { return buffer.data == nullptr; }
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -268,6 +436,29 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+        RangeObject<StructType> const total_range{info.total_elements, buffer.attributes};
+        if (scenario < 0) {
+            return total_range;
+        }
+        if (info.elements_per_scenario < 0) {
+            return RangeObject<StructType>{total_range.begin() + buffer.indptr[scenario],
+                                           total_range.begin() + buffer.indptr[scenario + 1]};
+        }
+        return RangeObject<StructType>{total_range.begin() + info.elements_per_scenario * scenario,
+                                       total_range.begin() + info.elements_per_scenario * (scenario + 1)};
```
mgovers, line 460: please re-try with the template function to combine the logic for row-based and columnar. I think you should be OK now that the bug is fixed

mgovers (commented):
I won't approve cause I did a lot of the work, but here's some feedback still

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: It was used in old (now obsolete) implementation. I will remove now.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done. Now it works seamlessly.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -259,14 +438,23 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+            return {};
         }
-        return std::span<StructType>{ptr + info.elements_per_scenario * scenario,
-                                     ptr + info.elements_per_scenario * (scenario + 1)};
+        // return span based on uniform or non-uniform buffer
+        ComponentInfo const& info = dataset_info_.component_info[component_idx];
+        Buffer const& buffer = buffers_[component_idx];
+        assert(is_columnar(buffer));
+        RangeObject<StructType> const total_range{info.total_elements, buffer.attributes};
+        return get_span_impl<RangeObject<StructType>>(buffer, info, scenario, total_range);
```
mgovers, line 457: optional nitpick: use template argument deduction

```suggestion
        return get_span_impl(buffer, info, scenario, total_range, RangeObject<StructType>{info.total_elements, buffer.attributes});
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -259,14 +438,23 @@ template <dataset_type_tag dataset_type_> class Dataset {
         ComponentInfo const& info = dataset_info_.component_info[component_idx];
         Buffer const& buffer = buffers_[component_idx];
         auto const ptr = reinterpret_cast<StructType*>(buffer.data);
-        if (scenario < 0) {
-            return std::span<StructType>{ptr, ptr + info.total_elements};
-        }
-        if (info.elements_per_scenario < 0) {
-            return std::span<StructType>{ptr + buffer.indptr[scenario], ptr + buffer.indptr[scenario + 1]};
+        auto const total_range = std::span<StructType>{ptr, ptr + info.total_elements};
+        return get_span_impl<std::span<StructType>>(buffer, info, scenario, total_range);
```
mgovers, line 442: optional nitpick: use template argument deduction
```suggestion
        return get_span_impl(buffer, info, scenario, std::span<StructType>{ptr, ptr + info.total_elements});
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -250,6 +414,21 @@ template <dataset_type_tag dataset_type_> class Dataset {
         buffers_.push_back(Buffer{});
     }
 
+    template <class RangeType>
+    RangeType get_span_impl(Buffer const& buffer, ComponentInfo const& info, Idx scenario,
+                            RangeType const& total_range) const {
```
mgovers, line 419: it's common to start with the most important argument, then other relevant arguments, and then metadata. Compare e.g. in Python: positional arguments followed by arguments (`*args`) and then keyword arguments (`**kwargs`)

```suggestion
    RangeType get_span_impl(RangeType const& total_range, Idx scenario, IdxVector const& indptr, ComponentInfo const& info) const {
```

Note: not relevant for this PR but if you have one or more output targets (captured by non-const `&`), then it's common to put them last in the list

mgovers (commented):
only a couple nitpicks. please let someone else do the final review

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -36,16 +38,127 @@ static_assert(is_indptr_mutable_v<writable_dataset_t>);
...
 struct DatasetInfo {
-    bool is_batch;
-    Idx batch_size; // for single dataset, the batch size is one
-    MetaDataset const* dataset;
+    bool is_batch{false};
+    Idx batch_size{1}; // for single dataset, the batch size is one
+    MetaDataset const* dataset{nullptr};
     std::vector<ComponentInfo> component_info;
 };
+template <typename Data> struct AttributeBuffer {
```
TonyXiang8787, line 51: use `class` instead of `typename` to be consistant with the rest of the code. Please do so in all the places.

tests/cpp_unit_tests/test_dataset.cpp
```diff
@@ -115,9 +174,161 @@ DatasetType create_dataset(bool const is_batch, Idx const batch_size, MetaDatase
     CHECK(info.component_info.empty());
     return dataset;
 };
+
+void check_nan_or_equal(double first, double second) {
+    CHECK(((first == second) || (is_nan(first) == is_nan(second))));
+}
+
+void check_nan(double value) { CHECK(is_nan(value)); }
```
nitbharambe, line 182: This function can be directly used

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -104,6 +222,12 @@ template <dataset_type_tag dataset_type_> class Dataset {
     Buffer const& get_buffer(std::string_view component) const { return get_buffer(find_component(component, true)); }
     Buffer const& get_buffer(Idx i) const { return buffers_[i]; }
 
+    constexpr bool is_columnar(std::string_view component) const {
+        return is_columnar(find_component(component, true));
```
TonyXiang8787, line 226: I do not agree with the logic here. It throws an exception when the component is not found. This requires extra work later in the main model to check if the component exist or not.

My proposal: if the component is not found, always return false is `is_columnar`. The row-base `get_span` will automatically return an empty range so no extra handeling in main model is needed.

tests/cpp_unit_tests/test_dataset.cpp
mgovers: well, yes but it looks nicer this way because otherwise, when it's used, there's a mixture of `CHECK(foo(bar))` and `check_foo(bar)`. I prefer to have them all be `CHECK` or `check_`, but not a mixture. Optional though

figueroa1395: I will keep `check_nan` because of Martjin's reason above. Since the `CHECK`s are done through auxiliary functions like this one, even though redundant, the same style/convention is mantained througout the tests.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Agree. Done.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: Done.

figueroa1395 (approved):

Issue #601, opened 2024-05-16
CalculationType and Symmetry as a MainModel::Option


mgovers:
move C API code concerning `opt->calculation_type` and `opt->symmetric` also into the `MainModel::Options`.

This is follow-up on the cleanup done in #599 .

It is a good idea to also merge the `calculate_power_flow` and `calculate_state_estimation` exposition functions

_Originally posted by @mgovers in https://github.com/PowerGridModel/power-grid-model/pull/599#discussion_r1603188302_

PR #681, opened 2024-08-01
Feature/calculation type in options

mgovers:
Part 1 of #601 

I do not want to touch the tests at the same time as the core implementation, which is why the follow-up that migrates the tests to the main model will be a separate PR (that will use this PR as a base if it's not merged before then)


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -57,6 +57,64 @@ template <> struct output_type_getter<SolverOutput<asymmetric_t>> {
     template <class T> using type = meta_data::asym_output_getter_s<T>;
 };
 
+struct power_flow_t {};
+struct state_estimation_t {};
+struct short_circuit_t {};
```
Jerry-Jinfeng-Guo, line 62: What are these supposed to mean exactly?

mgovers: these are tags used by the (runtime) function selector to choose the individual implementation functions for the different calculation types

Jerry-Jinfeng-Guo: So `t` stands for tags. Can we call them `_tag` directly? `_t` implies something different to me.

mgovers: Cfr `symmetric_t` vs `symmetry_tag`: `_t` stands for the concrete type while `_tag` stands for the "it is something that looks like a xyz...", i.e. the concept that acts upon a tag type.

We can talk about the conventions but i think that's a whole separate topic altogether. I'd propose to keep it according to the existing conventions in this PR

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -651,106 +710,64 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-            output_result(math_output, result_data, pos);
-        }
-    }
+        if (options.calculation_type == CalculationType::short_circuit) {
+            auto const faults = state_.components.template citer<Fault>();
+            auto const is_three_phase = std::ranges::all_of(
+                faults, [](Fault const& fault) { return fault.get_fault_type() == FaultType::three_phase; });
+            options.calculation_symmetry =
+                is_three_phase ? CalculationSymmetry::symmetric : CalculationSymmetry::asymmetric;
+        };
```
figueroa1395, line 747: I know this is implementation details, but since it is only relevant for short circuit calculations, I think it should be moved to `short_circuit_calculation`, maybe within its own function. Also, the logic would be more straightforward, as in the other two types of calculations the symmetry is obtained there.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -651,106 +710,64 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-    // Batch load flow calculation, propagating the results to result_data
-    template <symmetry_tag sym>
-    BatchParameter calculate_power_flow(Options const& options, MutableDataset const& result_data,
-                                        ConstDataset const& update_data) {
-        return batch_calculation_(
-            [&options](MainModelImpl& model, MutableDataset const& target_data, Idx pos) {
-                auto sub_opt = options; // copy
-                sub_opt.err_tol = pos != ignore_output ? options.err_tol : std::numeric_limits<double>::max();
-                sub_opt.max_iter = pos != ignore_output ? options.max_iter : 1;
+        calculation_type_symmetry_func_selector(
```
figueroa1395, line 748: I think it is worth to briefly discuss the implementation details of how the `func_selectors` are used and work within the main mode, and how they reduced and simplified a bunch of code, to be used within the `calculate` function, in a knowledge sharing session.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -51,18 +51,18 @@ template <class T> constexpr CType ctype_v = ctype_t<T>::value;
...
+        return std::forward<Functor>(f).template operator()<double>(std::forward<Args>(args)...);
     case c_double3:
-        return f.template operator()<RealValue<asymmetric_t>>(std::forward<Args>(args)...);
+        return std::forward<Functor>(f).template operator()<RealValue<asymmetric_t>>(std::forward<Args>(args)...);
     case c_int8:
-        return f.template operator()<int8_t>(std::forward<Args>(args)...);
+        return std::forward<Functor>(f).template operator()<int8_t>(std::forward<Args>(args)...);
     case c_int32:
-        return f.template operator()<int32_t>(std::forward<Args>(args)...);
+        return std::forward<Functor>(f).template operator()<int32_t>(std::forward<Args>(args)...);
```
figueroa1395, line 65: Was this refactor done here to align with the other implementations of `func_selector`s? Because this doesn't necessarily aligns with this scope, does it?

figueroa1395 (commented):
Everything looks fine to me. It can go to the merge queue.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: The problem is that this was part of its own function before, but the logic would become convoluted. The whole point of this PR is to isolate the differences between the calculation types etc. in one place. In the new state after this PR, it is **very** clear that specifically short circuit calculations ignore the provided calculation symmetry. Earlier, that would be hidden behind the fact that some functions are templated and some are not.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: before or after merge?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
mgovers: > Was this refactor done here to align with the other implementations of func_selectors?

yes. sonar cloud was complaining for the other PR (because we didn't do perfect forwarding even though we could. I also updated this PR to ensure that now as well, it is perfectly forwarded. This comes with 2 benefits:

* performance benefits, because the whole function does not need to be copy-pasted every single time
* usability benefits, because lifetimes are preserved when using perfect forwarding

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: Understood. Now it is indeed very straight forward when one sees this.

figueroa1395 (approved):

nitbharambe (approved):

PR #668, opened 2024-07-09
Feature / Binary search in Automatic Tap Changer

Jerry-Jinfeng-Guo:
This PR adds binary search to the tap changer optimization process. The follow tasks are carried out:
- [x] Add `fast_any` tap changing strategy
- [x] Unit test benchmark the number of runs
- [x] Implement binary search in the `fast_any` tap changing strategy
- [x] Implement binary search in the `global_max` and `global_min` tap changing strategy
- [x] Existing unit tests
- [x] Existing validation tests

PR #684, opened 2024-08-06
remove erroneously added unused and test logic after #668

mgovers:
Partially reverts #668 

* Reverts adding unused exception type
* Reverts adding test logic that causes actual issues to remain hidden
* Reverts adding unused test parameters

### DIFF
--- a/tests/unit/test_0Z_model_validation.py
+++ b/tests/unit/test_0Z_model_validation.py
@@ -175,23 +175,13 @@ def test_batch_validation(
     for update_data, reference_result in zip(update_list, reference_output_list):
         model_copy = copy(model)
         model_copy.update(update_data=update_data)
-        try:
-            result = calculation_function(
-                model_copy, **supported_kwargs(kwargs=base_kwargs, supported=calculation_args)
-            )
-        except Exception as e:
-            print(f"An error occurred during calculation: {e}")
-        else:
-            compare_result(result, reference_result, rtol, atol)
+        result = calculation_function(model_copy, **supported_kwargs(kwargs=base_kwargs, supported=calculation_args))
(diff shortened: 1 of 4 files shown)


tests/unit/test_0Z_model_validation.py
```diff
@@ -175,23 +175,13 @@ def test_batch_validation(
...
-            compare_result(result, reference_result, rtol, atol)
+        result = calculation_function(model_copy, **supported_kwargs(kwargs=base_kwargs, supported=calculation_args))
+        compare_result(result, reference_result, rtol, atol)
 
     # execute in batch one go
     for threading in [-1, 0, 1, 2]:
         kwargs = dict(base_kwargs, update_data=update_batch, threading=threading)
-        try:
-            result_batch = calculation_function(model, **supported_kwargs(kwargs=kwargs, supported=calculation_args))
-        except Exception as e:
```
figueroa1395, line 192: Was this exception thrown here before because of the incompatibility added initially (not anymore) when the binary search was introduced for the tap changer strategy? So this no longer is needed, right?

Jerry-Jinfeng-Guo (approved):
Funny the small unused imports are not discovered by the autoation

tests/unit/test_0Z_model_validation.py
Jerry-Jinfeng-Guo: At Python API level the combination error will never be hit and hence not exposed. This change from the binary PR (https://github.com/PowerGridModel/power-grid-model/pull/684) could potentially hide issues with future added features. In this PR this exception handling logic is removed as a patch to https://github.com/PowerGridModel/power-grid-model/pull/684

PR #685, opened 2024-08-06
Implementation of TypeVar in more scripts

### LINKED ISSUE #465: Add type hinting to Python API

zhen0427:
Improve type hinting that is present to Python 3.10 standards (e.g. replace Union with |)
Add type hinting to places where it’s not present yet (make use of TypeVar)

* part 2 of #465

Closes #465


src/power_grid_model/core/buffer_handling.py
```diff
@@ -233,7 +235,7 @@ def get_buffer_view(data: np.ndarray | Mapping[str, np.ndarray], schema: Compone
 
 def create_buffer(
     properties: BufferProperties, schema: ComponentMetaData
-) -> np.ndarray | dict[ComponentType, np.ndarray]:
+) -> np.ndarray | dict[ComponentTypeVar, np.ndarray]:
```
mgovers, line 238: for return types, being specific is fine
```suggestion
) -> np.ndarray | dict[ComponentType, np.ndarray]:
```

src/power_grid_model/core/data_handling.py
```diff
@@ -19,6 +19,8 @@
 from power_grid_model.enum import CalculationType
 from power_grid_model.typing import ComponentAttributeMapping, _ComponentAttributeMappingDict
 
+ComponentTypeVar = TypeVar("ComponentTypeVar", bound=ComponentType | str)
```
mgovers, line 22:
```suggestion
from power_grid_model.typing import ComponentAttributeMapping, _ComponentAttributeMappingDict
from power_grid_model.data_types import ComponentTypeVar
```

src/power_grid_model/core/data_handling.py
```diff
@@ -8,7 +8,7 @@
 
 
 from enum import Enum
-from typing import Mapping
+from typing import Mapping, TypeVar
```
mgovers, line 11:
```suggestion
from typing import Mapping
```

src/power_grid_model/core/data_handling.py
```diff
@@ -147,22 +149,22 @@ def create_output_data(
...
+        available_components (list[ComponentTypeVar]):  all components available in model instance
 
     Raises:
         ValueError: when the type for output_comoponent_types is incorrect
         KeyError: with "unknown component" for any unknown components
         KeyError: with "unknown attributes" for any unknown attributes for a known component
 
     Returns:
-        _OutputComponentTypeDict: processed output_component_types in a dictionary
+        _OutputComponentTypeVarDict: processed output_component_types in a dictionary
```
mgovers, line 167:
```suggestion
        _OutputComponentTypeDict: processed output_component_types in a dictionary
```

src/power_grid_model/core/power_grid_dataset.py
```diff
@@ -244,7 +244,8 @@ def get_buffer_views(self) -> list[CBuffer]:
         return self._buffer_views
 
     def _add_data(
-        self, data: Mapping[ComponentType, np.ndarray] | Mapping[ComponentType, np.ndarray | Mapping[str, np.ndarray]]
+        self,
+        data: Mapping[ComponentType, np.ndarray] | Mapping[ComponentType, np.ndarray | Mapping[str, np.ndarray]],
```
mgovers, line 248:
```suggestion
        data: Mapping[ComponentTypeVar, np.ndarray] | Mapping[ComponentTypeVar, np.ndarray | Mapping[str, np.ndarray]],
```

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -8,7 +8,7 @@
 
 from dataclasses import dataclass
 from enum import IntEnum
-from typing import Any
+from typing import Any, TypeVar
```
mgovers, line 11:
```suggestion
from typing import Any
```

src/power_grid_model/core/power_grid_meta.py
```diff
@@ -20,6 +20,8 @@
 )
 from power_grid_model.core.power_grid_core import AttributePtr, ComponentPtr, DatasetPtr, power_grid_core as pgc
 
+ComponentTypeVar = TypeVar("ComponentTypeVar", bound=ComponentType | str)
+
```
mgovers, line 24:
```suggestion
from power_grid_model.data_types import ComponentTypeVar

```

src/power_grid_model/core/buffer_handling.py
```diff
@@ -8,7 +8,7 @@
 
 
 from dataclasses import dataclass
-from typing import Mapping, Optional
+from typing import Mapping, Optional, TypeVar
```
mgovers, line 11:
```suggestion
from typing import Mapping, Optional
```

src/power_grid_model/core/buffer_handling.py
```diff
@@ -18,6 +18,8 @@
 from power_grid_model.core.power_grid_core import IdxPtr, VoidPtr
 from power_grid_model.core.power_grid_meta import ComponentMetaData
 
+ComponentTypeVar = TypeVar("ComponentTypeVar", bound=ComponentType | str)
```
mgovers, line 21: idem

src/power_grid_model/data_types.py
```diff
@@ -13,7 +13,7 @@
 
 from power_grid_model.core.dataset_definitions import ComponentType
 
-ComponentTypeVar = TypeVar("ComponentTypeVar", ComponentType, str)
+ComponentTypeLike = TypeVar("ComponentTypeLike", bound=ComponentType | str)
```
mgovers, line 16: Technically breaking, but I don't think any of our users uses this as of now
I can also do
```suggestion
ComponentTypeLike = TypeVar("ComponentTypeLike", bound=ComponentType | str)
ComponentTypeVar = ComponentTypeVar
```

src/power_grid_model/core/data_handling.py
```diff
@@ -147,14 +147,14 @@ def create_output_data(
...
-    available_components: list[ComponentType],
+    available_components: list[ComponentTypeLike],
 ) -> _ComponentAttributeMappingDict:
     """Checks valid type for output_component_types. Also checks for any invalid component names and attribute names
 
     Args:
         output_type (OutputType): the type of output that the user will see (as per the calculation options)
         output_component_types (OutputComponentNamesType):  output_component_types provided by user
-        available_components (list[ComponentType]):  all components available in model instance
+        available_components (list[ComponentTypeVar]):  all components available in model instance
```
figueroa1395, line 157: Given the current implementation, it should be `list[ComponentTypeLike]`, instead of `list[ComponentTypeVar]`, shouldn't it?

src/power_grid_model/data_types.py
Jerry-Jinfeng-Guo: Why does it become necessary now to use `bound`? Also your can-also-do has a typo.

figueroa1395: If `ComponentTypeVar` was removed altogether, why not just name this `ComponentTypeVar` instead. That shouldn't be breaking, right?

figueroa1395: @Jerry-Jinfeng-Guo I believe `bound` must be together with `|` to indicate that it can not only be `ComponentType` or `str`, but also subsets of them and subsets of their `union`. If one was to remove bound, I am confident that Pylint, or some other linter, would complain if the user inputs `ComponentType`s with both `ComponentType` and `str`, and that is why initially (before Python 3.10) we had to explicitly write all the type combinations, which made it really ugly and long.

mgovers: > If ComponentTypeVar was removed altogether, why not just name this ComponentTypeVar instead. That shouldn't be breaking, right?

I originally had

```suggestion
ComponentTypeVar = TypeVar("ComponentTypeVar", bound=ComponentType | str)
```

(your suggestion) but i have a problem with how it's used in the code base, not with the object. However, renaming would be maybe breaking if the users are already using it.

>> Why does it become necessary now to use bound?
>
> I believe bound must be together with | to indicate that it can not only be ComponentType or str, but also subsets of them and subsets of their union

@figueroa1395 is right: `bounds` enables subsets of types. Strict matching would prevent mixed combinations and actually also makes deductions from subsets trickier. `bounds` can do both in one go

> Also your can-also-do has a typo.

whoops, thanks

src/power_grid_model/core/data_handling.py
mgovers: yes

src/power_grid_model/data_types.py
figueroa1395: > (your suggestion) but i have a problem with how it's used in the code base, not with the object. However, renaming would be maybe breaking if the users are already using it.

So I understand that the problem is how we, as devs., use it internally, correct? Hence why this naming convention suits it best within the codebase itself?

If so, doing `ComponentTypeVar: TypeAlias = ComponentTypeLike`, as you suggested, should keep it clean within the codebase and prevent breaking changes for the users (I would say let's assume they already use it, just to be safe).

figueroa1395: I believe the problem comes from the fact that the user could do things like:
```
line = initialize_array(DatasetType.input, ComponentType.line, 1)
...

input_data = {
    "line": line,
    ...
}
```
or
```
input_data = {
    ComponentType.node: node,
    "line": line,
}
```
Hence mixing `literals` with the new `ComponenType`s. Such mixing should be supported, but linters usually hate it, even though it should be straight forward. Introducing `TypeVar` with `bound` will make sure no linter complains and that users can freely mix notation (They shouldn't, but it is probably going to happen).

mgovers: FYI: i applied above suggestion. if agree with it, please resolve the comment and approve. otherwise, let's have a quick call

figueroa1395 (approved):

src/power_grid_model/data_types.py
figueroa1395: I approved changes. If @Jerry-Jinfeng-Guo is okay with these answers, please resolve.

Jerry-Jinfeng-Guo: Agree to disagree. Please proceed without my agreement.

mgovers: Potential compromise after continuing on #663

mgovers: cfr. offline sync: this is the solution we go with

```py
ComponentTypeLike = ComponentType | str
ComponentTypeVar = TypeVar("ComponentTypeVar", bound=ComponentTypeLike)  # helper used for type deduction
```

Jerry-Jinfeng-Guo (dismissed):

src/power_grid_model/core/buffer_handling.py
```diff
@@ -231,9 +230,7 @@ def get_buffer_view(data: np.ndarray | Mapping[str, np.ndarray], schema: Compone
     return _get_sparse_buffer_view(data, schema)
 
 
-def create_buffer(
-    properties: BufferProperties, schema: ComponentMetaData
-) -> np.ndarray | dict[ComponentType, np.ndarray]:
+def create_buffer(properties: BufferProperties, schema: ComponentMetaData) -> np.ndarray | dict[str, np.ndarray]:
```
mgovers, line 233: this was an actual bug: `buffer` has type `np.ndarray` or sparse buffer, which has type `{"data": np.ndarray, "indptr": np.ndarray}`. It should not have been `ComponentType`

nitbharambe: Should we introduce another enum :P

src/power_grid_model/typing.py
```diff
@@ -5,8 +5,8 @@
...
 Type hints for PGM. This includes all miscellaneous type hints not under dataset or dataset_definitions categories
 """
-from power_grid_model.core.dataset_definitions import ComponentType
+from power_grid_model.data_types import ComponentTypeLike
 
-_ComponentAttributeMappingDict = dict[ComponentType, set[str] | list[str] | None]
+_ComponentAttributeMappingDict = dict[ComponentTypeLike, set[str] | list[str] | None]
 
-ComponentAttributeMapping = set[ComponentType] | list[ComponentType] | None | _ComponentAttributeMappingDict
+ComponentAttributeMapping = set[ComponentTypeLike] | list[ComponentTypeLike] | None | _ComponentAttributeMappingDict
```
mgovers, line 12: actually, we can revert this to the original state. This is new behavior and we therefore can assume that people using it have migrated to the modern way of doing things

src/power_grid_model/core/dataset_definitions.py
```diff
@@ -72,32 +72,33 @@ class ComponentType(str, Enum, metaclass=_MetaEnum):
...
-    if isinstance(component, str):
-        return ComponentType[component]
-    return component
+    if isinstance(component, ComponentType):
+        return component
+    return ComponentType[component]
 
 
-def _map_to_component_types(data: Mapping[Any, Any]) -> dict[ComponentType, Any]:
+def _map_to_component_types(data: Mapping[ComponentTypeVar, Any]) -> dict[ComponentType, Any]:
```
mgovers, line 102: be as flexible as you need to be on the input side, and as restrictive as you can be on the output side

mgovers: NOTE: since this only affects private core functionality (private because it's prefixed `_`; core because it's not exposed in the `__init__` or other public modules outside the `core` module), this does not affect the user

code_generation/templates/src/power_grid_model/core/dataset_class_maps.py.jinja
```diff
@@ -10,12 +10,12 @@
 {%- set components = all_map['input'].keys() %}
 
 from enum import Enum, EnumMeta
-from typing import Any, Mapping
-
-# pylint: disable=invalid-name
+from typing import Any, Mapping, TypeAlias, TypeVar
```
figueroa1395, line 13: Why is this needed? What is Pylint complaining about?

src/power_grid_model/validation/utils.py
```diff
@@ -162,7 +168,7 @@ def errors_to_string(
     return msg
 
 
-def nan_type(component: str | ComponentType, field: str, data_type: DatasetType = DatasetType.input):
+def nan_type(component: str | ComponentTypeLike, field: str, data_type: DatasetType = DatasetType.input):
```
figueroa1395, line 171: Shouldn't this just be 
```
def nan_type(component: ComponentTypeLike, field: str, data_type: DatasetType = DatasetType.input):
```

figueroa1395 (dismissed):
Two very minor things, but it looks good to me.

src/power_grid_model/data_types.py
```diff
@@ -11,12 +11,10 @@
 
 import numpy as np
 
-from power_grid_model.core.dataset_definitions import ComponentType
-
-ComponentTypeVar = TypeVar("ComponentTypeVar", ComponentType, str)
+from power_grid_model.core.dataset_definitions import ComponentTypeLike  # pylint: disable=unused-import
```
nitbharambe, line 14: What's the purpose of importing btw?

mgovers: interface

code_generation/templates/src/power_grid_model/core/dataset_class_maps.py.jinja
mgovers: what do you mean? is this about the correct line?

src/power_grid_model/core/data_handling.py
```diff
@@ -184,7 +184,8 @@ def process_output_component_types(
     for comp_name, attrs in output_component_types.items():
         if attrs is None:
             continue
-        diff = set(attrs).difference(output_meta[comp_name].dtype.names)
+        attr_names = output_meta[comp_name].dtype.names
+        diff = set(attrs).difference(attr_names) if attr_names is not None else set(attrs)
```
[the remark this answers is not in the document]
mgovers: it **should** never be `None` in our case but NumPy disagrees. Using `assert`s in Python in production code is bad practice, because it suggests test code in production code to most people. IMO it can't hurt to have an additional `None` check

src/power_grid_model/typing.py
```diff
@@ -5,8 +5,8 @@
 """
 Type hints for PGM. This includes all miscellaneous type hints not under dataset or dataset_definitions categories
 """
-from power_grid_model.core.dataset_definitions import ComponentType
+from power_grid_model.core.dataset_definitions import ComponentTypeVar
 
-_ComponentAttributeMappingDict = dict[ComponentType, set[str] | list[str] | None]
+_ComponentAttributeMappingDict = dict[ComponentTypeVar, set[str] | list[str] | None]
```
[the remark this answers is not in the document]
mgovers: `ComponentAttributeMapping` can ingest a `ComponentAttributeMappingDict`. I don't think this is internal use only with the whole columnar thing. As a result, this should be `ComponentTypeVar`

unless we decide that we don't support `str` at all to users (we can still decide now because this is new functionality)

tests/unit/test_serialization.py
mgovers, line 1: maybe these changes should be reverted because this involves test code only

src/power_grid_model/validation/utils.py
mgovers: yes

src/power_grid_model/data_types.py
mgovers: no, correction, you are right: we want them to use `ComponentType` or if really needed they can specify their own type, in which case it's entirely on them. will remove

code_generation/templates/src/power_grid_model/core/dataset_class_maps.py.jinja
figueroa1395: Oops. Added a comment on the correct line.

code_generation/templates/src/power_grid_model/core/dataset_class_maps.py.jinja
```diff
@@ -10,12 +10,12 @@
...
 
 # fmt: off
 
+# pylint: disable=invalid-name
```
figueroa1395, line 17: > what do you mean? is this about the correct line?

This is the correct line. I don't know what happened yesterday.

src/power_grid_model/validation/utils.py
figueroa1395: Resolve when you make the change. I won't so you don't forget.

tests/unit/test_serialization.py
figueroa1395: If the tests reflect the use of the newly introduced data types it should be fine. It is probably just preference, right?

code_generation/templates/src/power_grid_model/core/dataset_class_maps.py.jinja
mgovers: i just changed the order and added a whiteline

src/power_grid_model/typing.py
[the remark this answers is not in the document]
nitbharambe: Its marked as _. 
We convert `ComponentAttributeMapping` to `_ComponentAttributeMappingDict` internally and check if valid things are provided in `ComponentAttributeMapping`. And internally we can cast the string to ComponentType.

The users would not have any use for it. Either way, its really minor so you can resolve the comment.

nitbharambe (approved):

PR #687, opened 2024-08-09
columnar data python api serialization

nitbharambe:
Part of https://github.com/PowerGridModel/power-grid-model/issues/548

- Modify serialization and deserilization per step 3 and 4 of python api in https://github.com/PowerGridModel/power-grid-model/issues/548
- Points marked as TODO need to be addressed before merge


src/power_grid_model/_utils.py
```diff
@@ -305,20 +306,83 @@ def copy_output_to_columnar_dataset(
...
+        output_component_types (OutputComponentNamesType):  output_component_types provided by user
+        available_components (list[ComponentType]):  all components available in model instance
+
+    Returns:
+        _OutputComponentTypeDict: processed output_component_types in a dictionary
+    """
+    # limit all component count to user specified component types in output and convert to a dict
+    if data_filter is None:
+        data_filter = {k: None for k in available_components}
+    elif data_filter == ...:
```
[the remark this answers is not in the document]
nitbharambe: Changed

src/power_grid_model/_utils.py
```diff
@@ -305,20 +306,83 @@ def copy_output_to_columnar_dataset(
...
+    """Raise error if some specified components or attributes are unknown
+
+    Args:
+        data_filter (OutputType): Component to attribtue dictionary
+        dataset_type (DatasetType):  Type of dataset
+
+    Raises:
+        ValueError: when the type for output_comoponent_types is incorrect
+        KeyError: with "unknown component" for any unknown components
+        KeyError: with "unknown attributes" for any unknown attributes for a known component
```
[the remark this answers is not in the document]
nitbharambe: Changed to "unknown component types"

src/power_grid_model/core/serialization.py
```diff
@@ -94,10 +100,16 @@ class Serializer(ABC):
...
-    def __new__(cls, data: Dataset, serialization_type: SerializationType, dataset_type: Optional[DatasetType] = None):
+    def __new__(cls, data: Dataset, serialization_type: SerializationType, dataset_type: DatasetType | None = None):
         instance = super().__new__(cls)
 
-        instance._data = data
+        if not data:
+            instance._data = {}
+        else:
+            dataset_type = dataset_type if dataset_type is not None else get_dataset_type(data)
+            instance._data = copy_to_row_or_columnar_dataset(
```
mgovers, line 110: why is a copy always required? it should only be required when dealing with columnar datasets. and even then, i don't think the copy should be done here but in the dataset itself, because that is something that will change when the API starts supporting columnar data

mgovers (commented):
i did not do an in-depth review yet due to lack of time. please ask someone else if needed

src/power_grid_model/core/serialization.py
nitbharambe: If the input of copy_to_row_or_columnar_dataset is row based and requested output is row based too, a copy is not created. Ill add that to the docstring too.

mgovers: then the name is not good. maybe you can go with `compatibility_convert_to_row_based`

nitbharambe: It should not matter where this copy is being done since it is temporary. This was a place where changes were minimal now and later when we have to remove it.

tests/unit/test_serialization.py
```diff
@@ -34,6 +36,70 @@ def from_msgpack(data):
...
+def is_columnar_filter(data_filter, component):
+    """A function to find if a data_filter will give out row or columnar format for a component"""
+    if data_filter is None:
+        return False
+    if data_filter is Ellipsis:
+        return True
+    return data_filter[component] is not None
+
+
+def is_filtered_out(data_filter, component, attribute=None):
```
figueroa1395, line 70: Please split into `is_component_filtered_out()` and `is_attribute_filtered_out()`.

tests/unit/test_serialization.py
```diff
@@ -34,6 +36,70 @@ def from_msgpack(data):
...
+    if attribute is None:
+        return component not in data_filter
+    return not (
+        data_filter[component] is None or data_filter[component] is Ellipsis or attribute in data_filter[component]
+    )
+
+
+def is_sparse_data_input(serialized_input_data, component):
+    """Checks if serialized_input_data will be of sparse format after deserialization.
+    If there are uneven ids for scenarios in the serialized input, that would result in sparse data"""
```
figueroa1395, line 82: I would argue that any uneven attribute, not just ids, leads to sparse data.

tests/unit/test_serialization.py
```diff
@@ -34,6 +36,70 @@ def from_msgpack(data):
...
+            if is_non_compact_list(scenario_comp_name_comp_idx):
+                scenario_ids.add(scenario_comp_name_comp_idx["id"])
+            else:
+                id_attr_idx = serialized_input_data["attributes"][component].index("id")
+                scenario_ids.add(scenario_comp_name_comp_idx[id_attr_idx])
+        all_scenarios_ids.append(scenario_ids)
+    first_scenario_ids = next(iter(all_scenarios_ids))
+
+    return any(first_scenario_ids != scenario_ids for scenario_ids in all_scenarios_ids)
+
```
figueroa1395, line 101: In connection with above comment: if all the components must have a valid id, always, then this check is enough, but if not, shouldn't we check sparcity by checking that each component of the same type has the same number of attributes and the same attribute types defined accross the whole dataset? It would involve changing this a lot, but I want to confirm.

tests/unit/test_serialization.py
```diff
@@ -267,105 +375,167 @@ def assert_almost_equal(value: np.ndarray, reference: Any):
...
     serialized_dataset: Mapping[str, Any],
     sparse_components: list[ComponentType],
+    data_filter,
 ):
+    """Check a SingleDataset's individual components for correctness"""
+    # TODO What is the purpose of this check? Implement without using derserialized_dataset
     for key in serialized_dataset["data"]:
-        if key not in deserialized_dataset:
+        if key not in deserialized_dataset and not is_filtered_out(data_filter, key):
             assert key in sparse_components
```
figueroa1395, line 388: Weird check. I would assume that this is to check that if a componenet is filtered out, then it must live somewhere (`sparse_components`), but if you check that it is not in the `deserialized_dataset` but it was not filtered out either, why would it be in `sparse_components` then? Is this check actually being made?

tests/unit/test_serialization.py
```diff
@@ -267,105 +375,167 @@ def assert_almost_equal(value: np.ndarray, reference: Any):
...
-                for attr, values in input_entry.items():
-                    assert attr in result_entry.dtype.names
-                    assert_almost_equal(result_entry[attr], values)
+    for component, component_input in serialized_dataset["data"].items():
+        if is_filtered_out(data_filter, component):
+            assert component not in deserialized_dataset
+            continue
+
+        component_result = deserialized_dataset[component]
+        assert isinstance(component_input, list)  # TODO Why?
```
figueroa1395, line 397: I this the example you provided to differentiate compact list from non-compact helps understand why this check?
```
{
    "version": "1.0", 
    "type": dataset_type, 
    "is_batch": False,
    "attributes": {
        "sym_load": ["status", "p_specified", "q_specified"],
        "source": ["status"],
    }
    "data": {
        "line": [{}, {"from_status": 1, "to_status": 1}],
        "source": [[1], [0], {"status": 1, "u_ref": 1.03}],
        "sym_load": [[9, 1.01e6, 0.21e6], [10, 1.02e6, 0.22e6]],
}
```
If you see the components in data, everything is organized in a list. So I guess this is to check that the serialization format is correct.

tests/unit/test_serialization.py
```diff
@@ -267,105 +375,167 @@ def assert_almost_equal(value: np.ndarray, reference: Any):
...
+
+        # Complete component data checks
+        if is_columnar_filter(data_filter, component):
+            assert isinstance(component_result, dict)
+            assert all(len(v) == len(component_input) for v in component_result.values())
+        else:
+            assert isinstance(component_result, np.ndarray)
```
figueroa1395, line 404: Isn't this `assert isinstance(component_result, SingleArray)`?

tests/unit/test_serialization.py
```diff
@@ -267,105 +375,167 @@ def assert_almost_equal(value: np.ndarray, reference: Any):
         np.testing.assert_almost_equal(value, reference)
 
 
-def assert_scenario_correct(
-    deserialized_dataset: Mapping[ComponentType, np.ndarray],
+def assert_single_dataset_entries(
```
figueroa1395, line 378: I would be in favor of splitting this function if possible, since several different checks are done at once, it becomes very cognitively complex.

tests/unit/test_serialization.py
```diff
@@ -267,105 +375,167 @@ def assert_almost_equal(value: np.ndarray, reference: Any):
...
+        # Individual data entry checks
+        for comp_idx, input_entry in enumerate(component_input):
+            if is_non_compact_list(input_entry):
+                for attr in input_entry:
+                    if is_columnar_filter(data_filter, component):
+                        if is_filtered_out(data_filter, component, attr):
+                            assert attr not in component_result
+                            continue
+                        assert attr in component_result.keys()
+                        assert_almost_equal(component_result[attr][comp_idx], component_input[comp_idx][attr])
```
figueroa1395, line 416: Nitpick: Maybe change names of `component_result` and `component_input` to reflect that the second comes from deserialized data and the first from serialized one?

tests/unit/test_serialization.py
```diff
@@ -267,105 +375,167 @@ def assert_almost_equal(value: np.ndarray, reference: Any):
...
+                            assert attr not in component_result
+                            continue
+                        assert attr in component_result.keys()
+                        assert_almost_equal(component_result[attr][comp_idx], component_input[comp_idx][attr_idx])
+                    else:
+                        assert attr in component_result[comp_idx].dtype.names
+                        assert_almost_equal(component_result[comp_idx][attr], component_input[comp_idx][attr_idx])
+
+
+def assert_single_dataset_structure(deserialized_dataset, data_filter):
```
figueroa1395, line 434: I will leave this here for my reference. I will continue the review tomorrow.

figueroa1395 (commented):
Partially finished. Will continue in the morning

src/power_grid_model/core/serialization.py
```diff
@@ -94,10 +100,16 @@ class Serializer(ABC):
     _dataset: CConstDataset
     _serializer: SerializerPtr
 
-    def __new__(cls, data: Dataset, serialization_type: SerializationType, dataset_type: Optional[DatasetType] = None):
+    def __new__(cls, data: Dataset, serialization_type: SerializationType, dataset_type: DatasetType | None = None):
```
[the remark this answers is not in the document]
nitbharambe: Yup, no good reason I did it. Its just a personal preference to not have an additional keyword here

tests/unit/test_serialization.py
nitbharambe: Yes, its equivalent. The isinstance would work in this case where SingleArray is just a np.ndarray.
However it would not work when we try something like: isinstance(data, SingleDataset).
Hence I personally don't like to use it. Is it specially recommended by standard to use it?

tests/unit/test_serialization.py
nitbharambe: Yes indeed. The goal is to deduce if the input to tests, which is serialized in this case, would be converted to a sparse or a dense dataset.
The logic is only applicable for this test file.

And when its uneven content inside each batch, its sparse dataset. "id" is the common attribute in all cases here, hence used "id" to differentiate between them.

tests/unit/test_serialization.py
```diff
@@ -267,105 +385,167 @@ def assert_almost_equal(value: np.ndarray, reference: Any):
...
-        assert_scenario_correct(
+        assert_single_dataset_entries(
             deserialized_dataset,  # type: ignore[arg-type]
             serialized_dataset,
-            [key for key, value in deserialized_dataset.items() if not isinstance(value, np.ndarray)],
+            [],
+            # [
+            #     key for key, value in deserialized_dataset.items() if not isinstance(value, np.ndarray)
+            # ],
+            # # TODO Why check sparse components for non batch scenarios
```
figueroa1395, line 523: You don't as per our discussion.

figueroa1395 (commented):
I completed the review.

src/power_grid_model/core/serialization.py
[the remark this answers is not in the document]
figueroa1395: I would guess that since we had a PR that involved moving forward towards the use of `|`, it would be better to keep it, but if we don't have any concrete guidelines, either should be fine.

tests/unit/test_serialization.py
nitbharambe: Exactly. I removed the check now. In the main as well it never made it to this control loop

tests/unit/test_serialization.py
nitbharambe: As you are aware, tried and reverted as mypy was causing issued.

figueroa1395 (approved):
Looks good to me.

PR #690, opened 2024-08-12
Bugfix/dont crash on invalid homogeneous batch dataset

mgovers:
## The issue

Non-caught exceptions caused by real-world update scenarios thrown by a worker thread in parallel batch calculations cause fatal crashes.

## Considerations

After investigation related to exceptions during thread creation/execution, the following findings were made:

* [x] It is possible that the main thread throws an error when creating a thread
  * [x] If this happens, stack unwinding is done
  * [x] then a joinable `std::thread` will be destructed
  * [x] this causes `std::terminate()`
  * [x] This is OK because it is of the highest severity and the PowerGridModel instance cannot continue with its current tasks
    * [x] Decision: intended behavior. the users' program crashes and we do not catch this to throw a `PowerGridError`
* [x] It is possible that a worker thread raises a exception
  * [x] Option 1: in a scenario
    * [x] this should be handled by the scenario runner and instead a `PowerGridBatchError` should be raised
    * [x] This is intended behavior
  * [x] Option 2: during creation of the thread
    * [x] Option 2a: during copy of MainModel, e.g.
      * [x] This points at a severe issue like running out of memory
      * [x] This causes `std::terminate()`
      * [x] This is intended behavior (see above)
    * [x] Option 2b: during other caching mechanisms
      * [x] This may be caused by a real-life scenario (see below)
      * [x] This causes `std::terminate()` (fatal crash).
      * [x] This is a bug: instead, it should cause an actual `PowerGridError`
      * [x] This is solved by this PR

## Preconditions

The above-mentioned bug solved in this PR arises when all of the following conditions are satisfied:
1. There is a batch calculation
2. Running in parallel with at least 2 threads (`threading=0` or `threading=x` where `x >= 2`)
3. With at least 2 scenarios
4. With homogeneous update data (all scenarios touch the same IDs)
5. And at least 1 ID in the update data is not present in the input data

These conditions trigger the following contributing factors:
* The 5th condition triggers the exception
* The 4th condition triggers throwing the exception in the setup part instead of in a scenario
* The 1st, 2nd and 3rd condition triggers throwing the exception in a worker thread

## The solution

The caching in which the exception was moved from the worker-thread setup part to all-scenario combined setup. This prevents non-handled exceptions from being raised in the worker thread

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -528,22 +528,27 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         // cache component update order if possible
         bool const is_independent = MainModelImpl::is_update_independent(update_data);
 
-        SequenceIdx scenario_sequence = is_independent ? get_sequence_idx_map(update_data) : SequenceIdx{};
+        std::shared_ptr<SequenceIdx const> all_scenarios_sequence{
+            is_independent ? std::make_shared<SequenceIdx const>(get_sequence_idx_map(update_data)) : nullptr};
 
-        return [&base_model, &exceptions, &infos, &calculation_fn, &result_data, &update_data, is_independent,
-                scenario_sequence](Idx start, Idx stride, Idx n_scenarios) {
+        return [&base_model, &exceptions, &infos, &calculation_fn, &result_data, &update_data, all_scenarios_sequence,
+                is_independent](Idx start, Idx stride, Idx n_scenarios) {
             assert(n_scenarios <= narrow_cast<Idx>(exceptions.size()));
             assert(n_scenarios <= narrow_cast<Idx>(infos.size()));
+            assert(is_independent == (all_scenarios_sequence != nullptr));
 
             Timer const t_total(infos[start], 0000, "Total in thread");
 
-            auto model = [&base_model, &infos](Idx scenario_idx) {
+            auto const copy_model = [&base_model, &infos](Idx scenario_idx) {
                 Timer const t_copy_model(infos[scenario_idx], 1100, "Copy model");
                 return MainModelImpl{base_model};
-            }(start);
+            };
+            auto model = copy_model(start);
 
-            auto [setup, winddown] =
-                scenario_update_restore(model, update_data, is_independent, scenario_sequence, infos);
+            SequenceIdx cacheable_scenario_sequence = SequenceIdx{};
+            auto const& scenario_sequence = is_independent ? *all_scenarios_sequence : cacheable_scenario_sequence;
+            auto [setup, winddown] = scenario_update_restore(model, update_data, scenario_sequence,
+                                                             cacheable_scenario_sequence, is_independent, infos);
 
             auto calculate_scenario = MainModelImpl::call_with<Idx>(
                 [&model, &calculation_fn, &result_data, &infos](Idx scenario_idx) {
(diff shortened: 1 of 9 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -528,23 +528,27 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+        return [&base_model, &exceptions, &infos, &calculation_fn, &result_data, &update_data, all_scenarios_sequence,
                 is_independent](Idx start, Idx stride, Idx n_scenarios) {
             assert(n_scenarios <= narrow_cast<Idx>(exceptions.size()));
             assert(n_scenarios <= narrow_cast<Idx>(infos.size()));
+            assert(is_independent == (all_scenarios_sequence != nullptr));
 
             Timer const t_total(infos[start], 0000, "Total in thread");
 
-            auto copy_model = [&base_model, &infos](Idx scenario_idx) {
+            auto const copy_model = [&base_model, &infos](Idx scenario_idx) {
```
TonyXiang8787, line 542: As we agreed. Remove this named lambda. Just create the lambda and immediately execute it. The name `copy_model` is confusing.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -528,23 +528,27 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         // cache component update order if possible
         bool const is_independent = MainModelImpl::is_update_independent(update_data);
 
-        return [&base_model, &exceptions, &infos, &calculation_fn, &result_data, &update_data,
+        std::shared_ptr<SequenceIdx const> all_scenarios_sequence{
```
TonyXiang8787, line 531: Why use a shared pointer? The thread does not need to own the resource right? We can just create an object at the main thread and capture the reference into the thread lambda. The main thread always out-lives the work threads.

mgovers: If we make it as a member variable of the `MainModel` instance in the main thread, then it looks weird because we reference a cache object that's not part of `this`.

On the other hand, if we define it locally in this function, then it will go out of lifetime the moment this function returns its lambda. The lambda then contains a reference to an out-of-lifetime object, which causes segmentation violations.

The only other potentially nice solution I see that does not use `shared_ptr` would have both the `all_scenarios_sequence` and the `cacheable_scenario_sequence` declared as member variables; the former in the main thread, and the latter in the worker thread

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
TonyXiang8787: then name it like `copy_model_functor`

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
TonyXiang8787: Why will the local object go out of life before the work threads finish?

You first declare and create the local object `all_scenarios_sequence` locally in the function entry of `calculate`. And capture it by reference in the lambda's. The work threads will finish before the `calculate` finish.

TonyXiang8787: Ah I see the problem. This function is not the main entry point, but just a sub-routine to return the lambdas.

So my proposal:

1. Declare `all_scenarios_sequence` the entry point of `calculate`.
2. Pass `all_scenarios_sequence` by rvalue reference (which will be mutated) into this function `sub_batch_calculation_` returning the lambda.
3. Capture `all_scenarios_sequence` by const reference in the lambda.
4. Keep `all_scenarios_sequence` alive in the main entry point function until threads are finished.

mgovers: step 2 should be lvalue reference, otherwise ownership will be transferred to this function, after which it will go out of scope

TonyXiang8787: Yes, adjusted

TonyXiang8787 (approved):

PR #691, opened 2024-08-14
find boost using CONFIG

mgovers:
Conforms to https://cmake.org/cmake/help/latest/policy/CMP0167.html

Boost 1.70 (the first version that supports `CONFIG` mode for `find_package`) was released in 2019. It is reasonable to assume all people who want to build themselves have bumped beyond that


mgovers:
`libboost-headers` is not shipped via `conda-forge` with the `BoostConfig.cmake` file.

Likely cause in https://github.com/conda-forge/boost-feedstock/blob/main/recipe/install-lib.bat

This **may** be fixed by changing to `boost-cpp`, but this is out of scope for now. Closing PR.

PR #694, opened 2024-08-15
fix compiler flag with new setuptools

TonyXiang8787:


TonyXiang8787:
@mgovers it is now compatible with both `setuptools` versions.

mgovers (approved):

Issue #662, opened 2024-07-04
[FEATURE] Automatic tap changer in step-up transformer support 


Jerry-Jinfeng-Guo:
What need to happen:
- [x] Figure out the meeth
- [x] Control side at the node with higher rated voltage 
- [x] Tap side and the control side at the same side

Jerry-Jinfeng-Guo:
Most changes will happen in `power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp`

What needs to happen:
* [ ] validation to check for unsupported scenarios when it comes to enabled tap controllers:
 * [ ] more than one enabled tap controller on the same transformer
 * [ ] an enabled tap controller is coupled with a not-energized transformer 
 * [ ] an enabled tap controller is coupled with a two/three winding transformer with disconnected tap side or control side
 * [ ] an enabled tap controller is coupled with a two winding transformer which are not fully connected
 * [ ] an enabled tap controller is coupled with a three winding transformer with two sides disconnected
 * [ ] an enabled tap controller is coupled with a transformer which is not reachable by transformer ranking graph, see logic below
* [ ] un-regulated transformers given weight of `0` instead of `1`:
      in lines 119-120 and 157-158, from
       ```c++
            add_to_edge(state, edges, edge_props, from_node, to_node, {unregulated_idx, 1});
            add_to_edge(state, edges, edge_props, to_node, from_node, {unregulated_idx, 1});
       ```
       to
       ```c++
            add_to_edge(state, edges, edge_props, from_node, to_node, {unregulated_idx, 0);
            add_to_edge(state, edges, edge_props, to_node, from_node, {unregulated_idx, 0});
       ```
* [ ] logic determining direction of transformer tap/control sides needs update:
  * [ ] two winding transformer in lines 151-153
  * [ ] three winding transformer in lines 107-109 
* [ ] after assigning ranks of all regulated transformers, one round of check, between line 274 and 276
  * [ ] two winding transformers: for every enabled such transformer, rank at the control side (node connected to it) should be 1 greater than the other side, throw exception if not the case.
  * [ ] three winding transformers: 
    * [ ] control side == tap side: rank at the control side (node connected to it) should be 1 greater than any one of the two secondary sides, throw exception if not the case.
    * [ ] control side != tap side: rank at the control side (node connected to it) should be 1 greater than the secondary side with tap, throw exception if not the case.
* [ ] Adjust control logic 
[...]

PR #698, opened 2024-08-20
Feature / Updated transformer weight before transformer ranking

Jerry-Jinfeng-Guo:
This PR implements the updated transformer ranking in https://github.com/PowerGridModel/power-grid-model/issues/662

* [x] Unregulated transformer edges get weight `0` instead of `1`
* [x] Update unit tests
  * [x] Add unit tests with un-regulated transformers
  * [x] Update existing tests based on new correct results
  * [x] Update docs


power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -113,11 +113,14 @@ inline void process_trafo3w_edge(main_core::main_model_state_c auto const& state
             auto const& non_tap_side_node = tap_at_first_side ? to_node : from_node;
             // add regulated idx only when the first side node is tap side node.
             // This is done to add only one directional edge with regulated idx.
-            Idx2D const regulated_idx = from_node == tap_side_node ? unregulated_idx : trafo3w_idx;
```
figueroa1395, line 116: Just for context, what does `trafo3w_idx` stand for? Transformer three winding index?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -113,11 +113,14 @@ inline void process_trafo3w_edge(main_core::main_model_state_c auto const& state
...
-            add_to_edge(state, edges, edge_props, tap_side_node, non_tap_side_node, {regulated_idx, 1});
+            if (from_node == tap_side_node) {
+                add_to_edge(state, edges, edge_props, tap_side_node, non_tap_side_node, {unregulated_idx, 0});
+            } else {
+                add_to_edge(state, edges, edge_props, tap_side_node, non_tap_side_node, {trafo3w_idx, 1});
```
figueroa1395, line 119: Why this change to if-else? Is it because unlike the `unregulated_idx`, the `trafo3w_idx = 1` for the ranking weight (didn't change)?

figueroa1395 (commented):
Looks good to me, but just to double check, some Python tests don't need to be adjusted (besides the validation part) in the same way as the Cpp one?

Note: VS-Code was stroking on me when I tried to review from there, so duplicated comments may randomly appear at some point. If I catch them first I will delete, if someone else does, please delete or resolve.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
nitbharambe: Yes

mgovers (commented):
LGTM. not yet approving due to remaining open comment, but otherwise OK

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: Addressed. For readability, a variable is used.

mgovers (approved):

PR #699, opened 2024-08-21
Feature/columnar dataset python extension

mgovers:
Adds the actual support for columnar dataset types in the Python API

Follow-up to #663

* [x] TODO's related to columnar data to be resolved


src/power_grid_model/core/buffer_handling.py
```diff
@@ -64,7 +91,23 @@ def _get_raw_data_view(data: np.ndarray, schema: ComponentMetaData) -> VoidPtr:
...
+    Get a raw view on the data.
+
+    Args:
+        data: the data.
+        schema: the schema the raw buffer should obey.
+
+    Returns:
+        a raw view on the data set.
+    """
+    return _get_raw_data_view(data, dtype=schema.dtype[attribute])
```
nitbharambe, line 110:
```suggestion
    return _get_raw_data_view(data, dtype=schema.dtype[attribute].base)
```
This caused asym tests to fail.

nitbharambe:
When deserializing using an all columnar filter, the deserializer does not know what attributes are present in the given binary buffer. Hence buffers for all attributes are generated when `...` data_filter is used.

tests/unit/utils.py
```diff
@@ -226,6 +226,9 @@ def save_json_data(json_file: str, data: Dataset):
 
 def compare_result(actual: SingleDataset, expected: SingleDataset, rtol: float, atol: float | dict[str, float]):
     for key, expected_data in expected.items():
+        if not isinstance(expected_data, np.ndarray):
+            raise NotImplementedError()  # TODO(mgovers): Add support for columnar data
+
```
mgovers, line 231: shouldn't we fix this too?

tests/unit/test_serialization.py
```diff
@@ -401,12 +471,22 @@ def assert_single_dataset_entries(
         # Complete component data checks
         if is_columnar_filter(data_filter, component):
             assert isinstance(deserialized_output, dict)
+            # Assert number of batches are equal for all attributes
             assert all(len(v) == len(serialized_input) for v in deserialized_output.values())
+            # Assert number of attributes are not more than input for all batches
+            # Discrepencies in equal / less than input captured in individual check
+            # assert all(len(v) >= len(deserialized_output) for v in serialized_input)
```
mgovers, line 478: forgotten comment?

```suggestion
            # Assert number of attributes are not more than input for all batches
            # Discrepencies in equal / less than input captured in individual check
            assert all(len(v) >= len(deserialized_output) for v in serialized_input)
```

tests/unit/test_serialization.py
mgovers, line 1: this file has a lot of reformatting without other changes. was that intended?

tests/unit/test_internal_utils.py
```diff
@@ -533,3 +540,29 @@ def test_copy_output_to_columnar_dataset(output_component_types, expected):
...
+
+
+@pytest.mark.parametrize(
+    ("data",),
+    [
+        pytest.param(np.empty(shape=3), id="row based single"),
+        pytest.param({"u": np.empty(shape=3)}, id="columnar single"),
+    ],
+)
+def test_get_batch_size__non_batch(data):
```
mgovers, line 566: can you rename this to something more descriptive? i was confused when i first read it. maybe

```suggestion
def test_get_batch_size__single_dataset_is_not_supported(data):
```

src/power_grid_model/validation/utils.py
```diff
@@ -103,7 +103,22 @@ def update_input_data(input_data: SingleDataset, update_data: SingleDataset):
...
+    component: ComponentTypeLike, input_data: SingleComponentData, update_data: SingleComponentData
+) -> None:
+    """
+    Update the data in a single component data set, with another single component data set,
+    indexed on the "id" field and only non-NaN values are overwritten.
+    """
+    if isinstance(input_data, np.ndarray) and isinstance(update_data, np.ndarray):
+        return _update_component_array_data(component=component, input_data=input_data, update_data=update_data)
+
+    raise NotImplementedError()  # TODO(mgovers): add support for columnar data
```
mgovers, line 116: @figueroa1395 is this still necessary?

src/power_grid_model/validation/rules.py
```diff
@@ -715,6 +723,9 @@ def all_finite(data: SingleDataset, exceptions: Optional[dict[ComponentType, lis
     """
     errors = []
     for component, array in data.items():
+        if not isinstance(array, np.ndarray):
+            raise NotImplementedError()  # TODO(mgovers): add support for columnar data
```
mgovers, line 727: @figueroa1395 is this still necessary?

src/power_grid_model/validation/rules.py
```diff
@@ -373,6 +373,9 @@ def none_match_comparison(  # pylint: disable=too-many-arguments
     if default_value_2 is not None:
         set_default_value(data=data, component=component, field=field, default_value=default_value_2)
     component_data = data[component]
+    if not isinstance(component_data, np.ndarray):
+        raise NotImplementedError()  # TODO(mgovers): add support for columnar data
```
mgovers, line 377: @figueroa1395 is this still necessary?

src/power_grid_model/validation/rules.py
mgovers, line 1: i think most of the validation changes can be reverted after you `update with base branch`

src/power_grid_model/utils.py
```diff
@@ -52,20 +59,29 @@ def get_dataset_scenario(dataset: BatchDataset, scenario: int) -> SingleDataset:
         The dataset for a specific scenario
     """
 
+    def _get_dense_scenario(arr: np.ndarray) -> np.ndarray:
+        return arr[scenario]
+
+    def _get_sparse_scenario(arr: np.ndarray, indptr: IndexPointer) -> np.ndarray:
+        return arr[indptr[scenario] : indptr[scenario + 1]]
+
```
mgovers, line 67:
```suggestion


    def _get_sparse_scenario(arr: np.ndarray, indptr: IndexPointer) -> np.ndarray:
        return arr[indptr[scenario] : indptr[scenario + 1]]


```

src/power_grid_model/utils.py
```diff
@@ -52,20 +59,29 @@ def get_dataset_scenario(dataset: BatchDataset, scenario: int) -> SingleDataset:
...
+            return _get_dense_scenario(component_scenarios)
 
-        indptr = component_scenarios["indptr"]
-        data = component_scenarios["data"]
-        if isinstance(indptr, np.ndarray) and isinstance(data, np.ndarray):
-            return data[indptr[scenario] : indptr[scenario + 1]]
+        if "indptr" in component_scenarios:
+            indptr = component_scenarios["indptr"]
+            data = component_scenarios["data"]
+            if isinstance(indptr, np.ndarray):
```
mgovers, line 75: this should raise a `ValueError`; the `else` case is also not supported

src/power_grid_model/data_types.py
```diff
@@ -92,39 +92,50 @@
     - The first element and last element will therefore be 0 and the size of the data, respectively.
 """
 
-SparseBatchArray = dict[SparseDataComponentType, IndexPointer | SingleArray]
-"""
-A sparse batch array is a dictionary containing the keys `indptr` and `data`.
 
-- data: a :class:`SingleArray`. The exact dtype depends on the type of component.
-- indptr: an :class:`IndexPointer` representing the start and end indices for each batch scenario.
+class SparseBatchArray(TypedDict):
```
nitbharambe, line 96: Special attention here. Maybe doing this is not okay

src/power_grid_model/utils.py
mgovers, line 75: this stuff has happened soooooo often now, it's probably a good idea to create a utility function. this would make a lot of our code a lot simpler

```py
def _extract_indptr_data(data: SparseBatchData) -> tuple[np.ndarray, SingleArray | SingleColumnarData]:
    indptr = data["indptr"]
    sub_data = data["data"]

    if isinstance(indptr, np.ndarray):
        return indptr, data

    raise ValueError("some error message")
```

if you want to, you can even go as far as to make it a template thingy using `TypeVar` so that you can even deduce the return type of `data`, depending on whether it's a `SparseBatchData`, a `SparseBatchArray` or a `SparseBatchColumnarData`

src/power_grid_model/data_types.py
mgovers, line 96: 😮 didn't know this existed. good find!

mgovers (commented):
preliminary part of everything except the `src/power_grid_model/core/` and `src/power_grid_model/_utils`

tests/unit/utils.py
nitbharambe:
```suggestion
        if not isinstance(expected_data, np.ndarray):
            raise NotImplementedError("Validation is not implemented for columnar data")

```
Its not necessary to implement this. Validation tests would always convert to row based data.

src/power_grid_model/_utils.py
```diff
@@ -439,11 +482,45 @@ def validate_data_filter(
...
 
 
 def is_columnar(component_data: ComponentData) -> bool:
     """Check if component_data is columnar or row based"""
     if is_sparse(component_data):
         return not isinstance(component_data["data"], np.ndarray)
     return not isinstance(component_data, np.ndarray)
+
+
+def validate_component_data(component_data: ComponentData, component=None) -> None:
```
mgovers, line 498: "validate" is a little bit ambiguous cause we kind of reserve that for the stuff in the data validator. it's probably a good idea to think about this properly

src/power_grid_model/_utils.py
```diff
@@ -439,11 +482,45 @@ def validate_data_filter(
...
+    """Checks if component_data is of ComponentData and raises ValueError if its not"""
+    component_name = f"'{component}'" if component is not None else ""
+    err_msg = f"Invalid data for {component_name} component. " "{0}"
+    err_msg_suffixed = err_msg + "It should be a 1D/2D Numpy structured array or a dictionary of such."
+
+    if is_sparse(component_data):
+        indptr = component_data["indptr"]
+        if not isinstance(indptr, np.ndarray):
+            raise TypeError(err_msg_suffixed.format(f"Invalid indptr type {type(indptr).__name__}. "))
+        sub_data = component_data["data"]
```
mgovers, line 508: idem to my comment on the `utils` about making a function that does this for you

mgovers (commented):
part 2 of the review done. will do another review after lunch on the subset of commits after you took over

src/power_grid_model/validation/utils.py
figueroa1395: It shouldn't be, as all the validation is done in row-based format. As long as we can slice columnar batch data into single scenarios of columnar, this is not needed; and this is only need for the improvement in the batch validation.

src/power_grid_model/validation/rules.py
figueroa1395: For the same reason, all validation is done in row format, it shouldn't be either.

However, before merging, we can test by re-basing with main (which already has the PR adding the support for columnar validation), and I can briefly make a branch on top of this to do the validation upgrade, and then we are sure we don't miss anything. @mgovers @nitbharambe what do you think?

Jerry-Jinfeng-Guo (commented):
Minor comments

tests/unit/test_serialization.py
nitbharambe: This is the test statement where the new 2 variations of ellipsis can be tested. 
Ill mark it as TODO.

src/power_grid_model/core/buffer_handling.py
```diff
@@ -37,23 +37,53 @@ class BufferProperties:
...
+
+    Args:
+        data: the data.
+        dtype: the dtype the raw buffer should obey.
+
+    Returns:
+        a raw view on the data set.
+    """
+    x = np.ascontiguousarray(data, dtype=dtype)
+    assert data is x
```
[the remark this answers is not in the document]
nitbharambe: Whoops. Missed it. I shall remove it.

It was how one of the issues was detected btw. 
Earlier we had `np.ascontiguousarray(data, dtype=dtype_of_data["attribute"])`
For asymmetric data type, we needed the base type, ie. float int etc without the shape:
Hence now: `np.ascontiguousarray(data, dtype=dtype_of_data["attribute"].base)`

src/power_grid_model/_utils.py
```diff
@@ -125,64 +129,83 @@ def get_and_verify_batch_sizes(batch_data: BatchDataset) -> int:
...
         return [data]
-    if data.ndim == 2:
-        return [data[i, :] for i in range(data.shape[0])]
-    raise TypeError(
-        f"Invalid data dimension {data.ndim} in batch data for '{component}' "
-        "(should be a 1D/2D Numpy structured array)."
-    )
+    if data.ndim in [2, 3]:
+        return [data[i, ...] for i in range(data.shape[0])]
+    raise ValueError("Dimension of the component data is invalid.")
```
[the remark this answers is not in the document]
nitbharambe: Its possible dimension of the numpy array actually. Really different for each case. We need to document it:

Dimensions for single, batch dataset:
Dense Row: 1, 2
Sparse Row: indptr: 1, 1 and  data: 1, 1
Dense Columnar: 1 for sym or 2 for asym, 2 for sym or 3 for asym
Sparse Columnar: indptr: 1, 1 and data: 1 for sym or 2 for asym, 2 for sym or 3 for asym

tests/unit/test_serialization.py
nitbharambe: @mgovers This file's changes are all formatting changes that you did. And one additional test case of simple_asym. 

I think the changes came from the fact that the dictionaries now have an extra comma at the end. (Hence black makes it in newline for each

src/power_grid_model/_utils.py
```diff
@@ -439,11 +482,45 @@ def validate_data_filter(
...
+            raise TypeError(err_msg_suffixed.format(""))
+        for attribute, attribute_array in sub_data.items():
+            if not isinstance(attribute_array, np.ndarray):
+                raise TypeError(err_msg_suffixed.format(f"'{attribute}' attribute. "))
+            if attribute_array.ndim not in [1, 2, 3]:
+                raise TypeError(err_msg_suffixed.format(f"Invalid dimension: {attribute_array.ndim }"))
+    elif not isinstance(sub_data, np.ndarray):
+        raise TypeError(err_msg_suffixed.format(f"Invalid data type {type(sub_data).__name__} "))
+    elif isinstance(sub_data, np.ndarray) and sub_data.ndim not in [1, 2]:
+        raise TypeError(err_msg_suffixed.format(f"Invalid dimension: {sub_data.ndim}. "))
```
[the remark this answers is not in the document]
nitbharambe: It is indeed ugly. We really need to get rid of all these error checks at every function.

src/power_grid_model/data_types.py
nitbharambe: Yeah on the retrospect we could have avoided the TypeVar thing.
We colud have had:

class ComponentType(total=False):
    node: SingleArray
    line: SingleArray
    ...

The totality makes these members optional.

mgovers: but that would replace dictionary datasets entirely, right?

tests/unit/test_serialization.py
mgovers: ohhhh right, no, i think it's the other way around: jerry said `black` needed to be updated so I did and that probably caused this

Jerry-Jinfeng-Guo: Hmm, interesting, I only knew about pylint

mgovers: oh then forget what i said 😬

Jerry-Jinfeng-Guo: _This message has been deleted._

src/power_grid_model/data_types.py
nitbharambe: Oh, right. 
```
class Dataset(total=False):
    node: ComponentData
    line: ComponentData
    ...
```
I see now that we can't improve anything with this with the current implementation.

src/power_grid_model/_utils.py
nitbharambe: Renamed to component_data_checks

src/power_grid_model/utils.py
nitbharambe: Should we do this in this PR? 
The final goal is to add more extraction functions where the checks happen. Then we can remove the big component_data_checks and all places where those checks are manually done. We had to keep them because of mypy. (@Jerry-Jinfeng-Guo ref to your other comment)

I have now added _extract_indptr_data, _extract_columnar_data, _extract_row_based_data as an example. They arent used anywhere as of yet.

src/power_grid_model/_utils.py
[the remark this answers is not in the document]
nitbharambe: Added this note to the documentation ticket

src/power_grid_model/_utils.py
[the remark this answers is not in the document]
nitbharambe: Added functions which shall make it nicer. Shall be used in the next PR.

src/power_grid_model/_utils.py
nitbharambe: Added functions which shall make it nicer. Shall be used in the next PR.

src/power_grid_model/validation/rules.py
mgovers: let's leave it as is and remove it as a follow-up. it's fine. your original PR should already ensure you don't end up here, but leaving it at least makes it clear what is hit and when.

src/power_grid_model/_utils.py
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/issues/749

src/power_grid_model/_utils.py
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/issues/749

src/power_grid_model/validation/rules.py
mgovers: idem: out of scope of this PR; to be removed in follow-up

src/power_grid_model/utils.py
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/issues/749

src/power_grid_model/validation/utils.py
mgovers: idem: to be removed in follow-up PR

src/power_grid_model/core/buffer_handling.py
```diff
@@ -127,121 +211,197 @@ def _get_sparse_buffer_properties(data: SparseBatchArray) -> BufferProperties:
...
-        raise ValueError(f"Data array can only be 1D. {VALIDATOR_MSG}")
+    if not isinstance(indptr, np.ndarray):
+        raise TypeError("indptr must be of type IndexPointer [np.ndarray]")
+
+    ndim = 1
+    columns: list[AttributeType] | None = None
+    if isinstance(contents, np.ndarray):
+        shape: tuple[int, ...] = contents.shape
+    elif not contents:
+        raise ValueError("Empty columnar buffer is ambiguous. {VALIDATOR_MSG}")
```
mgovers, line 227: forgotten
```suggestion
        raise ValueError(f"Empty columnar buffer is ambiguous. {VALIDATOR_MSG}")
```

please also go over all other instances

src/power_grid_model/core/buffer_handling.py
```diff
@@ -127,121 +211,197 @@ def _get_sparse_buffer_properties(data: SparseBatchArray) -> BufferProperties:
...
-    is_sparse = True
+    is_sparse_property = True
 
     contents = data["data"]
     indptr = data["indptr"]
 
-    if contents.ndim != 1:
-        raise ValueError(f"Data array can only be 1D. {VALIDATOR_MSG}")
+    if not isinstance(indptr, np.ndarray):
+        raise TypeError("indptr must be of type IndexPointer [np.ndarray]")
```
mgovers, line 220:
```suggestion
        raise TypeError(f"indptr must be of type IndexPointer [np.ndarray]. {VALIDATOR_MSG}")
```

src/power_grid_model/validation/utils.py
Jerry-Jinfeng-Guo: Can you guys agree on this point and resolve the thread (and below)? @mgovers @figueroa1395

src/power_grid_model/data_types.py
Jerry-Jinfeng-Guo: Should a Todo be added here for that matter? I am not sure what to make of this comment

mgovers (commented):
great cleanup. a couple remarks left, you can close the rest

src/power_grid_model/validation/utils.py
figueroa1395: I'm double checking at the moment. I'll get back in a few.

src/power_grid_model/data_types.py
nitbharambe: Dont think there are other places for it. Lets move forward with this for now then.

src/power_grid_model/validation/utils.py
figueroa1395: All this are okay to close, double checked all.

Jerry-Jinfeng-Guo (dismissed):

figueroa1395 (approved):

mgovers:
@nitbharambe CI is failing. do you know why?

also, why is there a commit from this morning called
[revert to original functionality](https://github.com/PowerGridModel/power-grid-model/pull/699/commits/36128989ee13a46ba9d064e1237b79a126420fbf)?
* [ ] Is this functionality not ready then?
* [ ] If so, can you explain why and convert to draft?

nitbharambe:
> @nitbharambe CI is failing. do you know why?
> 
> also, why is there a commit from this morning called
> [revert to original functionality](https://github.com/PowerGridModel/power-grid-model/pull/699/commits/36128989ee13a46ba9d064e1237b79a126420fbf)?
> * [ ] Is this functionality not ready then?
> * [ ] If so, can you explain why and convert to draft?

The single batch is really tricky. Before this it was changing how the row based situation gets handled. Hence lets tackle that in the separate PR. We can only support some part. Ill have this part including the tests in a separate PR.

figueroa1395 (dismissed):

figueroa1395 (approved):

figueroa1395 (dismissed):

figueroa1395 (approved):

Issue #702, opened 2024-08-24
[BUG] Access violation due to un-connected nodes visited when producing solver output result after power flow with tap position optimization


Jerry-Jinfeng-Guo:
**Description**

Tap optimization visits isolated nodes in the grid when producing solver results after power flow run, causing access violation. 

**Input Data Validity**

Isolated nodes are nodes when connection edges (link, line and transformer) have one or both `from_status` and `to_status` having value `0`. This input is supported, therefore validation won't throw any exception. The logic inside the tap optimizer should be updated to handle this more appropriately.

**To Reproduce**

This can be reproduced by running power flow calculation with tap optimization enabled on a grid with isolated nodes.

**Expected behavior**

Exception access violation thrown.

PR #703, opened 2024-08-26
BugFix / Handle access violation to unconnected-regulated transformers

Jerry-Jinfeng-Guo:
This PR fixes https://github.com/PowerGridModel/power-grid-model/issues/702


Jerry-Jinfeng-Guo:
NB, clang on Windows is broken atm.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers, line 1: i know you're probably already working on this but just leaving this comment so I don't forget myself:
please provide a test case

mgovers (commented):
i think the changes make sense to me. let's hope this fixes the issue

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -967,6 +978,9 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
 
         regulator.transformer.apply([&](transformer_c auto const& transformer) { // NOSONAR
             using TransformerType = std::remove_cvref_t<decltype(transformer)>;
+            if (!is_regulated_transformer_connected<TransformerType>(regulator, state)) {
```
nitbharambe, line 981: Maybe we shouldremove the unconnected regulated transformer from the ranking itself. Hence, this connected by topology check instead of the status checks. This would alter the ranking of transformers but for the better.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: Test added.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: More digging gives the surprising reality check that the `topo_comp_coup` member of `state` remains `empty` during trafo ranking and is only updated after/during iterations. This renders the impossibility to retrieve necessary information in the transformer ranking stage. Therefore, unless we do more fundamental changes, the current way might be the only reasonable solution given the present state of code.

mgovers: the entire topology step is done only during the first calculation. we can change this, of course, with significant changes (my guess is that these are what you mean)

however, it should be possible to determine which nodes/regulators are not energized because they obtain infinite path length to the nearest source during transformer ranking (with the only overhead compared to skipping them entirely being the creation)

mgovers (commented):
i'm investigating what causes clang-cl to fail

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: while something we need to keep in our minds, it is out of scope of this PR

mgovers (dismissed):
Approved but i have one very small suggestion remaining

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -269,6 +269,9 @@ inline auto get_edge_weights(TransformerGraph const& graph) -> TrafoGraphEdgePro
             continue;
         }
         auto edge_res = std::min(vertex_distances[boost::source(e, graph)], vertex_distances[boost::target(e, graph)]);
+        if (edge_res == infty) {
+            continue;
+        }
         result.push_back({graph[e].regulated_idx, edge_res});
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Addressed in the latest commit.

mgovers (approved):

Jerry-Jinfeng-Guo:
This PR also fixes the validation of issue https://github.com/PowerGridModel/power-grid-model/issues/662

PR #708, opened 2024-08-30
Feature/columnar data serialization

mgovers:
Step 5 & 6 of Implementation #548


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -739,49 +744,67 @@ class Deserializer {
...
+            ValueVisitor<T> visitor{
+                {}, [&]() -> T& {
+                    if (buffer.buffer->data != nullptr) {
+                        return attribute.get_attribute<T>(component.advance_ptr(buffer.buffer->data, buffer.idx));
+                    }
+                    if (auto it =
+                            std::ranges::find_if(buffer.buffer->attributes,
+                                                 [attribute_ptr = &attribute](AttributeBuffer<void> const& attr_buf) {
+                                                     return attr_buf.meta_attribute == attribute_ptr;
+                                                 });
```
mgovers, line 803: this is a potentially expensive search in the hot loop. Unfortunately, I have tried many different things but haven't been able to come up with a satisfactory solution except for making 2 completely separate control flows (one for row-based, one for columnar).

@TonyXiang8787 what do you think?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -495,17 +518,61 @@ class Serializer {
...
+                          MetaAttribute const& attribute) {
+        if (element_buffer.buffer->data != nullptr) {
+            RawElementPtr element_ptr = component.advance_ptr(element_buffer.buffer->data, element_buffer.idx);
+            return ctype_func_selector(attribute.ctype, [element_ptr, &attribute]<class T> {
+                return is_nan(attribute.get_attribute<T const>(element_ptr));
+            });
+        } else if (auto it = std::ranges::find_if(element_buffer.buffer->attributes,
+                                                  [&attribute](auto const& attribute_buffer) {
+                                                      return attribute_buffer.meta_attribute == &attribute;
+                                                  });
```
mgovers, line 531: this is a potentially expensive search in the hot loop. Unfortunately, I have tried many different things but haven't been able to come up with a satisfactory solution except for making 2 completely separate control flows (one for row-based, one for columnar).

@TonyXiang8787 what do you think?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -495,17 +518,61 @@ class Serializer {
...
+        if (element_buffer.buffer->data != nullptr) {
+            RawElementPtr element_ptr = component.advance_ptr(element_buffer.buffer->data, element_buffer.idx);
+            ctype_func_selector(attribute.ctype, [this, element_ptr, &attribute]<class T> {
+                packer_.pack(attribute.get_attribute<T const>(element_ptr));
+            });
+        } else if (auto it = std::ranges::find_if(element_buffer.buffer->attributes,
+                                                  [&attribute](auto const& attribute_buffer) {
+                                                      return attribute_buffer.meta_attribute == &attribute;
+                                                  });
+                   it != element_buffer.buffer->attributes.end()) {
```
mgovers, line 561: this is a potentially expensive search in the hot loop. Unfortunately, I have tried many different things but haven't been able to come up with a satisfactory solution except for making 2 completely separate control flows (one for row-based, one for columnar).

@TonyXiang8787 what do you think?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: # Let me start with some observations:

1. For a given element, if the element is a `dict` with verbose attribute names, the name look-up per attribute per element is not avoidable, regardless of if we have columnar or row-based format. We just need to make sure the look-up does not happen twice.
2. For a given element, if the element is a `list` with pre-defined attribute names in the header, your current implementation is very inefficient for columnar format as we still do the name look-up per attribute per element, which is not needed for row-based format.

# Now my proposal:

## Columnar buffer order

At the beginning of the `parse` function, we need to define another function called `map_columnar_attributes` or something, it should per component check if the provided buffer is columnar. If yes, it maps in one go the sequence number of the provided columnar buffers regards to the sequence of the pre-defined attributes, with `-1` meaning the attribute buffer is not provided. 

For example, for the following attributes list

```json
{
  "version": "1.0",
  "type": "sym_output",
  "is_batch": false,
  "attributes": {
       "node": ["id", "u_pu", "u_angle"]
   },
  "data": {}
}
```

If the user provide columnar buffers in the order `["u_angle", "u", "u_pu"]`, the `map_columnar_attributes` should return `{"node": [-1, 2, 1]}`.

## Parse map element

In the `parse_map_element`, if the buffer is columnar, you look-up the attribute names directly from the attribute list of provided columnar buffer. If the buffer is row-based, you look-up the attribute names from the metadata as currently.

## Parse array element

In the `parse_array_element`, if the bufer is columnar, you refer the sequence from `map_columnar_attributes` to retrieve directly the relevant buffer position without to do name look-up again.

## Parse attribute

You need two versions of `parse_attribute`, for row-based and columnar.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
TonyXiang8787: We can use similar strategy as in de-serializer. At the beginning we pre-map all the columnar buffers to the sequence number of all the attributes of a component. Then we do not do the name look-up afterwards.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -364,6 +373,17 @@ class Deserializer {
     using DataByteMeta = std::vector<std::vector<ComponentByteMeta>>;
     using AttributeByteMeta = std::vector<std::pair<std::string_view, std::vector<std::string_view>>>;
 
+    struct BufferView {
+        WritableDataset::Buffer const* buffer{nullptr};
+        Idx idx{0};
+        std::unordered_map<MetaAttribute const*, AttributeBuffer<void> const*> const* attribute_order{nullptr};
```
TonyXiang8787, line 379: I do not think we need a `map` or `unordered_map`. A `vector` should be enough. It's a index lookup.

mgovers: i initially had done it with `std::vector` as well. however, using `std::unordered_map` means it can be used to speed up map-like as well as array-like representations.

Using the `std::vector` version would only speed up the array-like representation, while the map-like representation remains linear in complexity.

Since, on average, look-up is `O(1)` when using `std::unordered_map`, i don't see any disadvantages with using `std::unordered_map` compared to `std::vector`. In addition, the creation of the look-up table is much simpler this way.

TonyXiang8787: What do you mean by map-like elements? We are talking about the condense format right?

The non-condense format with named attribute, you need a name lookup anyway. That's a separate routine with its own data structure.

In general, `vector` is much faster than `std::unordered_map`. The `O(1)` does not take the constant into account.

mgovers: i've rewritten it to use vector. it does make the logic a lot more convoluted, though. the i've had to add a lot of asserts to even convince myself that what i need is indeed valid in that location. fortunately, asserts are free.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -364,6 +367,17 @@ class Deserializer {
     using DataByteMeta = std::vector<std::vector<ComponentByteMeta>>;
     using AttributeByteMeta = std::vector<std::pair<std::string_view, std::vector<std::string_view>>>;
 
+    struct BufferView {
+        WritableDataset::Buffer const* buffer{nullptr};
+        Idx idx{0};
+        std::vector<AttributeBuffer<void> const*> const* attribute_order{nullptr};
```
TonyXiang8787, line 373:
```suggestion
        std::vector<AttributeBuffer<void>> attribute_order{};
```

`AttributeBuffer` is quite small, store in value makes more sense rather than all pointer stuff.
We also do not need an additional indirection of `attribute_order`. Just store it by value. If the component is not columnar, it is just an empty `vector`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -709,27 +725,29 @@ class Deserializer {
             }
             return found->second;
         }();
+        auto const attribute_buffer_order = get_attribute_buffer_mapping(*buffer.buffer, attributes);
+        buffer.attribute_order = &attribute_buffer_order;
```
TonyXiang8787, line 729:
```suggestion
        buffer.attribute_order = get_attribute_buffer_mapping(*buffer.buffer, attributes);
```

As suggested above, assign directly by value.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -739,49 +757,102 @@ class Deserializer {
...
+            parse_attribute(buffer, component, *attributes[attribute_number_]);
         }
         attribute_number_ = -1;
     }
 
-    void parse_attribute(void* element_pointer, MetaAttribute const& attribute) {
+    void parse_attribute(BufferView const& buffer, MetaComponent const& component, MetaAttribute const& attribute) {
+        assert(buffer.buffer != nullptr);
+
+        if (buffer.buffer->data != nullptr) {
```
TonyXiang8787, line 804: This is very tight loop to do duplicated checking on columnar data. We need to abstract this in the `parse_component` level. You can make `parse_scenario`, `parse_map_element` and `parse_array_element` by having a template argument to solve it compile time.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: then the `BufferView` approach is not possible anymore. We currently copy for every individual component and only update the `idx`. Copying the entire attribute buffer makes this quite expensive if there are e.g. 20 columns.

In fact, this was my first implementation: directly copying the `Buffer`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -739,49 +757,102 @@ class Deserializer {
...
+        } else if (auto it = std::ranges::find_if(buffer.buffer->attributes,
+                                                  [&attribute](auto const& attribute_buffer) {
+                                                      return attribute_buffer.meta_attribute == &attribute;
+                                                  });
+                   it != buffer.buffer->attributes.end()) {
+            parse_attribute(*it, buffer.idx);
+        } else {
+            parse_skip();
+        }
+    }
```
TonyXiang8787, line 848: I feel this logic as not necessary in this level. In this level, we should directly have a columnar buffer and an `MetaAttribute` reference. Nothing else.

In the `parse_map_element` and `parse_array_element`, the logic should already divide to resolve different path.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: If we're going to copy `AttributeBuffer`s, then this is no longer an expensive step. copying the buffers will be more expensive

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: 20 columns is still  `20 * sizeof(AttributeBuffer) = 640` bytes, which is relatively negligible to the data volume, which is MBs to GBs.

mgovers: but the copy is made for every single component, not only per component type. so if there are 10k components, then there are ~10k copies. It's part of the hotloop

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: Copying `AttributeBuffer`s is like copying 640 bytes of data even with 20 columns. This tight loop check is about 1k scenario X 1k element X 20 attributes = 20 million checks per component.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: Am I misunderstanding here? You only do this at the beginning of `parse_component`, which is valid for all scenarios and element within that component type. So it only happens once per component type for the entire dataset.

mgovers: this is also done for each scenario in the current implementation

mgovers: I think i found a workaround. Let me dig into it first

mgovers: right, sorry, i was confused cause i'm working on 2 different things simultaneously. it is however true for the serializer.

again. I'll probably need to re-start this PR from scratch

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -739,49 +757,102 @@ class Deserializer {
...
         offset_ = 0;
     }
 
-    void parse_map_element(void* element_pointer, Idx map_size, MetaComponent const& component) {
+    void parse_map_element(BufferView const& buffer, Idx map_size, MetaComponent const& component) {
         while (map_size-- != 0) {
             attribute_key_ = parse_string();
-            Idx const found_idx = component.find_attribute(attribute_key_);
-            if (found_idx < 0) {
+            Idx const component_attribute_idx = component.find_attribute(attribute_key_);
```
TonyXiang8787, line 776: The logic here needs to be split, even with a template argument.

For row-based buffer, you try to find the `MetaAttribute` from our internal meta data.

For column-base buffer, you try to find the `MetaAttribute` from the user provided list of columnar buffers.

mgovers: really, there's too much specialization going on here. i'm going to re-implement this entire PR with the knowledge I have now and hope I can write it in such a way that it is more efficient. However, I'm not particularly fond of having to add 8 overloads because for every different combination of `{serializer, deserializer} x {row_based, columnar} x {map, array}` the implementation, caching, buffer mapping and even terminology is completely different and it's quite confusing even for someone like me who has worked with the (de)serializer before. It will be even worse for someone who hasn't.

I'm going to re-implement and I'm first going to spend time on homogenizing the caching mechanism

TonyXiang8787: Yes, I agree to re-think about the PR. But still, efficiency/performance is the key here for (de-)serialization, which is more important than the "beauty` of the code. The (de-)serialization is the most complicated and hard to read part anyway (in the sense of C++ features and abstraction). That's almost always the case. We need to accept that fact sometimes...

mgovers: i think i've processed all your comments on the deserializer. if you have time, can you do another check?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: i think i've processed all your comments on the deserializer. if you have time, can you do another check?

TonyXiang8787: The new implementation looks good. I hereby resolve all de-serializer comments.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -816,6 +904,28 @@ class Deserializer {
...
+                ctype_func_selector(attribute_buffer.meta_attribute->ctype, [&attribute_buffer, &info]<typename T> {
+                    std::ranges::fill(std::span{reinterpret_cast<T*>(attribute_buffer.data),
+                                                narrow_cast<size_t>(info.total_elements)},
+                                      nan_value<T>);
+                });
+            }
+        }
+    }
+
+    static BufferView advance(BufferView buffer, Idx offset) {
```
TonyXiang8787, line 924: Now I see your point here about copying of the `std::vector<AttributeBuffer<void>>`. In the `advance` you made implicit copy of `BufferView` which copies the whole `reordered_attribute_buffers`. This `advance` is called for every scenario and every element, which indeed make it very expensive copy. Maybe we still need to change that....

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -364,6 +367,18 @@ class Deserializer {
     using DataByteMeta = std::vector<std::vector<ComponentByteMeta>>;
     using AttributeByteMeta = std::vector<std::pair<std::string_view, std::vector<std::string_view>>>;
 
+    using Buffer = WritableDataset::Buffer;
+    struct BufferView {
+        Buffer const* buffer{nullptr};
+        Idx idx{0};
+        std::vector<AttributeBuffer<void>> reordered_attribute_buffers;
```
TonyXiang8787, line 374:
```suggestion
        std::span<AttributeBuffer<void> const> reordered_attribute_buffers;
```

Maybe make them a span and store the owned copy in `parse_component` which outlives the whole parsing of the component.

mgovers: why? this is very closely related to the actual buffer. After the buffer is gone, there's no point in keeping this mapping. The only real reason I can think of is that the serializer (not the deserializer) does things scenario-by-scenario, and this would keep the similarity between the two. however, there are already so many differences between the serializer and deserializer that i'm not sure whether this will make it more readable.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -699,37 +733,44 @@ class Deserializer {
...
+                return it->second;
             }
-            return found->second;
+            return {};
         }();
+        BufferView const buffer_view{.buffer = &buffer,
+                                     .idx = 0,
+                                     .reordered_attribute_buffers =
+                                         detail::is_columnar_v<row_or_column_t>
+                                             ? detail::reordered_attribute_buffers(buffer, attributes)
```
TonyXiang8787, line 748:
```suggestion
        auto const reordered_attribute_buffers = detail::reordered_attribute_buffers(buffer, attributes);
        BufferView const buffer_view{.buffer = &buffer,
                                     .idx = 0,
                                     .reordered_attribute_buffers = reordered_attribute_buffers};
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -699,37 +733,44 @@ class Deserializer {
...
+                                             : std::vector<AttributeBuffer<void>>{}};
+
```
TonyXiang8787, line 750:
```suggestion
        auto const reordered_attribute_buffers = detail::reordered_attribute_buffers(buffer, attributes);
        BufferView const buffer_view{.buffer = &buffer,
                                     .idx = 0,
                                     .reordered_attribute_buffers = reordered_attribute_buffers};
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: See my other comment 

https://github.com/PowerGridModel/power-grid-model/pull/708#discussion_r1743568744

Previously I though this `vector` will not be copied around so the memory and CPU overhead is negligible. However, in your `advance` function you implicitly copy the `BufferView`. And `advance` is called for every single element and scenario.

mgovers: EDIT: apologies, i missed your other comment. Yes, this is a good solution for the case that there are very many scenarios with few components (e.g. N-1)

mgovers: i'd propose keeping it next to `attributes`, which does not outlive `parse_component` but has the same lifetime

TonyXiang8787: You also call `advance` in all elements. So the number of copy operations of `BufferView` is the same as the number of scenarios times number of elements per scenario in one component type.

TonyXiang8787: > i'd propose keeping it next to `attributes`, which does not outlive `parse_component` but has the same lifetime

That's should be a good proposal.

TonyXiang8787:
The deserialization part is looking good now.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
TonyXiang8787, line 1: @mgovers the serializer also looks good to me. But I see another thing, the private static function `json_convert_inf` and `json_inf_to_string` in the `Serializer` is not used anymore. We should delete them.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -330,8 +335,21 @@ class Serializer {
...
+                    if (!attribute.check_all_nan(
+                            buffer.component->advance_ptr(buffer.buffer.buffer->data, buffer.buffer.idx),
+                            buffer.size)) {
+                        attributes.push_back(&attribute);
+                    }
+                } else if (auto it = std::ranges::find_if(buffer.buffer.buffer->attributes,
+                                                          [&attribute](auto const& attribute_buffer) {
+                                                              return attribute_buffer.meta_attribute == &attribute;
+                                                          });
+                           it != buffer.buffer.buffer->attributes.end()) {
```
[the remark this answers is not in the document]
mgovers: i have renamed all occurrences of `BufferView buffer` to `BufferView buffer_view` and `ComponentBuffer buffer` to `ComponentBuffer component_buffer`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -330,8 +335,21 @@ class Serializer {
             for (auto const& attribute : buffer.component->attributes) {
                 // if not all the values of an attribute are nan
                 // add this attribute to the list
-                if (!attribute.check_all_nan(buffer.data, buffer.size)) {
-                    attributes.push_back(&attribute);
+                assert(buffer.buffer.buffer != nullptr);
+                if (buffer.buffer.buffer->data != nullptr) {
```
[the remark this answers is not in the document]
mgovers: added `static constexpr` `is_row_based` and `is_columnar` functions to make the if statements and assertions more clear

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -495,17 +518,61 @@ class Serializer {
...
+                        MetaAttribute const& attribute) {
+        if (element_buffer.buffer->data != nullptr) {
+            RawElementPtr element_ptr = component.advance_ptr(element_buffer.buffer->data, element_buffer.idx);
+            ctype_func_selector(attribute.ctype, [this, element_ptr, &attribute]<class T> {
+                packer_.pack(attribute.get_attribute<T const>(element_ptr));
+            });
+        } else if (auto it = std::ranges::find_if(element_buffer.buffer->attributes,
+                                                  [&attribute](auto const& attribute_buffer) {
+                                                      return attribute_buffer.meta_attribute == &attribute;
+                                                  });
```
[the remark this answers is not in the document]
mgovers: now it's only used once in the serializer and once in the deserializer 😬 Are you OK with closing the issue?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -816,6 +840,25 @@ class Deserializer {
...
+                    std::ranges::fill(std::span{reinterpret_cast<T*>(attribute_buffer.data) + buffer.idx,
+                                                narrow_cast<size_t>(info.total_elements)},
+                                      nan_value<T>);
+                });
+            }
+        }
+    }
+
+    static BufferView advance(BufferView const& buffer, Idx offset) {
+        return {.buffer = buffer.buffer, .idx = buffer.idx + offset};
```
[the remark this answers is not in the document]
mgovers: yes indeed. this is extremely little data to copy. made the function `constexpr` though

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -330,8 +335,21 @@ class Serializer {
             for (auto const& attribute : buffer.component->attributes) {
                 // if not all the values of an attribute are nan
                 // add this attribute to the list
-                if (!attribute.check_all_nan(buffer.data, buffer.size)) {
-                    attributes.push_back(&attribute);
+                assert(buffer.buffer.buffer != nullptr);
+                if (buffer.buffer.buffer->data != nullptr) {
+                    if (!attribute.check_all_nan(
+                            buffer.component->advance_ptr(buffer.buffer.buffer->data, buffer.buffer.idx),
+                            buffer.size)) {
```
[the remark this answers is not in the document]
mgovers: because we introduced `buffer.idx`

TonyXiang8787 (approved):

TonyXiang8787:
@mgovers PR now in the merge queue. After merge, maybe you can run a serialization benchmark (using your old dataset) to compare the old and new version of row-based (de-)serialization performance. It should not have significant change.

mgovers:
bad news: the parsing step in the deserializer takes excessively longer. will 

update dataset shape `(800, 800)` for all component types with random values for all attributes. Notice that `total` for deserialization is increadibly long but that that is mainly related to the python string stuff and maybe something we need to have a look at

## v1.9.35

| msgpack [compact] | 0.0005251000402495265 s | 2.2746466000098735 s | 2.8601861000061035 s | 0.33473320002667606 s | 55.92428519995883 s | 56.386925800004974 s |
| msgpack [regular] | 0.0005136000690981746 s | 2.6831328000407666 s | 3.5524667999707162 s | 0.5777812999440357 s | 99.07666269992478 s | 99.91340609989129 s |
| json [compact] | 0.0007071000291034579 s | 29.271804499905556 s | 60.09442940005101 s | 12.122917199973017 s | 56.96022929996252 s | 70.54345009999815 s |
| json [regular] | 0.0004412999842315912 s | 36.51309969997965 s | 76.11392769997474 s | 15.74001260008663 s | 100.83267979999073 s | 119.05708439997397 s |


## v1.9.34

| data | serializer creation | serialization | serialization total | deserializer creation + pre-parse | deserializer parse | deserializer total |
| -- | -- | -- | -- | -- | -- |-- |
| msgpack [compact] | 0.000408099964261055 s | 2.672189600067213 s | 3.146100400015712 s | 0.2728536000940949 s | 2.623856300022453 s | 3.0367689999984577 s |
| msgpack [regular] | 0.000543000060133636 s | 2.9165597000392154 s | 3.8790119000477716 s | 0.5125961999874562 s | 5.055858200066723 s | 5.831057699979283 s |
| json [compact] | 0.0004981999518349767 s | 29.134515200043097 s | 60.77616540004965 s | 12.264068600023165 s | 2.503174800076522 s | 16.27681280008983 s |
| json [regular] | 0.0005403000395745039 s | 35.86926089995541 s | 75.49933109991252 s | 15.36376619990915 s | 4.866490800050087 s | 22.552572900080122 s |

## file sizes

| data | size |
| --- | --- |
| msgpack [compact] | 356'456 KB |
| msgpack [regular] | 744'580 KB |
| json [compact] | 902'956 KB |
| json [regular] | 1'431'704 KB |

I will follow-up on this

TonyXiang8787:
@mgovers  please for the table of 1.9.35

mgovers:
benchmark script

```py
# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
#

from pathlib import Path
from timeit import timeit
from typing import Callable
import numpy as np

from power_grid_model import initialize_array
from power_grid_model.core.dataset_definitions import ComponentType, DatasetType
from power_grid_model.utils import json_deserialize_from_file, msgpack_deserialize_from_file, msgpack_serialize_to_file, json_serialize_to_file


# def generate(shape):
#     data = {}
#     for comp in ComponentType:
#         arr = initialize_array(DatasetType.update, comp, shape=shape, empty=True)
#         for attribute in arr.dtype.names:
#             arr[attribute] = np.random.rand(*arr[attribute].shape)
#         data[comp] = arr

#     print("done generating")
#     return data


def serialize(func: Callable, file_path: Path, data, **kwargs):
    def execute():
        print(f"serializing: {file_path}")
        func(file_path, data, **kwargs)
        print(f"done serializing: {file_path}")

    print(f"timed {func.__name__}({file_path}): {timeit(execute, number=1)} s")


def deserialize(func: Callable, file_path: Path):
    def execute():
        print(f"deserializing: {file_path}")
        func(file_path)
        print(f"done deserializing: {file_path}")
    
    print(f"timed {func.__name__}({file_path}): {timeit(execute, number=1)} s")


if __name__ == "__main__":
    shape = (800, 800)
    # update_data = generate(shape)
    update_data = msgpack_deserialize_from_file(Path("tmp/data/base_data.msgpack"))

    serialize(msgpack_serialize_to_file, Path("tmp/data/all_random_compact.msgpack"), update_data, use_compact_list=True)
    deserialize(msgpack_deserialize_from_file, Path("tmp/data/all_random_compact.msgpack"))
    serialize(msgpack_serialize_to_file, Path("tmp/data/all_random.msgpack"), update_data)
    deserialize(msgpack_deserialize_from_file, Path("tmp/data/all_random.msgpack"))
    serialize(json_serialize_to_file, Path("tmp/data/all_random_compact.json"), update_data, use_compact_list=True)
    deserialize(json_deserialize_from_file, Path("tmp/data/all_random_compact.json"))
    serialize(json_serialize_to_file, Path("tmp/data/all_random.json"), update_data)
    deserialize(json_deserialize_from_file, Path("tmp/data/all_random.json"))
```

test script
```cpp
// SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
//

#include <power_grid_model/auxiliary/input.hpp>
#include <power_grid_model/auxiliary/update.hpp>

#include <fstream>
#include <sstream>

namespace {
} // namespace

int main() {
    std::string str = [] {
        std::ifstream f{"<some_file>.json"};
        std::stringstream sstr;
        sstr << f.rdbuf();
        return sstr.str();
    }();

    auto deserializer = Deserializer{from_json, str, meta_data_gen::meta_data};
    auto dataset = deserializer.get_dataset_info();
    auto info = dataset.get_description();
    std::vector<std::vector<char>> buffers{};
    for (Idx idx = 0; idx < meta_data_gen::meta_data.get_dataset("update").n_components(); ++idx) {
        buffers.push_back(
            std::vector<char>(info.component_info[idx].total_elements * info.component_info[idx].component->size));
        dataset.set_buffer(info.component_info[idx].component->name, nullptr, buffers[idx].data());
    }
    deserializer.parse();
    return 0;
}
```

TonyXiang8787:
@mgovers  there is a major overhead (>90%) in the new de-serializer. We need to dump the data into a file and write a cpp executable to run the data. Then we can profile the overhead.

mgovers:
i found out what's wrong. attempting to fix now

mgovers:
same test script but using msgpack:

```cpp
// SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
//

#include <power_grid_model/auxiliary/input.hpp>
#include <power_grid_model/auxiliary/update.hpp>

#include <fstream>
#include <sstream>

namespace {
} // namespace

int main() {
    std::vector<char> result = [] {
        constexpr auto file_path = "<msgpack_file>";
        std::vector<char> result(std::filesystem::file_size(file_path));
        std::ifstream f{file_path, std::ios::binary};
        f.read(result.data(), result.size());
        return result;
    }();

    auto deserializer = Deserializer{from_msgpack, result, meta_data_gen::meta_data};
    auto dataset = deserializer.get_dataset_info();
    auto info = dataset.get_description();
    std::vector<std::vector<char>> buffers{};
    for (Idx idx = 0; idx < meta_data_gen::meta_data.get_dataset("update").n_components(); ++idx) {
        buffers.push_back(
            std::vector<char>(info.component_info[idx].total_elements * info.component_info[idx].component->size));
        dataset.set_buffer(info.component_info[idx].component->name, nullptr, buffers[idx].data());
    }
    deserializer.parse();
    return 0;
}
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -685,11 +701,31 @@ class Deserializer {
     }
 
     void parse_component(Idx component_idx) {
+        if (dataset_handler_.is_row_based(component_idx)) {
+            parse_component(row_based, component_idx);
+        } else {
+            parse_component(columnar, component_idx);
+        }
```
mgovers, line 708: I found out that this call will cause major overhead when no attributes are present. That means that the logic concerning empty attribute buffers is incorrect

```cpp
        } else if (!dataset_handler_.get_buffer(component_idx).attributes.empty()) {
            parse_component(columnar, component_idx);
        }
```

This is not a sustainable fix (it points at the fact that there is some other problem) but will resolve the issue in the short-term

| data | serializer creation | serialization | serialization total | deserializer creation + pre-parse | deserializer parse | deserializer total |
| -- | -- | -- | -- | -- | -- |-- |
msgpack [compact] | 0.0004923000233247876 s | 2.3669713999843225 s | 3.0552195999771357 s | 0.28873490006662905 s | 2.858401000034064 s | 3.285264000063762 s
msgpack [regular] | 0.0007754000835120678 s | 2.9432286999654025 s | 3.9057689999463037 s | 0.5173835000023246 s | 5.383461399935186 s | 6.166236900025979 s
json [compact] | 0.00043050001841038465 s | 29.926429499988444 s | 63.13568380009383 s | 14.445208700024523 s | 3.2074557000305504 s | 19.59107560000848 s
json [regular] | 0.0008066999725997448 s | 37.490393099957146 s | 78.83358660002705 s | 18.796793300076388 s | 5.285955099971034 s | 26.85202120000031 s

TonyXiang8787: What has changed in your suggestion to avoid the overhead?

TonyXiang8787: Your test scripts are testing row-based de-serialization right? So the `else` or `else if` in the segment is never reached.

mgovers: turns out it is... there's definitely an issue. I have very little time left before i need to go, though. Should we roll-back?

mgovers: ok found the problem in my test script (`auto` where there should've been `auto&` when declaring the dataset)

TonyXiang8787: I see. So, the buffers are never set in the de-serializer. And all the components are treated column-based  with no attributes.

PR #709, opened 2024-09-02
Update quickstart.md

PaulSchulz:
Need to import numpy for code to run.

I needed to add this to get the example code to run. I am not completely familiar with using numpy.
(Other examples have this included.)

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -165,5 +165,5 @@ update_data = {
     "sym_load": sym_load_update
 }
 
-model = PowerGridModel(input_data=input_data, update_data=update_data)
+model.calculate_power_flow(update_data=update_data)
 ```
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -125,6 +125,8 @@ Both input and output data are dictionaries of structured `numpy` arrays.
 We can use `pandas` to convert them to data frames and print them.
 
 ```python
+import numpy as np
+
 print('Node Input')
 print(pd.DataFrame(input_data['node']))
 print('Node Result')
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -125,6 +125,8 @@ Both input and output data are dictionaries of structured `numpy` arrays.
 We can use `pandas` to convert them to data frames and print them.
 
 ```python
+import numpy as np
```
mgovers, line 128: Instead of `numpy`, it is `pandas` that is required here.

```suggestion
import pandas as pd
```

mgovers (commented):
Hi @PaulSchulz , Thank you for your contribution! You're 100% correct that the Quick Start guide is not a working example. Looking at the rest of the Quick Start guide, there are more things that need to be improved/fixed as well. I will add them to this PR, as I think it is the correct location to do so.

docs/quickstart.md
```diff
@@ -164,4 +166,4 @@ update_data = {
 }
 
 model = PowerGridModel(input_data=input_data, update_data=update_data)
```
mgovers, line 168:
```suggestion
model.calculate_power_flow(update_data=update_data)
```

mgovers (commented):
EDIT: i will not edit everything myself because it is on your `main` branch. I will just propose the most important changes that enable accepting this PR instead

mgovers (approved):
I hereby approve this proposal.

Note that the Quick Start guide is still not in the best possible state cfr. the following observations:

* [ ] `self_test` is imported twice
* [ ] the Batch data example script will result in an error, because:
  * [ ] the IDs of the update data are not set
  * [ ] the input data contains only 1 `sym_load`, while the update data has 4 `sym_load`s per batch scenario. This will result in errors, even if the IDs are set. Instead, either the input data needs to be updated, or the update data needs to have a number of components per scenario equal to 1.

This is, however, out of scope of this PR. I will create a new issue for this.

Again, thank you for your contribution - we value that a lot! If you ever feel like contributing in the future, please do so. We're happy to support where needed.

PR #711, opened 2024-09-03
Feature / Columnar data C API

Jerry-Jinfeng-Guo:
Implements C API changes in https://github.com/PowerGridModel/power-grid-model/issues/548
* [x] interface
* [x] add to c++ api wraper
* [x] ~~c~~ api tests
  * [x] api model
  * [x] serialization 
* [x] fix ... things that inevitably need fixing


power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787, line 1: You also need the function `PGM_dataset_writable_set_attribute_buffer` for the deserializer.

Jerry-Jinfeng-Guo: Will continue once blocker https://github.com/PowerGridModel/power-grid-model/pull/712 is merged

Jerry-Jinfeng-Guo: @TonyXiang8787 There is a disparity between the names of `set_attribute_buffer` (API) and `add_attribute_buffer` in the dataset implementation. Could you check the name of the function?

TonyXiang8787: This is on-purpose. For writable dataset we use `set_buffer`, for mutable and const dataset we use `add_buffer.

figueroa1395: Is it intended for `PGM_dataset_writable_set_buffer` of the C API to be able to only set a buffer if it has already been added to the dataset? i.e. only when such a buffer has been created and added in the `WritableDataset` by the `Deserializer` itself (before parsing of course)? Because this is current behavior. Or should  `PGM_dataset_writable_set_buffer` be able to add the buffer and then set it if it is not present in the `WritableDataset` at this point? I believe the second should be supported as well.

This is relevant because `PGM_dataset_writable_set_attribute_buffer ` should behave the same way as `PGM_dataset_writable_set_buffer`.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -302,20 +302,16 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+    void add_attribute_buffer(std::string_view component, std::string_view attribute, Data* data)
+        requires(!is_indptr_mutable_v<dataset_type>)
+    {
+        add_attribute_buffer_impl(component, attribute, data);
+    }
+
+    void set_attribute_buffer(std::string_view component, std::string_view attribute, Data* data)
+        requires is_indptr_mutable_v<dataset_type>
+    {
+        add_attribute_buffer_impl(component, attribute, data);
```
mgovers, line 314: @Jerry-Jinfeng-Guo after offline discussion with Tony, we decided to go with the same behavior between `add_attribute_buffer` and `set_attribute_buffer` (but different entrypoints). The behavior of `set_attribute_buffer` therefore differs from the one of `set_buffer`. The reasoning is as follows:

* For components:
  * the deserializer tells the user via the dataset info that a certain component is present in the serialized data.
  * It is possible to efficiently determine whether that is the case.
  * The user can then only call `set_buffer` for those components that are already present
* For attributes:
  * the deserializer would need to go over the entire dataset to look for components with the map serialization representation to determine whether an attribute is present.
  * this is expensive
  * the deserializer therefore cannot let the user know beforehand which attributes are present
  * `set_attribute_buffer` therefore should only be called if it has not been set yet.
  * this is the same behavior as `add_attribute_buffer`

tests/c_api_tests/test_cpp_api_model.cpp
```diff
@@ -180,7 +194,7 @@ TEST_CASE("C++ API Model") {
     }
 
     SUBCASE("Copy model") {
-        Model model_copy{model};
+        Model const& model_copy{model};
```
mgovers, line 197: this results in a reference, not a copy.
```suggestion
        Model model_copy{model};
```

If `clang-tidy` complains, this is where you can use `NOLINT(<identifier>)`

mgovers (commented):
great cleanup on the clang-tidy

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
Jerry-Jinfeng-Guo: In that case, shouldn't there be something in place to make sure 
> ... therefore should only be called if it has not been set yet.
is actually guarded?

mgovers: it is in https://github.com/PowerGridModel/power-grid-model/pull/711/files#diff-2cc40425b5c3604ed2e2401defd34d90a6fb7050406a81139fc1be14bbfda661R447-R451

```cpp
        if (std::ranges::find_if(buffer.attributes, [&attribute](auto const& buffer_attribute) {
                return buffer_attribute.meta_attribute->name == attribute;
            }) != buffer.attributes.end()) {
            throw DatasetError{"Cannot have duplicated attribute buffers!\n"};
        }
```

figueroa1395: `set_attribute_buffer` should then be carefully documented, as it behaves different to `set_buffer`, but it make sense due to the expensiveness of searching for specific attributes.

Jerry-Jinfeng-Guo: Considering renaming it to `set_attribute_buffer_but_not_really`

figueroa1395: If the renaming to `set_attribute_buffer_but_not_really` doesn't follow through (or even if it does), please add a comment briefly explaining why this doesn't really set anything. I can see one of us very confused in a couple of months about this. A link to this comments would suffice.

mgovers: from the users' perspective, a dataset has all attributes declared for any given component. this keeps the symmetry between row-based and columnar. in a way, we could also rename `add_attribute_buffer` to `set_attribute_buffer`, but then you encounter the exact same naming problem that triggered this discussion in the first place

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp
```diff
@@ -13,7 +13,7 @@
 namespace power_grid_model_cpp {
 class PowerGridError : public std::exception {
   public:
-    PowerGridError(const std::string& message) : message_(message) {}
+    PowerGridError(std::string message) : message_(std::move(message)) {}
```
figueroa1395, line 16: Nice job in catching all these details here and in `meta_data.cpp`.

tests/c_api_tests/test_cpp_api_model.cpp
```diff
@@ -83,34 +93,49 @@ TEST_CASE("C++ API Model") {
...
     load_buffer.set_value(PGM_def_input_sym_load_q_specified, &load_q_specified, -1);
 
-    // add buffers
-    input_dataset.add_buffer("node", 1, 1, nullptr, node_buffer);
+    // add buffers - row
     input_dataset.add_buffer("sym_load", 1, 1, nullptr, load_buffer);
     input_dataset.add_buffer("source", 1, 1, nullptr, source_buffer);
 
+    // add buffers - columnar
+    input_dataset.add_buffer("node", 2, 2, nullptr, nullptr);
```
figueroa1395, line 109: After seeing this, it makes me wonder if `add_buffer` should have an overload specific for columnar like
```
void add_buffer(std::string const& component, Idx elements_per_scenario, Idx total_elements, Idx const* indptr) const {
        add_buffer(*this, component, elements_per_scenario, total_elements, indptr, nullptr);
    }
```
Later it can be removed as this API is not stable, but it may make the tests cleaner and easier to write for the user. Just food for thought.

tests/c_api_tests/test_cpp_api_model.cpp
```diff
@@ -134,14 +159,22 @@ TEST_CASE("C++ API Model") {
...
     load_updates_buffer.set_value(PGM_def_update_sym_load_q_specified, load_updates_q_specified.data(), 0, -1);
     load_updates_buffer.set_value(PGM_def_update_sym_load_q_specified, load_updates_q_specified.data(), 1, -1);
-
     // dataset
-    DatasetConst single_update_dataset{"update", 0, 1};
+    DatasetConst const single_update_dataset{"update", 0, 1};
     single_update_dataset.add_buffer("source", 1, 1, nullptr, source_update_buffer);
     single_update_dataset.add_buffer("sym_load", 1, 1, nullptr, load_updates_buffer.get());
-    DatasetConst batch_update_dataset{"update", 1, 2};
+    // update containing columnar data still fail
```
figueroa1395, line 166: I hope this line is to be removed?

tests/c_api_tests/test_cpp_api_serialization.cpp
```diff
@@ -54,12 +54,12 @@ TEST_CASE("C++ API Serialization and Deserialization") {
     std::vector<Idx> const total_elements = {1, 2};
 
     SUBCASE("Serializer") {
-        DatasetConst dataset{"input", is_batch, batch_size};
+        DatasetConst const dataset{"input", is_batch, batch_size};
```
figueroa1395, line 57: Thanks again! Same for all below.

tests/c_api_tests/test_cpp_api_serialization.cpp
```diff
@@ -130,7 +130,7 @@ TEST_CASE("C++ API Serialization and Deserialization") {
         auto check_deserializer = [&](Deserializer& deserializer) {
             // get dataset
             auto& dataset = deserializer.get_dataset();
-            auto& info = dataset.get_info();
+            auto const& info = dataset.get_info();
```
figueroa1395, line 133: Commenting on this line for reference. Shouldn't we test that we can `add_attribute_buffer` to `writable_dataset` when generated by the `deserializer`? Or is it enough with the unit tests that cover the (de)serializer? I think it is a good idea that we add a specific test for this here since it was introduced both in the C and Cpp APIs.

figueroa1395 (commented):
Looks good to me overall, just minor comments and the possible addition of tests for `add_attribute_buffer` for deserialized data, may be needed.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -302,20 +302,16 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
     }
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Good point. Martijn reverted the `set` function from my commit. I will add a docstring.

tests/c_api_tests/test_cpp_api_model.cpp
Jerry-Jinfeng-Guo: Indeed.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -147,6 +147,18 @@ PGM_API void PGM_dataset_const_add_buffer(PGM_Handle* handle, PGM_ConstDataset*
...
+/**
+ * @brief Add a attribute buffer to an instance of PGM_ConstDataset/component.
+ * @param handle
+ * @param dataset The pointer to the PGM_ConstDataset.
+ * @param component The name of the component.
+ * @param attribute The name of the attribute.
+ * @param data A void pointer to the buffer data.
+ * @return
+ */
+PGM_API void PGM_dataset_const_add_attribute_buffer(PGM_Handle* handle, PGM_ConstDataset* dataset,
```
mgovers, line 159: Please also extend the documentation on the `*_(add|set)_buffer` (particularly that nullptr is now supported)

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -720,9 +720,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
             return true;
         }
 
-        auto const is_component_update_independent = [&update_data]<typename CT>() -> bool {
-            // get span of all the update data
-            auto const all_spans = update_data.get_buffer_span_all_scenarios<meta_data::update_getter_s, CT>();
+        auto const process_span = []<typename CT>(auto const& all_spans) -> bool {
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: I think `process_span` works better here for its sole purpose in the lambda `is_component_update_independent`. Perhaps `process_buffer_span`, which could be both better and redundant at the same time.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
Jerry-Jinfeng-Guo: Apart from the docstring that needs addition, I consider this thread resolvable. Could someone involved do it? ;)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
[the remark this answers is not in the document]
figueroa1395: As pointed out by Martjin (something we both forgot), remeber to update documentation regarding `set/add_attribute_buffer` along with the comment here.

tests/c_api_tests/test_cpp_api_model.cpp
Jerry-Jinfeng-Guo: Having an overload is a valid approach, but here in this case I think the argument behind is that the user could simply provide a buffer for row based and a `nullptr` for columnar ones. The benefit of this one can be that the columnar input is explicit assumed and that debugging might be easier if not seeing corresponding `add_attr_buffer` calls later on. The `nullptr` here explicitly acts as a token for it.

figueroa1395: Just something to think about. But it can be brought up again when migrating the tests to the API, if needed. Feel free to resolve.

tests/c_api_tests/test_cpp_api_model.cpp
Jerry-Jinfeng-Guo: Addressed in the latest commit. Comment resolved.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Addressed in the latest commit. Comment resolved.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Addressed in the latest commit. Comment resolved.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
Jerry-Jinfeng-Guo: Addressed in the latest commit. Comment resolved.

tests/c_api_tests/test_cpp_api_model.cpp
Jerry-Jinfeng-Guo: I understand. Since we still have time before we publish this, what do you think? @TonyXiang8787

tests/c_api_tests/test_cpp_api_serialization.cpp
TonyXiang8787: You cannot test `set_attribute_buffer` separately. There is no way to go to the opaque struct to see if the attribute buffer is actually set.

You can only test this function together with de-serializer to see if the columnar data are parsed and written into the buffer.

tests/c_api_tests/test_cpp_api_model.cpp
TonyXiang8787: The CPP wrapper is not stable anyway. I propose to leave it out of the scope of this PR, which is the C-API adjustment.

@figueroa1395 if you think this is helpful for the test, feel free to add this function in your PR.

TonyXiang8787 (changes requested):
@Jerry-Jinfeng-Guo please resolve the merge conflict and we should be good to go

tests/c_api_tests/test_cpp_api_serialization.cpp
figueroa1395: That is what I meant, hence why I commented on the serialization test file. Isn't this still needed? Because else the C and Cpp API call of `set_attribute_buffer` isn't being tested.

TonyXiang8787: Ah, now I get it. 

@Jerry-Jinfeng-Guo, indeed we need to create a columnar version of test de-serializer in this test case, to test the `set_attribute_buffer` C-API function.

tests/c_api_tests/test_cpp_api_serialization.cpp
TonyXiang8787: I actually see something weird here. The `add_buffer`/`add_attribute_buffer` functions in the CPP wrapper of const and mutable dataset are `const` functions. This is for me very confusing.

We are literally mutating the layout of the dataset, why are they const functions? In my opinion those functions should be declared as non-const, and this const specifier should be gone.

figueroa1395: They can be (and are) const-qualified because the wrapper object `Dataset*` itself does not mutate, since it only holds smart pointers to the core objects that contain the actual dataset (which does mutate). However, following the logic and intention, probably they should not be `const` functions. Nevertheless, since the `const` qualifier here is used by the compiler for optimizations, I wonder if we should still keep them. @Jerry-Jinfeng-Guo any thoughts on this?

Jerry-Jinfeng-Guo: I was indeed simply following what was implemented in the core. I see this as extra encapsulation, not that I am in favor. If needed be, I could also apply the changes (which can be numerous).

tests/c_api_tests/test_cpp_api_model.cpp
Jerry-Jinfeng-Guo: Issue already fixed. Comment resolved.

tests/c_api_tests/test_cpp_api_serialization.cpp
TonyXiang8787: I do think we should remove the `const` function for this. `const` should be also expressing intentions. Like for `std::vector`, which also only stores a pointer to the actual array buffer. What if they allow this? It is theoretically possible but extremely confusing.

```c++
std::vector<int> const data{1, 2, 3, 4, 5};
data[2] = 100;
```

So yes, my proposal is to remove the const function and adjust all the tests. Instead of doing other way around.

tests/native_api_tests/test_api_model.cpp
```diff
@@ -83,34 +93,49 @@ TEST_CASE("API Model") {
...
-    single_output_dataset.add_buffer("node", 1, 1, nullptr, node_output);
-    Buffer node_batch_output{PGM_def_sym_output_node, 2};
+    DatasetMutable const single_output_dataset{"sym_output", 0, 1};
+    single_output_dataset.add_buffer("node", 2, 2, nullptr, node_output);
+    Buffer node_batch_output{PGM_def_sym_output_node, 4};
     node_batch_output.set_nan();
-    DatasetMutable batch_output_dataset{"sym_output", 1, 2};
-    batch_output_dataset.add_buffer("node", 1, 2, nullptr, node_batch_output);
+    DatasetMutable const batch_output_dataset{"sym_output", 1, 2};
+    batch_output_dataset.add_buffer("node", 2, 4, nullptr, node_batch_output);
```
Jerry-Jinfeng-Guo, line 127: @TonyXiang8787 in the core, the downstream function calls after `add_buffer` for output do not verify whether the shapes match between the output buffer and the actual result. For example, if we run batch size 4 with 10 nodes provided in columnar format, we would be expecting a 4x10 or a 1x40 holder for the result. However, in reality, we could also provide e.g., 1x10 to it and only the first 10 results will be filled to the holder. Is this behavior intentional?

tests/c_api_tests/test_cpp_api_serialization.cpp
Jerry-Jinfeng-Guo: > `const` should be also expressing intentions

Strongly for this. Changes incoming.

tests/native_api_tests/test_api_model.cpp
TonyXiang8787: I guess we can add extra checks in the main model and throw errors if the size does not match.

However, this does not change the fact that users can still provide crazy C-API input and we cannot do anything about it. The user can create a buffer of size `10` and tell the C-API dataset that the size is `100`. We do not have any way to verify that.

As with any C-API, we can only document the expected input requirements, and if user does not specify, it is UB.

Jerry-Jinfeng-Guo: Created an issue: https://github.com/PowerGridModel/power-grid-model/issues/722

tests/c_api_tests/test_cpp_api_serialization.cpp
Jerry-Jinfeng-Guo: Added in the newest commit.

TonyXiang8787:
Hmmm. buffer overflow...

Jerry-Jinfeng-Guo:
> Hmmm. buffer overflow...

Fixed... `|complete json data| < |json data|`

TonyXiang8787 (approved):

PR #714, opened 2024-09-06
don't parse_skip() for very many components

mgovers:
Fixes very slow row-based deserialization cfr. https://github.com/PowerGridModel/power-grid-model/pull/708#discussion_r1746985416

**NOTE:** columnar deserialization is still potentially slow due to very many `parse_skip` calls when a certain column is not present

Issue was introduced in #708 

A couple issues are not yet explained and also need further investigation but are out of scope of the immediate regression mitigation:

* why does the python deserializer end up in this edge case? does it call `parse` multiple times on different components?
* how can we improve the efficiency of the deserializer when the columnar dataset contains few columns?

PR #716, opened 2024-09-09
more efficient columnar deserialization

mgovers:
Makes deserialization more efficient compared to #708 as follows:

* for each component type in the data set
  * if there are no row-based buffers or attribute buffers set to a `WritableDataset` component buffer:
    * we skip the component entirely
    * because it is neither `row_based(*)` nor `columnar(*, with_attribute_buffers=true)` and therefore there's nothing to do
  * otherwise, for each scenario
    * for each element of the current component type in that scenario
      * if a row-based buffer is set
        * parse the element as before
      * otherwise, peek whether the current element in the serialized data
        * if it is map-like
          * enter the element and parse as before
        * otherwise, it is array-like.
          * if none of the provided attribute buffers are present in the header,
            * skip this element
          * otherwise, enter the element and parse as before
            * there is no more efficient way to do this, because it's no more efficient way to skip, e.g., 3 attributes in the msgpack data

This provides a sustainable solution compared to the one proposed in #714 

NOTE: the check `if none of the provided attribute buffers are present in the header` is done when reordering the attribute buffers and is reduced to a simple check whether it is empty. this is the best we can do because there's no more efficient way to read only a subset of the msgpack array while skipping all the rest

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/common.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/common.hpp
@@ -46,6 +46,11 @@ reordered_attribute_buffers(BufferType& buffer, std::span<MetaAttribute const* c
             }
             return AttributeBuffer<void>{};
         });
+
+    if (std::ranges::all_of(result, [](auto const& attribute_buffer) { return attribute_buffer.data == nullptr; })) {
+        result = {};
+    }
+
     return result;
 }
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
@@ -703,7 +703,7 @@ class Deserializer {
     void parse_component(Idx component_idx) {
         if (dataset_handler_.is_row_based(component_idx)) {
             parse_component(row_based, component_idx);
-        } else {
+        } else if (dataset_handler_.is_columnar(component_idx, true)) {
             parse_component(columnar, component_idx);
         }
     }
@@ -713,7 +713,7 @@ class Deserializer {
         auto const& buffer = dataset_handler_.get_buffer(component_idx);
 
         assert(dataset_handler_.is_row_based(buffer) == detail::is_row_based_v<row_or_column_t>);
-        assert(dataset_handler_.is_columnar(buffer) == detail::is_columnar_v<row_or_column_t>);
+        assert(dataset_handler_.is_columnar(buffer, true) == detail::is_columnar_v<row_or_column_t>);
         assert(is_row_based(buffer) == detail::is_row_based_v<row_or_column_t>);
         assert(is_columnar(buffer) == detail::is_columnar_v<row_or_column_t>);
 
@@ -744,9 +744,9 @@ class Deserializer {
             }
             return {};
         }();
-        auto const reordered_attribute_buffers = detail::is_columnar_v<row_or_column_t>
-                                                     ? detail::reordered_attribute_buffers(buffer, attributes)
-                                                     : std::vector<AttributeBuffer<void>>{};
+        auto reordered_attribute_buffers = detail::is_columnar_v<row_or_column_t>
+                                               ? detail::reordered_attribute_buffers(buffer, attributes)
+                                               : std::vector<AttributeBuffer<void>>{};
 
         BufferView const buffer_view{
             .buffer = &buffer, .idx = 0, .reordered_attribute_buffers = reordered_attribute_buffers};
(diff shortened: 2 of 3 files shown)


mgovers:
test script:
```cpp
// SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
//

#include <power_grid_model/auxiliary/input.hpp>
#include <power_grid_model/auxiliary/update.hpp>

#include <fstream>
#include <sstream>

namespace {
} // namespace

int main() {
    std::vector<char> serialized_data = [] {
        constexpr auto file_path = "<msgpack_data>";
        std::vector<char> result(std::filesystem::file_size(file_path));
        std::ifstream f{file_path, std::ios::binary};
        f.read(result.data(), result.size());
        return result;
    }();

    auto deserializer = Deserializer{from_msgpack, serialized_data, meta_data_gen::meta_data};
    auto& dataset = deserializer.get_dataset_info();
    auto const& info = dataset.get_description();

    std::vector<std::vector<std::byte>> row_buffers{};
    std::vector<std::vector<std::vector<std::byte>>> column_buffers{};

    auto const n_components = meta_data_gen::meta_data.get_dataset("update").n_components();
    for (Idx idx = 0; idx < n_components; ++idx) {
        auto const& meta_component = *info.component_info[idx].component;
        auto const buffer_size = info.component_info[idx].total_elements * meta_component.size;
        if (idx < n_components / 4) {
            row_buffers.push_back(std::vector<std::byte>(buffer_size));
            dataset.set_buffer(meta_component.name, nullptr, row_buffers.back().data());
        } else if (idx < n_components / 2) {
            auto& buffer = column_buffers.emplace_back();

            for (auto const& meta_attribute : meta_component.attributes) {
                buffer.push_back(std::vector<std::byte>(buffer_size));
                dataset.add_attribute_buffer(meta_component.name, meta_attribute.name, buffer.back().data());
            }
        }
    }
    deserializer.parse();
    return 0;
}
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -703,7 +703,7 @@ class Deserializer {
     void parse_component(Idx component_idx) {
         if (dataset_handler_.is_row_based(component_idx)) {
             parse_component(row_based, component_idx);
-        } else {
+        } else if (dataset_handler_.is_columnar(component_idx, true)) {
```
mgovers, line 706: this skips components without any buffers provided by the user

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/common.hpp
```diff
@@ -46,6 +46,11 @@ reordered_attribute_buffers(BufferType& buffer, std::span<MetaAttribute const* c
             }
             return AttributeBuffer<void>{};
         });
+
+    if (std::ranges::all_of(result, [](auto const& attribute_buffer) { return attribute_buffer.data == nullptr; })) {
+        result = {};
```
mgovers, line 51: this is the pre-calculate part that checks whether the intersection is empty between
* the set of all attributes listed in the serialized data header for array-like elements; and
* the set of all attribute buffers provided by the user

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -779,23 +779,48 @@ class Deserializer {
...
+
+        auto const element_visitor = parse_map_array<visit_map_array_t, stay_offset>();
+        if (element_visitor.is_map) {
+            parse_map_array<visit_map_array_t, move_forward>();
+            parse_map_element(tag, buffer_view, element_visitor.size, component);
+        } else if (!buffer_view.reordered_attribute_buffers.empty()) {
+            parse_map_array<visit_map_array_t, move_forward>();
+            parse_array_element(tag, buffer_view, element_visitor.size, component, attributes);
+        } else {
+            parse_skip();
```
mgovers, line 820: the peeking pattern enables to `skip` components that are specified using the array serialization schema, but do not have any attribute buffers specified that are deserialized

TonyXiang8787:
This is a nice improvement. But I still need to understand why `parse_skip` is so slow compared to `parse_bool` or `parse_attribute`. This needs attention before we just bypass the issue.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/common.hpp
TonyXiang8787: Please put an in-code documentation here to specify the intention.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: that's good improvement

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/common.hpp
mgovers: added docstring

TonyXiang8787 (approved):

mgovers:
> This is a nice improvement. But I still need to understand why `parse_skip` is so slow compared to `parse_bool` or `parse_attribute`. This needs attention before we just bypass the issue.

cfr. offline discussions + benchmarking, `parse_skip` is not slower than `parse_bool` or `parse_attribute`. Hence, this PR is ready for merge

PR #718, opened 2024-09-11
Feature/Re route validation tests

### LINKED ISSUE #486: [FEATURE] Optimize compilation speed by reorganize headers

figueroa1395:
Re-route the validation tests through the C API:
- [x] Move the call for the tests to the C++ wrapper of the C API.
- [x] Remove the PGM core dependencies.

~~Re-route the benchmarks through the C API:~~
~~- [ ] Move the call for the benchmark to the C++ wrapper of the C API.~~
~~- [ ] Remove the PGM core dependencies.~~

Note: The re-routing of the benchmark won't be addressed here anymore, since it requieres more effort than initially anticipated. For the moment, it is taken out of the CI build and addressed on the nightly build #736 

Closes #486


TonyXiang8787:
@figueroa1395 should the name of this PR be changed to only re-route validation tests?

figueroa1395:
> @figueroa1395 should the name of this PR be changed to only re-route validation tests?

PR edited accordingly.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -2,27 +2,70 @@
...
+
+#include "power_grid_model_cpp.hpp"
+
 #include <power_grid_model/auxiliary/dataset.hpp>
 #include <power_grid_model/auxiliary/meta_data_gen.hpp>
 #include <power_grid_model/auxiliary/serialization/deserializer.hpp>
 #include <power_grid_model/container.hpp>
 #include <power_grid_model/main_model.hpp>
 
+#include <Eigen/Dense>
```
TonyXiang8787, line 15: Please do not include eigen in this. We don't need `eigen` for validation test.

For the asymmetric attributes, just use `std::array<double,3>`;

tests/cpp_validation_tests/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ add_executable(power_grid_model_validation_tests ${PROJECT_SOURCES})
 target_link_libraries(power_grid_model_validation_tests
     PRIVATE
         power_grid_model
```
TonyXiang8787, line 14: Remove this one
Add a dependency for `nolhmann_json`

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -2,27 +2,70 @@
...
 
 namespace {
+// taken from common.hpp
+using IdxVector = std::vector<Idx>;
+
+// taken from three_phase_tensor.hpp
+template <class T> using Eigen3Vector = Eigen::Array<T, 3, 1>;
+
+// is nan. Taken from three_phase_tensor.hpp
+template <class Derived> inline bool is_nan(Eigen::ArrayBase<Derived> const& x) { return x.isNaN().all(); }
```
TonyXiang8787, line 42: Just use `std::array<double,3>`

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -2,27 +2,70 @@
...
+template <class Functor, class... Args> decltype(auto) pgm_type_func_selector(Idx type, Functor&& f, Args&&... args) {
+    switch (type) {
+    case Idx{0}: // PGM_int32
+        return std::forward<Functor>(f).template operator()<int32_t>(std::forward<Args>(args)...);
+    case Idx{1}: // PGM_int8
+        return std::forward<Functor>(f).template operator()<int8_t>(std::forward<Args>(args)...);
+    case Idx{2}: // PGM_double
+        return std::forward<Functor>(f).template operator()<double>(std::forward<Args>(args)...);
+    case Idx{3}: // PGM_double3
+        return std::forward<Functor>(f).template operator()<Eigen3Vector<double>>(std::forward<Args>(args)...);
```
TonyXiang8787, line 64: same here

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: Okay, will do.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -2,27 +2,70 @@
...
+inline bool is_nan(int8_t x) { return x == std::numeric_limits<int8_t>::min(); }
+template <class Enum>
+    requires std::same_as<std::underlying_type_t<Enum>, int8_t>
+inline bool is_nan(Enum x) {
+    return static_cast<int8_t>(x) == std::numeric_limits<int8_t>::min();
+}
+
+template <class Functor, class... Args> decltype(auto) pgm_type_func_selector(Idx type, Functor&& f, Args&&... args) {
+    switch (type) {
+    case Idx{0}: // PGM_int32
```
TonyXiang8787, line 57: why use `Idx{0}`? What's the problem to use `PGM_int32` directly?

tests/cpp_validation_tests/CMakeLists.txt
figueroa1395: I will after I have removed all the dependencies on the core. Leaving it for now. I will resolve this comment once it is fully removed.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: Because the attribute type as returned by `power_grid_model_cpp::MetaData::attribute_ctype` is a raw C `enum`.

figueroa1395:
Since I have started working on this PR again, this is very outdated at the moment. Many things have and will keep changing.

tests/cpp_validation_tests/test_validation.cpp
TonyXiang8787: That should not hurt right? A raw C enum is just an integer literal, you can use it directly.

figueroa1395:
At this stage all the single validation tests have been migrated to the Cpp API and pass. Next I will fix the batch validation.

TonyXiang8787:
@figueroa1395 you need to resolve merge conflict, otherwise the CI will not start.

figueroa1395:
@TonyXiang8787 Thanks, I did not know. I am doing a bit of refactoring and once I finish that, I will solve the conflict.

TonyXiang8787:
@figueroa1395  I would suggest you to first do a down-merge to resolve the merge conflict, then refactoring. After refactoring the conflict will be more difficult to resolve.

figueroa1395:
@TonyXiang8787 Merge conflict resolved and refactoring done. Just one thing remains: Previously, the batch validation tests were tested not only for a whole update (which is the current state), but also on a per-scenario basis, is this still needed or is it sufficient to validate the results after the whole update has been done?

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,236 +87,301 @@ auto read_json(std::filesystem::path const& path) {
...
+                        } else {
+                            ref_buffer.get_value(attribute_meta, &ref_attribute_value,
+                                                 (elements_per_scenario * scenario_idx) + obj, 0);
+                            buffer.get_value(attribute_meta, &attribute_value,
+                                             (elements_per_scenario * scenario_idx) + obj, 0);
+                            ref_buffer.get_value(possible_attr_meta, &ref_possible_attribute_value,
+                                                 (elements_per_scenario * scenario_idx) + obj, 0);
+                            buffer.get_value(possible_attr_meta, &possible_attribute_value,
+                                             (elements_per_scenario * scenario_idx) + obj, 0);
+                        }
```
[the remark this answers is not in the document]
figueroa1395: I cleaned it up in the last commit. It should be better now.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,236 +87,301 @@ auto read_json(std::filesystem::path const& path) {
...
     }
-    assert(angle_attr.ctype == CType::c_double3);
-    assert(mag_attr.ctype == CType::c_double3);
-    return check_angle_and_magnitude<asymmetric_t>(reference_result_ptr, result_ptr, angle_attr, mag_attr, atol, rtol,
-                                                   obj);
 }
 
 // assert single result
-void assert_result(ConstDataset const& result, ConstDataset const& reference_result,
+void assert_result(OwningDataset const& owning_result, OwningDataset const& owning_reference_result,
```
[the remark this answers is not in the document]
figueroa1395: I personally wouldn't, unfortunately. Even though it is humongoes, there are just many things that must be done in order to check the results, and at least this way is very linear, because it will be very complex nevertheless.

On the other hand, I could just make the lambdas into proper functions, which would reduce the length of this one by a lot and not necessarily increase complexity, but that comes at the cost of having huge and ugly signatures for the resulting functions. Is that something we want to do?

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -323,6 +432,7 @@ struct CaseParam {
     bool is_batch{};
     double rtol{};
     bool fail{};
+    // to remove batch parameter in the end, most likely
     [[no_unique_address]] BatchParameter batch_parameter{};
```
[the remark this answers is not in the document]
figueroa1395: I removed it altogether. It is not needed anymore indeed.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -531,21 +644,31 @@ void execute_test(CaseParam const& param, T&& func) {
...
+        // create and set options
+        power_grid_model_cpp::Options options{};
+        options.set_calculation_type(calculation_type_mapping.at(param.calculation_type));
+        options.set_calculation_method(calculation_method_mapping.at(param.calculation_method));
+        options.set_symmetric(param.sym ? 1 : 0);
+        options.set_err_tol(1e-8);
+        options.set_max_iter(20);
+        options.set_threading(-1);
+        options.set_short_circuit_voltage_scaling(sc_voltage_scaling_mapping.at(param.short_circuit_voltage_scaling));
+        options.set_tap_changing_strategy(optimizer_strategy_mapping.at(param.tap_changing_strategy));
```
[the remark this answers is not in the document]
figueroa1395: I added a separate `get_options` function as suggested. However, the magic numbers still remain (and were there originally) since the `CaseParam`does not contain them because they can not be extracted from the data files (they are just not present there).

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -456,16 +567,18 @@ void add_cases(std::filesystem::path const& case_dir, std::string const& calcula
...
-    auto const output_type = get_output_type(param.calculation_type, param.sym);
-
+ValidationCase create_validation_case(CaseParam const& param, std::string const& output_type) {
     // input
-    ValidationCase validation_case{.param = param, .input = load_dataset(param.case_dir / "input.json")};
+    ValidationCase validation_case{.param = param,
+                                   .input = load_dataset(param.case_dir / "input.json"),
+                                   .output = std::nullopt,
+                                   .update_batch = std::nullopt,
+                                   .output_batch = std::nullopt};
```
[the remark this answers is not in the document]
figueroa1395: GCC complains, it wants everything to be initialized even if it is optional, hence the `.* = std::nullopt`. So as much as I dislike this, I think it needs to stay here.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: This has been addressed now.

TonyXiang8787:
> @TonyXiang8787 Merge conflict resolved and refactoring done. Just one thing remains: Previously, the batch validation tests were tested not only for a whole update (which is the current state), but also on a per-scenario basis, is this still needed or is it sufficient to validate the results after the whole update has been done?

No, I think that's not needed.

tests/cpp_validation_tests/test_validation.cpp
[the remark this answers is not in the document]
figueroa1395: Cleaned it up

tests/cpp_validation_tests/CMakeLists.txt
```diff
@@ -11,7 +11,8 @@ add_executable(power_grid_model_validation_tests ${PROJECT_SOURCES})
 
 target_link_libraries(power_grid_model_validation_tests
     PRIVATE
-        power_grid_model
+        power_grid_model_c
```
mgovers, line 14: you can optionally delete this cause `power_grid_model_cpp` already exposes `power_grid_model_c` in the interface
```suggestion
```

mgovers (commented):
great progress. will review the rest on monday when you're done cleaning up

tests/cpp_validation_tests/CMakeLists.txt
figueroa1395: I removed it. No point in keeping the C API if we have the Cpp API.

tests/cpp_validation_tests/CMakeLists.txt
figueroa1395: All the dependencies have been removed. Resolving.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -2,27 +2,59 @@
...
+namespace power_grid_model_cpp {
 
 namespace {
+inline bool is_nan(std::floating_point auto x) { return std::isnan(x); }
+template <std::floating_point T> inline bool is_nan(std::complex<T> const& x) {
+    return is_nan(x.real()) || is_nan(x.imag());
+}
+inline bool is_nan(int32_t x) { return x == std::numeric_limits<int32_t>::min(); }
+inline bool is_nan(int8_t x) { return x == std::numeric_limits<int8_t>::min(); }
+template <typename T, std::size_t N> inline bool is_nan(std::array<T, N> const& array) {
```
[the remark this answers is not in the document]
TonyXiang8787: By the way, please limit the cpp wrapper to c++17 features. Using c++20 features will be too new.

TonyXiang8787:
> The only things remaining at this point are doing a full manual clang tidy run and figuring out what is going on with the MacOS build.

@figueroa1395 @mgovers 
After debugging in my Mac I have found the bug. The `std::aligned_alloc` has two requirements for unix-like system which we do not enforce:

* Alignment should be multiple of `sizeof(void*)`
* Size should be multiple of alignment

I have now pushed the fix.

figueroa1395:
@TonyXiang8787 Thanks! Great find and fix.

tests/cpp_validation_tests/test_validation.cpp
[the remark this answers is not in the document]
figueroa1395: After this PR is merged, I will then create a small separate PR to add `is_nan` and `pgm_type_func_selector` to the wrapper.

power_grid_model_c/power_grid_model_c/src/buffer.cpp
```diff
@@ -20,10 +20,15 @@ using meta_data::RawDataPtr;
...
+    size_t const alignment = std::max(component->alignment, sizeof(void*));
+    // total bytes should be multiple of alignment
+    size_t const requested_bytes = component->size * size;
+    size_t const rounded_bytes = ((requested_bytes + alignment - 1) / alignment) * alignment;
 #ifdef _WIN32
-    return _aligned_malloc(component->size * size, component->alignment);
+    return _aligned_malloc(rounded_bytes, alignment);
 #else
-    return std::aligned_alloc(component->alignment, component->size * size);
+    return std::aligned_alloc(alignment, rounded_bytes);
```
mgovers, line 31: isn't this an actual bugfix to the C API? should we do it in a separate PR?

tests/cpp_validation_tests/test_validation.cpp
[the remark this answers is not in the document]
mgovers: Manual run started in https://github.com/PowerGridModel/power-grid-model/actions/runs/11210360513/job/31157118018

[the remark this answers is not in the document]
mgovers: manual run failed. Please resolve the errors. NOTE: this might not be an exhaustive list, so you will need to run it again later. 
image

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,238 +86,263 @@ auto read_json(std::filesystem::path const& path) {
...
-        Idx* indptr_data = buffer.indptr.empty() ? nullptr : buffer.indptr.data();
-
-        info.set_buffer(component_info.component->name, indptr_data, buffer.ptr.get());
-        buffers.push_back(std::move(buffer));
+        auto const& component_name = info.component_name(component_idx);
+        auto const& component_meta = MetaData::get_component_by_name(dataset_name, component_name);
+        Idx const component_elements_per_scenario = info.component_elements_per_scenario(component_idx);
+        Idx const component_size = info.component_total_elements(component_idx);
+
+        auto& current_intptr = owning_dataset.storage.indptrs.emplace_back(
```
mgovers, line 114: indptr for index pointer, not for integer pointer 😬 

```suggestion
        auto& current_indptr = owning_dataset.storage.indptrs.emplace_back(
```

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,238 +86,263 @@ auto read_json(std::filesystem::path const& path) {
...
+            info.component_elements_per_scenario(component_idx) < 0 ? batch_size + 1 : 0);
+        bool const empty_indptr = current_intptr.empty();
```
mgovers, line 116: feels excessive. you can call `current_intptr.empty()` without overhead (because the compiler will optimize it away)

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,238 +86,263 @@ auto read_json(std::filesystem::path const& path) {
...
 }
 
-auto construct_individual_scenarios(OwningDataset& owning_dataset) {
-    for (Idx scenario_idx{}; scenario_idx < owning_dataset.dataset.batch_size(); ++scenario_idx) {
-        owning_dataset.batch_scenarios.push_back(owning_dataset.const_dataset.get_individual_scenario(scenario_idx));
+OwningDataset create_result_dataset(OwningDataset const& input, std::string const& dataset_name, Idx is_batch = 0,
+                                    Idx batch_size = 1) {
+    OwningDataset owning_dataset{.dataset{DatasetMutable{dataset_name, is_batch, batch_size}},
+                                 .const_dataset = std::nullopt};
+    auto const& info_input = input.const_dataset.value().get_info();
```
mgovers, line 135: readability

```suggestion
    auto const& input_info = input.const_dataset.value().get_info();
```

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,238 +86,263 @@ auto read_json(std::filesystem::path const& path) {
...
+                   std::string const& component_name, std::string const& attribute_name, double const& dynamic_atol,
+                   double const& rtol) {
+    bool const match =
+        is_angle
+            ? check_angle_and_magnitude<decltype(ref_attribute_value)>(ref_attribute_value, attribute_value,
+                                                                       ref_possible_attribute_value,
+                                                                       possible_attribute_value, dynamic_atol, rtol)
+            : compare_value<decltype(ref_attribute_value)>(ref_attribute_value, attribute_value, dynamic_atol, rtol);
+    if (match) {
+        CHECK(match);
```
Jerry-Jinfeng-Guo, line 234: This line is redundant

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,238 +86,263 @@ auto read_json(std::filesystem::path const& path) {
...
+        return;
     }
-    assert(angle_attr.ctype == CType::c_double3);
-    assert(mag_attr.ctype == CType::c_double3);
-    return check_angle_and_magnitude<asymmetric_t>(reference_result_ptr, result_ptr, angle_attr, mag_attr, atol, rtol,
-                                                   obj);
+    // for angle attribute, also check the magnitude available
+    if (is_angle && is_nan(ref_possible_attribute_value)) {
+        return;
+    }
```
Jerry-Jinfeng-Guo, line 278: Why is it that in a `check*` function some `nan` would only yeild `void return`s but nothing else?

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,238 +86,263 @@ auto read_json(std::filesystem::path const& path) {
...
                 for (auto const& [reg, value] : atol) {
-                    if (std::regex_match(attr.name, std::regex{reg})) {
+                    if (std::regex_match(attribute_name, std::regex{reg})) {
                         dynamic_atol = value;
                         break;
                     }
                 }
                 // for other _angle attribute, we need to find the magnitue and compare together
                 std::regex const angle_regex("(.*)(_angle)");
                 std::smatch angle_match;
```
Jerry-Jinfeng-Guo, line 329: These two fellas could live out side of the for loops

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -40,238 +86,263 @@ auto read_json(std::filesystem::path const& path) {
...
+            auto const& ref_buffer = reference_storage.buffers.at(component_idx);
+            auto const& buffer = storage.buffers.at(component_idx);
+            Idx const elements_per_scenario = reference_result_info.component_elements_per_scenario(component_idx);
+            CHECK(elements_per_scenario >= 0);
+            // loop through all attributes
+            for (Idx attribute_idx{}; attribute_idx < MetaData::n_attributes(component_meta); ++attribute_idx) {
+                auto const* const attribute_meta = MetaData::get_attribute_by_idx(component_meta, attribute_idx);
+                auto attribute_type = MetaData::attribute_ctype(attribute_meta);
+                auto const& attribute_name = MetaData::attribute_name(attribute_meta);
                 // TODO skip u angle, need a way for common angle
```
Jerry-Jinfeng-Guo, line 315: Not from this PR, but got curious: was it skipping that is TODO, or a way for common angle TODO? If latter:
```suggestion
                // TODO need a way for common angle: u angle skipped for now
```

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -335,29 +400,17 @@ struct CaseParam {
...
-        }
-        return model.calculate(options, dataset, update_dataset);
-    };
+Options get_options(CaseParam const& param, Idx threading = -1) {
+    Options options{};
+    options.set_calculation_type(calculation_type_mapping.at(param.calculation_type));
+    options.set_calculation_method(calculation_method_mapping.at(param.calculation_method));
+    options.set_symmetric(param.sym ? 1 : 0);
+    options.set_err_tol(1e-8);
+    options.set_max_iter(20);
```
Jerry-Jinfeng-Guo, line 409: Redundant?

Jerry-Jinfeng-Guo (commented):
Couple minor comments.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: I don't think so. If the tolerance is reached before 20 iterations, then it is; however, if the tolerance is not reached, the maximum number of iterations will be the safeguard (specially when creating new tests and they don't seem to converge).

mgovers (commented):
halfway through the review. continuing now but then you can already process what i have so far

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: Should they? Isn't it best practice to shorten lifetime of variables as much as possible into their relevant scope? Since both the fellas are used within the for loop, maybe it is best to keep them in? I am not totally sure what the best practice here is though.

tests/cpp_validation_tests/test_validation.cpp
mgovers: the latter

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: Because if the reference attributes are `nan`, nothing is to be checked, they can be skipped. This is needed because of the way the reference result dataset is created. This way we only check for the results we actually want to check.

tests/cpp_validation_tests/test_validation.cpp
mgovers: > Isn't it best practice to shorten lifetime of variables as much as possible into their relevant scope?

yes indeed. however, since the regex is contant, you can also make it `constexpr` (iff std::regex supports that), and/or make it module-level static, in which case it's important that you put it in an anonymous namespace.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: In the context of did this test fail or not, yes, but it is very useful for debugging, so I would advocate to leave it in.

tests/cpp_validation_tests/test_validation.cpp
Jerry-Jinfeng-Guo: Not if you reuse them this frequently ;)

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -321,11 +384,13 @@ struct CaseParam {
     std::string calculation_method;
     std::string short_circuit_voltage_scaling;
     std::string tap_changing_strategy;
+    double err_tol = 1e-8;
+    Idx max_iter = 20;
+    Idx threading = -1;
     bool sym{};
```
Jerry-Jinfeng-Guo, line 390: > I don't think so. If the tolerance is reached before 20 iterations, then it is; however, if the tolerance is not reached, the maximum number of iterations will be the safeguard (specially when creating new tests and they don't seem to converge).

@figueroa1395 are you sure?

tests/cpp_validation_tests/test_validation.cpp
Jerry-Jinfeng-Guo: See the other comment.

tests/cpp_validation_tests/test_validation.cpp
Jerry-Jinfeng-Guo: In that case, perhaps we could do a simple `should_skip_check` function to do these two `if` statements and only do the check if otherwise:
```
    // check attributes
    if (!should_skip_check(ref_attribute_value, is_angle, ref_possible_attribute_value)) {
        check_results(ref_attribute_value, attribute_value, ref_possible_attribute_value, possible_attribute_value,
                      is_angle, scenario_idx, obj, component_name, attribute_name, dynamic_atol, rtol);
    }
}
```

tests/cpp_validation_tests/test_validation.cpp
Jerry-Jinfeng-Guo: I think you meant to say `current_indptr.empty()`? @mgovers

tests/cpp_validation_tests/test_validation.cpp
Jerry-Jinfeng-Guo: I am not sure that's a convincing argument. You could also simply do:
```suggestion
    if (!match) {
        //...
```

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: Oops. Thanks for catching.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: After thinking again about this, I think you are right, this is just redundant. Change will be reflected in following commit.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: We were talking about different types of redundancy. Thanks for pointing this out. Changes will show in following commit. Resolving.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: Making appropriate changes. Resolving.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: I like this approach, a lot more readable. Resolving now, the implementation will be reflected in the following commit.

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: > since the regex is contant, you can also make it `constexpr` (iff std::regex supports that), and/or make it module-level static

`std::regex` doesn't support `consexpr` and I would avoid making it module-level static because the readability would be heavily affected.

> In this particular case, I would buy in this for the following two reasons:
> 
> * readability
> * relatively cheap construction & destruction

I advocate to leave it as is for these two reasons specially. Taking it to the outter scope really will affect readability, specially in such complex checking logic.

@Jerry-Jinfeng-Guo @mgovers Do we agree to keep it as is in this particular case?

tests/cpp_validation_tests/test_validation.cpp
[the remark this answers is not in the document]
figueroa1395: It should be fixed now. However, I will shortly do another manual run after I have pushed the changes addressing comments.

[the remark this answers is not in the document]
figueroa1395: Current manual run: https://github.com/PowerGridModel/power-grid-model/actions/runs/11213443536

Still waiting for results.

tests/cpp_validation_tests/test_validation.cpp
Jerry-Jinfeng-Guo: Yup, let's leave it like this.

tests/cpp_validation_tests/test_validation.cpp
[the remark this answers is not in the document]
TonyXiang8787: It's passed now.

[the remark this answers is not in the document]
figueroa1395: Since it passes now, I will resolve this.

TonyXiang8787 (approved):
I see no open points and therefore approve the PR.

Feel free to put it in the merge queue.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -132,16 +140,20 @@ OwningDataset create_result_dataset(OwningDataset const& input, std::string cons
                                     Idx batch_size = 1) {
     OwningDataset owning_dataset{.dataset{DatasetMutable{dataset_name, is_batch, batch_size}},
                                  .const_dataset = std::nullopt};
-    auto const& info_input = input.const_dataset.value().get_info();
 
-    for (Idx component_idx{}; component_idx != info_input.n_components(); ++component_idx) {
-        auto const& component_name = info_input.component_name(component_idx);
+    if (!input.const_dataset.has_value()) {
+        throw OptionalNotInitialized("DatasetConst");
```
mgovers, line 145: here and below `UnreachableHit` could also do the trick but it's fine to keep as is

figueroa1395: I see that `UnreachableHit` lives in `power_grid_model` only, so I will keep as is for now. If error management becomes an issue later, we can open an issue when/if that happens.

mgovers: OK

PR #732, opened 2024-09-23
MutableDataset to ConstDataset for columnar data

nitbharambe:


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -209,7 +209,15 @@ template <dataset_type_tag dataset_type_> class Dataset {
         : meta_data_{&other.meta_data()}, dataset_info_{other.get_description()} {
         for (Idx i{}; i != other.n_components(); ++i) {
             auto const& buffer = other.get_buffer(i);
-            buffers_.push_back(Buffer{.data = buffer.data, .indptr = buffer.indptr});
+            Buffer new_buffer{.data = buffer.data, .indptr = buffer.indptr};
+            for (auto const& attribute_buffer : buffer.attributes) {
+                AttributeBuffer<Data> const new_attribute_buffer{
+                    .data = attribute_buffer.data,
+                    .meta_attribute = &dataset_info_.component_info[i].component->get_attribute(
+                        attribute_buffer.meta_attribute->name)};
```
TonyXiang8787, line 217: Why does it need to be so complicated? Can we just do:

```suggestion
                    .meta_attribute =attribute_buffer.meta_attribute;
```

figueroa1395: You are totally right and we just overlooked it. Fixed in the next commit.

TonyXiang8787 (approved):

PR #734, opened 2024-09-25
Feature/Update validation for columnar

figueroa1395:
Update validation functions to support columnar data:

- [x] Update `validate_input_data` function.
- [x] Update `validate_batch_data` function.
- [x] Update tests:
  - [x] `test_input_validation.py`
  - [x] `test_batch_validation.py`

### DIFF
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -307,16 +307,12 @@ def compatibility_convert_row_columnar_dataset(
     dataset_type: DatasetType,
     available_components: list[ComponentType] | None = None,
 ) -> Dataset:
-    """Temporary function to copy row based dataset to a column based dataset as per the data_filter.
-    The purpose of this function is to mimic columnar data without any memory footprint benefits.
-    Note: If both the input and requested output are row based, the same dataset is returned without a copy.
+    """Temporary function to transform row, column or mixed based datasets to a full row or column based dataset as per
+    the data_filter. The purpose of this function is to mimic columnar data and transform back to row data without any
+    memory footprint benefits.
+    Note: Copies are made in a per-component basis; if a component is row based in both the input and the requested
+    output, that componened is returned without a copy.
 
-    Args:
-        data (Dataset):
-        component_types (_ComponentAttributeMappingDict):
-
-    Returns:
-        Dataset: converted dataset
     Args:
         data (Dataset): dataset to convert
         data_filter (ComponentAttributeMapping): desired component and attribute mapping
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -310,7 +310,7 @@ def compatibility_convert_row_columnar_dataset(
     """Temporary function to transform row, column or mixed based datasets to a full row or column based dataset as per
     the data_filter. The purpose of this function is to mimic columnar data and transform back to row data without any
     memory footprint benefits.
-    Note: Copies are made in a per-component basis; if a component is row based in both the input and the requested
+    Note: Copies are made on a per-component basis; if a component is row based in both the input and the requested
     output, that componened is returned without a copy.
 
     Args:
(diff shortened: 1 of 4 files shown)


figueroa1395:
At this moment, the columnar based batch data is validated by converting the full batch to row based, which may have a big memory overhead. However, once #699 is finished, this will be upgraded to do the conversion per individual batches.

src/power_grid_model/_utils.py
```diff
@@ -307,16 +307,12 @@ def compatibility_convert_row_columnar_dataset(
     dataset_type: DatasetType,
     available_components: list[ComponentType] | None = None,
 ) -> Dataset:
-    """Temporary function to copy row based dataset to a column based dataset as per the data_filter.
-    The purpose of this function is to mimic columnar data without any memory footprint benefits.
-    Note: If both the input and requested output are row based, the same dataset is returned without a copy.
+    """Temporary function to transform row, column or mixed based datasets to a full row or column based dataset as per
+    the data_filter. The purpose of this function is to mimic columnar data and transform back to row data without any
+    memory footprint benefits.
+    Note: Copies are made in a per-component basis; if a component is row based in both the input and the requested
```
mgovers, line 313:
```suggestion
    Note: Copies are made on a per-component basis; if a component is row based in both the input and the requested
```

src/power_grid_model/validation/validation.py
```diff
@@ -131,8 +135,12 @@ def validate_batch_data(
 
     input_errors: list[ValidationError] = list(validate_unique_ids_across_components(input_data))
 
+    # Convert to row based if in columnar format
+    # TODO(figueroa1395): transform to columnar per single batch once the columnar dataset python extension is finished
```
mgovers, line 139:
```suggestion
    # TODO(figueroa1395): transform to columnar per single batch scenario once the columnar dataset python extension is finished
```

this is not strictly necessary; just a speed-up

tests/unit/validation/test_batch_validation.py
```diff
@@ -53,6 +54,16 @@ def batch_data() -> dict[str, np.ndarray]:
     return {"line": line, "asym_load": asym_load}
 
 
+@pytest.fixture
+def original_batch_data_columnar(original_batch_data):
+    return compatibility_convert_row_columnar_dataset(original_batch_data, ..., DatasetType.update)
```
mgovers, line 59: it's a nitpick, but `...` looks weird here. Maybe it's a good idea to be explicit.

```suggestion
    return compatibility_convert_row_columnar_dataset(original_batch_data, Ellipsis, DatasetType.update)
```

it's equivalent, but more clearly tells the reader that it's a deliberate special value that is different than just "everything"/"et cetera" (which is what the ellipsis stands for)

the way i read the current state is also:
* "return (for compatibility) a converted row/columnar data set from the original batch data etc. as an update data set type".

as opposed to
* "return (for compatibility) a converted row/columnar data set from the original batch, using ellipsis and update data set type as parameters"

alternatively, you can also use explicit keyword arguments

tests/unit/validation/test_input_validation.py
```diff
@@ -274,6 +275,16 @@ def input_data() -> dict[ComponentType, np.ndarray]:
     return data
 
 
+@pytest.fixture
+def original_data_columnar(original_data):
+    return compatibility_convert_row_columnar_dataset(original_data, ..., DatasetType.input)
```
mgovers, line 280: idem

src/power_grid_model/validation/validation.py
figueroa1395: Yup, but since everything validation related is done on the Python side, we may as well try to do it as "fast" as possible. Hence, even though this is probably mergable, I will wait a bit since there is no rush.

mgovers (approved):
since validation for columnar data is not blocked by the Python API (since we already have preliminary python support), it's also OK to merge this. your call

figueroa1395:
Merging now and doing the incremental upgrade to per-batch scenario validation later. This way the functionality is already provided to the users.

PR #736, opened 2024-09-25
Feature/add nightly

### LINKED ISSUE #735: [FEATURE] Nightly build

Add a Nightly build that runs a full clang-tidy compilation on all targets, including the benchmark, package tests, etc.

## Background
Due to compilation time, the following decisions were made in the past:

* In #345 , the clang-tidy build was restricted to the C API, and building clang-tidy on the tests was disabled from every PR and `main` build.
* During #718 , it was decided to remove the benchmark build from every PR and from every `main` build.

The downside of those decisions is, that there is the danger that the omitted parts of the compilation break and that that is unnoticed.

This feature will introduce a Nightly build that re-enables those checks

## Desired implementation

* [x] exclude build target `power_grid_model_benchmark_cpp` from the `ALL` target using `EXCLUDE_FROM_ALL`
* [x] add `all power_grid_model_benchmark_cpp` to the `options` for the `target` `inputs` for the `workflow_dispatch` in `.github/workflows/clang-tidy.yml`
  * [x] This is possible cfr. https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-build-t
* [x] enable `on` `workflow_call` for `.github/workflows/nightly.yml` and `.github/workflows/clang-tidy.yml` cfr. https://docs.github.com/en/actions/sharing-automations/reusing-workflows
* [x] add `.github/workflows/nightly.yml`
  * [x] call `.github/workflows/main.yml` and `.github/workflows/clang-tidy.yml` from this workflow cfr. https://docs.github.com/en/actions/sharing-automations/reusing-workflows

mgovers:
* [x] Adds support for nightly build
* [x] Disables benchmark from the `all` target

Closes #735

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,8 @@ on:
   pull_request:
   # run pipeline on merge queue
   merge_group:
+  # run pipeline from another workflow
+  workflow_call:
   # run this workflow manually from the Actions tab
   workflow_dispatch:
     inputs:
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -13,6 +13,13 @@ on:
   pull_request:
   # run pipeline on merge queue
   merge_group:
+  # run pipeline from another workflow
+  workflow_call:
+    inputs:
+      target:
+        type: string
+        description: "The CMake target to run (e.g.: all)"
+        required: true
   # run this workflow manually from the Actions tab
   workflow_dispatch:
     inputs:
@@ -24,7 +31,7 @@ on:
           - all
           - power_grid_model_c
           - power_grid_model_cpp
-          - install power_grid_model_benchmark_cpp
+          - all power_grid_model_benchmark_cpp
         required: true
 
 concurrency:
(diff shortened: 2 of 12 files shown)


TonyXiang8787:
Why are there so many code change? I thought this is only adding new workflow?

mgovers:
> Why are there so many code change? I thought this is only adding new workflow?

to prove that the workflow does indeed call the correct parameters. I would prefer the first version of the workflow to pass, and then we can start there.

however, there's no target that currently passes in clang-tidy other than the `power_grid_model_c` target. This adds `power_grid_model_cpp` to `clang-tidy` as well.

.github/workflows/main.yml
```diff
@@ -14,6 +14,8 @@ on:
   pull_request:
   # run pipeline on merge queue
   merge_group:
+  # run pipeline from another workflow
+  workflow_call:
```
TonyXiang8787, line 18: Why do we need to run `main.yml` in nightly build? We just need `clang-tidy` right?

mgovers: we don't need to but i'd consider it best practice.

in addition, there's no harm in enabling this option now. if you have strong feelings about it, i can remove it

TonyXiang8787: It's fine to also run main

figueroa1395 (approved):

PR #750, opened 2024-09-30
add check code quality + citations to nightly

mgovers:
* [x] test run in https://github.com/PowerGridModel/power-grid-model/actions/runs/11104508116

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -25,6 +25,9 @@ jobs:
   check-code-quality:
     uses: "./.github/workflows/check-code-quality.yml"
 
+  reuse-compliance:
+    uses: "./.github/workflows/reuse-compliance.yml"
+
   clang-tidy:
     uses: "./.github/workflows/clang-tidy.yml"
     with:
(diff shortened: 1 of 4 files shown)


TonyXiang8787 (dismissed):

.github/workflows/nightly.yml
```diff
@@ -22,7 +22,13 @@ jobs:
     with:
```
TonyXiang8787, line 22: why do we give the permission write here?

mgovers: it is needed by the Release job in `main.yml` (note that the actual release is skipped)

TonyXiang8787: If the release step is skipped. We don't need that permission right?

mgovers: theoretically, we don't, but the full sub-workflow needs the same permissions and that can't depend on the fact that some step is skipped somewhere later in the action. just like how in C++, a functions' `const`-ness can't depend on its arguments:

if you want to remove these permissions, the release job needs to be in a separate workflow

TonyXiang8787 (approved):

mgovers:
adding auto-merge even though the test run has not fully finished yet, because the jobs that haven't finished yet are things that are not affected by this PR

Issue #729, opened 2024-09-23
[FEATURE] Provide a generic_branch


petersalemink95:
Following the discussion in #642 it was concluded that the implementation of a generic branch could be an addition to Power Grid Model.

We can make a new component `generic_branch`, which models branches as a Pi model; i.e. the user can specify:
- k: transformer ratio (float)
- theta0 / theta1 / theta2: angle shift
- r0_series / r1_series / r2_series
- x0_series / x1_series / x2_series
- g0_shunt / g1_shunt / g2_shunt
- b0_shunt / b1_shunt / b2_shunt

Note: except for `k`, all attributes are given for zero, positive and negative sequence.

##steps to follow
In order to add a new component to Power Grid Model the following steps need to be taken:

- [ ] Check if the available input/update/output data suffices (check `power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/` `input.hpp` / `update.hpp` / `output.hpp` or in the documentation directly)
- [ ] If not, add the new data format to `code_generation/data/attribute_classes/` `input.json` / `update.json` / `output.json`  + run `code_generation/code_gen.py`
- [ ] Create a new component in a new `power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/component/[component].hpp` file that at least inherits from Base, but in this case `GenericBranch` should inherit from `Branch`
- [ ] If necessary: add new enums or exceptions
- [ ] Create the necessary unit tests in `power-grid-model/tests/cpp_unit_tests/test_[component].cpp`
- [ ] Add the `test_[component].cpp` to `power-grid-model/tests/cpp_unit_tests/CMakeLists.txt`
- [ ] Add component to `power_grid_model_c/power_grid_model/include/power_grid_model/all_components.hpp`
- [ ] Not necessary for this component (If necessary update `main_core/topology.hpp` / `input.hpp` / `output.hpp` / `update.hpp`)
- [ ] Add component to `code_generation/data/dataset_class_maps/dataset_definitions.json` + re-run `code_generation/code_gen.py`
- [ ] Add validation test cases to `tests/data`
- [ ] Update input/update data validator for the new component: `src/power_grid_model/validation/validation.py` + add corresponding tests

Note: the order is recommended, but not necessary

TonyXiang8787:
[...]

PR #752, opened 2024-09-30
create generic_branch #729

sudo-ac:
**Fixes issue**: `#729 [FEATURE] Provide a generic_branch `

- added Component generic_branch as diskussed in #729 
- added testcase generic_branch  (self test of the new component)
- added testcase compare_generic_branch  (compare a generic_branch with identical transformer)


TonyXiang8787:
Hi @sudo-ac,

Thanks for submitting the PR. Before I go in-depth of the content, I would like to give compliment for passing almost all the CI checks for the first try. This rarely happens, even for the active PGM developer.

mgovers:
Again: thank you for creating the PR. It looks quite good at first sight. I've already started going over this PR, but something else came up. Will continue with the review after that's solved.

power_grid_model_c/power_grid_model/include/power_grid_model/component/genericbranch.hpp
TonyXiang8787, line 1: To make the file name consistent, please change this file name to `generic_branch.hpp`

code_generation/data/attribute_classes/input.json
```diff
@@ -127,6 +127,33 @@
...
+                },
+                {
+                    "data_type": "double",
+                    "names": "k",
+                    "description": "transformer ratio, default = 1.0"
+                },
+                {
+                    "data_type": "double",
+                    "names": "theta",
+                    "description": "angle shift"
```
TonyXiang8787, line 147: We do not use degree anywhere in PGM interface. Please emphasize in the description that this is in radian.

power_grid_model_c/power_grid_model/include/power_grid_model/component/genericbranch.hpp
```diff
@@ -0,0 +1,84 @@
...
+          sn_{genericbranch_input.sn},
+          r1_{genericbranch_input.r1},
+          x1_{genericbranch_input.x1},
+          g1_{genericbranch_input.g1},
+          b1_{genericbranch_input.b1},
+          k_{std::isnan(genericbranch_input.k) ? 1.0 : genericbranch_input.k},
+          theta_{std::isnan(genericbranch_input.theta) ? 0.0 : std::fmod(genericbranch_input.theta, 2 * pi)},
+          base_i_from_{base_power_3p / u1_rated / sqrt3},
+          base_i_to_{base_power_3p / u2_rated / sqrt3},
+          base_y_{base_i_to_ / (u2_rated / sqrt3)} {
```
TonyXiang8787, line 50: Good one. As conventional, the generical branch parameters are calculated with respect in the `to` side of the branch. 

Please also document this in the documentation.

code_generation/data/attribute_classes/input.json
```diff
@@ -127,6 +127,33 @@
...
+            "attributes": [
+                {    
+                    "data_type": "double",
+                    "names": ["r1","x1","g1","b1"],
+                    "description": "positive sequence parameters"
+                },
+                {
+                    "data_type": "double",
+                    "names": "k",
+                    "description": "transformer ratio, default = 1.0"
```
TonyXiang8787, line 142: What is the definition of user-provided transformer ratio?

Is it regular transformer ratio or off-nominal ratio? 

For example, given a 155/11 kV transformer connecting to the nodes with rated voltage of 150kV and 10kV.

The regular ratio is: 155/11= 14.09
The off-nominal ratio is: (155/11) / (150/10) = 0.9393

Based on your implementation below, you seem to have chosen for the off-nominal ratio. Please also document this clearly in the attribute description and component documentation.

power_grid_model_c/power_grid_model/include/power_grid_model/component/genericbranch.hpp
```diff
@@ -0,0 +1,84 @@
...
+    static constexpr char const* name = "generic_branch";
+
+    explicit GenericBranch(GenericBranchInput const& genericbranch_input, double u1_rated, double u2_rated)
+        : Branch{genericbranch_input},
+          sn_{genericbranch_input.sn},
+          r1_{genericbranch_input.r1},
+          x1_{genericbranch_input.x1},
+          g1_{genericbranch_input.g1},
+          b1_{genericbranch_input.b1},
+          k_{std::isnan(genericbranch_input.k) ? 1.0 : genericbranch_input.k},
```
TonyXiang8787, line 46: The `k_` in the class should always be off-nominal transformer ratio. Given the code you wrote here, the user input `k` is also off-nominal transformer ratio (instead of regular transformer ratio). If this is a deliberate choice, please document it.

src/power_grid_model/validation/validation.py
```diff
@@ -534,6 +535,12 @@ def validate_line(data: SingleDataset) -> list[ValidationError]:
     return errors
 
 
+def validate_genericbranch(data: SingleDataset) -> list[ValidationError]:
```
TonyXiang8787, line 538: Use `generic_branch`.

tests/cpp_unit_tests/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ set(PROJECT_SOURCES
     "test_three_phase_tensor.cpp"
     "test_node.cpp"
     "test_line.cpp"
+    "test_genericbranch.cpp"
```
TonyXiang8787, line 17: same here, change the name to `generic_branch.cpp`

TonyXiang8787 (changes requested):
Nice work, some comments about clarity.

The documentation still has to be added, specifically two parts:

* Add `generic_branch` in the [component hierarchy](https://github.com/PowerGridModel/power-grid-model/blob/main/docs/user_manual/data-model.md) 
* Add `generic_branch` in the [component attribute documentation](https://github.com/PowerGridModel/power-grid-model/blob/main/docs/user_manual/components.md). Pay attention to explain two important notes:
  * The attributes of impedance and admittance should be calculated in the `to` side.
  * The definition of `k`, is that regular ratio or off-nominal ratio?

mgovers (commented):
since tony also already reviewed, here's my intermediate review as well. I'll continue after all comments up til now are resolved

power_grid_model_c/power_grid_model/include/power_grid_model/component/genericbranch.hpp
```diff
@@ -0,0 +1,84 @@
...
+          base_y_{base_i_to_ / (u2_rated / sqrt3)} {
+
+        y1_series_ = 1.0 / (r1_ + 1.0i * x1_) / base_y_;
+        y1_shunt_ = (g1_ + 1.0i * b1_) / base_y_;
+    }
+
+    // override getter
+    double base_i_from() const override { return base_i_from_; }
+    double base_i_to() const override { return base_i_to_; }
+    double loading(double max_s, double /*max_i*/) const override { return std::isnan(sn_) ? 0.0 : (max_s / sn_); };
```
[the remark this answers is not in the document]
mgovers: ok, makes sense. we can always add it later as a separate feature request. resolving this comment

mgovers:
> One test failed, but we have no control over it:
> 
> > >
> 
> ************* Module power_grid_model.validation.validation /opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/site-packages/power_grid_model/validation/validation.py:1:0: C0302: Too many lines in module (1001/1000) (too-many-lines)

Hi @sudo-ac , i'll see if i can come up with a good suggestion on this. The length of the module has always bothered me, but  we've never had the time to refactor it. Unfortunately, you're the first one to experience a failing CI because of it. I'll dig into this on Monday.

TonyXiang8787:
@mgovers My proposal is to add a linter exception for this file? To refactor the whole validation function is out of the scope.

Maybe we need to define all the rules in `json` files. The validation function load the `json` file and perform certain check.

docs/user_manual/data-model.md
```diff
@@ -20,6 +20,7 @@ graph LR
       branch-->line
       branch-->link
       branch-->transformer
+      branch-->generic_branch
```
TonyXiang8787, line 23: You also need to add `generic_branch` in the list below in line 43.

TonyXiang8787:
> @TonyXiang8787,
> 
> but I added the requested documentation. What was wrong?

Yes, indeed. Sorry I did not pay attention. I just put one more comment.

TonyXiang8787 (changes requested):
Hi @sudo-ac,

Thanks for the great work so far. The final step to merge this PR is to add a validation test cases in the `tests/data` folder. Please have a look at the example of validation test case of `line`.

https://github.com/sudo-ac/power-grid-model/tree/feature/symmetric-generic-branch/tests/data/power_flow/pandapower/components/symmetric/line

You can build a simple network which involves a `generic_branch`. Manually calculate the power flow results and put both input and output in a validation test case.

mgovers:
> @mgovers My proposal is to add a linter exception for this file? To refactor the whole validation function is out of the scope.

Thank you @TonyXiang8787 for picking that up. I indeed agree that silencing for this PR is the way to go for now. Let's revisit the module after this PR is merged and try to come up with a sustainable solution.

> Maybe we need to define all the rules in `json` files. The validation function load the `json` file and perform certain check.

I believe the entire validation functionality requires some rework. JSON files sound good, but more complex logic will be difficult to implement that way. Take for example the logic surrounding the default value of the tap position and the one concerning the control side of the transformer tap regulator

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -166,6 +166,11 @@ class DuplicativelyRegulatedObject : public PowerGridError {
     }
 };
 
+class NotImplementedError : public PowerGridError {
```
mgovers, line 169: this needs to be added to `src/power_grid_model/errors.py` and `src/power_grid_model/core/error_handling.py`

mgovers (commented):
great work so far! I have a couple more additions to Tony's review but nothing major.

power_grid_model_c/power_grid_model/include/power_grid_model/all_components.hpp
```diff
@@ -25,8 +26,9 @@
 
 namespace power_grid_model {
 
-using AllComponents = ComponentList<Node, Line, Link, Transformer, ThreeWindingTransformer, Shunt, Source, SymGenerator,
-                                    AsymGenerator, SymLoad, AsymLoad, SymPowerSensor, AsymPowerSensor, SymVoltageSensor,
-                                    AsymVoltageSensor, Fault, TransformerTapRegulator>;
+using AllComponents =
+    ComponentList<Node, Line, Link, GenericBranch, Transformer, ThreeWindingTransformer, Shunt, Source, SymGenerator,
+                  AsymGenerator, SymLoad, AsymLoad, SymPowerSensor, AsymPowerSensor, SymVoltageSensor,
+                  AsymVoltageSensor, Fault, TransformerTapRegulator>;
```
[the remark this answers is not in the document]
TonyXiang8787: We never guarantee the order of components. So it is not breaking change. No futher actions needed.

docs/user_manual/components.md
```diff
@@ -224,6 +224,48 @@ $$
...
+| `k`     | `double`  | -                | off-nominal ratio             | &#10060; default `1.0` | &#10060; |                 |
+| `theta` | `double`  | radian           | angle shift                   | &#10060; default `0.0` | &#10060; |                 |
+| `sn`    | `double`  | volt-ampere (VA) | rated power                   | &#10060; default `0.0` | &#10060; | `>= 0`          |
+
+```{note} 
+The impedance (`r1`, `x1`) and admittance (`g1`, `b1`) attributes are calculated with reference to the "to" side of the branch.
+
+The parameter `k` represents the **off-nominal ratio**, not the nominal voltage ratio. This means that `k` must be explicitly provided by the user, particularly in cases involving a tap changer or a voltage transformer. The program does not automatically calculate `k` based on the nominal voltage levels of the connected nodes.
+
+Asymmetric calculation is not supported for `generic_branch`.
```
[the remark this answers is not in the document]
TonyXiang8787: I agree. The parameter `k` defined in a certain way. In this block we explain the attribute. It is not warning to the user.

src/power_grid_model/errors.py
```diff
@@ -109,6 +109,10 @@ class PowerGridDatasetError(PowerGridError):
     """Error occurs during dataset handling."""
 
 
+class NotImplementedError(PowerGridError):
+    """The functionality is either not supported or not yet implemented."""
```
mgovers, line 113: hmm, I see pylint doesn't like this (https://github.com/PowerGridModel/power-grid-model/actions/runs/11215571356/job/31173041487?pr=752)

```
 ************* Module power_grid_model.errors
/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/site-packages/power_grid_model/errors.py:112:0: W0622: Redefining built-in 'NotImplementedError' (redefined-builtin)
************* Module power_grid_model.core.error_handling
/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/site-packages/power_grid_model/core/error_handling.py:16:0: W0622: Redefining built-in 'NotImplementedError' (redefined-builtin)
```

maybe you can rename it to `PowerGridNotImplementedError` (cfr. `PowerGridUnreachableHitError` below).

As an alternative, we could also use the built-in `NotImplementedError`, with the danger that it does not derive from `PowerGridError`, which could surprise users. I don't think that that would be the way to go here.

We could let `PowerGridNotImplementedError` derive from both `PowerGridError` and `NotImplementedError`, but multiple-inheritance exception handling may also result in unexpected behavior

```py
>>> class A(Exception):
...   pass
...
>>> class B(A):         
...   pass
...
>>> class C(A, NotImplementedError):
...   pass
...
>>> try:        
...   raise B()
... except A:
...   print("A")
... except NotImplementedError:
...   print("NIE")
...
A
>>> try:         
...   raise C() 
... except A:    
...   print("A")
... except NotImplementedError:
...   print("NIE")
...
A
>>> try:
...   raise C()
... except NotImplementedError:
...   print("NIE")
... except A:    
...   print("A")
...
NIE
```

tests/data/power_flow/generic_branch/params.json
```diff
@@ -0,0 +1,5 @@
+{
+  "calculation_method": ["newton_raphson"],
+  "rtol": 1e-05,
```
mgovers, line 3: this is quite large for a relative tolerance. E.g. the difference in the `u` of the two nodes in the `sym_output.json` is smaller (relatively), meaning the actual output could also have `u[1] > u[0]`, which would imply a sign inversion. I'm OK with keeping it as is, but if it also works with smaller rtol, then that would be great.

docs/user_manual/components.md
```diff
@@ -224,6 +224,52 @@ $$
 where $z_{\text{base}} = 1 / y_{\text{base}} = {u_{\text{2, rated}}}^2 / s_{\text{base}}$. 
 Here, $s_{\text{base}}$ is a constant value determined by the solver and $u_{\text{2, rated}}$ is rated voltage at `to_node`.
 
+
+### Generic_Branch
```
mgovers, line 228: our other sections on components consisting of 2 words (e.g. `three_winding_transformer` and `transformer_tap_regulator`) have spaces as a section header
```suggestion
### Generic Branch  
```

TonyXiang8787 (approved):
I think we are good to go!

@sudo-ac thanks for the nice contribution!

mgovers (commented):
LGTM, I have no further remarks left. @TonyXiang8787 can you do a final review and give the approval if it's the same for you?

mgovers (approved):
you just did 😬 Great work and thank you for your input!

PR #753, opened 2024-10-01
Add numpy to mypy check

mgovers:
apparently, we did not check for numpy types, so a lot of the mypy errors in #699 were hidden

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -55,7 +55,7 @@ jobs:
 
       - name: Install and run mypy
         run: |
-          pip install mypy
+          pip install mypy numpy
           mypy .
 
       - name: Install and run pylint
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -561,6 +561,8 @@ def _extract_columnar_data(data: ComponentData, is_batch: bool | None = None) ->
     Returns:
         ColumnarData: the contents of columnar data
     """
+    not_columnar_data_message = "Expected columnar data"
+
     if is_batch is not None:
         allowed_dims = [2, 3] if is_batch else [1, 2]
     else:
@@ -569,12 +571,12 @@ def _extract_columnar_data(data: ComponentData, is_batch: bool | None = None) ->
     sub_data = data["data"] if is_sparse(data) else data
 
     if not isinstance(sub_data, dict):
-        raise TypeError("Expected columnar data")
+        raise TypeError(not_columnar_data_message)
     for attribute, attribute_array in sub_data.items():
         if not isinstance(attribute_array, np.ndarray) or not isinstance(attribute, str):
-            raise TypeError("Expected columnar data")
+            raise TypeError(not_columnar_data_message)
         if attribute_array.ndim not in allowed_dims:
-            raise TypeError("Expected columnar data")
+            raise TypeError(not_columnar_data_message)
     return cast(ColumnarData, sub_data)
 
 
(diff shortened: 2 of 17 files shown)


.github/workflows/check-code-quality.yml
```diff
@@ -55,7 +55,7 @@ jobs:
 
       - name: Install and run mypy
         run: |
-          pip install mypy
+          pip install mypy numpy
```
mgovers, line 58: @TonyXiang8787 this is a serious side-effect of the fact that we don't do `pip install -e .[dev]` in this pipeline. should we?

src/power_grid_model/data_types.py
```diff
@@ -167,38 +167,40 @@ class SparseBatchColumnarData(TypedDict):
 Columnar data can be :class:`SingleColumnarData` or :class:`BatchColumnarData`.
 """
 
+_SingleComponentData = TypeVar("_SingleComponentData", SingleArray, SingleColumnarData)  # deduction helper
```
mgovers, line 170: i was only able to find this precisely because it suddenly started complaining on mypy on the `scripts/` dir when running locally. this almost broke mypy for our users

.github/workflows/check-code-quality.yml
mgovers: maybe out of scope of this PR

src/power_grid_model/_utils.py
```diff
@@ -151,7 +151,9 @@ def get_batch_size(batch_data: BatchComponentData) -> int:
         return indptr.size - 1
 
     data_to_check = (
-        next(iter(cast(DenseBatchColumnarData, batch_data).values())) if is_columnar(batch_data) else batch_data
+        next(iter(cast(DenseBatchColumnarData, batch_data).values()))
+        if is_columnar(batch_data)
+        else cast(DenseBatchArray, batch_data)
```
mgovers, line 156: so many of these kinds of statements. it may be a good idea to use a functional approach instead, e.g.

```py
def apply_row_based_or_columnar(
    data,
    *,
    row_based_func: Callable | None = None,
    columnar_func: Callable | None = None
):
    if is_columnar(data):
        if columnar_func is not None:
            columnar_func(data)
    elif row_based_func is not None:
        row_based_func(data)
```
we'll probably have to do some typevar magic to mimic the appropriate template behavior though, but it should get rid of a lot of the `cast`s

mgovers:
merging even though there are unresolved comments and pipeline is still running because it's to a base PR anyways

.github/workflows/check-code-quality.yml
TonyXiang8787: Can you elaborate what you mean?

mgovers: because `numpy` was not installed in this pipeline, the pipeline did not check mypy on any lines containing `numpy` arrays (which is quite a lot). as a result, we now did not catch stuff on mypy that we should have caught.

installing `numpy` into this pipeline enabled checking mypy on lines containing numpy arrays.

now, if, in the future, we also have other dependencies that we want to use in our own mypy checks, we would also have to add those to this pipeline.

However, if we have a step before running any of the code quality checks that does `pip install -e .[dev]`, then all those dev dependencies are already installed automatically, so we don't need to explicitly add them here.

TonyXiang8787: But you can change the order of stuff right. First install `.[dev]` then you have all the stuff.

mgovers: yes that's what I meant: do we want to do that?

TonyXiang8787: no problem

mgovers: creted #756

PR #754, opened 2024-10-01
Feature/Update reuse

figueroa1395:
There is a new release of the reuse-tool (https://github.com/fsfe/reuse-tool/releases/tag/v4.0.0), which makes `.reuse/dep5` deprecated and it is replaced by `REUSE.toml`.

This PR updates to to reuse 4.0.0 and adds `REUSE.toml` with `reuse convert-dep5` command.


REUSE.toml
TonyXiang8787, line 1: You forgot to add header to the `REUSE.toml` self. Haha

figueroa1395: Oops, how it is fixed now.

TonyXiang8787 (approved):

PR #759, opened 2024-10-01
remove experimental flag on columnar output

mgovers:
Remove experimental feature flag on columnar output

### DIFF
--- a/src/power_grid_model/core/power_grid_model.py
+++ b/src/power_grid_model/core/power_grid_model.py
@@ -5,6 +5,7 @@
 """
 Main power grid model class
 """
+
 from enum import IntEnum
 from typing import Optional, Type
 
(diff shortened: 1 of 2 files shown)


src/power_grid_model/core/power_grid_model.py
```diff
@@ -242,7 +244,7 @@ def _calculate_impl(  # pylint: disable=too-many-positional-arguments
         options: Options,
         continue_on_batch_error: bool,
         decode_error: bool,
-        experimental_features: _ExperimentalFeatures | str,  # pylint: disable=too-many-arguments
+        experimental_features: _ExperimentalFeatures | str,  # pylint: disable=too-many-arguments,unused-argument
```
Jerry-Jinfeng-Guo, line 247: space afte comma?

mgovers: nope, the pattern that pylint uses is without spaces

mgovers: either way, i'll just remove this cause sonar cloud is also complaining

mgovers: actually, i'll keep it, cause otherwise the experimental check is somewhere else

PR #761, opened 2024-10-04
move datasetype from dataset to _utils

nitbharambe:
Moves get_dataset_type and tests from power_grid_dataset to _utils

### DIFF
--- a/src/power_grid_model/_utils.py
+++ b/src/power_grid_model/_utils.py
@@ -40,6 +40,7 @@ from power_grid_model.data_types import (
     SparseBatchData,
 )
 from power_grid_model.enum import ComponentAttributeFilterOptions
+from power_grid_model.errors import PowerGridError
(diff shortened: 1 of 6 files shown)


src/power_grid_model/_utils.py
Jerry-Jinfeng-Guo, line 1: What necessitates the move from `power_grid_model_dataset` to here?

nitbharambe: Yes, for circular imports

Jerry-Jinfeng-Guo (approved):

figueroa1395:
> I will soon approve but just a quick question over the context: why?

Since this was answered in a[ comment](https://github.com/PowerGridModel/power-grid-model/pull/761#discussion_r1789902815),I will just re-phrase the answer here such that context is not lost: The change is done to avoid circular imports.

PR #767, opened 2024-10-07
Feature/Add utils to Cpp wrapper

figueroa1395:
Move the utility functions `is_nan` and `pgm_type_func_selector` from the validation tests in #718 to `utils.hpp` as part of the Cpp wrapper, while keeping the CPP API C++17 compliant.


figueroa1395:
As a side note: the `NOLINT`s here are used to keep the Cpp API C++17 compliant and clang-tidy quiet about it. Manual runs of cling-tidy for validation tests and api tests are being run:
- Api tests: https://github.com/PowerGridModel/power-grid-model/actions/runs/11215644660
- Validation tests: Will add when above is finished.

TonyXiang8787:
I have no preference where to put the functions. @mgovers  what do you think?

mgovers (dismissed):

mgovers:
> I have no preference where to put the functions. @mgovers what do you think?

also no objections/preferences. `utils.hpp` is alright, but since it affects the C++ API (experimental), I do think we need to think for a second to see if we can come up with a better name. I'll see if something comes to mind tonight

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <algorithm>
+#include <complex>
+#include <limits>
+#include <type_traits>
+
+namespace power_grid_model_cpp {
+template <typename T>
+inline typename std::enable_if<std::is_floating_point<T>::value, // NOLINT(modernize-use-constraints,
+                                                                 // modernize-type-traits)
+                               bool>::type
```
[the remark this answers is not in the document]
figueroa1395: Definitely. These old ways are hideous!

mgovers, line 21: now that i think about it, you may be able to disable the type hint warning by setting the `CXX_STANDARD` target property in CMake (https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html and https://cmake.org/cmake/help/latest/command/set_target_properties.html)

```cmake
set_target_properties(power_grid_model_cpp PROPERTIES CXX_STANDARD 17)
```

not sure about that, though. If it does not solve the problem, please keep as is.

TonyXiang8787: You cannot do that. In your test code which you consume the wrapper header, you specify c++20.

So this will be conflict.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
```diff
@@ -0,0 +1,65 @@
...
+is_nan(T x) {
+    return std::isnan(x);
+}
```
TonyXiang8787, line 24: Do we really need such complicated template stuff. Why not just

```suggestion
inline is_nan(double x) {
    return std::isnan(x);
}
```

TonyXiang8787: We do not have `float` type anyway. Making template is overkill.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
```diff
@@ -0,0 +1,65 @@
...
+template <typename T>
+inline typename std::enable_if<std::is_floating_point<T>::value, // NOLINT(modernize-use-constraints,
+                                                                 // modernize-type-traits)
+                               bool>::type
+is_nan(std::complex<T> const& x) {
+    return is_nan(x.real()) || is_nan(x.imag());
+}
```
TonyXiang8787, line 31:
```suggestion
inline bool is_nan(std::complex<double> const& x) {
    return is_nan(x.real()) || is_nan(x.imag());
}
```

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
```diff
@@ -0,0 +1,65 @@
...
+inline bool is_nan(int32_t x) { return x == std::numeric_limits<int32_t>::min(); }
+inline bool is_nan(int8_t x) { return x == std::numeric_limits<int8_t>::min(); }
+template <typename T, std::size_t N> inline bool is_nan(std::array<T, N> const& array) {
+    return std::any_of(array.begin(), array.end(), [](T const& element) { return is_nan(element); });
+}
```
TonyXiang8787, line 36:
```suggestion
inline bool is_nan(std::array<double, 3> const& array) {
    return is_nan(array[0]) || is_nan(array[1]) || is_nan(array[2]);
}
```

We will only have `std::array<double,3>` in the cpp wrapper. Why making stuff so difficult.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
TonyXiang8787: I doubt do we need `std::complex` at all in the cpp wrapper? We do not have any complex number in the C-API boundary

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
```diff
@@ -0,0 +1,65 @@
...
+template <typename T>
+inline typename std::enable_if<std::is_floating_point<T>::value, // NOLINT(modernize-use-constraints,
+                                                                 // modernize-type-traits)
+                               bool>::type
+is_nan(std::complex<T> const& x) {
+    return is_nan(x.real()) || is_nan(x.imag());
+}
+inline bool is_nan(int32_t x) { return x == std::numeric_limits<int32_t>::min(); }
+inline bool is_nan(int8_t x) { return x == std::numeric_limits<int8_t>::min(); }
```
mgovers, line 33: please add `Idx` and `ID` instead. @TonyXiang8787 should we also add an additional type `IntS` instead of `int8_t`?

mgovers: interestingly, there is no enum value for 64-bit ctype (`Idx`), only for 32-bit ctype (`ID`). is there any particular reason for that?

TonyXiang8787: We do not have `CType` of 64-bit integer in any attributes. `Idx` is not an attribute!

TonyXiang8787: So we do not need to add `Idx` in this overload set. It is never used.

TonyXiang8787: Define `IntS` is fine.

figueroa1395: All the `is_nan` overloads were taken from `three_phase_tensor.hpp`, and I removed the ones that I thought were not useful. The `Idx` overload sliped my mind, I will remove and add `IntS` as well.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
figueroa1395: I will simplify then. I didn't realice that only `double` is used, hence the template.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
figueroa1395: We don't have any complex number in the C-API boundary indeed, but this is very useful for the validation tests, since we do get the magnitude and angle of certain attributes. In any case, I can remove and just leave the specific overload in the validation tests themselves, but this may be useful for other tests as well.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
figueroa1395: It was done this way so it would also accept arrays of complex values, again, for it usability within the validation tests. I will add a simple overload for complex arrays then. Later if we deem it necessary, I can remove all the complex overloads of this function.

figueroa1395:
I left the complex overloads of `is_nan` on purpose, as they are useful in the validation tests. Let me know if I should remove them from here and make them validation tests exclusive.

Additionally, if any other name came to mind besides `utils`, please do tell.

TonyXiang8787 (approved):

mgovers:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=767) **Quality Gate failed**
> 
> Failed conditions  [70.6% Coverage on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=767&metric=new_coverage&view=list) (required ≥ 80%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=767)

you can ignore this. this is deliberate fallback code

figueroa1395:
@mgovers How do I bypass the CI in this case then?

mgovers:
> @mgovers How do I bypass the CI in this case then?

do nothing because `SonarCloud Code Analysis` is not a required status check. i added the comment for future reference

PR #769, opened 2024-10-08
Repro case for buffer alignment 

mgovers:
Repro case for the bug found and fixed in #718 (specifically https://github.com/PowerGridModel/power-grid-model/pull/718/commits/b3ca98f3a933306efea2afb752d3433aa43b93e8 )

* [x] alignment bug reproed in https://github.com/PowerGridModel/power-grid-model/actions/runs/11232655479/job/31224704499?pr=769 and https://github.com/PowerGridModel/power-grid-model/actions/runs/11233550245/job/31227932484
* [x] successful run after merge in https://github.com/PowerGridModel/power-grid-model/actions/runs/11232745138/job/31225179870?pr=769


TonyXiang8787:
What is the intention of this PR?

The bug can be reliably reproduced in `6099e4712dccb3f5f85f540eaa10935d649c1f47`

build.sh
```diff
@@ -70,8 +74,9 @@ if [[ "${COVERAGE}" ]];  then
     -d ${BUILD_DIR}/tests/cpp_unit_tests/CMakeFiles/power_grid_model_unit_tests.dir \
     -d ${BUILD_DIR}/tests/cpp_integration_tests/CMakeFiles/power_grid_model_integration_tests.dir \
     -d ${BUILD_DIR}/tests/cpp_validation_tests/CMakeFiles/power_grid_model_validation_tests.dir \
-    -d ${BUILD_DIR}/tests/c_api_tests/CMakeFiles/power_grid_model_c_api_tests.dir \
+    -d ${BUILD_DIR}/tests/native_api_tests/CMakeFiles/power_grid_model_api_tests.dir \
     -d ${BUILD_DIR}/power_grid_model_c/power_grid_model_c/CMakeFiles/power_grid_model_c.dir \
+    -d ${BUILD_DIR}/power_grid_model_c_example/CMakeFiles/power_grid_model_c_example.dir \
```
TonyXiang8787, line 79: Do not add example into the coverage report.

mgovers:
> What is the intention of this PR?
> 
> The bug can be reliably reproduced in `6099e4712dccb3f5f85f540eaa10935d649c1f47`

while that is indeed the case, i prefer not to make the validation cases the only location to have tests like this. i prefer to have smaller-sized simple unit tests that test exactly one thing in an isolated environment. IMO, the validation cases are tests that the actual calculation results are correct and not designed to "just test everything there".

Particularly this bug concerns a pitfall that any developer is likely to make more than once in their lifetime, and the fact that it can only be reproduced on one platform is a good indication that we need to be extra careful.

power_grid_model_c/power_grid_model_c/src/buffer.cpp
```diff
@@ -20,15 +20,10 @@ using meta_data::RawDataPtr;
...
-    size_t const alignment = std::max(component->alignment, sizeof(void*));
-    // total bytes should be multiple of alignment
-    size_t const requested_bytes = component->size * size;
-    size_t const rounded_bytes = ((requested_bytes + alignment - 1) / alignment) * alignment;
 #ifdef _WIN32
-    return _aligned_malloc(rounded_bytes, alignment);
+    return _aligned_malloc(component->size * size, component->alignment);
 #else
-    return std::aligned_alloc(alignment, rounded_bytes);
+    return std::aligned_alloc(component->alignment, component->size * size);
```
TonyXiang8787, line 26: I am confused here. Why do we revert the fix?

mgovers: Just to test, nothing more

TonyXiang8787 (changes requested):
I propose to add one additional check in the `buffer.cpp`.

The returned pointer by `alloc`, if it is a `nullptr` and `size > 0`, set an error code in the `handle` with error message `Buffer allocation failed.`.

TonyXiang8787:
> > I propose to add one additional check in the `buffer.cpp`.
> > The returned pointer by `alloc`, if it is a `nullptr` and `size > 0`, set an error code in the `handle` with error message `Buffer allocation failed.`.
> 
> are you sure it's desirable to set error to non-null when the result is nullptr? this is a behavioral change.
> 
> i'm fine with adding it, but i want to double-check that this is indeed desired behavior

You are right. This changes the behavior of the C-API. Let's roll back...

TonyXiang8787 (dismissed):

TonyXiang8787 (dismissed):

TonyXiang8787 (approved):

PR #770, opened 2024-10-08
Bump actions ubuntu 24.04

figueroa1395:
Partially solves #773. Only upgrading the clang-tidy workflow to `ubuntu-24.04` and clang-tidy itself to version 18.1.3 are left; however, discussion on how to resolve each resulting issue of this upgrade is out of scope of this PR.

-----
This in an experiment PR to try to bump to Ubuntu 24.04 in github actions.

In https://github.com/actions/runner-images/issues/10636, it is stated that `ubuntu-latest` won't fully migrate to 24.04 until October 30th, hence the explicit versioning in https://github.com/PowerGridModel/power-grid-model/pull/770/commits/6032f7295ce1bb9898e04f8676620071c4fc9133.

Relevant tools, packages, etc. are updated to the latest supported by this Ubuntu image as stated in https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -66,8 +66,8 @@ jobs:
 
       - name: Install and run clang-format
         run: |
-          sudo apt-get update && sudo apt-get install -y clang-format-15
-          find . -regex '.*\.\(h\|c\|cpp\|hpp\|cc\|cxx\)' -exec clang-format-15 -style=file -i {} \;
+          sudo apt-get update && sudo apt-get install -y clang-format-18
+          find . -regex '.*\.\(h\|c\|cpp\|hpp\|cc\|cxx\)' -exec clang-format-18 -style=file -i {} \;
 
       - name: If needed raise error
         run: |
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -66,8 +66,8 @@ jobs:
 
       - name: Install and run clang-format
         run: |
-          sudo apt-get update && sudo apt-get install -y clang-format-18
-          find . -regex '.*\.\(h\|c\|cpp\|hpp\|cc\|cxx\)' -exec clang-format-18 -style=file -i {} \;
+          sudo apt-get update && sudo apt-get install -y clang-format-15
+          find . -regex '.*\.\(h\|c\|cpp\|hpp\|cc\|cxx\)' -exec clang-format-15 -style=file -i {} \;
 
       - name: If needed raise error
         run: |
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -29,7 +29,7 @@ concurrency:
 jobs:
   check-code-quality:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     steps:
       - name: Checkout
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.12'
 
       - name: Upgrade pip
         run: pip install --upgrade pip
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
+          python-version: '3.10'
 
       - name: Upgrade pip
         run: pip install --upgrade pip
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,7 +11,7 @@ sphinx:
 build:
   os: "ubuntu-24.04"
   tools:
-    python: "3.12"
+    python: "3.11"
   jobs:
     post_install:
       # Build package with doc requirements from pyproject.optional-dependencies
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,7 +9,7 @@ sphinx:
   configuration: docs/conf.py
 
 build:
-  os: "ubuntu-24.04"
+  os: "ubuntu-22.04"
   tools:
     python: "3.11"
   jobs:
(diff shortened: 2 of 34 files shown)


figueroa1395:
Just to keep track of the experiment, after a full migration attempt (Ubuntu 24.04 with clang 18, python 3.12, etc.), there seem to be three major places that would need work:

- [Code quality check](https://github.com/PowerGridModel/power-grid-model/actions/runs/11236794230/job/31237705042?pr=770): Formatting of several files (mainly header files).
- [Clang tidy](https://github.com/PowerGridModel/power-grid-model/actions/runs/11236794224/job/31237705818?pr=770): About 60 errors (from warnings) from all over the core.
- [Validate citations](https://github.com/PowerGridModel/power-grid-model/actions/runs/11236794248/job/31237704644?pr=770): Citations are broken.

In addition, I would expect the nighly build to uncover even more errors from clang-tidy.

TonyXiang8787:
We need to discuss case-by-case about the `clang-tidy` stuff.

figueroa1395:
> We need to discuss case-by-case about the `clang-tidy` stuff.
We can add it to the knowledge sharing session this Thursday (if time allows), so we can all take a look. Should I tell Peter about it?

Side note: Force pushed some commits because I forgot to sign off the ones reverting some changes. Moreover, all changes were reverted (except for only bumping Ubuntu to 24.04), to see what would be the effect of this isolated change in our CI.

figueroa1395:
It seems like the "isolated" experiment of only bumping to Ubuntu 24.04 arguably breaks things even more.

mgovers:
Yeah we definitely need to talk about it then. This will become a blocker on oct 30th

Op 8 okt 2024 17:18 schreef Santiago Figueroa Manrique ***@***.***>:

It seems like the "isolated" experiment of only bumping to Ubuntu 24.04 arguably breaks things even more.

.github/workflows/check-code-quality.yml
```diff
@@ -29,7 +29,7 @@ concurrency:
 jobs:
   check-code-quality:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
TonyXiang8787, line 32: We don't need to do that. Just use `ubuntu-latest`. It is already default `24.04`.

.readthedocs.yaml
```diff
@@ -9,7 +9,7 @@ sphinx:
   configuration: docs/conf.py
 
 build:
-  os: "ubuntu-22.04"
+  os: "ubuntu-24.04"
```
TonyXiang8787, line 12: Don't do this.

ReadTheDocs uses completely different CI than GitHub Actions. We need to see the supported OS version of ReadTheDocs.

TonyXiang8787:
> It seems like the "isolated" experiment of only bumping to Ubuntu 24.04 arguably breaks things even more.

If you only bump Ubuntu 24.04, it still uses `clang-15` compiler but with `gcc-14` C++ std library. This causes the issue with many errors. The error should be less if you bump the clang version to `clang-18`.

.github/workflows/check-code-quality.yml
figueroa1395: I intentionally did this because `ubuntu-latest` is still migrating to 24.04 (see [**target date**](https://github.com/actions/runner-images/issues/10636)) and even though we had `ubuntu-latest` by default, we were running on 22.04 as you can see in this [example](https://github.com/PowerGridModel/power-grid-model/actions/runs/11234754635/job/31231229579).

figueroa1395:
> If you only bump Ubuntu 24.04, it still uses `clang-15` compiler but with `gcc-14` C++ std library. This causes the issue with many errors. The error should be less if you bump the clang version to `clang-18`.

I'll try this and fix the ReadTheDocs version as well. Additionally I may try a fix for the validate-citations action.

mgovers:
> > It seems like the "isolated" experiment of only bumping to Ubuntu 24.04 arguably breaks things even more.
> 
> If you only bump Ubuntu 24.04, it still uses `clang-15` compiler but with `gcc-14` C++ std library. This causes the issue with many errors. The error should be less if you bump the clang version to `clang-18`.

see the first experiment: bumping to clang-18 works but clang-tidy-18 complains. we need to decide on the path going forward and refine - most likely in the afternoon.

> We don't need to do that. Just use ubuntu-latest. It is already default 24.04.

it is released but not yet the default - migration from 22.04 to 24.04 is going gradually, with the migration expected to be finished on 2024 October 30. This experiment mainly checks whether it will work in the first place. Turns out it did for the `check-code-quality.yml` (so we can keep that action at `ubuntu-latest`) but not for the `main.yml` action (so we need to pin and migrate manually)

figueroa1395:
To keep track of the experiment:

- Not upgrading `clang-format` behaves as expected: The [code quality check](https://github.com/PowerGridModel/power-grid-model/actions/runs/11250947912/job/31280918179?pr=770) passes.
- Upgrading only `clang` to version 18.1.3. (but not `clang-tidy`) still [breaks](https://github.com/PowerGridModel/power-grid-model/actions/runs/11250947913/job/31280919270?pr=770) the `clang-tidy` action. However, `sonar-cloud` [doesn't complain](https://github.com/PowerGridModel/power-grid-model/actions/runs/11250947911/job/31280937676?pr=770) anymore.

Following I will upgrade `clang-tidy` (but not `clang-format`) and attempt to fix the failed _validate-citations_ action.

figueroa1395:
Conclusions at this point:

- https://github.com/PowerGridModel/power-grid-model/pull/770/commits/7f8ee76382f4335f8cfe4366e42a937f1c569f9b fixed the [validate-citations action](https://github.com/PowerGridModel/power-grid-model/actions/runs/11251673480/job/31283225521?pr=770).
- When upgrading `clang-tidy` and `clang` to version 18.1.3, the [clang-tidy action](https://github.com/PowerGridModel/power-grid-model/actions/runs/11251673483/job/31283225777?pr=770) breaks, and we need to address it case by case.

Furthermore, an [issue](https://github.com/PowerGridModel/power-grid-model/issues/773) was created as a result of this experiment.

TonyXiang8787:
@figueroa1395  please do a down merge before you run the `clang-format`

mgovers (approved):

figueroa1395:
I will add to auto-merge after the [manual nightly build](https://github.com/PowerGridModel/power-grid-model/actions/workflows/nightly.yml) finishes. Just to double check that these changes are not breaking.

Issue #774, opened 2024-10-09
[FEATURE] Explicitly disallow unsupported data types (undefined behavior)


mgovers:
**Describe the feature request**

Explicitly go from undefined behavior to disallowed behavior when unsupported data formats are provided to the user.

## Background

The `power-grid-model` only supports specific data types for the input, output and update datasets. This is true for both the Python API and the C API, and for both row-based and columnar data. The [`initialize_array`](https://power-grid-model.readthedocs.io/en/stable/api_reference/python-api-reference.html#power_grid_model.initialize_array) and `power_grid_meta_data` (undocumented)

Providing other data types results in undefined behavior (UB) for both the Python API and the C API.

However, the unknowing Python user may not realize that they are doing stuff that is not officially supported, because the current implementation happens to be a copy (because we use the `as_contiguous_array` to get the raw buffers). Even worse: the copy also implicitly casts data to their specified supported `dtype`, potentially losing or even incorrectly truncating data. This erroneous behavior should be explicitly disallowed

**NOTE:** THIS FUNCTIONALITY WAS NEVER OFFICIALLY SUPPORTED.

## Out of scope

If the user provides a dataset type that is equal (`==`) but not equivalent (`is`) to the `power_grid_meta_data` dtype, then a copy is done. While not recommended for performance reasons, this remains supported behavior.

## TODO

Since this is potentially breaking workflows of users that are incorrectly using the power grid model, we should follow a 2-step approach: first deprecating the functionality, and then explicitly disallowing it in 

### Step 1: deprecation

* [x] Add `power_grid_meta_data` to the Python API docs
* [x] In the dataset module, create a check that checks that all the dtypes of all dataset members' `dtype`s are equal (`==`; not `is`) to the respective ones in `power_grid_meta_data`.
  * [x] If not: add a deprecation warning (follow the same approach as for the deprecated functionality in `power_grid_model.utils`)

### Step 2: removal

* [ ] At the next (minor) version bump: throw an error instead of raising a deprecation warning.
* [ ] In the same PR: bump (minor) version
  * [ ] Since the behavior has always been undefined behavior, this is not a breaking change, and it is not required to do a major version bump.

TonyXiang8787:
I thought we agreed to immediately go to step 2. Is that correct? No warnings. Just in the next minor version, a hard error.

nitbharambe:
This and it also helps us separate this task into implementation and version bump.

[...]

PR #776, opened 2024-10-11
Add dtype compatibility check

nitbharambe:
**Fixes issue**: https://github.com/PowerGridModel/power-grid-model/issues/774

Add a warning for dtype check. To be changed to error before minor version bump.


src/power_grid_model/core/power_grid_dataset.py
```diff
@@ -266,6 +273,18 @@ def _validate_properties(self, data: ComponentData, schema: ComponentMetaData):
         if properties.batch_size != self._batch_size:
             raise ValueError(f"Dataset must have a consistent batch size across all components. {VALIDATOR_MSG}")
 
+    def _validate_dtypes_compatibility(self, data: ComponentData, schema: ComponentMetaData):
+        sub_data = _extract_data_from_component_data(data)
+        if is_columnar(data):
+            for attr, array in sub_data.items():
+                if schema.dtype.names is None or attr not in schema.dtype.names:
+                    raise Warning(f"Attribute {attr} is not in schema. {VALIDATOR_MSG}")
+                if array.dtype != schema.dtype[attr]:
```
nitbharambe, line 282: For columnar data, we now have dtype of an asym attribute as dtype("float64"). 
This comes from the fact that we use `np.ascontigiousarray(data, schema_dtype[attribute].base)` for columnar data in raw data view.

Is this the right approach to keep? Will there be any other way to do it? In schema the dtype for attribute is f8 and (3,). ie. including the shape.

nitbharambe, line 282: And this would fix the tests
```suggestion
                if array.dtype != schema.dtype[attr].base:
```

mgovers: i have dived into this some time ago as well. we might indeed need to work around this a little bit. an array with dtype `(f8, (3))` is different than the dtype of a column of a structured array:

```py
>>> from power_grid_model import DatasetType, ComponentType, initialize_array, power_grid_meta_data
>>> a = initialize_array(DatasetType.input, ComponentType.asym_load, 1)
>>> a.dtype.fields["p_specified"]    
(dtype(('<f8', (3,))), 16)
>>> a["p_specified"].dtype
dtype('float64')
>>> power_grid_meta_data[DatasetType.input][ComponentType.asym_load].dtype["p_specified"].base
dtype('float64')
```

the best we can do is check that the last dimension is equal to `3`

even worse (in our case): 
```py
>>> np.array(a["p_specified"], dtype=("<f8", (3,))).shape
(1, 3, 3)
>>> a["p_specified"].shape 
(1, 3)
```

mgovers: as i mentioned in the other comment: this + shape check is the best we can do here

src/power_grid_model/core/power_grid_dataset.py
```diff
@@ -266,6 +273,18 @@ def _validate_properties(self, data: ComponentData, schema: ComponentMetaData):
...
+                    raise Warning(f"Data type for attribute {attr} does not match schema. {VALIDATOR_MSG}")
+        else:
+            if sub_data.dtype != schema.dtype:
+                raise Warning("Data type does not match schema. {VALIDATOR_MSG}")
+
```
mgovers, line 287: this is the worst of both worlds because the user will get an exception (as in the sustainable check) but it's not the exception they expect.

Please have a look at the steps mentioned in #774 (especially about the existing solution used in the deprecated `import/export_*` functionality in `utils`. In particular, use `warnings.warn` with the correct category, rather than `raise` manually

src/power_grid_model/core/power_grid_dataset.py
mgovers, line 1: please also update docs etc. cfr. the steps listed in #774

src/power_grid_model/core/power_grid_dataset.py
```diff
@@ -266,6 +274,24 @@ def _validate_properties(self, data: ComponentData, schema: ComponentMetaData):
...
+                    or attr not in schema.dtype.names
+                    or (schema.dtype[attr].shape == (3,) and array.shape[-1] != 3)
+                ):
+                    raise TypeError("Given data has a different schema than supported.")
+                if array.dtype.base != schema.dtype[attr].base:
+                    warnings.warn(
+                        f"Data type for attribute {attr} does not match schema. {VALIDATOR_MSG}", DeprecationWarning
+                    )
+        else:
+            if sub_data.dtype != schema.dtype:
```
mgovers, line 292:
```suggestion
        elif sub_data.dtype != schema.dtype:
```

src/power_grid_model/core/power_grid_dataset.py
```diff
@@ -266,6 +274,23 @@ def _validate_properties(self, data: ComponentData, schema: ComponentMetaData):
...
+        if is_columnar(data):
+            for attr, array in sub_data.items():
+                if (
+                    schema.dtype.names is None
+                    or attr not in schema.dtype.names
+                    or (schema.dtype[attr].shape == (3,) and array.shape[-1] != 3)
+                ):
+                    raise TypeError("Given data has a different schema than supported.")
+                if array.dtype.base != schema.dtype[attr].base:
+                    warnings.warn(
```
TonyXiang8787, line 288: Do we need such a complicated loop? We can just check it in the data view right?

We check before the line:

https://github.com/PowerGridModel/power-grid-model/blob/f90cf37da600762552b6b028e5bf5cf1beb5aab0/src/power_grid_model/core/buffer_handling.py#L86

By

```python
if data.dtype != dtype:
    warning.warn(...)
```

mgovers:
This PR should not close #774 . That ticket should be closed when v1.10.0 is released

src/power_grid_model/core/buffer_handling.py
```diff
@@ -115,6 +118,8 @@ def _get_raw_attribute_data_view(data: np.ndarray, schema: ComponentMetaData, at
     Returns:
         a raw view on the data set.
     """
+    if schema.dtype[attribute].shape == (3,) and data.shape[-1] != 3:
+        raise TypeError("Given data has a different schema than supported.")
```
mgovers, line 122: Should probably be ValueError

src/power_grid_model/core/buffer_handling.py
```diff
@@ -115,6 +118,8 @@ def _get_raw_attribute_data_view(data: np.ndarray, schema: ComponentMetaData, at
     Returns:
         a raw view on the data set.
     """
+    if schema.dtype[attribute].shape == (3,) and data.shape[-1] != 3:
```
figueroa1395, line 121: May you elaborate on this addition please? Seems confusing to me.

src/power_grid_model/core/buffer_handling.py
nitbharambe:
```suggestion
        raise ValueError("Given data has a different schema than supported.")
```

src/power_grid_model/core/buffer_handling.py
nitbharambe: if its an asym attribute, its shape would be (3,) in the schema. And the condition is to check if it is not equal to the shape of data.

docs/api_reference/python-api-reference.md
```diff
@@ -11,6 +11,7 @@ SPDX-License-Identifier: MPL-2.0
 .. autoclass:: power_grid_model.PowerGridModel
    :show-inheritance:
 .. autofunction:: power_grid_model.initialize_array
+.. autoclass:: power_grid_model.power_grid_meta_data
```
mgovers, line 14: this becomes a mess. i'm working on adding this in #781 instead because the default docstring is unreadable

mgovers (commented):
i think you can remove the `python_grid_meta_data` again, or you'll have to wait a bit for me to find a way to document it in #781

mgovers (approved):

PR #778, opened 2024-10-11
Improvement/Integration tests for columnar data

figueroa1395:
This PR extends the integration tests to columnar based data in the Python side.

PR #779, opened 2024-10-11
Bug/Convert to/form row/col util function

figueroa1395:
Fixes a small bug (found in #778 ) and adds tests for it.

### DIFF
--- a/tests/unit/test_internal_utils.py
+++ b/tests/unit/test_internal_utils.py
@@ -828,9 +828,9 @@ def row_sparse_data():
     transformer["id"] = 1
     transformer["tap_pos"] = 3
 
-    sym_gen = initialize_array(DatasetType.update, ComponentType.sym_gen, (2, 4))
-    sym_gen["id"] = [[5, 6, 7, 8], [5, 6, 7, 8]]
-    sym_gen["q_specified"] = [[1.1, 2.2, 3.3, 4.4], [4.4, 3.3, 2.2, 1.1]]
+    sym_gen = initialize_array(DatasetType.update, ComponentType.sym_gen, 8)
+    sym_gen["id"] = [5, 6, 7, 8, 5, 6, 7, 8]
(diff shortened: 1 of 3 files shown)


tests/unit/test_internal_utils.py
```diff
@@ -803,3 +804,75 @@ def test_get_dataset_type__conflicting_data():
...
+            actual_component = actual_component["data"]
+            desired_component = desired_component["data"]
+        assert actual_component.dtype == desired_component.dtype
+        assert actual_component.shape == desired_component.shape
+
+        for field in actual_component.dtype.names:
+            assert np.all(
+                (np.isnan(actual_component[field]) & np.isnan(desired_component[field]))
+                | (actual_component[field] == desired_component[field])
+            )
```
nitbharambe, line 869: This condition seems sketchy. it shall pass when all nans match.

Try in the way of https://github.com/PowerGridModel/power-grid-model/blob/main/tests/unit/test_serialization.py#L779

tests/unit/test_internal_utils.py
```diff
@@ -803,3 +804,75 @@ def test_get_dataset_type__conflicting_data():
...
+    }
+
+
+@pytest.fixture
+def row_sparse_data():
+    transformer = initialize_array(DatasetType.update, ComponentType.transformer, 1)
+    transformer["id"] = 1
+    transformer["tap_pos"] = 3
+
+    sym_gen = initialize_array(DatasetType.update, ComponentType.sym_gen, (2, 4))
```
nitbharambe, line 831: Sparse data has a SingleArray under data. ie. `sym_gen = initialize_array(DatasetType.update, ComponentType.sym_gen, 8)`

nitbharambe: Can you add an exception for that too under https://github.com/PowerGridModel/power-grid-model/blob/main/src/power_grid_model/_utils.py#L420

TonyXiang8787 (approved):

PR #781, opened 2024-10-14
[Feature] columnar data Python API docs

mgovers:
Docs for columnar data support

Relates to #548 

Preview in https://power-grid-model--781.org.readthedocs.build/en/781/api_reference/python-api-reference.html#power_grid_model.ComponentAttributeMapping


src/power_grid_model/typing.py
```diff
@@ -23,13 +35,14 @@
 
 `ComponentAttributeMapping` can be one of the following:
 
-- A set of `ComponentTypeVar`
+- A set of :class:`ComponentType` or `str`
 
-- A list of `ComponentTypeVar`
+- A list of :class:`ComponentType`
```
nitbharambe, line 40:
```suggestion
- A list of :class:`ComponentType` or `str`
```

nitbharambe (approved):
Tiny modification, rest looks good.

.gitignore
```diff
@@ -513,6 +513,7 @@ PYPI_VERSION
 # docs builds
 docs/make.bat
 docs/Makefile
+docs/jupyter_execute/
```
nitbharambe, line 516: I mentioned command to build as `sphinx-build -b html . _build`. My bad. 
I see using `sphinx-build -b html . _build/html` is more appropriate in which case the jupyter_execute ends up inside _build.
```suggestion
```

docs/api_reference/python-api-reference.md
```diff
@@ -24,18 +32,10 @@ SPDX-License-Identifier: MPL-2.0
...
-.. autoclass:: power_grid_model.data_types.DenseBatchArray
-.. autoclass:: power_grid_model.data_types.SparseBatchArray
-.. autoclass:: power_grid_model.dataset_definitions.DatasetType
-.. autoclass:: power_grid_model.dataset_definitions.ComponentType
-.. autodata:: power_grid_model.typing.ComponentAttributeMapping
-   :annotation: ComponentAttributeMapping
+.. automodule:: power_grid_model.data_types
+   :undoc-members:
+   :show-inheritance:
+   :exclude-members: BatchList,NominalValue,RealValue,AsymValue,AttributeValue,Component,ComponentList,SinglePythonDataset,BatchPythonDataset,PythonDataset
```
figueroa1395, line 38: If we are not to expose these, should we think about moving them to a `_data_types.py` module or something along those lines?

docs/api_reference/python-api-reference.md
```diff
@@ -65,6 +84,8 @@ SPDX-License-Identifier: MPL-2.0
 ```{eval-rst}
 .. autofunction:: power_grid_model.utils.get_dataset_type
 .. autofunction:: power_grid_model.utils.get_dataset_scenario
+.. autofunction:: power_grid_model.utils.get_dataset_batch_size
```
figueroa1395, line 87: Should we also expose here the `compatibility_convert_row_columnar_dataset` function? It has proven to be really valuable for us, I bet it would be for the user.

docs/api_reference/python-api-reference.md
```diff
@@ -75,6 +96,5 @@ SPDX-License-Identifier: MPL-2.0
 .. autofunction:: power_grid_model.utils.msgpack_serialize_to_file
 .. autofunction:: power_grid_model.utils.import_json_data
 .. autofunction:: power_grid_model.utils.export_json_data
-.. autofunction:: power_grid_model._utils.get_and_verify_batch_sizes
-.. autofunction:: power_grid_model._utils.get_batch_size
```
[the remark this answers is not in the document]
figueroa1395: Maybe then a good idea to explicitly comment about this in the official release notes, similarly to how it was done for 1.9. This change may be breaking in the future only for people who are using it incorrectly, but a heads up would still be nice.

figueroa1395 (commented):
Looks good to me. Only very minor comments.

figueroa1395 (dismissed):
I forgot to approve before. Here it is.

docs/api_reference/python-api-reference.md
mgovers: then we should come up with a better name. any suggestions?

figueroa1395: Definetely needs a better name, but also probably needs to be discussed on the knowledge sharing (?). As for names, let me think, nothing pops to mind at the moment. Maybe we can just scope it out for know and tell Peter to add it to some agenda to discuss it with the team.

docs/api_reference/python-api-reference.md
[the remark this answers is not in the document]
mgovers: those symbols were already weren't part of the API reference anymore in `v1.9` (and even before that). no need to mention in the release notes.

.gitignore
mgovers: reverting revert. no need to be bothered with details

docs/api_reference/python-api-reference.md
mgovers: let's discuss later and scope out for now idd

docs/api_reference/python-api-reference.md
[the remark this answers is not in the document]
mgovers: EDIT: nvm. we can just add to the release notes: fixes and improvements to our documentation

figueroa1395 (approved):

PR #787, opened 2024-10-15
Feature/private core module

### LINKED ISSUE #784: [FEATURE] make python API core functionality private

Currently, the functionality in `src/power_grid_model/core` is not considered stable. However, it may not be directly clear to the user that that is the case. To show the intentionality, the core module should be marked as a private module by prefixing `_`, i.e., `src/power_grid_model/_core`.

However, because of the intentionality issue, users may not have realized that we do not consider functionality in `core` stable, so as an intermediate step, we should still expose the functions currently listed in `core` in the same way - otherwise, we may break users' workflow, even if it was never intended to be used that way.

## TODO

* [x] move `src/power_grid_model/core` to `src/power_grid_model/_core`
* [x] make a new module `src/power_grid_model/core` that contains the same files as `src/power_grid_model/_core` that import all existing symbols from the `_core` module.
  * [x] this ensures that new functionality is not exposed but existing functionality is not (yet) broken
* [x] ensure all other modules and tests import from `_core` (not from `core`)

mgovers:
Closes #784 

## Notes to reviewer:

review commit-by-commit. that's much easier

* [ ] reviewed [3aa2ee5](https://github.com/PowerGridModel/power-grid-model/pull/787/commits/3aa2ee56ee956eede2ad7d301d6ddc10dabbf0ee)
* [ ] reviewed [6ca7e5c](https://github.com/PowerGridModel/power-grid-model/pull/787/commits/6ca7e5cc775f1eee863ca74225c3a4df8c750db1)
* [ ] reviewed [ee1cdf9](https://github.com/PowerGridModel/power-grid-model/pull/787/commits/ee1cdf9d2f83fb2bcbc3ed78409a4291bcea6679)
* [ ] reviewed any newer commits


docs/examples/Validation Examples.ipynb
```diff
@@ -195,11 +195,11 @@
...
-      "   2. Field 'to_node' does not contain a valid node id for 1 line.\n",
-      "   3. Field 'power_sigma' is not greater than zero for 2 sym_power_sensors.\n",
-      "   4. Field 'measured_object' does not contain a valid line/transformer id for 1 sym_power_sensor. (measured_terminal_type=branch_from)\n",
-      "   5. Field 'measured_object' does not contain a valid source id for 1 sym_power_sensor. (measured_terminal_type=source)\n"
+      "   1. Fields ComponentType.line.id and ComponentType.sym_power_sensor.id are not unique for 2 ComponentsTypes.lines/ComponentsTypes.sym_power_sensors.\n",
+      "   2. Field 'to_node' does not contain a valid node id for 1 ComponentType.line.\n",
+      "   3. Field 'power_sigma' is not greater than zero for 2 ComponentsTypes.sym_power_sensors.\n",
+      "   4. Field 'measured_object' does not contain a valid line/generic_branch/transformer id for 1 ComponentType.sym_power_sensor. (measured_terminal_type=branch_from)\n",
+      "   5. Field 'measured_object' does not contain a valid source id for 1 ComponentType.sym_power_sensor. (measured_terminal_type=source)\n"
```
figueroa1395: I know you refer to components here in plural, but using `ComponentTypes` (with an `s`) may lead to confusion. The same with the names of the actual components, such as `lines`, `sym_power_sensors`, etc. It also happens in other places.

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -11,25 +11,25 @@
...
 )
-from power_grid_model.core.error_handling import PowerGridBatchError, assert_no_error, handle_errors
-from power_grid_model.core.index_integer import IdNp, IdxNp
-from power_grid_model.core.options import Options
-from power_grid_model.core.power_grid_core import ConstDatasetPtr, IDPtr, IdxPtr, ModelPtr, power_grid_core as pgc
+from power_grid_model._core.error_handling import PowerGridBatchError, assert_no_error, handle_errors
+from power_grid_model._core.index_integer import IdNp, IdxNp
+from power_grid_model._core.options import Options
+from power_grid_model._core.power_grid_core import ConstDatasetPtr, IDPtr, IdxPtr, ModelPtr, power_grid_core as pgc
+from power_grid_model._utils import compatibility_convert_row_columnar_dataset
```
figueroa1395, line 32: Here and before: Just out of curiosity, why was `power_grid_model._utils`'s position changed?

figueroa1395 (commented):
Reviewed [3aa2ee5](https://github.com/PowerGridModel/power-grid-model/pull/787/commits/3aa2ee56ee956eede2ad7d301d6ddc10dabbf0ee). Only minor comments.

src/power_grid_model/core/__init__.py
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+"""
+Power Grid Model core library
+
+[deprecated]
+"""
```
figueroa1395, line 9: Shouldn't we expose here the (unintentionally previously exposed) core functionalities as well? Or we leave this blank because, as before, the user has to manually import the specific module from core that is wanted?

figueroa1395 (commented):
Reviewed [6ca7e5c](https://github.com/PowerGridModel/power-grid-model/pull/787/commits/6ca7e5cc775f1eee863ca74225c3a4df8c750db1): Looks good, the only worry I have is that since this task can be very cherry-pickish (We have to cherry pick all, but still...), I hope we don't miss exposing one functionality from the core and hence break anyones workflow.

figueroa1395 (commented):
Reviewed [ee1cdf9](https://github.com/PowerGridModel/power-grid-model/pull/787/commits/ee1cdf9d2f83fb2bcbc3ed78409a4291bcea6679): All looks good, code generation check comming to the rescue here! ;)

TonyXiang8787:
Do we really need this? I propose to change it to `_core` immediately. No warnings, just error.

When we released PGM, we made it very clear that Python stuff which is not documented in the read the docs is not public API and not supported.

src/power_grid_model/_core/power_grid_model.py
nitbharambe: @figueroa1395 The imports are sorted alphabetically

mgovers:
> Do we really need this? I propose to change it to `_core` immediately. No warnings, just error.
> 
> When we released PGM, we made it very clear that Python stuff which is not documented in the read the docs is not public API and not supported.

i'm all for that but it may break users' code. i propose to remove this (temp) library at the same time when we actually do the version bump, so that users can specify that. let's discuss tomorrow @petersalemink95

docs/examples/Validation Examples.ipynb
figueroa1395: I think you answered another question here (?)

mgovers: yes, apologies. i thought something went wrong when i clicked submit but your comment is correct indeed.

src/power_grid_model/core/__init__.py
mgovers: this was already blank before, so there's no risk here. manual import, it is

mgovers:
cfr. discussions: this will be merged in the same PR that will bump the version to `v1.10.0`

figueroa1395 (approved):
Re re-ran all the examples and checked the changes again. All looks good.

PR #799, opened 2024-10-21
improve C API documentation

mgovers:
Very small improvements to C API documentation regarding columnar data.

Relates to #548

### DIFF
--- a/power_grid_model_c/power_grid_model_c/include/power_grid_model_c/buffer.h
+++ b/power_grid_model_c/power_grid_model_c/include/power_grid_model_c/buffer.h
@@ -28,7 +28,7 @@ extern "C" {
  * It is recommended to call PGM_buffer_set_nan() after you create an input or update buffer.
  * In this way all the attributes will be set to NaN.
  * Additionally, if there is a new optional attribute added in the future,
- * compatibility with your code is guaranteed: that optional attribute will be set to NaN 
+ * compatibility with your code is guaranteed: that optional attribute will be set to NaN
  * and the default value will be used.
  *
  * @param handle
--- a/power_grid_model_c/power_grid_model_c/include/power_grid_model_c/buffer.h
+++ b/power_grid_model_c/power_grid_model_c/include/power_grid_model_c/buffer.h
@@ -28,8 +28,8 @@ extern "C" {
  * It is recommended to call PGM_buffer_set_nan() after you create an input or update buffer.
  * In this way all the attributes will be set to NaN.
  * Additionally, if there is a new optional attribute added in the future,
- * compatibility with your code is guaranteed,
- * because that optional attribute will be set to NaN and the default value will be used.
+ * compatibility with your code is guaranteed: that optional attribute will be set to NaN 
+ * and the default value will be used.
  *
  * @param handle
  * @param component A component pointer.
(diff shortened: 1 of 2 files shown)


power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -140,7 +140,7 @@ PGM_API void PGM_destroy_dataset_const(PGM_ConstDataset* dataset);
  *     If the component is not uniform, indptr must point to an array of size (batch_size + 1).
  *         The values in the array must be not decreasing.
  *         And we must have indptr[0] = 0, indptr[batch_size] = total_elements.
- * @param data A void pointer to the row based buffer data or a nullptr for columnar data.
+ * @param data A void pointer to the row based buffer data or NULL for columnar data.
```
mgovers, line 143: C (until C23) has no `nullptr`

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/buffer.h
```diff
@@ -20,15 +20,15 @@ extern "C" {
...
+ * with size and alignment obtained from PGM_meta_component_size() and PGM_meta_component_alignment().
  * The buffer created by this function should be freed by PGM_destroy_buffer().
  *
  * It is recommended to call PGM_buffer_set_nan() after you create an input or update buffer.
  * In this way all the attributes will be set to NaN.
- * And if there is a new optional attribute added in the future.
- * You have garantee that your code is still compatible
+ * Additionally, if there is a new optional attribute added in the future,
+ * compatibility with your code is guaranteed,
  * because that optional attribute will be set to NaN and the default value will be used.
```
Jerry-Jinfeng-Guo, line 32:
```suggestion
 * compatibility with your code is guaranteed: that optional attribute will be set to NaN 
 * and the default value will be used.
```

mgovers:
> While going through this PR, two questions were raised:
> 
>     * Since we don't have a create buffer function for columnar buffers (since probably there is no value in it due to the nature of columnar data), we should document this somewhere as there is an "inconsistency" on how row and columnar data are managed at the C API level. Or at least those are my thought, it may lead to confusion from the user perspective.

I've also been thinking about that, but I don't know where to put it. Do you have a good proposal?

>     * Should we (at some point) make something like `is_columnar` available through the C API? It may not be needed, but given that we offer functions to check if something is batch and its size, it would probably go along with those (and may even provide value in the python side).

I think that makes sense, but maybe out of scope of `v1.10.0`. @TonyXiang8787 do you have any thoughts on this?

Jerry-Jinfeng-Guo (approved):
Ongoing discussions are deemed out of scope of this PR. Please create a dedicated thread to discuss them further.

figueroa1395:
> > While going through this PR, two questions were raised:
> > 
> > * Since we don't have a create buffer function for columnar buffers (since probably there is no value in it due to the nature of columnar data), we should document this somewhere as there is an "inconsistency" on how row and columnar data are managed at the C API level. Or at least those are my thought, it may lead to confusion from the user perspective.
> 
> I've also been thinking about that, but I don't know where to put it. Do you have a good proposal?

Not sure yet, it probably makes sense to directly comment something about it under `PGM_create_buffer`, but it should probably be standalone somewhere else.

mgovers:
continuing thread in https://github.com/PowerGridModel/power-grid-model/issues/548#issuecomment-2428421384

PR #807, opened 2024-10-25
Feature/improve docs

mgovers:
* Extend C API buffer usage documentation
* Explicitly document behavior after error during permanent update


docs/advanced_documentation/c-api.md
```diff
@@ -69,45 +69,86 @@ In the `PGM_calculate` function you need to pass a pointer to `PGM_Options`.
...
+For compatibility reasons, that format is dictated by the C API using the `PGM_meta_*` functions and `PGM_def_*` dataset definitions.
 
-The biggest challenge in the design of C API is the handling of input/output/update data buffers.
 We define the following concepts in the data hierarchy:
 
 * Dataset: a collection of data buffers for a given purpose. 
-At this moment, we have four dataset types: `input`, `update`, `sym_output`, `asym_output`.
-* Component: a homogeneous data buffer for a component in our [data model](../user_manual/components.md), e.g., `node`.
+  At this moment, we have four dataset types: `input`, `update`, `sym_output`, `asym_output`.
+* Component: the representation of attributes for the same component in our [data model](../user_manual/components.md), e.g., `node`.
```
figueroa1395, line 82: I don't think I like the wording here, it substracts importance to the Component itself and doesn't tie well with the Dataset definition. What about something like "A data buffer for a physical  grid components, characterized by its attributes, in our..."

docs/advanced_documentation/c-api.md
```diff
@@ -134,6 +175,16 @@ You do not need to call `PGM_buffer_set_nan` on output buffers,
...
 
+### Attribute buffers
+
+An attribute buffer contains data for a single component attribute and represents one column in a columnar representation of component data.
+A combination of attribute buffers with the same amount of elements has the power to carry the same information as row-based [component buffers](#component-buffers).
+
+The type (implying the size) of each attribute can be found using the `PGM_meta_attribute_ctype`.
+
+Since all attributes consist of primitive types, operations are straightforward.
+We therefore do not provide explicit interface functionality.
```
figueroa1395, line 186: Maybe it is a good idea to briefly mention something about the `*set_attribute_buffer()` functionality, or point towards it. Even thought this is more related with the datasets, this is basically the only way to "interact" with "attribute buffers". What do you think?

figueroa1395 (commented):
Looks very good, just a couple of minor comments.

docs/advanced_documentation/c-api.md
mgovers: Can you add that? Specifically below in the section on the dataset

mgovers: Actually, i completely forgot that part so feel free to update there

nitbharambe (approved):

PR #813, opened 2024-10-29
Feature / row based optional id field support in update data 

Jerry-Jinfeng-Guo:
This PR implements the optional id support for row based update data.

- [x] Internal logic to support optional `id` for row based update data
- [ ] From `is_update_independent` to `get_sequence_idx_map`
* [ ] Tests
  * [ ] API `test_api_model.cpp`
  * [ ] Integration `test_main_model.cpp`
  * [ ] Python

Nothing to see here yet.


mgovers (dismissed):
nice. let's indeed merge this into nitish' PR that re-adds the tests and then we can verify that everything works correctly and it should be good to go 🎉

tests/native_api_tests/test_api_model.cpp
```diff
@@ -494,14 +494,23 @@ TEST_CASE("API Model") {
         update_sym_load_buffer.set_value(PGM_def_update_sym_load_id, update_sym_load_id.data(), -1);
         update_sym_load_buffer.set_value(PGM_def_update_sym_load_q_specified, update_sym_load_q_specified.data(), -1);
 
+        // update dataset buffers - no ids
+        Buffer update_source_buffer_no_id{PGM_def_update_source, 2};
+        update_source_buffer_no_id.set_nan();
+        update_source_buffer_no_id.set_value(PGM_def_update_source_u_ref, update_source_u_ref.data(), -1);
+
+        Buffer update_sym_load_buffer_no_id{PGM_def_update_sym_load, 2};
+        update_sym_load_buffer_no_id.set_nan();
```
figueroa1395, line 503: Nitpick: I know it's good practice, but should we also test without setting first to `nan`? It should also work as intended, right?

figueroa1395 (commented):
One additional thing, integration tests for row based are still missing I believe. Just to be completely thorough.

figueroa1395:
> > One additional thing, integration tests (on c++) for row based are still missing I believe. Just to be completely thorough, as they are not in #817.
> 
> i don't think we need to dig too deeply into this. we're not going to support integration tests anyway in the long run. it's fine as is

I mention it because they are already in place for columnar, but you are right. Then the those can be ignored.

mgovers (approved):

PR #771, opened 2024-10-08
Feature / columnar data optional id field support in update data

Jerry-Jinfeng-Guo:
This PR implements step 7 of https://github.com/PowerGridModel/power-grid-model/issues/548

- [x] Validation on the combinations of different situations of `id`
- [x] Functionalities to generate sequential `id` for qualified input 
- [x] Update columnar component with generated sequence `id`
* [x] Basic tests
  * [x] Integration tests
  * [x] API tests
  * [x] Updated wrong `id` tests
  * [x] Python tests
* [x] Extensive tests ??
  * [x] Unit tests for new functionality
  * [x] ~~Validation for input without `id`~~
- [x] AddressSanitizer [https://github.com/PowerGridModel/power-grid-model/pull/810]
- [x] Last but not least: resolve merge conflicts with `main`
- [x] ~~Refactor the code to live outside `main_model_impl.hpp`~~

PR #814, opened 2024-10-30
Feature/Update documentation for optional update IDs

### LINKED ISSUE #548: [FEATURE] Add support columnar data buffer to save memory usage

figueroa1395:
Adds documentation for #771 and #813.

Closes #548 

Todo:

- [x] Update examples for row based data with optional update id (After #813 is ready. May be a good idea to rebase with that branch at that point.)
- [x] Update examples for columnar based data with optional update id
- [x] Add optional id documentation
- [ ] [Optional] Clean metadata


docs/examples/Asymmetric Calculation Example.ipynb
```diff
@@ -829,7 +825,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.13.0"
+   "version": "3.10.14"
```
mgovers, line 828: please use a more recent version. we will not support `python3.10` from next January onward and it's good to live at head (idem for the other notebooks)

docs/examples/Asymmetric Calculation Example.ipynb
```diff
@@ -46,11 +46,7 @@
    "source": [
     "# some basic imports\n",
     "import numpy as np\n",
-    "import warnings\n",
-    "\n",
-    "with warnings.catch_warnings(action=\"ignore\", category=DeprecationWarning):\n",
-    "    # suppress warning about pyarrow as future required dependency\n",
-    "    import pandas as pd\n",
```
mgovers, line 53: oh was this issue with pandas fixed? 😮

docs/examples/Power Flow Example.ipynb
```diff
@@ -663,6 +659,33 @@
...
+   "id": "7eabed0c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "line_update_dtype = power_grid_meta_data[DatasetType.update][ComponentType.line].dtype\n",
+    "columnar_no_ID_update_line = {\n",
+    "    # Update IDs are not specified\n",
+    "    \"from_status\": np.array(\n",
+    "        [0, 1, 1], dtype=line_update_dtype[\"from_status\"]\n",
+    "    ),  # The update for the whole column needs to be specified\n",
```
mgovers, line 682: it's probably good to also show what happens if you don't do that (or explicitly mention that)

figueroa1395:
Documentation and the columnar-data related example can be reviewed already. Meta data will be cleaned up at the end if necessary, but feel free to already point things out.

docs/examples/Validation Examples.ipynb
```diff
@@ -127,23 +127,10 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
```
mgovers, line 130: unfortunately, the validation examples need to be run entirely manually (or run with `--continue-on-error` or something similar).

docs/user_manual/components.md
```diff
@@ -18,7 +18,7 @@ The base type for all power-grid-model components.
...
 | ---- | --------- | ---- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- | :------: | :----------------------------------------------------------------------------: |
-| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (id needs to be specified in the update query, but cannot be changed) |
+| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (If a component update is uniform, the ID can be omitted or set to `nan`. Otherwise, the id needs to be specified in the update query, but cannot be changed) |
```
mgovers, line 21: i don't know whether a user understands what `uniform` means if you put it in the components like this.

you can also just say something like `(see below)` and then have a note block elaborate on the topic

docs/examples/Make Test Dataset.ipynb
mgovers, line 1: i don't think it's necessary to rerun all notebooks if there are no changes; especially not if it's a leap back in time 😬

docs/examples/Asymmetric Calculation Example.ipynb
figueroa1395: Yup, no I don't get an issue anymore, so we might as well remove it altogether.

docs/user_manual/components.md
figueroa1395: Good idea, I will add a block further explaining below.

docs/examples/Make Test Dataset.ipynb
figueroa1395: I may have accidentally done it when removing the Pandas warning, but in the end I could revert that specific notebook's metadata if needed. I will leave this open for now as a reminder in the end.

docs/examples/Validation Examples.ipynb
figueroa1395: I didn't notice this, good catch, I will fix it.

docs/examples/Asymmetric Calculation Example.ipynb
figueroa1395: I just tend to run things with the minimum supported version myself as a sort of safe guard. But I will definitely update then.

docs/user_manual/components.md
```diff
@@ -18,7 +18,9 @@ The base type for all power-grid-model components.
...
-| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (id needs to be specified in the update query, but cannot be changed) |
+| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (If a component update is uniform (See below), the ID can be omitted or set to `nan`. Otherwise, the id needs to be specified in the update query, but cannot be changed) |
+
+A component update is uniform when the number of elements is consistent across all scenarios and matches the number of elements in the input data for that component. Therefore, the ID is not optional for updates on sparse data, and when the ID of a component is ommited, the updated field must be explicit across all scenarios (even if it is not always updated). An example of the usage of optional IDs is given in [Power Flow Example](./Power%20Flow%20Example.ipynb)
```
Jerry-Jinfeng-Guo, line 23: > is uniform when the number of elements is consistent across all scenarios and matches the number of elements in the input data

Not sure if this **was** the case. Updates can all be updating to the same components but not all the components. 

> Therefore, the ID is not optional for updates on sparse data, and when the ID of a component is ommited

Being `uniform` and being `sparse` are two orthogonal concepts I believe. @mgovers could you confirm this?


> the updated field must be explicit across all scenarios (even if it is not always updated).

I do not quite understand this line

Jerry-Jinfeng-Guo: Oh, it's not yet in review. My apologies.

mgovers: > Being uniform and being sparse are two orthogonal concepts I believe. @mgovers could you confirm this?

see my post in teams in which i elaborate the existing conventions used in the docs 😬 if we can make clear boundaries on terminology, we may resolve this more straightforwardly

mgovers: short version: there's a difference between the API reference and the implementation in main model and we need to agree on one set of conventions

figueroa1395: > > is uniform when the number of elements is consistent across all scenarios and matches the number of elements in the input data
> 
> Not sure if this **was** the case. Updates can all be updating to the same components but not all the components.

Maybe my explanation wasn't clear, but this intended to reflect the fact that even if you are updating one line (out of 3 for instance), you still need to provide the whole column of the updated attribute in the scenario where it is updated. My intention was to "demonstrate" this in the example notebook.

> > Therefore, the ID is not optional for updates on sparse data, and when the ID of a component is ommited
> 
> Being `uniform` and being `sparse` are two orthogonal concepts I believe. @mgovers could you confirm this?

I see this now, I was following the documentation rather than the code logic. I will update.

> > the updated field must be explicit across all scenarios (even if it is not always updated).
> 
> I do not quite understand this line

I had a hard time trying to write this down concisely and clearly, but it looks like I failed. The line tries to explain what I replied on the first line above.

> short version: there's a difference between the API reference and the implementation in main model and we need to agree on one set of conventions

We can discuss tomorrow, but I believe we should go with the code here.

Please do give me suggestions if you feel I could have written something better, specially regarding the uniform part.

figueroa1395: I have adjusted in [5bc7cf5](https://github.com/PowerGridModel/power-grid-model/pull/814/commits/5bc7cf5b7b6a24ef5b5540d85703f61db432e830) according to the team discussions about the definitions.

docs/user_manual/components.md
```diff
@@ -18,7 +18,9 @@ The base type for all power-grid-model components.
...
 | ---- | --------- | ---- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- | :------: | :----------------------------------------------------------------------------: |
-| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (id needs to be specified in the update query, but cannot be changed) |
+| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (If a component update is uniform (see below), the ID can be omitted or set to `nan`. Otherwise, the id needs to be specified in the update query, but cannot be changed) |
```
Jerry-Jinfeng-Guo, line 21:
```suggestion
| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (if a component update is uniform (see below) and is updating all the elements with the same component type, IDs can be omitted or set to `nan`. In any other case, IDs need to be present in the update dataset for reference) |
```

docs/user_manual/components.md
```diff
@@ -18,7 +18,9 @@ The base type for all power-grid-model components.
...
+
+A uniform component update is defined as one in which every scenario modifies the same number of elements within that component. Therefore, a dense update is always uniform, hence the ID can always be optional; whereas a sparse update is only uniform if the index pointer is given as a linear sequence of elements (sparse representation of a dense buffer). An example of the usage of optional IDs is given in [Power Flow Example](./Power%20Flow%20Example.ipynb)
```
Jerry-Jinfeng-Guo, line 23:
```suggestion
Uniform component updates ones that update the same number of component elements across scenarios. By definition, a dense update is always uniform, hence the ID can always be optional; ...
```

``` a sparse update is only uniform if the index pointer is given as a linear sequence of elements```

I don't think this is necessarily true. Sequence `{1, 2, 3}` is linear, just like `{2, 4, 6}` is linear, and `{6, 4, 2}`.

File: docs/examples/Power Flow Example.ipynb [hunk 9]
```diff
@@ -638,17 +634,42 @@
...
    "metadata": {},
    "source": [
-    "Columnar data is also supported"
+    "Columnar data is also supported."
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 15,
+   "execution_count": null,
```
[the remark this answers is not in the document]
figueroa1395: Fixed it.

docs/user_manual/components.md
Jerry-Jinfeng-Guo: Maybe consider:
``` suggestion
... whereas a sparse update is only uniform if the index pointer is given as an arithmetic sequence of all elements (sparse representation of a dense buffer). An example of the usage of optional IDs is given in [Power Flow Example](./Power%20Flow%20Example.ipynb)
```

docs/examples/Power Flow Example.ipynb
```diff
@@ -660,7 +681,34 @@
...
-    "update_data = {ComponentType.sym_load: update_sym_load, ComponentType.line: columnar_update_line}"
+    "update_data_col = {ComponentType.sym_load: update_sym_load, ComponentType.line: columnar_update_line}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f5179288",
+   "metadata": {},
+   "source": [
+    "Optional IDs for an update in columnar-data format are also supported for uniform component updates."
```
mgovers, line 692: i'd inverse this statement:

```suggestion
    "If you are updating all components of the same type (uniformly) in the same order as in the input data, providing IDs is optional."
```

docs/user_manual/components.md
mgovers: i feel like this is too long. it's better to divert as soon to the comment below as possible and to keep the table as brief as possible

mgovers, line 21: the unicode for update &#10060; needs to be adjusted

Jerry-Jinfeng-Guo:
```suggestion
| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; See below for optional IDs. |
```

figueroa1395: Deemed correct after discussion.

mgovers (dismissed):

Jerry-Jinfeng-Guo (approved):

Jerry-Jinfeng-Guo (dismissed):
The automatic rebase to main didn't happen. Hence this manual op.

mgovers (dismissed):

docs/user_manual/components.md
```diff
@@ -18,7 +18,11 @@ The base type for all power-grid-model components.
...
+| `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (see below) |
+
+If a component update is uniform and is updating all the elements with the same component type, IDs can be omitted or set to `nan`s. In any other case, the IDs need to be present in the update dataset, but cannot be changed.
+
+Uniform component updates are ones that update the same number of component elements across scenarios. By definition, a dense update is always uniform, hence the IDs can always be optional; whereas for a sparse update, the IDs can only be optional if the index pointer is given as an arithmetic sequence of all elements (i.e. a sparse representation of a dense buffer). An example of the usage of optional IDs is given in [Power Flow Example](./Power%20Flow%20Example.ipynb)
```
mgovers, line 25: there is a bug when dealing with mixed optional ID, e.g.:

```py
update_sym_load_no_id = initialize_array(DatasetType.update, ComponentType.sym_load, (3, 2))
update_sym_load_no_id["p_specified"] = [30e6, 15e6]  # broadcast

update_line = initialize_array(DatasetType.update, ComponentType.line, (3, 1))
update_line["id"] = [[3], [5], [8]]  # broadcast
update_line["from_status"] = [0]

update_data_no_id = {ComponentType.sym_load: update_sym_load_no_id, ComponentType.line: update_line}
```

this will pass `assert_valid_batch_data` but will still throw `PowerGridBatchErrors`:
```
[IDNotFound('The id cannot be found: -2147483648\n'), IDNotFound('The id cannot be found: -2147483648\n'), IDNotFound('The id cannot be found: -2147483648\n')]
```

figueroa1395:
Only merge after #819 and #820 are merged.

docs/user_manual/components.md
mgovers: resolved in #819

mgovers, line 25:
```suggestion
Uniform component updates are ones that update the same number of component of the same type across scenarios. By definition, a dense update is always uniform, hence the IDs can always be optional; whereas for a sparse update, the IDs can only be optional if the index pointer is given as an arithmetic sequence of all elements (i.e. a sparse representation of a dense buffer). An example of the usage of optional IDs is given in [Power Flow Example](./Power%20Flow%20Example.ipynb)
```

mgovers (approved):

PR #817, opened 2024-11-04
Optional IDs tests

nitbharambe:
Tests for https://github.com/PowerGridModel/power-grid-model/pull/813

- Minor refactoring in a0b2c09
- C++ Tests added in 9e24f85
-  Tests for optional ids removed e5d5670 as there is no way to skip or xfail specific params. The commit can be reverted when those tests are needed.
- Python tests added. (Remove xfail for new features)
- Tests for buffer handling

### Additional points

- The location doesn't seem right. Maybe we should only keep errors here and rest to main model tests. 
- The `self contained model update error` seems to repeated.

### DIFF
--- a/src/power_grid_model/core/buffer_handling.py
+++ b/src/power_grid_model/core/buffer_handling.py
@@ -245,7 +245,7 @@ def _get_sparse_buffer_properties(
             if attribute_data.ndim != ndim + schema.dtype[attribute].ndim or attribute_data.shape[:ndim] != shape:
                 raise ValueError(f"Data buffers must be consistent. {VALIDATOR_MSG}")
 
-    contents_size = shape[0] if len(shape) == 1 else shape[0] * shape[1]
+    contents_size = shape[0]
     check_indptr_consistency(indptr, batch_size, contents_size)
 
     is_batch = True
--- a/src/power_grid_model/core/buffer_handling.py
+++ b/src/power_grid_model/core/buffer_handling.py
@@ -136,7 +136,7 @@ def _get_indptr_view(indptr: np.ndarray) -> IdxPtr:  # type: ignore[valid-type]
     return np.ascontiguousarray(indptr, dtype=IdxNp).ctypes.data_as(IdxPtr)
 
 
-def _get_uniform_buffer_properties(
+def _get_dense_buffer_properties(
     data: ComponentData,
     schema: ComponentMetaData,
     is_batch: bool | None,
@@ -284,7 +284,7 @@ def get_buffer_properties(
         the properties of the dataset component.
     """
     if not is_sparse(data):
-        return _get_uniform_buffer_properties(data=data, schema=schema, is_batch=is_batch, batch_size=batch_size)
+        return _get_dense_buffer_properties(data=data, schema=schema, is_batch=is_batch, batch_size=batch_size)
 
     if is_batch is not None and not is_batch:
         raise ValueError("Sparse data must be batch data")
@@ -334,7 +334,7 @@ def _get_uniform_buffer_view(
     Returns:
         the C API buffer view.
     """
-    properties = _get_uniform_buffer_properties(data, schema=schema, is_batch=is_batch, batch_size=batch_size)
+    properties = _get_dense_buffer_properties(data, schema=schema, is_batch=is_batch, batch_size=batch_size)
 
     return CBuffer(
         data=_get_raw_component_data_view(data=data, schema=schema),
--- a/src/power_grid_model/core/buffer_handling.py
+++ b/src/power_grid_model/core/buffer_handling.py
@@ -245,7 +245,7 @@ def _get_sparse_buffer_properties(
             if attribute_data.ndim != ndim + schema.dtype[attribute].ndim or attribute_data.shape[:ndim] != shape:
                 raise ValueError(f"Data buffers must be consistent. {VALIDATOR_MSG}")
 
-    contents_size = sum(shape)
+    contents_size = shape[0] if len(shape) == 1 else shape[0] * shape[1]
     check_indptr_consistency(indptr, batch_size, contents_size)
 
     is_batch = True
--- a/tests/native_api_tests/test_api_model.cpp
+++ b/tests/native_api_tests/test_api_model.cpp
@@ -538,4 +538,142 @@ TEST_CASE("API Model") {
     }
 }
 
+/*
+
+source_1 -- node_0 --load_2
+
+source and node inputs are row based.
+load input is either row based or columnar.
+load update is row based / columnar and densse / sparse.
+
+invalid_id_t tests are for testing the error handling of the model when the id is not found in the update dataset.
+optional_id_t tests are for testing the model when the id is not added to the update dataset.
+
+*/
+TEST_CASE_TEMPLATE(
+    "API update id tests", T, TypeCombo<row_t, row_t, dense_t, with_id_t>, TypeCombo<row_t, row_t, sparse_t, with_id_t>,
+    TypeCombo<columnar_t, columnar_t, dense_t, with_id_t>, TypeCombo<columnar_t, columnar_t, sparse_t, with_id_t>,
+    TypeCombo<columnar_t, row_t, dense_t, with_id_t>, TypeCombo<columnar_t, row_t, sparse_t, with_id_t>,
+    TypeCombo<row_t, columnar_t, dense_t, with_id_t>, TypeCombo<row_t, columnar_t, sparse_t, with_id_t>,
+    TypeCombo<row_t, row_t, dense_t, optional_id_t>, TypeCombo<row_t, row_t, sparse_t, optional_id_t>,
+    TypeCombo<columnar_t, columnar_t, dense_t, optional_id_t>,
+    TypeCombo<columnar_t, columnar_t, sparse_t, optional_id_t>, TypeCombo<columnar_t, row_t, dense_t, optional_id_t>,
+    TypeCombo<columnar_t, row_t, sparse_t, optional_id_t>, TypeCombo<row_t, columnar_t, dense_t, optional_id_t>,
+    TypeCombo<row_t, columnar_t, sparse_t, optional_id_t>, TypeCombo<row_t, row_t, dense_t, invalid_id_t>,
+    TypeCombo<row_t, row_t, sparse_t, invalid_id_t>, TypeCombo<columnar_t, columnar_t, dense_t, invalid_id_t>,
+    TypeCombo<columnar_t, columnar_t, sparse_t, invalid_id_t>, TypeCombo<columnar_t, row_t, dense_t, invalid_id_t>,
+    TypeCombo<columnar_t, row_t, sparse_t, invalid_id_t>, TypeCombo<row_t, columnar_t, dense_t, invalid_id_t>,
+    TypeCombo<row_t, columnar_t, sparse_t, invalid_id_t>) {
+
+    using namespace std::string_literals;
+    using input_type = typename T::input_type;
+    using update_type = typename T::update_type;
+    using sparsity_type = typename T::sparsity_type;
+    using id_check_type = typename T::id_check_type;
+
+    DatasetConst input_dataset{"input", 0, 1};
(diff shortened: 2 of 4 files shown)


tests/native_api_tests/test_api_model.cpp
```diff
@@ -192,57 +228,45 @@ TEST_CASE("API Model") {
...
+    SUBCASE("Single power flow") {
+        SUBCASE("Simple power flow") {
+            model.calculate(options, single_output_dataset);
+            node_output.get_value(PGM_def_sym_output_node_id, node_result_id.data(), -1);
+            node_output.get_value(PGM_def_sym_output_node_energized, node_result_energized.data(), 0, -1);
+            node_output.get_value(PGM_def_sym_output_node_u, node_result_u.data(), 0, 1, -1);
+            node_output.get_value(PGM_def_sym_output_node_u_pu, node_result_u_pu.data(), -1);
+            node_output.get_value(PGM_def_sym_output_node_u_angle, node_result_u_angle.data(), -1);
+            CHECK(node_result_u[0] == doctest::Approx(50.0));
+            CHECK(node_result_u_pu[0] == doctest::Approx(0.5));
```
mgovers, line 240: any particular reason why you check less now?

tests/native_api_tests/test_api_model.cpp
```diff
@@ -513,4 +537,133 @@ TEST_CASE("API Model") {
...
+*/
+TEST_CASE_TEMPLATE(
+    "API update id tests", T, TypeCombo<row_t, row_t, dense_t, with_id_t>, TypeCombo<row_t, row_t, sparse_t, with_id_t>,
+    TypeCombo<columnar_t, columnar_t, dense_t, with_id_t>, TypeCombo<columnar_t, columnar_t, sparse_t, with_id_t>,
+    TypeCombo<columnar_t, row_t, dense_t, with_id_t>, TypeCombo<columnar_t, row_t, sparse_t, with_id_t>,
+    TypeCombo<row_t, columnar_t, dense_t, with_id_t>, TypeCombo<row_t, columnar_t, sparse_t, with_id_t>,
+    TypeCombo<row_t, row_t, dense_t, invalid_id_t>, TypeCombo<row_t, row_t, sparse_t, invalid_id_t>,
+    TypeCombo<columnar_t, columnar_t, dense_t, invalid_id_t>, TypeCombo<columnar_t, columnar_t, sparse_t, invalid_id_t>,
+    TypeCombo<columnar_t, row_t, dense_t, invalid_id_t>, TypeCombo<columnar_t, row_t, sparse_t, invalid_id_t>,
+    TypeCombo<row_t, columnar_t, dense_t, invalid_id_t>, TypeCombo<row_t, columnar_t, sparse_t, invalid_id_t>) {
```
mgovers, line 560: it looks a bit heavy at first sight but looking at the rest of the tests, this is a significant improvement :+1:

tests/native_api_tests/test_api_model.cpp
```diff
@@ -513,4 +537,133 @@ TEST_CASE("API Model") {
...
+    if constexpr (std::is_same_v<input_type, row_t>) {
+        input_dataset.add_buffer("sym_load", 1, 1, nullptr, sym_load_buffer);
+    } else {
+        input_dataset.add_buffer("sym_load", 1, 1, nullptr, nullptr);
+        input_dataset.add_attribute_buffer("sym_load", "id", sym_load_id.data());
+        input_dataset.add_attribute_buffer("sym_load", "node", sym_load_node.data());
+        input_dataset.add_attribute_buffer("sym_load", "status", sym_load_status.data());
+        input_dataset.add_attribute_buffer("sym_load", "type", sym_load_type.data());
+        input_dataset.add_attribute_buffer("sym_load", "p_specified", sym_load_p_specified.data());
+        input_dataset.add_attribute_buffer("sym_load", "q_specified", sym_load_q_specified.data());
```
mgovers, line 619: instead, you can also add the `sym_load_buffer` initialization here. then, the only real difference between `row_t` and `columnar_t` is the call to `add_attribute_buffer` vs `set_value`

or will that cause `const` correctness clang-tidy warnings? I guess you can work around those by calling `set_nan` outside this code block

mgovers (commented):
great test, very clear what the differences are

tests/native_api_tests/test_api_model.cpp
nitbharambe: Oh, basically, I combined the checks down below which were repeated. 
Hence all the common checks are below the SUBCASE
I combined the get values too now.

mgovers: oh wait no, i think we had a team agreement to do common checks in a separate function (e.g. a lambda) and then call the function at the end of the subcase, because some people thought the logic flow is more difficult if there are checks after the subcases are exited. please follow that convention

tests/native_api_tests/test_api_model.cpp
nitbharambe: I can move the initialization of Buffer outside while set_value remains inside the block

src/power_grid_model/core/buffer_handling.py
```diff
@@ -245,7 +245,7 @@ def _get_sparse_buffer_properties(
             if attribute_data.ndim != ndim + schema.dtype[attribute].ndim or attribute_data.shape[:ndim] != shape:
                 raise ValueError(f"Data buffers must be consistent. {VALIDATOR_MSG}")
 
-    contents_size = sum(shape)
+    contents_size = shape[0]
```
nitbharambe, line 248: False Alarm, no bug.
The shape never goes to 2D, earlier catch was false too

mgovers: 😮

tests/native_api_tests/test_api_model.cpp
```diff
@@ -513,4 +538,133 @@ TEST_CASE("API Model") {
...
+
+source and node inputs are row based.
+load input is either row based or columnar.
+load update is row based / columnar and densse / sparse.
+
+invalid_id_t tests are for testing the error handling of the model when the id is not found in the update dataset.
+optional_id_t tests are for testing the model when the id is not added to the update dataset.
+
+*/
+TEST_CASE_TEMPLATE(
```
figueroa1395, line 553: Just commenting on the first line so it doesn't get too large...

This test is genius, thanks for such a big upgrade here, great effort. 

Personal request: please share in the tech review, I think it is worth it, we can all learn good stuff from here.

tests/unit/test_power_grid_model.py
```diff
@@ -260,3 +260,103 @@ def test_empty_input():
...
+    "minimal_update",
+    [
+        pytest.param(update_sym_load_row_optional_id(), id="update_dense_row"),
+        pytest.param(update_sym_load_col(update_sym_load_row_optional_id()), id="update_dense_col"),
+        pytest.param(update_sym_load_sparse(update_sym_load_row_optional_id()), id="update_sparse_row"),
+        pytest.param(
+            update_sym_load_col(update_sym_load_sparse(update_sym_load_row_optional_id())), id="update_sparse_col"
+        ),
+    ],
+)
```
figueroa1395, line 343: Shouldn't we just mark (for now), only the row tests as failed? Isn't this intented to be used as an additional sets of tests for columnar optional id initially?

tests/unit/test_buffer_handling.py
```diff
@@ -0,0 +1,85 @@
...
+        pytest.param(ComponentType.sym_load, True, False, id="sym_load-batch-row_based"),
+        pytest.param(ComponentType.sym_load, False, True, id="sym_load-single-columnar"),
+        pytest.param(ComponentType.sym_load, False, False, id="sym_load-single-row_based"),
+        pytest.param(ComponentType.asym_load, True, True, id="asym_load-batch-columnar"),
+        pytest.param(ComponentType.asym_load, True, False, id="asym_load-batch-row_based"),
+        pytest.param(ComponentType.asym_load, False, True, id="asym_load-single-columnar"),
+        pytest.param(ComponentType.asym_load, False, False, id="asym_load-single-row_based"),
+    ],
+)
+def test__get_uniform_buffer_properties(component_type, is_batch, is_columnar):
```
figueroa1395, line 45: Nitpick: Rename to `test__get_dense_buffer_properties`, since sparse can also be uniform.

tests/unit/test_buffer_handling.py
```diff
@@ -0,0 +1,85 @@
...
+    data = load_data(component_type, is_batch=is_batch, is_columnar=is_columnar, is_sparse=False)
+    schema = power_grid_meta_data[DatasetType.update][component_type]
+    batch_size = 2 if is_batch else None
+    properties = _get_uniform_buffer_properties(data, schema=schema, is_batch=is_batch, batch_size=batch_size)
```
figueroa1395, line 49: Related to above: is the name of this function the correct according to our renewed conventions? Also, is the raised error correct? That looks like a typo to me.

figueroa1395 (commented):
I had forgotten to take a look at the python tests, I just have minor comments. Great job, testing looks good.

tests/unit/test_power_grid_model.py
figueroa1395: Ignore this, I forgot this PR was from main. You can resolve.

figueroa1395:
I would assume that merge conflicts are going to arise when either this PR, #771 or #813 are merged into main. Also, some of the tests added in #771 and #813 should "comply" with the refactoring here (mainly talking about main model tests in c++). How are we approaching these to not make a mess of ourselves? Or am I making too big of a deal out of it?

Also, in addition to reverting [e5d5670](https://github.com/PowerGridModel/power-grid-model/pull/817/commits/e5d567007c4b667d13135ad0fc0b600c614078f3), let's not forget to enable the python tests marked as failed as well.

mgovers:
> I would assume that merge conflicts are going to arise when either this PR, #771 or #813 are merged into main. Also, some of the tests added in #771 and #813 should "comply" with the refactoring here (mainly talking about main model tests in c++). How are we approaching these to not make a mess of ourselves? Or am I making too big of a deal out of it?
> 
> Also, in addition to reverting [e5d5670](https://github.com/PowerGridModel/power-grid-model/pull/817/commits/e5d567007c4b667d13135ad0fc0b600c614078f3), let's not forget to enable the python tests marked as failed as well.

which ones specifically are you talking about? you can create a separate PR to this PR to try to fix whatever you mean

Jerry-Jinfeng-Guo:
Would this be part of tests for https://github.com/PowerGridModel/power-grid-model/pull/771 or https://github.com/PowerGridModel/power-grid-model/pull/813 or would it be merged to main as a standalone test PR? It now merges to main.

figueroa1395:
> > I would assume that merge conflicts are going to arise when either this PR, #771 or #813 are merged into main. Also, some of the tests added in #771 and #813 should "comply" with the refactoring here (mainly talking about main model tests in c++). How are we approaching these to not make a mess of ourselves? Or am I making too big of a deal out of it?
> > Also, in addition to reverting [e5d5670](https://github.com/PowerGridModel/power-grid-model/pull/817/commits/e5d567007c4b667d13135ad0fc0b600c614078f3), let's not forget to enable the python tests marked as failed as well.
> 
> which ones specifically are you talking about? you can create a separate PR to this PR to try to fix whatever you mean

I am talking about the main model tests of the API in the c++ side. I believe there were some tests added in the two mentioned PRs that would have different structure compared to this. I would prefer not to touch them at this moment since they are already being actively worked on, and perhaps more things are going to be added to those tests. However, let's just address if/when they happen. I'll create the necessary PR then. I was just pointing it out as a reference to everyone.

Jerry-Jinfeng-Guo (dismissed):
We should have this more often

src/power_grid_model/core/buffer_handling.py
mgovers, line 248: i actually wonder whether this is correct. the original one wasn't, but this is the full size of the buffer, right?

mgovers: ok, good one. i hereby also approve

mgovers (approved):

PR #819, opened 2024-11-06
Feature/support for optional IDs per component type

figueroa1395:
Provides additional tests for the optional IDs functionality.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -320,13 +320,14 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     update_component(ConstDataset const& update_data, Idx pos,
                      std::array<std::reference_wrapper<std::vector<Idx2D> const>, n_types> const& sequence_idx_map) {
         run_functor_with_all_types_return_void([this, pos, &update_data, &sequence_idx_map]<typename CT>() {
-            update_component<CT, CacheType>(update_data, pos, std::get<index_of_component<CT>>(sequence_idx_map).get());
+            this->update_component<CT, CacheType>(update_data, pos,
+                                                  std::get<index_of_component<CT>>(sequence_idx_map).get());
         });
     }
     template <cache_type_c CacheType>
     void update_component(ConstDataset const& update_data, Idx pos, SequenceIdx const& sequence_idx_map) {
         run_functor_with_all_types_return_void([this, pos, &update_data, &sequence_idx_map]<typename CT>() {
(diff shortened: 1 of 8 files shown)


figueroa1395:
The tests confirm that the optional ID feature only fails when mixing component updates that have no ID with the ones that have ID. Hence it is not a bug, but rather a missing feature. Knowing this, unnecesary tests will be removed and this PR will be cleaned up.

The missing feature is being addressed in #820

tests/unit/test_optional_ids.py
```diff
@@ -0,0 +1,175 @@
...
+def batch_update_sym_load_no_id_col():
+    source_attribute_dtypes = power_grid_meta_data[DatasetType.update][ComponentType.sym_load].dtype
+    return {
+        "p_specified": np.array([[30e6, 10e6], [30e6, 15e6]], dtype=source_attribute_dtypes["p_specified"]),
+    }
+
+
+@pytest.fixture(
+    params=[
+        "batch_update_sym_load_row",
```
nitbharambe, line 108:
```suggestion
        "batch_update_sym_load_row",
```
Can you use really short names? They would appear bigger  in the sidebar.
If function names look ugly, you can also use `@pytest.fixture(name= ...)` to rename fixtures instead of functions but maybe it wont be helpful.

maybe we can also use a symbol instead full name.

figueroa1395: [ed9c9b9](https://github.com/PowerGridModel/power-grid-model/pull/819/commits/ed9c9b91a1cad074a12073ffaf9614ee937f97ae) should be better

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -709,25 +720,30 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
     }
 
-    static auto scenario_update_restore(MainModelImpl& model, ConstDataset const& update_data,
-                                        SequenceIdx const& scenario_sequence, SequenceIdx& scenario_sequence_cache,
-                                        bool is_independent, std::vector<CalculationInfo>& infos) noexcept {
-        bool const do_update_cache = !is_independent;
+    static auto scenario_update_restore(
+        MainModelImpl& model, ConstDataset const& update_data,
+        std::array<std::reference_wrapper<std::vector<Idx2D> const>, n_types> const& scenario_sequence,
+        SequenceIdx& current_scenario_sequence_cache, ComponentFlags const& is_independent,
```
nitbharambe, line 726: nitpick: dont change now, adding it for refactoring
```suggestion
        SequenceIdx& current_scenario_sequence_cache, ComponentFlags const& independence_flags,
``` 
Was wondering how one can tranasform a boolean haha

mgovers (approved):

PR #822, opened 2024-11-07
make validation core functions private

mgovers:
Idem to #787 but now for validation core

relates to #784


src/power_grid_model/validation/rules.py
```diff
@@ -70,12 +70,12 @@
...
 )
 from power_grid_model.validation.utils import (
-    eval_expression,
-    get_indexer,
-    get_mask,
-    get_valid_ids,
-    nan_type,
-    set_default_value,
+    _eval_expression,
+    _get_indexer,
```
figueroa1395, line 74: Isn't this one intended to be public?

figueroa1395, line 74: Isn't this one intended to be public?

src/power_grid_model/validation/rules.py
```diff
@@ -70,12 +70,12 @@
...
+    _get_mask,
+    _get_valid_ids,
+    _nan_type,
```
figueroa1395, line 77: Would this be useful for the users to be public? Or is there a python side function that does something similar already?

figueroa1395 (approved):
Just double checking a couple.

src/power_grid_model/validation/rules.py
mgovers: we want people to use `PowerGridModel.get_indexer()` instead (https://power-grid-model.readthedocs.io/en/stable/api_reference/python-api-reference.html#power_grid_model.PowerGridModel.get_indexer). 

In addition, the `power_grid_model.validation.utils` module where it is declared has a module-level docstring that says
```
Utilities used for validation. Only errors_to_string() is intended for end users.
```

This file actually exposes these utilities even though it only wishes to use it internally

src/power_grid_model/validation/rules.py
mgovers: we have https://power-grid-model.readthedocs.io/en/stable/api_reference/python-api-reference.html#power_grid_model.typing.ComponentMetaData.nans

src/power_grid_model/validation/rules.py
figueroa1395: Oops I confused this one with the other `get_indexer`. Great find.

nitbharambe: @zhen0427 You can implement this in sensitivity directly instead of importing.

figueroa1395: @nitbharambe @zhen0427 but `power_grid_model.get_indexer` is still public. Or is it different to the other indexer in the validation functions?

PR #824, opened 2024-11-08
Feature / use span instead of reference wrapper for cache sequence

mgovers:
Part 1 of refactoring after optional ID


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -435,7 +432,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
                                      ComponentFlags const& to_store) const {
         // TODO: (jguo) this function could be encapsulated in UpdateCompIndependence in update.hpp
         return run_functor_with_all_types_return_array([this, scenario_idx, &update_data, &to_store]<typename CT>() {
-            if (!to_store[index_of_component<CT>]) {
+            if (!std::get<index_of_component<CT>>(to_store)) {
```
figueroa1395, line 435: Nitpick: Rename `to_store` in this function to something more expressive. Maybe `available_components`, or `components_to_store`?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -445,15 +442,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-        // TODO: (jguo) this function could be encapsulated in UpdateCompIndependence in update.hpp
-        return run_functor_with_all_types_return_array([this, &update_data, &to_store]<typename CT>() {
-            if (!to_store[index_of_component<CT>]) {
-                return std::vector<Idx2D>{};
-            }
-            auto const independence = check_components_independence<CT>(update_data);
-            validate_update_data_independence(independence);
-            return get_component_sequence<CT>(update_data, 0, independence);
-        });
+        return get_sequence_idx_map(update_data, 0, to_store);
```
figueroa1395, line 445: This is basically for permanent updates, right? Or for scenario 0 only. If so, shouldn't we edit as well the comment explaining the function?

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -720,28 +702,41 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+            return run_functor_with_all_types_return_array(
+                [&all_scenario_sequence, &current_scenario_sequence_cache, &is_independent]<typename CT>() {
+                    constexpr auto comp_idx = index_of_component<CT>;
+                    if (std::get<comp_idx>(is_independent)) {
+                        return std::span<Idx2D const>{std::get<comp_idx>(all_scenario_sequence)};
+                    }
+                    return std::span<Idx2D const>{std::get<comp_idx>(current_scenario_sequence_cache)};
+                });
+        };
+
```
figueroa1395, line 726: Curiousity: Why was this moved here from sub_batch_calculation? Just to reduce complexity on the later? Besides that, the "new" logic is clear.

figueroa1395 (commented):
Looks good to me, but I prefer to not approve since we did the changes "together" in pair programming.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: no, this is the entry point used when caching the entire batch after determining which component types are independent. the one below is for permanent updates.

will remove this as indeed it is very unclear as to what happens

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: > Just to reduce complexity on the latter?

yes and also to combine all the logic concerning which cache to use when into one single location.

figueroa1395 (approved):
I'm re-approving since @nitbharambe had approved already and the changes after are minor.

PR #826, opened 2024-11-11
Step-up transformer tap changer support: feature branch

Jerry-Jinfeng-Guo:
This is the feature branch for step-up transformer tap changer support. Closes https://github.com/PowerGridModel/power-grid-model/issues/662. 

Dedicated tests:
* [x] Validation test case [PowerGridModel/power-grid-model/pull/868] [PowerGridModel/power-grid-model/pull/912]
* [x] Meshed grid test case [PowerGridModel/power-grid-model/pull/777]

Feature additions (w/ unit tests):
* [x] Transformer weight `1` to `0` [PowerGridModel/power-grid-model/pull/698]
* [x] Additional validation (core+py) [PowerGridModel/power-grid-model/pull/829]
* [x] Updated ranking [PowerGridModel/power-grid-model/pull/832]
* [x] Updated unit test for tap position optimizer [PowerGridModel/power-grid-model/pull/892]
* [x] Updated control logic for linear search [PowerGridModel/power-grid-model/pull/893]
* [x] Updated control logic for binary search [PowerGridModel/power-grid-model/pull/894]


power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -270,9 +303,33 @@ inline auto get_edge_weights(TransformerGraph const& graph) -> TrafoGraphEdgePro
...
+        // |   A   |    C    | [A->B], [A->C], [B<->C] |
+        // |   B   |    A    | [B->A], [C<->A], [B->C] |
+        // |   B   |    B    | [A->B], [C<->A], [C->B] |
+        // |   B   |    C    | [B->A], [C<->A], [B->C] |
+        // |   C   |    A    | [A<->B], [C->A], [C->B] |
+        // |   C   |    B    | [A<->B], [C->A], [C->B] |
+        // |   C   |    C    | [A<->B], [A->C], [A->B] |
+        // In two winding trafo, the edge is always pointing to the control side; in three winding trafo edges, the
+        // edges are always pointing parallel to the control side: meaning that for delta configuration ABC, the above
+        // situations can happen.
```
mgovers, line 326: in meshed grids, this can also happen for regular transformers. The logic below still holds, but the comment is a little bit misleading

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -270,9 +303,33 @@ inline auto get_edge_weights(TransformerGraph const& graph) -> TrafoGraphEdgePro
...
+        if (edge_src_rank != edge_tgt_rank - 1) {
+            throw AutomaticTapInputError(
+                "Control side of a transformer should be the relatively further side to a source.\n");
```
mgovers, line 329:
```suggestion
                "The control side of a transformer regulator should be relatively further away from the source than the tap side.\n");
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -669,7 +746,8 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
             //  - tap_max > tap_min && strategy_max == true
             //  - tap_max < tap_min && strategy_max == false
             // Upper bound should be updated to the current tap position if the rest is the case.
-            if (tap_reverse_ == strategy_max) {
+            bool const invert_strategy = control_at_tap_side_ ? !strategy_max : strategy_max;
```
mgovers, line 749: optionally for consistency, but feel free to resolve because it is still readable
```suggestion
            bool const invert_strategy = control_at_tap_side_ != strategy_max;
```

src/power_grid_model/_core/error_handling.py
```diff
@@ -71,6 +72,8 @@
 _AUTOMATIC_TAP_CALCULATION_ERROR_RE = re.compile(
     r"Automatic tap changing regulator with tap_side at LV side is not supported. Found at id (-?\d+)\n"
 )
+_AUTOMATIC_TAP_INPUT_ERROR_RE = re.compile(r"Automatic tap changer has invalid configuration\..*")
```
mgovers, line 75: not sure but i believe this also should work ~(but please only apply the suggestion after checking that there is a test for it and/or manual verification)~ EDIT: there is a unit test for this ✅ 

```suggestion
_AUTOMATIC_TAP_INPUT_ERROR_RE = re.compile(r"Automatic tap changer has invalid configuration")
```

the difference prevents potential combinatorial blow-ups (sonar cloud sometimes catches regex blow-ups but not always)

mgovers (commented):
a couple minor suggestions after all PRs have been reviewd, but nothing major. Also feel free to resolve without applying if you disagree.

power_grid_model_c/power_grid_model/include/power_grid_model/common/enum.hpp
```diff
@@ -20,6 +20,8 @@ enum class BranchSide : IntS { from = 0, to = 1 };
 
 enum class Branch3Side : IntS { side_1 = 0, side_2 = 1, side_3 = 2 };
 
+enum class TapSide : IntS { from = 0, to = 1, side_1 = 0, side_2 = 1, side_3 = 2 };
```
petersalemink95, line 23: I tried finding usage of `TapSide` somewhere in the code, but cannot find it. Is this an unused enum or did I miss something?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -270,9 +307,34 @@ inline auto get_edge_weights(TransformerGraph const& graph) -> TrafoGraphEdgePro
         if (graph[e].regulated_idx == unregulated_idx) {
             continue;
         }
-        auto edge_res = std::min(vertex_distances[boost::source(e, graph)], vertex_distances[boost::target(e, graph)]);
+        // control_side, two winding or three winding, is always parallel to the
```
petersalemink95, line 310: What is meant by "parallel" here?

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -336,11 +401,19 @@ constexpr IntS one_step_tap_down(transformer_c auto const& transformer) {
     return tap_min < tap_max ? tap_pos - IntS{1} : tap_pos + IntS{1};
 }
 // higher voltage at control side => lower voltage at tap side => lower tap pos
-constexpr IntS one_step_control_voltage_up(transformer_c auto const& transformer) {
+constexpr IntS one_step_control_voltage_up(transformer_c auto const& transformer, bool control_at_tap_side) {
+    if (control_at_tap_side) {
+        // control side is tap side, voltage up requires tap up
```
petersalemink95, line 406: I don't get `control_at_tap-side`? What is meant here and why is it used? It sounds like control side and tap side are at the same side, which is not possible

petersalemink95:
```
    bool control_at_tap_side() const {
        return static_cast<IntS>(regulator.get().control_side()) == transformer.tap_side();
    }
```
This does not make sense to me

mgovers: Control side and tap side will be at the same side for step-up transformers, because the tap side will still be the HV case (away from source, just like the control side)

figueroa1395: For reference, you can take a look at the validation cases introduced in https://github.com/PowerGridModel/power-grid-model/pull/912 where control and type side on the same side are tested.

petersalemink95: N.v.m. was to quick with this comment. Thanks

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: Below there is a detailed breakdown of all directions, where the notion of "parallel" is explained.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -737,25 +820,27 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
-        bool tap_reverse_{false};    // tap range normal or reversed
-        bool inevitable_run_{false}; // inevitable run
+        IntS lower_bound_{};              // tap position lower bound
+        IntS upper_bound_{};              // tap position upper bound
+        IntS current_{0};                 // current tap position
+        bool last_down_{false};           // last direction
+        bool last_check_{false};          // last run checked
+        bool tap_reverse_{false};         // tap range normal or reversed
+        bool inevitable_run_{false};      // inevitable run
+        bool control_at_tap_side_{false}; // regulator control side is at tap side
```
Jerry-Jinfeng-Guo, line 843: @mgovers To continue the discussion from https://github.com/PowerGridModel/power-grid-model/pull/894#discussion_r1957986478 in https://github.com/PowerGridModel/power-grid-model/pull/894: 

> rather than `control_at_tap_side`, maybe you can combine this with the logic surrounding `tap_reverse`?

I'd still argue against this for the same reason: the intended logic became convoluted and hard to follow

power_grid_model_c/power_grid_model/include/power_grid_model/common/enum.hpp
Jerry-Jinfeng-Guo: This was introduced to differentiate from `ControlSide`. However, existing code base uses predominantly `ControlSide` for tap side, which extends out of scope. We can keep this line here and create a dedicated issue for that. Or we can remove this line and call it done.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
figueroa1395: I think that leaving `control_at_tap_side` makes sense here. It is more readable this way as the intention is clear at first sight.

power_grid_model_c/power_grid_model/include/power_grid_model/common/enum.hpp
figueroa1395: Since both `Enum`s are identical, would it make sense to just keep one but rename it to something like `ControlAndTapSide`?

petersalemink95: Existing code does not use `ControlSide` "predominantly" -> `TapSide` is never used.

If you think `TapSide` should be used somewhere you can make an issue and we can discuss if that is indeed the case. Anyway, since `TapSide` is never used in the code it should be removed here.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
petersalemink95: I don't get the usage of parallel here (and below), hence the question

Jerry-Jinfeng-Guo: Reworded the comment below and removed comment from this place where it does not belong.

power_grid_model_c/power_grid_model/include/power_grid_model/common/enum.hpp
Jerry-Jinfeng-Guo: > Existing code does not use ControlSide "predominantly" -> TapSide is never used.

Now it's exclusive

figueroa1395: After a bit of thought and `TapSide` being removed, maybe it is a good thing to just keep one. Then it is easier to recognize whether the tap and control sides are on the same or different sides, and which ones as well. I will not create a further issue about this.

figueroa1395 (dismissed):
There are no further comments from me. Hereby I give my approval.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -326,7 +324,8 @@ inline auto get_edge_weights(TransformerGraph const& graph) -> TrafoGraphEdgePro
         // |   C   |    B    | [A<->B], [C->A], [C->B] |
         // |   C   |    C    | [A<->B], [A->C], [A->B] |
         // In two winding trafo, the edge is always pointing to the control side; in three winding trafo edges, the
-        // edges are always pointing parallel to the control side: meaning that for delta configuration ABC, the above
+        // one directional edges are always pointing towards the control side or the node connected to the control
+        // side via the two directional edge if there exists one. For delta configuration ABC, the above
```
mgovers, line 328: for readability
```suggestion
        // unidirectional edges are always pointing towards the control side and the node connected to the control
        // side via the bidirectional edge (if it exists). For delta configuration ABC, the above
```

mgovers (dismissed):
only one more readability comment. Otherwise approved.

nitbharambe (approved):

PR #827, opened 2024-11-11
[Feature] add check-blocking-labels github action

mgovers:
this action was directly copy-pasted from the one in PGM-IO in https://github.com/PowerGridModel/power-grid-model-io/blob/main/.github/workflows/check-blocking-labels.yml

### DIFF
--- a/.github/workflows/check-blocking-labels.yml
+++ b/.github/workflows/check-blocking-labels.yml
@@ -1,0 +1,46 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+
+name: Check Blocking Labels
+
+on:
+  # run pipeline on pull request
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - labeled
+      - unlabeled
+  # run pipeline on merge queue
+  merge_group:
+  # run this workflow manually from the Actions tab
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


TonyXiang8787 (approved):

.github/workflows/check-blocking-labels.yml
```diff
@@ -0,0 +1,46 @@
...
+  # run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-blocking-labels
+  cancel-in-progress: true
+
+jobs:
+  check-blocking-labels:
+    runs-on: ubuntu-latest
```
figueroa1395, line 27: If the CI is passing, does this mean that ubuntu-latest has finally migrated to 24.04? Then we can update it everywhere else, right?

figueroa1395 (approved):
Looks good, just minor comment.

figueroa1395:
I removed from merge queue because for some reason the opened comment was bypassed when I approved. Is that normal?

.github/workflows/check-blocking-labels.yml
mgovers: this check only does a simple shell-check. no Ubuntu 24.04 features (like a specific compiler version) are used. If this action were to be affected, then the entire GitHub is broken (shell incompatibility on the very simplest level) and I don't think a lot of people would like that. I think it's safe to assume that it's no issue here 😬

PR #828, opened 2024-11-11
Feature/Optional ID structural refactor

figueroa1395:
Continues the refactor started in #824. The scope of this PR is as follows:

- [ ] Remove hard-coded `-1`s in `main_model_impl.hpp`.
- [ ] Minor renaming.
- [ ] Refactor `check_components_independence`.
- [ ] Rename `update_components` overload for single components.
- [ ] Move independence related functionality to  `update.hpp`.

Out of scope:
- Move `update_components` out of `main_model_impl.hpp`.
- Reduce `update_components` overload number.
- Variadic template with `V` suffix.
- Homogenize `CT`, `CompType` and `ComponentType`.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -171,6 +171,9 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
 
     static constexpr Idx ignore_output{-1};
     static constexpr Idx invalid_index{-1};
+    static constexpr Idx isolated_component{-1};
+    static constexpr Idx not_connected{-1};
+    static constexpr Idx sequential{-1};
 
   protected:
     // run functors with all component types
@@ -552,7 +555,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     template <typename Calculate>
         requires std::invocable<std::remove_cvref_t<Calculate>, MainModelImpl&, MutableDataset const&, Idx>
     BatchParameter batch_calculation_(Calculate&& calculation_fn, MutableDataset const& result_data,
-                                      ConstDataset const& update_data, Idx threading = -1) {
+                                      ConstDataset const& update_data, Idx threading = sequential) {
         // if the update dataset is empty without any component
         // execute one power flow in the current instance, no batch calculation is needed
         if (update_data.empty()) {
@@ -1080,7 +1083,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         // loop all branch
         for (Idx i = 0; i != static_cast<Idx>(state_.comp_topo->branch_node_idx.size()); ++i) {
             Idx2D const math_idx = state_.topo_comp_coup->branch[i];
-            if (math_idx.group == -1) {
+            if (math_idx.group == isolated_component) {
                 continue;
             }
             // assign parameters
@@ -1090,7 +1093,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         // loop all branch3
         for (Idx i = 0; i != static_cast<Idx>(state_.comp_topo->branch3_node_idx.size()); ++i) {
             Idx2DBranch3 const math_idx = state_.topo_comp_coup->branch3[i];
-            if (math_idx.group == -1) {
+            if (math_idx.group == isolated_component) {
                 continue;
             }
             // assign parameters, branch3 param consists of three branch parameters
@@ -1103,7 +1106,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         // loop all shunt
         for (Idx i = 0; i != static_cast<Idx>(state_.comp_topo->shunt_node_idx.size()); ++i) {
             Idx2D const math_idx = state_.topo_comp_coup->shunt[i];
-            if (math_idx.group == -1) {
+            if (math_idx.group == isolated_component) {
                 continue;
             }
             // assign parameters
@@ -1113,7 +1116,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         // loop all source
         for (Idx i = 0; i != static_cast<Idx>(state_.comp_topo->source_node_idx.size()); ++i) {
             Idx2D const math_idx = state_.topo_comp_coup->source[i];
-            if (math_idx.group == -1) {
+            if (math_idx.group == isolated_component) {
                 continue;
             }
             // assign parameters
@@ -1132,7 +1135,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
                     Idx2D const math_idx =
                         state.topo_comp_coup
                             ->branch[main_core::get_component_sequence<Branch>(state, changed_component_idx)];
-                    if (math_idx.group == -1) {
+                    if (math_idx.group == isolated_component) {
                         return;
                     }
                     // assign parameters
@@ -1141,7 +1144,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
                     Idx2DBranch3 const math_idx =
                         state.topo_comp_coup
                             ->branch3[main_core::get_component_sequence<Branch3>(state, changed_component_idx)];
-                    if (math_idx.group == -1) {
+                    if (math_idx.group == isolated_component) {
                         return;
                     }
                     // assign parameters, branch3 param consists of three branch parameters
@@ -1154,7 +1157,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
                     Idx2D const math_idx =
                         state.topo_comp_coup
                             ->shunt[main_core::get_component_sequence<Shunt>(state, changed_component_idx)];
-                    if (math_idx.group == -1) {
+                    if (math_idx.group == isolated_component) {
                         return;
                     }
                     // assign parameters
@@ -1236,7 +1239,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         for (Idx i = 0, n = narrow_cast<Idx>(components.size()); i != n; ++i) {
             if (include(i)) {
                 Idx2D const math_idx = components[i];
-                if (math_idx.group != -1) {
+                if (math_idx.group != isolated_component) {
                     auto const& component = get_component_by_sequence<ComponentIn>(state, i);
                     CalcStructOut& math_model_input = calc_input[math_idx.group];
                     std::vector<CalcParamOut>& math_model_input_vect = math_model_input.*comp_vect;
@@ -1256,7 +1259,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         for (Idx i = 0, n = narrow_cast<Idx>(components.size()); i != n; ++i) {
             if (include(i)) {
                 Idx2D const math_idx = components[i];
-                if (math_idx.group != -1) {
+                if (math_idx.group != isolated_component) {
                     auto const& component = get_component_by_sequence<ComponentIn>(state, i);
                     CalcStructOut& math_model_input = calc_input[math_idx.group];
                     std::vector<CalcParamOut>& math_model_input_vect = math_model_input.*comp_vect;
@@ -1290,7 +1293,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
                                      std::vector<StateEstimationInput<sym>>& input) {
         for (Idx i = 0, n = narrow_cast<Idx>(objects.size()); i != n; ++i) {
             Idx2D const math_idx = objects[i];
-            if (math_idx.group == -1) {
+            if (math_idx.group == isolated_component) {
                 continue;
             }
             (input[math_idx.group].*component)[math_idx.pos] =
@@ -1398,7 +1401,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
             }
         }
 
-        auto fault_coup = std::vector<Idx2D>(state_.components.template size<Fault>(), Idx2D{-1, -1});
+        auto fault_coup = std::vector<Idx2D>(state_.components.template size<Fault>(), Idx2D{isolated_component, not_connected});
         std::vector<ShortCircuitInput> sc_input(n_math_solvers_);
 
         for (Idx i = 0; i != n_math_solvers_; ++i) {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -439,7 +439,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
             if (!std::get<index_of_component<CT>>(to_store)) {
                 return std::vector<Idx2D>{};
             }
-            auto const independence = check_components_independence<CT>(update_data);
+            auto const independence = check_component_independence<CT>(update_data);
             validate_update_data_independence(independence);
             return get_component_sequence<CT>(update_data, scenario_idx, independence);
         });
@@ -707,7 +707,8 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     }
 
     static auto scenario_update_restore(MainModelImpl& model, ConstDataset const& update_data,
-                                        ComponentFlags const& independence_flags, SequenceIdx const& all_scenario_sequence,
+                                        ComponentFlags const& independence_flags,
+                                        SequenceIdx const& all_scenario_sequence,
                                         SequenceIdx& current_scenario_sequence_cache,
                                         std::vector<CalculationInfo>& infos) noexcept {
         auto do_update_cache = [&independence_flags] {
@@ -782,79 +783,76 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
   public:
     template <class Component> using UpdateType = typename Component::UpdateType;
 
-    template <class CompType>
-    UpdateCompProperties check_components_independence(ConstDataset const& update_data) const {
-        auto const comp_count_in_base = this->component_count<CompType>();
+    template <typename T> bool check_id_na(T const& obj) const {
+        if constexpr (requires { obj.id; }) {
+            return is_nan(obj.id);
+        } else if constexpr (requires { obj.get().id; }) {
+            return is_nan(obj.get().id);
+        } else {
+            throw UnreachableHit{"check_component_independence", "Only components with id are supported"};
+        }
+    }
 
-        auto const check_id_na = [](auto const& obj) -> bool {
-            if constexpr (requires { obj.id; }) {
-                return is_nan(obj.id);
-            } else if constexpr (requires { obj.get().id; }) {
-                return is_nan(obj.get().id);
-            } else {
-                throw UnreachableHit{"check_components_independence", "Only components with id are supported"};
-            }
-        };
+    template <typename CompType>
+    void process_buffer_span(auto const& all_spans, UpdateCompProperties& properties) const {
+        properties.ids_all_na = std::ranges::all_of(all_spans, [this](auto const& vec) {
+            return std::ranges::all_of(vec, [this](auto const& item) { return this->template check_id_na(item); });
+        });
+        properties.ids_part_na = std::ranges::any_of(all_spans,
+                                                     [this](auto const& vec) {
+                                                         return std::ranges::any_of(vec, [this](auto const& item) {
+                                                             return this->template check_id_na(item);
+                                                         });
+                                                     }) &&
+                                 !properties.ids_all_na;
 
-        auto const process_buffer_span = [check_id_na](auto const& all_spans, UpdateCompProperties& result) {
-            result.ids_all_na = std::ranges::all_of(
-                all_spans, [&check_id_na](auto const& vec) { return std::ranges::all_of(vec, check_id_na); });
-            result.ids_part_na =
-                std::ranges::any_of(
-                    all_spans, [&check_id_na](auto const& vec) { return std::ranges::any_of(vec, check_id_na); }) &&
-                !result.ids_all_na;
+        if (all_spans.empty()) {
+            properties.update_ids_match = true;
+        } else {
+            // Remember the begin iterator of the first scenario, then loop over the remaining scenarios and
+            // check the ids
+            auto const first_span = all_spans[0];
+            // check the subsequent scenarios
+            // only return true if all scenarios match the ids of the first batch
+            properties.update_ids_match =
+                std::ranges::all_of(all_spans.cbegin() + 1, all_spans.cend(), [&first_span](auto const& current_span) {
+                    return std::ranges::equal(current_span, first_span,
+                                              [](UpdateType<CompType> const& obj, UpdateType<CompType> const& first) {
+                                                  return obj.id == first.id;
+                                              });
+                });
+        }
+    }
 
-            if (all_spans.empty()) {
-                result.update_ids_match = true;
-            } else {
-                // Remember the begin iterator of the first scenario, then loop over the remaining scenarios and
-                // check the ids
-                auto const first_span = all_spans[0];
-                // check the subsequent scenarios
-                // only return true if all scenarios match the ids of the first batch
-                result.update_ids_match = std::ranges::all_of(
-                    all_spans.cbegin() + 1, all_spans.cend(), [&first_span](auto const& current_span) {
-                        return std::ranges::equal(current_span, first_span,
-                                                  [](UpdateType<CompType> const& obj,
-                                                     UpdateType<CompType> const& first) { return obj.id == first.id; });
-                    });
-            }
-        };
+    template <class CompType> UpdateCompProperties check_component_independence(ConstDataset const& update_data) const {
+        UpdateCompProperties properties;
+        properties.name = CompType::name;
+        auto const component_idx = update_data.find_component(properties.name, false);
+        properties.is_columnar = update_data.is_columnar(properties.name);
+        properties.dense = update_data.is_dense(properties.name);
+        properties.uniform = update_data.is_uniform(properties.name);
+        properties.has_any_elements =
+            component_idx != invalid_index && update_data.get_component_info(component_idx).total_elements > 0;
+        properties.elements_ps_in_update =
+            properties.uniform ? update_data.uniform_elements_per_scenario(properties.name) : invalid_index;
+        properties.elements_in_base = this->component_count<CompType>();
 
-        auto const check_each_component = [&update_data, &process_buffer_span,
-                                           &comp_count_in_base]() -> UpdateCompProperties {
-            // get span of all the update data
-            auto const comp_index = update_data.find_component(CompType::name, false);
-            UpdateCompProperties result;
-            result.name = CompType::name;
-            result.is_columnar = update_data.is_columnar(result.name);
-            result.dense = update_data.is_dense(result.name);
-            result.uniform = update_data.is_uniform(result.name);
-            result.has_any_elements =
-                comp_index != invalid_index && update_data.get_component_info(comp_index).total_elements > 0;
+        if (properties.is_columnar) {
+            process_buffer_span<CompType>(
+                update_data.template get_columnar_buffer_span_all_scenarios<meta_data::update_getter_s, CompType>(),
+                properties);
+        } else {
+            process_buffer_span<CompType>(
+                update_data.template get_buffer_span_all_scenarios<meta_data::update_getter_s, CompType>(), properties);
+        }
 
-            result.elements_ps_in_update =
-                result.uniform ? update_data.uniform_elements_per_scenario(result.name) : invalid_index;
-
-            result.elements_in_base = comp_count_in_base;
-
-            if (result.is_columnar) {
-                process_buffer_span(
-                    update_data.get_columnar_buffer_span_all_scenarios<meta_data::update_getter_s, CompType>(), result);
-            } else {
-                process_buffer_span(update_data.get_buffer_span_all_scenarios<meta_data::update_getter_s, CompType>(),
-                                    result);
-            }
-            return result;
-        };
-
-        return check_each_component();
+        return properties;
     }
 
     UpdateCompIndependence check_components_independence(ConstDataset const& update_data) const {
         // check and return indenpendence of all components
         return run_functor_with_all_types_return_array(
-            [this, &update_data]<typename CT>() { return this->check_components_independence<CT>(update_data); });
+            [this, &update_data]<typename CT>() { return this->check_component_independence<CT>(update_data); });
     }
 
     ComponentFlags is_update_independent(ConstDataset const& update_data) {
@@ -1271,24 +1269,6 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         }
     }
 
-    template <calculation_input_type CalcInputType>
-    static auto calculate_param(auto const& c, auto const&... extra_args)
-        requires requires {
-            { c.calc_param(extra_args...) };
-        }
-    {
-        return c.calc_param(extra_args...);
-    }
-
-    template <calculation_input_type CalcInputType>
-    static auto calculate_param(auto const& c, auto const&... extra_args)
-        requires requires {
-            { c.template calc_param<typename CalcInputType::sym>(extra_args...) };
-        }
-    {
-        return c.template calc_param<typename CalcInputType::sym>(extra_args...);
-    }
-
     template <symmetry_tag sym, IntSVector(StateEstimationInput<sym>::*component), class Component>
     static void prepare_input_status(MainModelState const& state, std::vector<Idx2D> const& objects,
                                      std::vector<StateEstimationInput<sym>>& input) {
@@ -1402,7 +1382,8 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
             }
         }
 
-        auto fault_coup = std::vector<Idx2D>(state_.components.template size<Fault>(), Idx2D{isolated_component, not_connected});
+        auto fault_coup =
+            std::vector<Idx2D>(state_.components.template size<Fault>(), Idx2D{isolated_component, not_connected});
         std::vector<ShortCircuitInput> sc_input(n_math_solvers_);
 
         for (Idx i = 0; i != n_math_solvers_; ++i) {
@@ -1501,5 +1482,23 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         last_updated_calculation_symmetry_mode_ = is_symmetric_v<sym>;
     }
 };
+
+template <calculation_input_type CalcInputType>
+static auto calculate_param(auto const& c, auto const&... extra_args)
+    requires requires {
+        { c.calc_param(extra_args...) };
+    }
+{
+    return c.calc_param(extra_args...);
+}
+
+template <calculation_input_type CalcInputType>
+static auto calculate_param(auto const& c, auto const&... extra_args)
+    requires requires {
+        { c.template calc_param<typename CalcInputType::sym>(extra_args...) };
+    }
+{
+    return c.template calc_param<typename CalcInputType::sym>(extra_args...);
+}
 
 } // namespace power_grid_model
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -169,6 +169,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     using ComponentFlags = std::array<bool, n_types>;
     using ComponentCountInBase = std::pair<std::string, Idx>;
 
+    // TODO: (figueroa1395) probably move to common.hpp or somewhere else
     static constexpr Idx ignore_output{-1};
     static constexpr Idx invalid_index{-1};
     static constexpr Idx isolated_component{-1};
@@ -706,21 +707,21 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     }
 
     static auto scenario_update_restore(MainModelImpl& model, ConstDataset const& update_data,
-                                        ComponentFlags const& is_independent, SequenceIdx const& all_scenario_sequence,
+                                        ComponentFlags const& independence_flags, SequenceIdx const& all_scenario_sequence,
                                         SequenceIdx& current_scenario_sequence_cache,
                                         std::vector<CalculationInfo>& infos) noexcept {
-        auto do_update_cache = [&is_independent] {
+        auto do_update_cache = [&independence_flags] {
             ComponentFlags result;
-            std::ranges::transform(is_independent, result.begin(), std::logical_not<>{});
+            std::ranges::transform(independence_flags, result.begin(), std::logical_not<>{});
             return result;
         }();
 
         auto const scenario_sequence = [&all_scenario_sequence, &current_scenario_sequence_cache,
-                                        &is_independent]() -> SequenceIdxView {
+                                        &independence_flags]() -> SequenceIdxView {
             return run_functor_with_all_types_return_array(
-                [&all_scenario_sequence, &current_scenario_sequence_cache, &is_independent]<typename CT>() {
+                [&all_scenario_sequence, &current_scenario_sequence_cache, &independence_flags]<typename CT>() {
                     constexpr auto comp_idx = index_of_component<CT>;
-                    if (std::get<comp_idx>(is_independent)) {
+                    if (std::get<comp_idx>(independence_flags)) {
                         return std::span<Idx2D const>{std::get<comp_idx>(all_scenario_sequence)};
                     }
                     return std::span<Idx2D const>{std::get<comp_idx>(current_scenario_sequence_cache)};


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -169,8 +169,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     using ComponentFlags = std::array<bool, n_types>;
     using ComponentCountInBase = std::pair<std::string, Idx>;
 
+    // TODO: (figueroa1395) probably move to common.hpp or somewhere else
```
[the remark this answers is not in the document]
figueroa1395: `invalid_index` will probably be moved to `.../main_core/update.hpp` since it is only going to be used there after the last step of refactoring. The rest are only used in `main_model_impl.hpp` (at least for now), so I will leave them be until the bigger refactoring happens, as they may be moved around or removed alltogether later.

figueroa1395:
Up until [44b4233](https://github.com/PowerGridModel/power-grid-model/pull/828/commits/44b4233dea43f17fbaeac51e1c5d1007731121cc) everything works fine. I am figuring out why can't I get the [move get_sequence_idx_map](https://github.com/PowerGridModel/power-grid-model/pull/828/commits/e4bf1a391b457e066c5df74ba372bf1d0664106a) correctly.

figueroa1395:
I just realized that `check_id_na` was supposed to be moved to `dataset.hpp`. I will address that on Monday, but besides that, everything else is ready.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -314,21 +293,21 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+                                                              std::get<index_of_component<CT>>(sequence_idx_map));
+            });
     }
-    template <cache_type_c CacheType> void update_component(ConstDataset const& update_data) {
-        update_component<CacheType>(update_data, 0, get_sequence_idx_map(update_data.get_individual_scenario(0)));
+    // update all components in the first scenario (e.g. permanent update)
+    template <cache_type_c CacheType> void update_components(ConstDataset const& update_data) {
+        auto const sequence_idx_map = main_core::update_independence::get_sequence_idx_map<ComponentType...>(
+            state_, update_data.get_individual_scenario(0));
+        update_components<CacheType>(update_data, 0, sequence_idx_map);
```
nitbharambe, line 310:
```suggestion
    template <cache_type_c CacheType> void update_components(ConstDataset const& update_data) {
        auto const sequence_idx_map = main_core::update_independence::get_sequence_idx_map<ComponentType...>(
            state_, update_data);
        update_components<CacheType>(update_data, 0, sequence_idx_map);
```

Also i think renaming it to get_all_sequence_idx_map would make it more readable instead

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
...
+                                                 [](auto const& vec) {
+                                                     return std::ranges::any_of(vec, [](auto const& item) {
+                                                         return detail::check_id_na(item);
+                                                     });
+                                                 }) &&
+                             !properties.ids_all_na;
+
+    if (all_spans.empty()) {
+        properties.update_ids_match = true;
+    } else {
```
nitbharambe, line 195: From a previous comment on old PR, early return here please. :D

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
...
+        return get_component_sequence<CompType>(state, std::begin(span), std::end(span), n_comp_elements);
+    };
+    if (update_data.is_columnar(CompType::name)) {
+        auto const buffer_span =
+            update_data.get_columnar_buffer_span<meta_data::update_getter_s, CompType>(scenario_idx);
+        return get_sequence(buffer_span);
+    }
+    auto const buffer_span = update_data.get_buffer_span<meta_data::update_getter_s, CompType>(scenario_idx);
+    return get_sequence(buffer_span);
+}
```
nitbharambe, line 281: I wonder if we can move the if columnar - else logic to dataset as `get_row_or_columnar buffer_span` since we use it everywhere. 
Maybe this is asked before but refresh my memory please.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
...
+
+// main core utils
+template <class CompType, class... ComponentTypes>
+constexpr auto index_of_component = main_core::utils::index_of_component<CompType, ComponentTypes...>;
+template <class... ComponentTypes> using SequenceIdx = main_core::utils::SequenceIdx<ComponentTypes...>;
+template <class... ComponentTypes> using ComponentFlags = main_core::utils::ComponentFlags<ComponentTypes...>;
+
+// get sequence idx map of a certain batch scenario
+template <typename CompType, class ComponentContainer>
+std::vector<Idx2D> get_component_sequence(MainModelState<ComponentContainer> const& state,
```
nitbharambe, line 268: We have many `get_component_sequence` defined at different places doing different things. Lets rename those too.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: Then we might as well call it direct `get_buffer_span` ;)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
...
+        detail::process_buffer_span<CompType>(
+            update_data.template get_columnar_buffer_span_all_scenarios<meta_data::update_getter_s, CompType>(),
+            properties);
+    } else {
+        detail::process_buffer_span<CompType>(
+            update_data.template get_buffer_span_all_scenarios<meta_data::update_getter_s, CompType>(), properties);
+    }
+
+    return properties;
+}
```
Jerry-Jinfeng-Guo, line 236: This is already inside `::detail`?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: Do you mean to put this early return at the very beginning of the function and return for empty input? Then we need to revisit and make sure `ids_all_na` and `ids_part_na` has good defaults.

If not, early return here is not needed.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
...
+    }
+    auto const elements_ps = comp.get_n_elements();
+    assert(comp.uniform || elements_ps < 0);
+
+    if (elements_ps >= 0 && comp.elements_in_base < elements_ps) {
+        throw DatasetError("Update data has more elements per scenario than input data for component " + comp.name +
+                           "!");
+    }
+    if (comp.ids_part_na) {
+        throw DatasetError("Some IDs are not valid for component " + comp.name + " in update data!");
```
Jerry-Jinfeng-Guo, line 250: Now that I re-read this, `valid` can be vague here. I would argue that we go back towards the old one, clear albeit mouthful.
```suggestion
        throw DatasetError("IDs contain both numbers and NANs for component " + comp.name + " in update data!");
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -218,6 +184,18 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
     }
 
+    // helper function to get the number of components per type
+    std::array<Idx, n_types> get_n_components_per_type() const {
+        std::array<Idx, n_types> result{};
+        size_t idx{};
+        main_core::utils::run_functor_with_all_types_return_void<ComponentType...>(
+            [&result, this, &idx]<typename CompType>() {
+                result[idx] = this->component_count<CompType>();
+                ++idx;
```
Jerry-Jinfeng-Guo, line 194: I know this is against all 'best practices' and 'recommendations', but just for the beauty:
```suggestion
                result[idx++] = this->component_count<CompType>();
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: The problematic one lives in `state_querries.hpp` and returns "sequence idx based on id or idx_2d", which is integer rather than a list of them. Other ones are doing the same thing. I would argue that the one in `state_querries.hpp` is named incorrectly. Let's fix that one in a separate PR maybe?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: I actually like your suggestion better. More verbose, but clearer.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
                                      get_component_sequence<Component>(state, begin, end));
 }
 
+namespace update_independence {
+namespace detail {
```
figueroa1395, line 139: Maybe (or maybe not) a better name for this namespace is needed. Suggestions are welcome.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: I already tried it when implementing `get_columnar_buffer_span` a few months ago, as I thought it would be more neat, but failed after several attempts. The biggest problem is that in C++ we can't have two different return types for the same overload. To solve this, one can use `std::any` or `std::variant`, but those come with additional overhead, which we want to avoid at this low level. Hence I was suggested to just use two separate overloads.

Maybe some sort of template shenanigan could be implemented, but my feeling is that the change would be not only difficult but may involve changing several other things. So, we could open some discussion about it if needed, but definitely scope it out of this PR.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
...
+template <class... ComponentTypes>
+ComponentFlags<ComponentTypes...> is_update_independent(ConstDataset const& update_data,
+                                                        std::span<Idx const> relevant_component_count) {
+    ComponentFlags<ComponentTypes...> result{};
+    size_t comp_idx{};
+    utils::run_functor_with_all_types_return_void<ComponentTypes...>([&result, &relevant_component_count, &update_data,
+                                                                      &comp_idx]<typename CompType>() {
+        Idx const n_component = relevant_component_count[comp_idx];
+        result[comp_idx] = detail::check_component_independence<CompType>(update_data, n_component).is_independent();
+        ++comp_idx;
```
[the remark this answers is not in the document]
figueroa1395: I personally prefer it as is. I find it more readably. But this is a matter of style only I suppose, right?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: See other comment.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: I am confused, the function is marked as `void`. What do you mean?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: `detail` is fine I think

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,4 +135,194 @@ inline void update_inverse(MainModelState<ComponentContainer> const& state, Forw
                                      get_component_sequence<Component>(state, begin, end));
 }
 
+namespace update_independence {
```
Jerry-Jinfeng-Guo, line 138: I would instead argue for a different name here
```suggestion
namespace update {
```

Jerry-Jinfeng-Guo:
> I just realized that `check_id_na` was supposed to be moved to `dataset.hpp`. I will address that on Monday, but besides that, everything else is ready.

Is this still relevant?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: I thought about `update` as well, but then the other functions in the file may as well go in there and I wanted to differentiate them. What do you think about that? Do you think renaming as `update` is good if we keep the function locations as is?

figueroa1395:
> > I just realized that `check_id_na` was supposed to be moved to `dataset.hpp`. I will address that on Monday, but besides that, everything else is ready.
> 
> Is this still relevant?

Yes. Thanks for the reminder, I had forgotten.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: It's 'just for the beauty' ;)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: In that case, I would think that moving the following three functions to live outside the namespaces is better:  `get_component_sequence`, `get_all_sequence_idx_map` x2.

figueroa1395: I actually really like this. `get_component_sequence` from `update.hpp` is actually now only used in `get_all_sequence_idx_map`. So I am going to move all flavors of `get_component_sequence` (and try to remove some if possible) into `namespace detail` and leave `get_all_sequence_idx_map` in the main `namespace update`. That way we also remove the ambiguity from `get_component_sequence` from `state_queries.hpp` which, on the contrary, is used across the whole codebase. In that sense, `namespace update_independence` can safely be renamed to `update`. I'll tag the commit later here.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
nitbharambe:
```suggestion
    if (all_spans.empty()) {
        properties.update_ids_match = true;
        return;
    }
```
And remove bracket below.
And @Jerry-Jinfeng-Guo , you are right. Hence, early return at this place itself.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
nitbharambe: @Jerry-Jinfeng-Guo Indeed!
Even without considering that problematic one, we have 2 variations here. 

I have a personal preference to avoid relying on function overloading unless for type variations or for special cases of templating etc.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: See [67929c5](https://github.com/PowerGridModel/power-grid-model/pull/828/commits/67929c5d2e9879f28f75620ce7de412ecaa0ce92)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: See [67929c5](https://github.com/PowerGridModel/power-grid-model/pull/828/commits/67929c5d2e9879f28f75620ce7de412ecaa0ce92). Let me know what you think.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -26,14 +38,150 @@ inline void iterate_component_sequence(Func&& func, ForwardIterator begin, Forwa
...
+                                                     return std::ranges::any_of(vec, [](auto const& item) {
+                                                         return detail::check_id_na(item);
+                                                     });
+                                                 }) &&
+                             !properties.ids_all_na;
+
+    if (all_spans.empty()) {
+        properties.update_ids_match = true;
+        return;
+    } else {
```
nitbharambe, line 98: Now this else statement and its { } is not necessary.

nitbharambe (approved):
LGTM

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/utils.hpp
Jerry-Jinfeng-Guo, line 1: Can we name this file something like `core_utils.hpp`? I understand that this is the first name in the core with this name, (which is actually a surprise) just for the sake of being more specific and not dragging into another duplication hell.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -53,14 +198,62 @@ inline void get_component_sequence(MainModelState<ComponentContainer> const& sta
 template <component_c Component, class ComponentContainer,
           forward_iterator_like<typename Component::UpdateType> ForwardIterator>
     requires model_component_state_c<MainModelState, ComponentContainer, Component>
-inline std::vector<Idx2D> get_component_sequence(MainModelState<ComponentContainer> const& state, ForwardIterator begin,
-                                                 ForwardIterator end, Idx n_comp_elements = na_Idx) {
+inline std::vector<Idx2D> get_component_sequence_by_it(MainModelState<ComponentContainer> const& state,
```
Jerry-Jinfeng-Guo, line 201: Please rename this to `get_component_sequence_by_iter` / `get_component_sequence_by_range` or anything that does not remotely imply a terrifying figure...

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -53,14 +198,62 @@ inline void get_component_sequence(MainModelState<ComponentContainer> const& sta
...
+// The sequence idx map of the batch is the same as that of the first scenario in the batch (assuming homogeneity)
+// This is the entry point for permanent updates.
+template <class... ComponentTypes, class ComponentContainer>
+SequenceIdx<ComponentTypes...> get_all_sequence_idx_map(MainModelState<ComponentContainer> const& state,
+                                                        ConstDataset const& update_data) {
+    constexpr ComponentFlags<ComponentTypes...> all_true = [] {
+        ComponentFlags<ComponentTypes...> result{};
+        std::ranges::fill(result, true);
+        return result;
+    }();
```
Jerry-Jinfeng-Guo, line 253: Should this be dependent on the `update_data` rather than the whole picture?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -691,8 +691,8 @@ TEST_CASE("Test Tap position optimizer") {
         auto const& transformers_dataset =
             update_dataset.get_buffer_span<meta_data::update_getter_s, MockTransformer>();
         auto changed_components = std::vector<Idx2D>{};
-        main_core::update_component<MockTransformer>(state, transformers_dataset.begin(), transformers_dataset.end(),
-                                                     std::back_inserter(changed_components));
+        main_core::update::update_component<MockTransformer>(
+            state, transformers_dataset.begin(), transformers_dataset.end(), std::back_inserter(changed_components));
     };
```
Jerry-Jinfeng-Guo, line 696: Wondering if we could make use of the ranges stuff to pass in directly here `transformers_dataset` rather than explicitly specify `begin` and `end`

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -300,8 +278,9 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
         }
     }
 
+    // helper for row or columnar-based components
     template <class CompType, cache_type_c CacheType>
-    void update_component(ConstDataset const& update_data, Idx pos, std::span<Idx2D const> sequence_idx) {
+    void update_component_row_col(ConstDataset const& update_data, Idx pos, std::span<Idx2D const> sequence_idx) {
```
Jerry-Jinfeng-Guo, line 283: Why is this necessary?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -26,14 +38,147 @@ inline void iterate_component_sequence(Func&& func, ForwardIterator begin, Forwa
...
+
+    if (all_spans.empty()) {
+        properties.update_ids_match = true;
+        return;
+    }
+    // Remember the begin iterator of the first scenario, then loop over the remaining scenarios and
+    // check the ids
+    auto const first_span = all_spans[0];
+    // check the subsequent scenarios
+    // only return true if all scenarios match the ids of the first batch
```
Jerry-Jinfeng-Guo, line 103:
```suggestion
    // only return true if ids of all scenarios match the ids of the first batch
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: It isn't. I was trying to remove so many overloads, hence I "added" `update_components` and this one. Since the whole purpose of this one is only to encapsulate the logic where one updates a columnar or a row-based component, I thought it was a good naming convention.

Do you think is more readable if we just change it back to `update_component`? Different name maybe? Or do you have something else in mind?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -128,58 +129,22 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-        constexpr bool is_independent() const { return qualify_for_optional_id() || provided_ids_valid(); }
-        constexpr Idx get_n_elements() const {
-            assert(uniform || elements_ps_in_update == invalid_index);
-
-            return qualify_for_optional_id() ? elements_ps_in_update : na_Idx;
-        }
-    };
-    using UpdateCompIndependence = std::array<UpdateCompProperties, n_types>;
-    using ComponentFlags = std::array<bool, n_types>;
     using ComponentCountInBase = std::pair<std::string, Idx>;
```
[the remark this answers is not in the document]
figueroa1395: It isn't used in `update.hpp` at all, as I removed string lookup. Hence, now it is only needed for `all_component_count`, which actually isn't even used anymore at all in `main_model_impl.hpp` itself, but only in `main_model.hpp`; however, I can't find it being used or called at all anywhere else in the codebase.

So two action points here:

- I will remove `ComponentCountInBase` altogether as it doesn't bring anything.
- Should we remove `all_component_count` from both `main_model_impl.hpp` and `main_model.hpp`, or is this one of those functions that may be useful later on so it is better it keep it? If the later is correct, then we should at least remove it from `main_model.hpp`. What do you think?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
figueroa1395: We might, but that would probably involve changing the implementation of most (if not all) `update_component` functions, which is out of scope here.

On the other hand, this may be a good thing to do "everywhere" when doing the major overhaul of the core: just use ranges stuff everywhere when possible and avoid being so explicit with iterators. Would you agree?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: Because this overload of `get_all_sequence_idx_map` is only used for permanent updates, where only the first scenario is considered, hence it is always independent.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: I got you!

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: Permanent updates are not necessarily updates of all components, right? What I meant to say is that yes it should be considered independent, but only for the components contained in the update data.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
Jerry-Jinfeng-Guo: > Do you think is more readable if we just change it back to `update_component`? 

I would argue changing it back but for a different reason: having it specified as `*_row_col` indicates that there are other counterparts. I think if you put some comment to explain the overload functions, that would already help a lot. E.g., "this is the entry point for update all sort of components in update data", "this is the overload to update ***", etc.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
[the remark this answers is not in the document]
figueroa1395: I agree. Specially since this is not just a common util. I will remove entirely then.

Jerry-Jinfeng-Guo:
General question regarding this: is there a follow up to this topic regarding the refactoring? If there is, we can leave some of the minor comments I have for that one since this PR is more about moving stuff around -- re-structuring

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: You are right and I will take a look at it. However, is it worth to do it for specific components in this case though? Since it they are all going to be independent anyway, why search for the specific update ones?

figueroa1395: After looking at implementation, this would mean that we would need to iterate over each component type and then do a string look-up (via `update_data.find_component`) to determine which components from input are present in the update, and then only update those, which becomes unnecesarily expensive. 

So in this case, I will add a comment explaining why (it is easier and faster) this choice. Would that suffice?

Jerry-Jinfeng-Guo: I don't think it's more expensive than the full list 'expansion'. My argument is that this way a vague communication between functions is introduced. Either we create a special state stating that it is `{'all', independent}` like in the very original implementation, or we always tailor the result. Think for example in the future when we debug this.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: Let's make that one of our todo points to see in the future if/when we can do it.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Then perhaps it could simply only live in tests? XD

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
figueroa1395: I have added it to the discussion about possible future changes so it lives somewhere and we don't forget: https://github.com/orgs/PowerGridModel/discussions/39#discussioncomment-11332885. Closing comment now.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: the check is compile-time resolved. yes, you do do a loop over all component types, but also immediately check that the size is zero and early-out. checking for every component type whether it is in fact in the update data would be like implementing the same code twice. if there's any performance overhead, it should be negligible.

i prefer to keep the code as simple as possible. if changing the entry-point would make it more readable without significant performance overhead, go ahead, but otherwise, i see no reason to change this.

mgovers:
> > General question regarding this: is there a follow up to this topic regarding the refactoring? If there is, we can leave some of the minor comments I have for that one since this PR is more about moving stuff around -- re-structuring
> 
> To my understanding there isn't a direct follow-up refactor as of now, besides the big one about the whole core. So I would say that if it is related with this scope, independence or update-ish, I can try to fix it at once here so we don't keep postponing it.
> 
> If it is surrounding this and smallish, I can just create a separate small PR and tackle it right away since it is fresh.
> 
> Else, we should probably scope it out.

agreed

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: That still does not address my question of returning something that does not truly reflect the state of the update data.

Jerry-Jinfeng-Guo:
> > > General question regarding this: is there a follow up to this topic regarding the refactoring? If there is, we can leave some of the minor comments I have for that one since this PR is more about moving stuff around -- re-structuring
> > 
> > 
> > To my understanding there isn't a direct follow-up refactor as of now, besides the big one about the whole core. So I would say that if it is related with this scope, independence or update-ish, I can try to fix it at once here so we don't keep postponing it.
> > If it is surrounding this and smallish, I can just create a separate small PR and tackle it right away since it is fresh.
> > Else, we should probably scope it out.
> 
> agreed: the refactoring is already big as it is. splitting up is the way to go, regardless of whether there are follow-ups or not

How do you split things up when there is not follow-up?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/utils.hpp
Jerry-Jinfeng-Guo: Is this comment resolved because the suggestion is taken or...?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: > Edit: I see it is actually used in some tests (which I missed before). So I will leave it be, for now.

Hmm..

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/utils.hpp
figueroa1395: Yes, hence the thumbs up. See [4022bac](https://github.com/PowerGridModel/power-grid-model/pull/828/commits/4022bac6480b3c64e0ff33ca6ceec878cbde1585)

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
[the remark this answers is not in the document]
figueroa1395: I will push in a few and you'll see the changes reflected. The tests were actually just testing the `all_component_count` function itself. So everything is removed altogether.

mgovers:
> > > > General question regarding this: is there a follow up to this topic regarding the refactoring? If there is, we can leave some of the minor comments I have for that one since this PR is more about moving stuff around -- re-structuring
> > > 
> > > 
> > > To my understanding there isn't a direct follow-up refactor as of now, besides the big one about the whole core. So I would say that if it is related with this scope, independence or update-ish, I can try to fix it at once here so we don't keep postponing it.
> > > If it is surrounding this and smallish, I can just create a separate small PR and tackle it right away since it is fresh.
> > > Else, we should probably scope it out.
> > 
> > 
> > agreed: the refactoring is already big as it is. splitting up is the way to go, regardless of whether there are follow-ups or not
> 
> How do you split things up when there is no follow-up?

you don't, that's my point. at some point, the scope is "large enough" and you should focus on what's currently already in scope. anything new that pops up can be moved to a follow-up (even if it's only minor). if there are no follow-ups: great

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: soooo you're worried about the name `result`?
```suggestion
    constexpr ComponentFlags<ComponentTypes...> all_true = [] {
        ComponentFlags<ComponentTypes...> to_be_filled{};
        std::ranges::fill(to_be_filled, true);
        return to_be_filled;
    }();
```
maybe i'm missing which part of your question is about returning?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/utils.hpp
Jerry-Jinfeng-Guo: Well, you will have to blame github; it was still the old version when I refreshed the page before leaving this comment.

Jerry-Jinfeng-Guo:
> you don't, that's my point. at some point, the scope is "large enough" and you should focus on what's currently already in scope. anything new that pops up can be moved to a follow-up (even if it's only minor). if there are no follow-ups: great

This is self contradictory statement.

mgovers:
> > you don't, that's my point. at some point, the scope is "large enough" and you should focus on what's currently already in scope. anything new that pops up can be moved to a follow-up (even if it's only minor). if there are no follow-ups: great
> 
> This is self contradictory statement.

which part?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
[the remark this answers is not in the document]
figueroa1395: On the contrary, I do like it, that is if we use it responsibly. Let's keep it and I'll personally keep a look at it. I think there should be a place for common core stuff, provided we actually only dump core utils stuff there.

Jerry-Jinfeng-Guo:
> > > you don't, that's my point. at some point, the scope is "large enough" and you should focus on what's currently already in scope. anything new that pops up can be moved to a follow-up (even if it's only minor). if there are no follow-ups: great
> > 
> > 
> > This is self contradictory statement.
> 
> which part?

"anything new that pops up can be moved to a follow-up (even if it's only minor)" vs "if there are no follow-ups: great"
My original question is tied to this PR, not a theoretical question. You agreed to scope this PR, sure, no problem, on the condition the mentioned things will be addressed in a follow-up, which is the premise of the whole discussion.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: 100% agree with this.

a potential solution would probably look similar to the component type functor application, but, for now, it's wayyyyyy overkill

tests/cpp_integration_tests/test_main_model.cpp
```diff
@@ -808,7 +808,7 @@ TEST_CASE_TEMPLATE("Test main model - unknown id", settings, regular_update, cac
     std::vector<SourceUpdate> const source_update2{SourceUpdate{100, true, nan, nan}};
     ConstDataset update_data{false, 1, "update", meta_data::meta_data_gen::meta_data};
     update_data.add_buffer("source", source_update2.size(), source_update2.size(), nullptr, source_update2.data());
-    CHECK_THROWS_AS((main_model.update_component<typename settings::update_type>(update_data)), IDNotFound);
+    CHECK_THROWS_AS((main_model.update_components<typename settings::update_type>(update_data)), IDNotFound);
```
mgovers, line 811: yay, merge conflicts 🫠 (easy to resolve though, just resolve this comment)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -26,14 +38,148 @@ inline void iterate_component_sequence(Func&& func, ForwardIterator begin, Forwa
...
+                                                 }) &&
+                             !properties.ids_all_na;
+
+    if (all_spans.empty()) {
+        properties.update_ids_match = true;
+        return;
+    }
+    // Remember the begin iterator of the first scenario, then loop over the remaining scenarios and
+    // check the ids
+    auto const first_span = all_spans[0];
```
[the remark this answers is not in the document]
figueroa1395: Actually, it is `auto const first_span = all_spans.front();`

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -53,14 +199,64 @@ inline void get_component_sequence(MainModelState<ComponentContainer> const& sta
...
+            update_data.get_columnar_buffer_span<meta_data::update_getter_s, CompType>(scenario_idx);
+        return get_sequence(buffer_span);
+    }
+    auto const buffer_span = update_data.get_buffer_span<meta_data::update_getter_s, CompType>(scenario_idx);
+    return get_sequence(buffer_span);
+}
+} // namespace detail
+
+template <class... ComponentTypes, class ComponentContainer>
+SequenceIdx<ComponentTypes...> get_all_sequence_idx_map(MainModelState<ComponentContainer> const& state,
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: I kinda wanted to mention this as well but hesitated because of the scope. It's a sequence of `Idx2D`, since we do have `idx`, to be more clear, all this chain of functions should be named `get_*Idx2D_sequence`.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -1499,4 +1306,22 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
+}
+
+template <calculation_input_type CalcInputType>
+static auto calculate_param(auto const& c, auto const&... extra_args)
+    requires requires {
+        { c.template calc_param<typename CalcInputType::sym>(extra_args...) };
+    }
+{
+    return c.template calc_param<typename CalcInputType::sym>(extra_args...);
+}
```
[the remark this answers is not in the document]
figueroa1395: Moved them to `detail` as suggested. 

On the other hand, I had not noticed what you point out and, after a bit of reading, I found that this only works because the `calculate_param` functions are templated. Why? Because the compiler first parses the whole file, but only instantiates them once the template type is known later on. So as long as the functions are templated and within the same translation unit where they are called, there is no need to forward declare them. Pretty cool find :).

Edit: Funny enough, if I add it to another namespace, i.e. detail, it doesn't work anymore. Hence I just moved it all the way up.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
```diff
@@ -41,7 +41,7 @@ constexpr Idx get_component_group_idx(MainModelState<ComponentContainer> const&
 
 template <typename ComponentType, class ComponentContainer>
     requires model_component_state_c<MainModelState, ComponentContainer, ComponentType>
-inline Idx get_component_sequence(MainModelState<ComponentContainer> const& state, auto const& id_or_index) {
+inline Idx get_component_sequence_idx(MainModelState<ComponentContainer> const& state, auto const& id_or_index) {
```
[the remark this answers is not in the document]
figueroa1395: For this PR I'll leave this one as is, but I will add it to the discussion for the major overhaul, because if we dive deep into this, we may as well unify it accross the whole codebase.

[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model/issues/839

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model/issues/839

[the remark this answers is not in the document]
figueroa1395: > #839

Resolving due to this.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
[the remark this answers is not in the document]
figueroa1395: > #839

Resolving due to this.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -4,12 +4,24 @@
...
+#include <map>
+
+namespace power_grid_model::main_core::update {
+
+constexpr Idx invalid_index{-1};
+
+// main core utils
+template <class CompType, class... ComponentTypes>
+constexpr auto index_of_component = main_core::utils::index_of_component<CompType, ComponentTypes...>;
+template <class... ComponentTypes> using SequenceIdx = main_core::utils::SequenceIdx<ComponentTypes...>;
```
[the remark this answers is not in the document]
figueroa1395: I just forgot to clean up after I moved things to utils. Done now :)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: good test case for a follow-up

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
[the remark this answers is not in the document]
mgovers: welp 🤦

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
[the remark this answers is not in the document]
mgovers: yeah i realized as well but didn't feel like editing :joy: you knew what i meant

figueroa1395:
[c56adb2](https://github.com/PowerGridModel/power-grid-model/pull/828/commits/c56adb238362b2a292d1de8378fbe4d2489d57fb) Still not finished as now there are some memory leaks. Either it is something very "simple" I missed, or something very complex and then we may want a follow up to separate from this already large PR.

TonyXiang8787:
@mgovers, @figueroa1395, @Jerry-Jinfeng-Guo 

I see the PR was initially green but after some commits it turns to red.

Maybe we can revert some refactors and merge this?

mgovers:
> @mgovers, @figueroa1395, @Jerry-Jinfeng-Guo
> 
> I see the PR was initially green but after some commits it turns to red.
> 
> Maybe we can revert some refactors and merge this?

i think so as well. it's very easy to revert the last 2 commits and then reapply in a new PR or cherry-pick the changes.

Jerry-Jinfeng-Guo:
> > @mgovers, @figueroa1395, @Jerry-Jinfeng-Guo
> > I see the PR was initially green but after some commits it turns to red.
> > Maybe we can revert some refactors and merge this?
> 
> i think so as well. it's very easy to revert the last 2 commits and then reapply in a new PR or cherry-pick the changes.

Let's indeed make the additional commits a follow-up PR and merge this one for now.

figueroa1395:
> @mgovers, @figueroa1395, @Jerry-Jinfeng-Guo
> 
> I see the PR was initially green but after some commits it turns to red.
> 
> Maybe we can revert some refactors and merge this?

Reverted the last "problematic" commit. Will merge after CI passes and then open another PR.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -1440,8 +1265,11 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
 
             // Check the branch and shunt indices
             constexpr auto branch_param_in_seq_map =
-                std::array{index_of_component<Line>, index_of_component<Link>, index_of_component<Transformer>};
-            constexpr auto shunt_param_in_seq_map = std::array{index_of_component<Shunt>};
+                std::array{main_core::utils::index_of_component<Line, ComponentType...>,
+                           main_core::utils::index_of_component<Link, ComponentType...>,
+                           main_core::utils::index_of_component<Transformer, ComponentType...>};
```
mgovers, line 1270: oof :joy: (maybe in the follow-up PR) can you make it a shorthand within this function? (may not be possible because of template stuff) or something along the lines of

```cpp
constexpr auto index_of = []<typename CompType> { return index_of_component<CompType, ComponentType...>; };
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -53,14 +193,65 @@ inline void get_component_sequence(MainModelState<ComponentContainer> const& sta
...
+            update_data.get_columnar_buffer_span<meta_data::update_getter_s, CompType>(scenario_idx);
+        return get_sequence(buffer_span);
+    }
+    auto const buffer_span = update_data.get_buffer_span<meta_data::update_getter_s, CompType>(scenario_idx);
+    return get_sequence(buffer_span);
+}
+} // namespace detail
+
+template <class... ComponentTypes, class ComponentContainer>
+main_core::utils::SequenceIdx<ComponentTypes...>
```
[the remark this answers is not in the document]
figueroa1395: Follow-up PR.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -128,58 +149,13 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-    // run functors with all component types
-    template <class Functor> static constexpr void run_functor_with_all_types_return_void(Functor functor) {
-        (functor.template operator()<ComponentType>(), ...);
-    }
-    template <class Functor> static constexpr auto run_functor_with_all_types_return_array(Functor functor) {
-        return std::array { functor.template operator()<ComponentType>()... };
-    }
+    static constexpr Idx isolated_component{-1};
+    static constexpr Idx not_connected{-1};
+    static constexpr Idx sequential{-1};
```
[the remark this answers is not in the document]
figueroa1395: Follow-up PR.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: Follow-up PR.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -128,58 +149,13 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
 
-    template <class CT>
-    static constexpr size_t index_of_component = container_impl::get_cls_pos_v<CT, ComponentType...>;
-
-    static constexpr size_t n_types = sizeof...(ComponentType);
-
-    using SequenceIdx = std::array<std::vector<Idx2D>, n_types>;
-    using SequenceIdxView = std::array<std::span<Idx2D const>, n_types>;
-
+    using SequenceIdxView = std::array<std::span<Idx2D const>, main_core::utils::n_types<ComponentType...>>;
```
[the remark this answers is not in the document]
figueroa1395: Follow-up PR.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "main_model_impl.hpp"
+#include <memory>
```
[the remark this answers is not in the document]
figueroa1395: Follow-up PR.

figueroa1395:
Merge only after https://github.com/PowerGridModel/power-grid-model/pull/833 is merged, just for safety. That is why I added the `do-not-merge` label.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: this conversation is still open

mgovers (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: Scoped out for a follow up PR

Issue #786, opened 2024-10-15
[FEATURE] Columnar data creation helper function


nitbharambe:
A `initialize_column(data_type, component, shape, attributes, empty=False)` helper function which gives out columnar data ie. dict of arrays.

## Motivation:

We have `initialize_array` helper function to make users not bother about the dtype. A same one for columnar data.

If the user has to create a columnar data now, they would have to do :
`{k: v for k,v in initialize_array(dataset_type, component_type, shape).items()}` 
or 
`{k: np.empty(shape=...,  dtype=power_grid_meta_data[dataset_type][component_type][k]) for k in attributes}` 
or some similar combination

Also in first way, user creates a huge array; subset of which is then converted to dictionary. 

Options:

initialize_column(data_type, component, shape, attribute, empty=False) -> np.ndarray

initialize_columnar_data(data_type, component, shape, attributes, empty=False) -> dict[str, np.ndarray]

mgovers:
`initialize_array(data_type, component, shape, empty=False, *, attributes=None)`:

petersalemink95:
> decision: use `initialize_array(data_type, component, shape, empty=False, *, attributes=None)` with additional kwarg

To be clear: `attributes=None` should refer to the row based initialization, that already is present. For columnar people should specify a "filter" in `attributes=`, similar as in the data filter

Jerry-Jinfeng-Guo:
@nitbharambe Is this still open?

PR #830, opened 2024-11-13
Feature / Columnar Data helper function

nitbharambe:
Implements https://github.com/PowerGridModel/power-grid-model/issues/786

### DIFF
--- a/src/power_grid_model/_core/power_grid_meta.py
+++ b/src/power_grid_model/_core/power_grid_meta.py
@@ -8,7 +8,7 @@ Load meta data from C core and define numpy structured array
 
 from dataclasses import dataclass
 from enum import IntEnum
-from typing import Any
+from typing import Any, Iterable
 
 import numpy as np
 
@@ -21,7 +21,8 @@ from power_grid_model._core.dataset_definitions import (
     _str_to_datatype,
 )
 from power_grid_model._core.power_grid_core import AttributePtr, ComponentPtr, DatasetPtr, power_grid_core as pgc
-from power_grid_model.data_types import DenseBatchArray, SingleArray
+from power_grid_model.data_types import DenseBatchData, SingleComponentData
+from power_grid_model.errors import PowerGridUnreachableHitError
 
 
 # constant enum for ctype
@@ -178,7 +179,9 @@ def initialize_array(
     component_type: ComponentTypeLike,
     shape: tuple | int,
     empty: bool = False,
-) -> SingleArray | DenseBatchArray:
+    *,
+    attributes: Iterable[str] | None = None,
+) -> SingleComponentData | DenseBatchData:
     """
     Initializes an array for use in Power Grid Model calculations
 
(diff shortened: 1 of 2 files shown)


src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -197,11 +200,29 @@ def initialize_array(
...
+        return np.full(
+            shape=shape,
+            fill_value=component_meta.nan_scalar,
+            dtype=component_meta.dtype,
+            order="C",
+        )
+    data = {}
+    for attribute in attributes:
+        if component_meta.dtype.names is None:
+            raise PowerGridUnreachableHitError
```
mgovers, line 216: you need to `raise` an object, not a type
```suggestion
            raise PowerGridUnreachableHitError()
```

src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -197,11 +200,29 @@ def initialize_array(
...
+            raise ValueError(f"Attribute {attribute} is not available for {component_type}")
+        if empty:
+            data[attribute] = np.empty(shape=shape, dtype=component_meta.dtype[attribute], order="C")
+        else:
+            data[attribute] = np.full(
+                shape=shape,
+                fill_value=component_meta.nan_scalar[attribute],
+                dtype=component_meta.dtype[attribute],
+                order="C",
+            )
```
mgovers, line 227: do you think it makes to put this in a (private) function since it's the same logic as before but with a different dtype?

tests/unit/test_meta_data.py
```diff
@@ -20,6 +20,20 @@ def test_initialize_array():
     assert np.all(np.isnan(arr_2d["u_rated"]))
 
 
+def test_initialize_array__columnar():
+    some_attributes = ["id", "from_node", "r1"]
+    actual = initialize_array("input", "line", 3, attributes=some_attributes)
```
mgovers, line 25:
```suggestion
    actual = initialize_array(DatasetType.input, ComponentType.line, 3, attributes=some_attributes)
```

tests/unit/test_meta_data.py
```diff
@@ -20,6 +20,20 @@ def test_initialize_array():
...
+    assert isinstance(actual, dict)
+    assert actual.keys() == set(some_attributes)
+    assert all(v.shape == (3,) for v in actual.values())
+    assert np.all(np.isnan(actual["r1"]))
+
+    actual_2d = initialize_array("input", "line", (2, 3), attributes=some_attributes)
```
mgovers, line 31:
```suggestion
    actual_2d = initialize_array(DatasetType.input, ComponentType.line, (2, 3), attributes=some_attributes)
```

src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -178,7 +179,9 @@ def initialize_array(
     component_type: ComponentTypeLike,
     shape: tuple | int,
     empty: bool = False,
-) -> SingleArray | DenseBatchArray:
+    *,
+    attributes: Iterable[str] | None = None,
```
mgovers, line 183: do we also want to support `everything` (genuine question, feel free to close it if not)?

src/power_grid_model/_core/power_grid_meta.py
nitbharambe: Thought about it, as it is all same logic. But they are essentially all single lines.
I think it would not benefit readability much. Hence kept it as such

nitbharambe:
Issue with type hinting:

`initialize_array` now gives out row based data, hence the type is `np.ndarray`.
When we assign like how it is done in example, `node["id"] = [1,2,3]`, we expect to use assignment of attribute within the array.

If we are to implement this feature, the return type is `np.ndarray  | dict[str, np.ndarray]`. 
If user tries to do the same assignment in case of the columnar data, ie. when  node is `dict[str, np.ndarray]`, the `[]` operator refers to dictionary and thus a numpy array gets replaced with a list. Hence, this correctly gives out error with mypy.

Thus this helper function might end up confusing the user. If the user does the assigning / creation etc themselves via power_grid_meta_data, such operations might be more clear to them.

figueroa1395:
@nitbharambe Looks like you have to manually delete the branch due to unmerged commits. Don't forget :)

PR #834, opened 2024-11-18
Feature/attribute-dtype-helper-function

nitbharambe:
**Fixes issue**: https://github.com/PowerGridModel/power-grid-model/issues/786

New helper functions for columnar data attribute_dtype and attribute_empty_value


src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -205,3 +205,41 @@ def initialize_array(
...
+    component_type = _str_to_component_type(component_type)
+    return power_grid_meta_data[data_type][component_type].dtype[attribute]
+
+
+def attribute_empty_value(
+    data_type: DatasetTypeLike, component_type: ComponentTypeLike, attribute: AttributeType
+) -> np.ndarray:
+    """
+    Returns the empty value for a specific attribute in the Power Grid Model.
+
```
nitbharambe, line 233:
```suggestion
```
Extra space

src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -205,3 +205,40 @@ def initialize_array(
         dtype=power_grid_meta_data[data_type][component_type].dtype,
         order="C",
     )
+
+
+def attribute_dtype(
```
Jerry-Jinfeng-Guo, line 210: I recall we agreed to  name it `get_attribute_dtype`?

TonyXiang8787: I recall we agreed to name it as `attribute_dtype`

docs/examples/Power Flow Example.ipynb
```diff
@@ -403,7 +402,7 @@
       "List of attribute types in line result\n",
       "['id', 'p_from']\n",
       "------line result------\n",
-      "{'id': array([3, 5, 8], dtype=int32), 'p_from': array([17360100.20222374, -3365613.74450156, 13854413.52498137])}\n"
+      "{'id': array([3, 5, 8]), 'p_from': array([17360100.20222374, -3365613.74450156, 13854413.52498137])}\n"
```
mgovers, line 405: is this change right? (either way, it's weird, because `ID` should be `int64`...)

docs/examples/Power Flow Example.ipynb
```diff
@@ -1470,7 +1469,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.13.0"
+   "version": "3.12.7"
```
mgovers, line 1472: can you please upgrade to python 3.13? it's better to be as recent as possible

src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -205,3 +205,40 @@ def initialize_array(
...
+
+    Returns:
+        np.dtype: The dtype of the specified attribute
+    """
+    data_type = _str_to_datatype(data_type)
+    component_type = _str_to_component_type(component_type)
+    return power_grid_meta_data[data_type][component_type].dtype[attribute]
+
+
+def attribute_empty_value(
```
figueroa1395, line 228: Maybe consider naming this one to something more expressive? Like `set_empty_attribute` , `set_nan_attribute`, or something along those lines?

src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -205,3 +205,40 @@ def initialize_array(
         dtype=power_grid_meta_data[data_type][component_type].dtype,
         order="C",
     )
+
+
+def attribute_dtype(
+    data_type: DatasetTypeLike, component_type: ComponentTypeLike, attribute: AttributeType
```
figueroa1395, line 211: I am unsure, but shouldn't we use `DatasetTypeVar` and `ComponentTypeLike` instead? Also, we expose these two in the docs, whereas the initial two we don't.

tests/unit/test_meta_data.py
```diff
@@ -51,14 +74,17 @@ def test_sensor_meta_data():
...
-        "input",
-        "update",
-        "sym_output",
-        "asym_output",
-        "sc_output",
+        DatasetType.input,
+        DatasetType.update,
+        DatasetType.sym_output,
+        DatasetType.asym_output,
+        DatasetType.sc_output,
```
mgovers, line 89: we need both the `str` and the `DatasetType` tests here, because it involves a backwards-compatibility check
```suggestion
    assert set(power_grid_meta_data.keys()) == {
        "input",
        "update",
        "sym_output",
        "asym_output",
        "sc_output",
        DatasetType.input,
        DatasetType.update,
        DatasetType.sym_output,
        DatasetType.asym_output,
        DatasetType.sc_output,
```

src/power_grid_model/_core/power_grid_meta.py
Jerry-Jinfeng-Guo: Wouldn't naming it per class be more intuitive since it's predicate here?

docs/examples/Power Flow Example.ipynb
TonyXiang8787: This is right. `ID` is always `int32`. We don't have a `int64` attribute yet.

mgovers: ohhh right, sorry, apologies. `Idx` is `int64`, `ID` is `int32`, my bad

docs/examples/Power Flow Example.ipynb
figueroa1395, line 1472: Nitpick: Keep 3.13

docs/examples/Power Flow Example.ipynb
```diff
@@ -1404,16 +1403,16 @@
...
+      "[[9.99401170e-001 9.92685785e-001 9.94521366e-001]\n",
+      " [9.99347687e-001 9.86226389e-001 9.89352855e-001]\n",
+      " [9.99288384e-001 9.79654011e-001 9.84095542e-001]\n",
+      " [6.95296562e-310 8.59448882e-312 6.95296562e-310]\n",
+      " [9.99151380e-001 9.66149483e-001 9.73298790e-001]\n",
+      " [9.99073166e-001 9.59205860e-001 9.67750710e-001]\n",
+      " [9.98988099e-001 9.52126208e-001 9.62096474e-001]\n",
+      " [8.59448882e-312 8.59448882e-312 6.95296562e-310]\n",
+      " [9.98796126e-001 9.37530046e-001 9.50447962e-001]\n",
+      " [9.98688504e-001 9.29997471e-001 9.44441670e-001]]\n",
```
figueroa1395, line 1415: The results are the same, but I am very curious why did the formatting change so much.

src/power_grid_model/_core/power_grid_meta.py
figueroa1395: You are right, then the naming makes perfect sense. Resolving.

docs/examples/Power Flow Example.ipynb
nitbharambe: Because the garbage values were 0.0 before now they are some e-312

src/power_grid_model/_core/power_grid_meta.py
nitbharambe: We have: 
```
DatasetTypeLike: TypeAlias = DatasetType | str
DatasetTypeVar = TypeVar("DatasetTypeVar", bound=DatasetTypeLike)  # helper used for type deduction

ComponentTypeLike: TypeAlias = ComponentType | str
ComponentTypeVar = TypeVar("ComponentTypeVar", bound=ComponentTypeLike)  # helper used for type deduction
```

I basically used the same as initialize_array.
Dont remember the exact reasoning back then.

src/power_grid_model/_core/power_grid_meta.py
nitbharambe: @Jerry-Jinfeng-Guo  As in source_dtype, line_dtype?
Its the same effort. No reason to add a whole lot of imports because of it.

docs/examples/Power Flow Example.ipynb
figueroa1395: but also the `e-001` ones. Just something funny I wanted to point out. I'll resolve.

src/power_grid_model/_core/power_grid_meta.py
nitbharambe: @figueroa1395 Do you have a preference?

figueroa1395: I prefer just to have one function with three arguments, as it is already.

Regarding the naming: Since it is intended to be used via the keyword argument `dtype`, I think `attribute_dtype` is fine, or else it is a bit long. If the usage was to be other, then I would prefer the use of the `get_*` prefix, but it is not the case.

figueroa1395 (approved):

TonyXiang8787 (approved):

docs/examples/Power Flow Example.ipynb
mgovers: could be python3.12 vs python3.13 or windows vs unix or arm vs x86

PR #841, opened 2024-11-25
Keep only important integration tests: step 2

mgovers:
Relates to #833 
Relates to #486

* [x] move all tests in `test_main_model_sc.cpp` to validation cases
* [x] remove already duplicate tests in `test_main_model.cpp`


TonyXiang8787:
@mgovers  one suggestion. Sensor output in power flow calculations is undefined. We should not test them. So remove the tests in the current `test_main_model.cpp`, and do not re-test them.

mgovers:
> @mgovers one suggestion. Sensor output in power flow calculations is undefined. We should not test them. So remove the tests in the current `test_main_model.cpp`, and do not re-test them.

while i'm OK with not testing, it does feel a little bit weird to call it UB, though, because we **do** by default output it if there are sensors in the grid

TonyXiang8787:
> > @mgovers one suggestion. Sensor output in power flow calculations is undefined. We should not test them. So remove the tests in the current `test_main_model.cpp`, and do not re-test them.
> 
> while i'm OK with not testing, it does feel a little bit weird to call it UB, though, because we **do** by default output it if there are sensors in the grid

UB means the C-API users are not supposed to provide a sensor output buffer in the dataset if you are running power flow. If you do, there is no guarantee what data will be written (or not at all) in the sensor output buffer. It just happens to have some non-sense output now in the sensor buffer. It can change to anything in the future.

NOTE: Python API never creates the sensor output buffer anyway. This is for C-API user.

mgovers:
> > > @mgovers one suggestion. Sensor output in power flow calculations is undefined. We should not test them. So remove the tests in the current `test_main_model.cpp`, and do not re-test them.
> > 
> > 
> > while i'm OK with not testing, it does feel a little bit weird to call it UB, though, because we **do** by default output it if there are sensors in the grid
> 
> UB means the C-API users are not supposed to provide a sensor output buffer in the dataset if you are running power flow. If you do, there is no guarantee what data will be written (or not at all) in the sensor output buffer. It just happens to have some non-sense output now in the sensor buffer. It can change to anything in the future.
> 
> NOTE: Python API never creates the sensor output buffer anyway. This is for C-API user.

ohhhh right, i completely forgot about the existence of https://github.com/PowerGridModel/power-grid-model/blob/main/src/power_grid_model/_core/power_grid_model.py#L171-L195

Agreed with the conclusion, although this still does feel weird
image

In addition, shouldn't we add it explicitly to the sensor output documentation?
image

TonyXiang8787 (approved):

Issue #806, opened 2024-10-25
[FEATURE] Refactor main model to move the optional id logic


Jerry-Jinfeng-Guo:
At the moment, the optional id logic to be added via https://github.com/PowerGridModel/power-grid-model/pull/771 is mainly implemented in `main_model_impl.hpp`. In this refactor, relevant logic is proposed to be moved to `update.hpp` and others accordingly.

PR #843, opened 2024-11-25
Optional ID refactor: part 3

figueroa1395:
Follow-up PR after https://github.com/PowerGridModel/power-grid-model/pull/828. Closes https://github.com/PowerGridModel/power-grid-model/issues/806


power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -221,35 +216,32 @@ std::vector<Idx2D> get_component_sequence(MainModelState<ComponentContainer> con
...
-                         main_core::utils::ComponentFlags<ComponentTypes...> const& components_to_store) {
+                         Idx scenario_idx, utils::ComponentFlags<ComponentTypes...> const& components_to_store,
+                         independence::UpdateIndependence<ComponentTypes...> const& independence, bool cached = true) {
     return utils::run_functor_with_all_types_return_array<ComponentTypes...>(
-        [&update_data, &components_to_store, &state, scenario_idx]<typename CompType>() {
-            if (!std::get<main_core::utils::index_of_component<CompType, ComponentTypes...>>(components_to_store)) {
+        [&state, &update_data, scenario_idx, &components_to_store, &independence, cached]<typename CompType>() {
+            auto const comp_properties = std::get<utils::index_of_component<CompType, ComponentTypes...>>(independence);
+            bool const is_comp_independent =
+                cached ? comp_properties.is_independent() : !comp_properties.is_independent();
```
figueroa1395, line 227: I am not very fond of this part of the new implementation... including the `cached = true` argument. However, I couldn't think of anything better. This is the comprimise I made to the the independence calculation outside the logic of this function.

figueroa1395:
The remaining problem comes from the fact that `assert(static_cast<ptrdiff_t>(sequence_idx.size()) == std::distance(begin, end));` in `update_component` within `main_model_impl.hpp` doesn't hold anymore for permanent updates with mixed optional ids and invalid ids. I think this can be traced to the new implementation of `get_all_sequence_idx_map`, as the sequence size is not the same for every case, as it was before. Hence, a suggestion that doesn't involve changing the whole `update_component` flow would be welcome.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -221,35 +216,32 @@ std::vector<Idx2D> get_component_sequence(MainModelState<ComponentContainer> con
...
-            components_to_store[comp_idx] = (update_data.find_component(CompType::name, false) != invalid_index);
-            ++comp_idx;
-        });
-    return get_all_sequence_idx_map<ComponentTypes...>(state, update_data, 0, components_to_store);
+utils::SequenceIdx<ComponentTypes...>
+get_all_sequence_idx_map(MainModelState<ComponentContainer> const& state, ConstDataset const& update_data,
+                         utils::ComponentFlags<ComponentTypes...> const& components_to_store,
+                         independence::UpdateIndependence<ComponentTypes...> const& independence) {
+    return get_all_sequence_idx_map<ComponentTypes...>(state, update_data, 0, components_to_store, independence);
 }
```
figueroa1395, line 245: This overload may as well be gone. What do you think? Or should we still keep an explicit entry point for permanent updates?

mgovers: as long as the entrypoint exists on the outside (which it does because it's part of the C API), i believe we should keep something similar. however, the internal implementation may indeed be refactored/removed as needed. since this is part of the internal implementation, you can remove it here, but please do make sure to keep the external entrypoint

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: yeah that's also why i had the logic with the whole `logical_not` etc. this function shouldn't have a notion as to what "cached" means. the `components_to_store` should already capture everything, right?

mgovers:
> The remaining problem comes from the fact that `assert(static_cast<ptrdiff_t>(sequence_idx.size()) == std::distance(begin, end));` in `update_component` within `main_model_impl.hpp` doesn't hold anymore for permanent updates with mixed optional ids and invalid ids. I think this can be traced to the new implementation of `get_all_sequence_idx_map`, as the sequence size is not the same for every case, as it was before. Hence, a suggestion that doesn't involve changing the whole `update_component` flow would be welcome.
> 
> Note: Only the API tests regarding what's mentioned above are failing. Every other one, including on the Python side, passes just fine.

how doesn't it hold up anymore? is that because you have 2 steps now; one involving the globally cached components, and another involving the locally cached ones?

if so, you can maybe go with one (or a combination) of the following solutions:

1. something along the lines of `for every component type that is cached: the assertion shall hold`
2. only call the concerning function if `sequence_idx.size() > 0`
3. add fallback to the concerning function if `sequence_idx.empty()`. I think this was the logic before the original optional ID implementation
4. split the function up into 2 logical steps
5. if none of these are right, feel free to remove the `assert` if there's a good reason to remove it. it was just a sanity check; nothing more. nothing will break when you remove an assert (the only thing that can happen is that potential future bugs will be slightly harder to detect)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -12,6 +12,8 @@
 
 namespace power_grid_model::main_core::utils {
 
+constexpr Idx invalid_index_v{-1};
```
Jerry-Jinfeng-Guo, line 15: What does `_v` imply?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -130,38 +126,37 @@ UpdateCompProperties check_component_independence(ConstDataset const& update_dat
     return properties;
 }
 
-inline void validate_update_data_independence(UpdateCompProperties const& comp) {
+template <class... ComponentTypes>
+using UpdateIndependence = std::array<UpdateCompProperties, utils::n_types<ComponentTypes...>>;
+
+inline void validate_update_data_independence(UpdateCompProperties const& comp, std::string const& comp_name) {
```
Jerry-Jinfeng-Guo, line 132: I understand the use of `string` inside `UpdateCompProperties` is suboptimal, but this change feels like un-necessary.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: I am also having trouble understand the logic here. I will have to read more appropriately in the code rather than on github...

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: This was introduced for one very specific reason. I wanted to make `UpdateCompProperties` trivially copyable, and the only way I saw was to remove string from there, hence why it is here (and elsewhere) now. And this came from the fact that I wanted to remove all "independence" logic from the `get_sequence_idx` functions.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -152,10 +152,10 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
     using OwnedUpdateDataset = std::tuple<std::vector<typename ComponentType::UpdateType>...>;
 
-    static constexpr Idx ignore_output{-1};
-    static constexpr Idx isolated_component{-1};
-    static constexpr Idx not_connected{-1};
-    static constexpr Idx sequential{-1};
+    static constexpr Idx ignore_output_v{-1};
+    static constexpr Idx isolated_component_v{-1};
+    static constexpr Idx not_connected_v{-1};
+    static constexpr Idx sequential_v{-1};
```
Jerry-Jinfeng-Guo, line 158: Again, `_v` voor?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
Jerry-Jinfeng-Guo: But you are not trying to extract `::value` from a template argument, right? That's the more appropriate context.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: Well, necessary trade-off I guess

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
Jerry-Jinfeng-Guo: Same as above. Closing this one and leaving the other one open.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
figueroa1395: You are right. I amended this in [c129a50](https://github.com/PowerGridModel/power-grid-model/pull/843/commits/c129a505abde61f497717f07ba7a2114328f1f09).

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: The overload (in [c129a50](https://github.com/PowerGridModel/power-grid-model/pull/843/commits/c129a505abde61f497717f07ba7a2114328f1f09)) in favor of explicit index specification, but the entry point via `main_model.hpp` was kept.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
figueroa1395: Tried to improve it a bit by adding a comment, renaming some things, and removing the default of `cached`. Take a look at [c129a50](https://github.com/PowerGridModel/power-grid-model/pull/843/commits/c129a505abde61f497717f07ba7a2114328f1f09).

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -221,35 +216,26 @@ std::vector<Idx2D> get_component_sequence(MainModelState<ComponentContainer> con
...
-        [&update_data, &components_to_store, &state, scenario_idx]<typename CompType>() {
-            if (!std::get<main_core::utils::index_of_component<CompType, ComponentTypes...>>(components_to_store)) {
+        [&state, &update_data, scenario_idx, &components_to_store, &independence, cached]<typename CompType>() {
+            auto const component_properties =
+                std::get<utils::index_of_component<CompType, ComponentTypes...>>(independence);
+            // The sequence for the independent components is cached (true). For the remaining components, the sequence
+            // cannot be cached (false), so the independence flags are inverted to not return an empty sequence when
+            // this is the case.
+            bool const component_independence =
+                !cached ? component_properties.is_independent() : !component_properties.is_independent();
```
Jerry-Jinfeng-Guo, line 231:
```suggestion
            bool const component_independence = cached ^ component_properties.is_independent();
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -388,9 +383,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
     // Entry point for main_model.hpp
-    template <class... ComponentTypes>
     main_core::utils::SequenceIdx<ComponentType...> get_all_sequence_idx_map(ConstDataset const& update_data) {
-        return main_core::update::get_all_sequence_idx_map<ComponentType...>(state_, update_data);
+        auto const components_to_update = get_components_to_update(update_data);
+        auto const update_independence =
+            main_core::update::independence::check_update_independence<ComponentType...>(state_, update_data);
+        return main_core::update::get_all_sequence_idx_map<ComponentType...>(
+            state_, update_data, 0, components_to_update, update_independence, false);
     }
```
Jerry-Jinfeng-Guo, line 392: Now this function is only called in the restore function in main_model.hpp, hmm

Jerry-Jinfeng-Guo:
Nothing major left, only minor comments like above.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: bitwise XOR is quite uncommon in modern C++ (except in microcontroller/compiler development), to the point that most new developers don't even know it exists. In particular, it is dangerous because `true` can be represented by any non-zero value. Please use `!=` instead

```suggestion
            bool const component_independence = cached != component_properties.is_independent();
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -301,8 +292,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
     // overload to update all components in the first scenario (e.g. permanent update)
     template <cache_type_c CacheType> void update_components(ConstDataset const& update_data) {
-        auto const sequence_idx_map =
-            main_core::update::get_all_sequence_idx_map<ComponentType...>(state_, update_data);
+
```
mgovers, line 295:
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: that's great news because that functionality is deprecated and removed in #848 😬

mgovers (commented):
no other comments from my end

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: (or, at least, public exposition of it)

Jerry-Jinfeng-Guo: mmmk

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
Jerry-Jinfeng-Guo: > In particular, it is dangerous because

Arguably anything is dangerous given certain context. In this particular case, the operands are two booleans. 

> bitwise XOR is quite uncommon in modern C++ (except in microcontroller/compiler development), to the point that most new developers don't even know it exists.

One could make an analogical argument that "arithmetic is quite uncommon in modern education, to the point that most new students don't even know how to do it". It is far fetched, funny but probably worth thinking.

figueroa1395: Since both suggestions are equally valid in this specific case, hence a matter of style, I went with `bool const component_independence = cached != component_properties.is_independent();` as it is more straightforward for me.

Jerry-Jinfeng-Guo (approved):

figueroa1395:
> > ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=843) **Quality Gate passed**
> > Issues  [2 New issues](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=843&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=843&issueStatuses=ACCEPTED)
> > Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=PowerGridModel_power-grid-model&pullRequest=843&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [96.8% Coverage on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=843&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=843&metric=new_duplicated_lines_density&view=list)
> > [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=843)
> 
> The two issues are actually quite trivial to fix, could you patch them? @figueroa1395

I had completely missed them. Thanks for the reminder. They have been addressed now.

Jerry-Jinfeng-Guo (approved):

PR #845, opened 2024-11-26
Adding message to `NotObservableError` when no voltage sensor present

Jerry-Jinfeng-Guo:
This PR adds message to the `NotObservableError` when no voltage sensor present.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
@@ -112,7 +112,12 @@ class SparseMatrixError : public PowerGridError {
 
 class NotObservableError : public PowerGridError {
   public:
-    NotObservableError() { append_msg("Not enough measurements available for state estimation.\n"); }
+    NotObservableError(std::string const& msg = "") {
+        append_msg("Not enough measurements available for state estimation.\n");
+        if (!msg.empty()) {
+            append_msg(msg + "\n");
+        }
+    }
 };
 
 class IterationDiverge : public PowerGridError {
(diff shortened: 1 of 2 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -112,7 +112,12 @@ class SparseMatrixError : public PowerGridError {
 
 class NotObservableError : public PowerGridError {
   public:
-    NotObservableError() { append_msg("Not enough measurements available for state estimation.\n"); }
+    NotObservableError(std::string const& msg = "") {
```
mgovers, line 115: might be good to double-check that the python exception wrapper still works as intended

mgovers, line 115: please also add it to the list of observability checks in https://power-grid-model.readthedocs.io/en/stable/user_manual/calculations.html#state-estimation

TonyXiang8787: It is already there. We need at least one voltage sensor.

TonyXiang8787: Based on the requirements of observability mentioned above, user needs to satisfy at least the following conditions for state estimation calculation in power-grid-model.

n_voltage_sensor >= 1

Jerry-Jinfeng-Guo: Message is appended to the back. re would still match.

TonyXiang8787 (approved):

PR #849, opened 2024-12-03
MessagePack deserialization: improve error messages

mgovers:


figueroa1395 (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -193,7 +193,7 @@ struct MapArrayVisitor : DefaultErrorVisitor<MapArrayVisitor<map_array>> {
     static constexpr bool enable_array =
         std::same_as<map_array, visit_array_t> || std::same_as<map_array, visit_map_array_t>;
     static constexpr std::string_view static_err_msg =
-        enable_map ? (enable_array ? "Expect a map or array." : "Expect a map.") : "Expect an array.";
+        enable_map ? (enable_array ? "Expected a map or array." : "Expected a map.") : "Expected an array.";
```
Jerry-Jinfeng-Guo, line 196: Now sure this is good English. You could either do `A map or an array expected` or `Expecting a map or an array`.

Jerry-Jinfeng-Guo (changes requested):
Please update the error messages: 
either "A/An {item 1}  or a/an {item 2} expected "
or 
"Expecting  a/an {item 1}  or  a/an {item 2}"

TonyXiang8787:
What are we fixing apart from the error messages?

mgovers:
> What are we fixing apart from the error messages?

read as (fix + improve) error messages. will update

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers:
```suggestion
        enable_map ? (enable_array ? "Expected a map or an array." : "Expected a map.") : "Expected an array.";
```

Jerry-Jinfeng-Guo:
> kept the `Expected ...` pattern. It doesn't look that weird if you read the full error image

That's exactly what I couldn't agree...

TonyXiang8787:
I still do not get the purpose of this PR. The original error message is already very clear. The changes in this PR are only about the tense.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -248,7 +248,7 @@ struct BoolVisitor : DefaultErrorVisitor<BoolVisitor> {
 template <class T> struct ValueVisitor;
 
 template <std::integral T> struct ValueVisitor<T> : DefaultErrorVisitor<ValueVisitor<T>> {
-    static constexpr std::string_view static_err_msg = "Expect an interger.";
+    static constexpr std::string_view static_err_msg = "Expected an integer.";
```
mgovers, line 251: This contained a typo `interger`, fixed to `integer`. Irena encountered that

mgovers:
> I still do not get the purpose of this PR. The original error message is already very clear. The changes in this PR are only about the tense.

In addition, while present tense was clear, it felt more like imperative tense (which it clearly isn't).

In this case, while we haven't dug into debugging yet, the JSON looked correct at first sight
* so the error message became confusing: "expect an integer" > "it tells me to expect something?"
* Past tense would have been more clear: "Expected an integer" > ah ok something went wrog. Maybe internally?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: This change I agree. But others not. 

Change from "expect" to "expected" is a bit confusing. I propose either keep the original, or changed them to "integer expected", as @Jerry-Jinfeng-Guo suggests.

TonyXiang8787 (approved):

Jerry-Jinfeng-Guo (approved):

PR #852, opened 2024-12-05
Sparse matrix error: repro case for not using numerical pivoting

TonyXiang8787:
The in-house LU Sparse solver does not do numerical pivoting for performance optimization. It assumes that the matrix is always diagonal major.

However, this assumption does not hold for certain real-world state estimation cases. This results in Sparse Matrix Error where the system is actually perfectly observable and solvable.

This PR reproduces the error using two cases.


tests/data/state_estimation/leaf-without-power-sensor/params.json
```diff
@@ -0,0 +1,8 @@
+{
+  "calculation_method": ["newton_raphson", "iterative_linear"],
```
mgovers, line 2: since the timeline on solving this is not yet known because of the nature of the problem, i propose to mark these test cases as `fail` and merge, and instead open an issue on this to reflect the progress on solving the bug.

TonyXiang8787:
@mgovers please go ahead to modify the files, merge the PR and create the issue.

mgovers:
Issue created in https://github.com/PowerGridModel/power-grid-model/issues/853

tests/data/state_estimation/leaf-without-power-sensor/params.json
mgovers: done

mgovers (approved):

Issue #853, opened 2024-12-06
[BUG] *descriptive_name*


mgovers:
**Describe the bug**

The in-house LU Sparse solver does not do numerical pivoting for performance optimization. It assumes that the matrix is always diagonal major.

However, this assumption does not hold for certain real-world state estimation cases. This results in Sparse Matrix Error where the system is actually perfectly observable and solvable.

**To Reproduce**

Run the test cases introduced in #852 . See that they run as expected fail (`xfail`) instead of pass.

**Expected behavior**

The test cases introduced in #852 pass

petersalemink95:
When this bug is fixed, we need to update the `SparseMatrixError` message, which was updated in #858

PR #854, opened 2024-12-08
Sparse matrix error: more complete repro case

TonyXiang8787:
Add a branch measurement also does not help the issue in #853


TonyXiang8787:
@mgovers, the code quality error seems to be un-related. Feel free to fix it.

mgovers (approved):
LGTM. FYI: i fixed mypy incompatibility with `numpy==2.2.0`, which caused the failing CI (both here and in the Nightly). This will also fix the nightly again

PR #858, opened 2024-12-15
Run full observability check for radial grid without phasor measurement

### LINKED ISSUE #838: [FEATURE] Check for necessary and sufficient condition for observability in radial grid
# Background

The full observability check for any grids and any sensors is described in #84, it is still a complicated task. In #508 we have implemented a simple necessary condition check so we can throw `NotObservableError` if we are sure there is not enough measurements.

The feature in #508 have greatly helped the real-time state estimation to give clear errors. However, there are still plenty of scenarios where the number of measurements seems to be sufficient, but they are not independent. It still results in `SparseMatrixError`.

# Proposal

We are still far from implementing the full observability check as in #84. This issue proposes to do another intermediate step to implement the full necessary and sufficient condition check if the data satisfy the following conditions:

1. The grid is in radial structure without cycles. Parrallel branches do not count as cycles.
2. There is no voltage phasor (and also no current phasor) measurements.

In this case, implementing a full necessary and sufficient condition check is relatively easy.

# Mathematics

See below the mathematical workout of proposal.

## Build graph

1. Build an undirected graph with vertices (bus) and edges (branch). Parallel branches will be merged into one edge.
2. Since the graph is radial and there is no parallel edges. We strictly have `n_edges = n_vertices - 1`.
3. Label all the vertices with either `injection_available` or `injection_unavailable`.
4. Lable all the edges with either `measured` (if there is at least one power sensor in one of the parallel branches), or `not_measured`.
5. Perform a DFS search from the source vertex. For each vertex, record the predecessor vertex, and predecessor edge.

## Check observability

Now we can check the observability. To do that, we traverse the vertices in the reverse order of the DFS results. For each vertex, we do the following:

1. If the predecessor edge is `not_measured` .
    1. If the current vertex has `injection_available`
        1. change the predecessor edge to `measured`.
    1. Else
        1. If the predecessor vertex has `injection_available`
            1. change the predecessor edge to `measured`.
            1. change the predecessor vertex to `injection_unavailable`.
1. Change the current vertex to `injection_unavailable`, regardless of the original state.

After traversal to all the vertices in reverse order, check if all the edges are now `measured`. If yes, the system is observable, otherwise it is not observable. This is a full necessary and sufficient condition check.

# Implementation

It seems that we need to build another graph to implement the proposed method. However, all the information is already in the `YBusStructure`. If the grid is radial, in the topological processing the DFS search is already done, and the vertices are already ordered in reverse of DFS result. The upper triangle of the `YBusStructure` contains all the information of the predecessor vertex and predecessor edge.

So the implementation proposal will be

1. In `MathModelTopology` add a member called `is_radial`. This is easy to set in the topological processing.
2. In the observability check, pass the `YBusStructure` into the check function, and perform the graph check as above.

The figure below shows a simple graph and its `YBusStructure`. In the upper triangle, there should be only one off-diagonal entry which is the predecessor vertex and predecessor edge.

IMG_6970

TonyXiang8787:
Closes #838

Check list:

- [ ] Implement the core
- [ ] Adjust unit test
- [ ] Validation test of a system with enough number of power sensors but they are not independent
- [ ] Adjust documentation, if needed

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -66,6 +66,42 @@ std::vector<int8_t> count_flow_sensors(MeasuredValues<sym> const& measured_value
     return flow_sensors;
 }
 
+// re-organize the flow sensor for radial grid without phasor measurement
+// this mutate the flow sensor vector to try to assign injection sensor to branch sensor
+// all the branch should be measured if the system is observable
+// this is a sufficient condition check
+void assign_injection_sensor(YBusStructure const& y_bus_structure, std::vector<int8_t>& flow_sensors) {
+    Idx const n_bus = std::ssize(y_bus_structure.row_indptr) - 1;
+    // loop the row without the last bus
+    for (Idx row = 0; row != n_bus - 1; ++row) {
+        Idx const current_bus = row;
+        Idx const bus_entry_current = y_bus_structure.bus_entry[current_bus];
+        Idx const branch_entry_upstream = bus_entry_current + 1;
+        // there should be only one upstream branch in the upper diagonal
+        // so the next of branch_entry_upstream is already the next row
+        // because the grid is radial
+        assert(y_bus_structure.row_indptr[current_bus + 1] == branch_entry_upstream + 1);
+        Idx const upstream_bus = y_bus_structure.col_indices[branch_entry_upstream];
+        Idx const bus_entry_upstream = y_bus_structure.bus_entry[upstream_bus];
+        // if the upstream branch is not measured
+        if (flow_sensors[branch_entry_upstream] == 0) {
+            if (flow_sensors[bus_entry_current] == 1) {
+                // try to steal from current bus
+                flow_sensors[bus_entry_current] = 0;
+                flow_sensors[branch_entry_upstream] = 1;
+            } else if (flow_sensors[bus_entry_upstream] == 1) {
+                // if not possible, steal from upstream bus
+                flow_sensors[bus_entry_upstream] = 0;
+                flow_sensors[branch_entry_upstream] = 1;
+            }
+        }
+        // remove the current bus injection regardless of the original state
+        flow_sensors[bus_entry_current] = 0;
+    }
+    // set last bus injection to zero
+    flow_sensors[y_bus_structure.bus_entry[n_bus - 1]] = 0;
+}
+
 } // namespace detail
 template <symmetry_tag sym>
 inline void necessary_observability_check(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
@@ -74,18 +110,31 @@ inline void necessary_observability_check(MeasuredValues<sym> const& measured_va
 
     auto const [n_voltage_sensor, n_voltage_phasor_sensor] = detail::count_voltage_sensors(n_bus, measured_values);
     if (n_voltage_sensor < 1) {
-        throw NotObservableError{"no voltage sensor found"};
+        throw NotObservableError{"No voltage sensor found!\n"};
     }
 
-    auto const flow_sensors = detail::count_flow_sensors(measured_values, topo, y_bus_structure);
+    std::vector<int8_t> flow_sensors = detail::count_flow_sensors(measured_values, topo, y_bus_structure);
     // count flow sensors, note we manually specify the intial value type to avoid overflow
     Idx const n_flow_sensor = std::reduce(flow_sensors.cbegin(), flow_sensors.cend(), Idx{}, std::plus<Idx>{});
 
+    // check nessessary condition for observability
     if (n_voltage_phasor_sensor == 0 && n_flow_sensor < n_bus - 1) {
         throw NotObservableError{};
     }
     if (n_voltage_phasor_sensor > 0 && n_flow_sensor + n_voltage_phasor_sensor < n_bus) {
         throw NotObservableError{};
+    }
+
+    // for radial grid without phasor measurement, try to assign injection sensor to branch sensor
+    // we can then check sufficient condition for observability
+    if (topo.is_radial && n_voltage_phasor_sensor == 0) {
+        detail::assign_injection_sensor(y_bus_structure, flow_sensors);
+        // count flow sensors again
+        Idx const n_flow_sensor_new = std::reduce(flow_sensors.cbegin(), flow_sensors.cend(), Idx{}, std::plus<Idx>{});
+        if (n_flow_sensor_new < n_bus - 1) {
+            throw NotObservableError{"The number of power sensors seems to be sufficient, but they are not independent "
+                                     "enough. The system is still not observable.\n"};
+        }
     }
 }
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -14,37 +14,6 @@ namespace power_grid_model::math_solver {
 namespace detail {
 
 template <symmetry_tag sym>
-Idx count_branch_sensors(const std::vector<BranchIdx>& branch_bus_idx, const Idx n_bus,
-                         const MeasuredValues<sym>& measured_values) {
-    Idx n_branch_sensor{};
-    std::vector<bool> measured_nodes(n_bus, false);
-    for (Idx branch = 0; branch != static_cast<Idx>(branch_bus_idx.size()); ++branch) {
-        auto const& [node_from, node_to] = branch_bus_idx[branch];
-        if (node_from == -1 || node_to == -1) {
-            continue;
-        }
-        if ((measured_values.has_branch_from(branch) || measured_values.has_branch_to(branch)) &&
-            !(measured_nodes[node_from] && measured_nodes[node_to])) {
-            n_branch_sensor++;
-            measured_nodes[node_from] = true;
-            measured_nodes[node_to] = true;
-        }
-    }
-    return n_branch_sensor;
-}
-
-template <symmetry_tag sym>
-Idx count_bus_injection_sensors(const Idx n_bus, const MeasuredValues<sym>& measured_values) {
-    Idx n_injection_sensor{};
-    for (Idx bus = 0; bus != n_bus; ++bus) {
-        if (measured_values.has_bus_injection(bus)) {
-            n_injection_sensor++;
-        }
-    }
-    return n_injection_sensor;
-}
-
-template <symmetry_tag sym>
 std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues<sym>& measured_values) {
     Idx n_voltage_sensor{};
     Idx n_voltage_phasor_sensor{};
@@ -59,27 +28,63 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
     return std::make_tuple(n_voltage_sensor, n_voltage_phasor_sensor);
 }
 
+// count flow sensors into ybus structure like
+// lower triangle part is always zero
+// for diagonal part, it will be one if there is bus injection
+// for upper triangle part, it will be one if there is branch flow sensor and the branch is fully connected
+template <symmetry_tag sym>
+std::vector<int8_t> count_flow_sensors(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
+                                       YBusStructure const& y_bus_structure) {
+    Idx const n_bus{topo.n_bus()};
+    std::vector<int8_t> flow_sensors(y_bus_structure.row_indptr.back(), 0); // initialize all to zero
+    for (Idx row = 0; row != n_bus; ++row) {
+        // lower triangle is ignored and kept as zero
+        // diagonal for bus injection measurement
+        if (measured_values.has_bus_injection(row)) {
+            flow_sensors[y_bus_structure.bus_entry[row]] = 1;
+        }
+        // upper triangle for branch flow measurement
+        for (Idx ybus_index = y_bus_structure.bus_entry[row] + 1; ybus_index != y_bus_structure.row_indptr[row + 1];
+             ++ybus_index) {
+            for (Idx element_index = y_bus_structure.y_bus_entry_indptr[ybus_index];
+                 element_index != y_bus_structure.y_bus_entry_indptr[ybus_index + 1]; ++element_index) {
+                YBusElement const& element = y_bus_structure.y_bus_element[element_index];
+                // shunt should not be considered
+                // if the branch is fully connected and measured, we consider it as a valid flow sensor
+                // we only need one flow sensor, so the loop will break
+                if (element.element_type != YBusElementType::shunt) {
+                    Idx const branch = element.idx;
+                    if ((measured_values.has_branch_from(branch) || measured_values.has_branch_to(branch)) &&
+                        topo.branch_bus_idx[branch][0] != -1 && topo.branch_bus_idx[branch][1] != -1) {
+                        flow_sensors[ybus_index] = 1;
+                        break;
+                    }
+                }
+            }
+        }
+    }
+    return flow_sensors;
+}
+
 } // namespace detail
 template <symmetry_tag sym>
 inline void necessary_observability_check(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
                                           YBusStructure const& y_bus_structure) {
-    (void)y_bus_structure; // TODO remove when used
     Idx const n_bus{topo.n_bus()};
-    std::vector<BranchIdx> const& branch_bus_idx{topo.branch_bus_idx};
 
     auto const [n_voltage_sensor, n_voltage_phasor_sensor] = detail::count_voltage_sensors(n_bus, measured_values);
     if (n_voltage_sensor < 1) {
         throw NotObservableError{"no voltage sensor found"};
     }
 
-    Idx const n_injection_sensor = detail::count_bus_injection_sensors(n_bus, measured_values);
-    Idx const n_branch_sensor = detail::count_branch_sensors(branch_bus_idx, n_bus, measured_values);
-    Idx const n_power_sensor = n_injection_sensor + n_branch_sensor;
+    auto const flow_sensors = detail::count_flow_sensors(measured_values, topo, y_bus_structure);
+    // count flow sensors, note we manually specify the intial value type to avoid overflow
+    Idx const n_flow_sensor = std::reduce(flow_sensors.cbegin(), flow_sensors.cend(), Idx{}, std::plus<Idx>{});
 
-    if (n_voltage_phasor_sensor == 0 && n_power_sensor < n_bus - 1) {
+    if (n_voltage_phasor_sensor == 0 && n_flow_sensor < n_bus - 1) {
         throw NotObservableError{};
     }
-    if (n_voltage_phasor_sensor > 0 && n_power_sensor + n_voltage_phasor_sensor < n_bus) {
+    if (n_voltage_phasor_sensor > 0 && n_flow_sensor + n_voltage_phasor_sensor < n_bus) {
         throw NotObservableError{};
     }
 }
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -61,9 +61,9 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
 
 } // namespace detail
 template <symmetry_tag sym>
-inline void necessary_observability_check(MeasuredValues<sym> const& measured_values,
-                                          MathModelTopology const& topo) {
-
+inline void necessary_observability_check(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
+                                          YBusStructure const& y_bus_structure) {
+    (void)y_bus_structure; // TODO remove when used
     Idx const n_bus{topo.n_bus()};
     std::vector<BranchIdx> const& branch_bus_idx{topo.branch_bus_idx};
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "measured_values.hpp"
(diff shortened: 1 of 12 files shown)


docs/user_manual/calculations.md
```diff
@@ -57,10 +57,10 @@ Output:
...
-raise a sparse matrix error; the matrix reprensentation of the system is too sparse to solve and particular so when it is
-singular. In short, meeting the requirement of observability indicates that the system is either an overdetermined system (when the number of measurements is larger than the number of
-unknowns) or a balanced system (when the number of measurements is equal to the number of unknowns). For each node, there are two unknowns, `u` and `u_angle`, so the following conditions should be met:
+In order to perform a state estimation, the system should be observable. If the system is not observable, the calculation will either raise a not observable error or 
+a sparse matrix error.
+In short, meeting the requirement of observability indicates that the system is either an overdetermined system (when the number of measurements is larger than the number of
+unknowns) or a determined system (when the number of measurements is equal to the number of unknowns). For each node, there are two unknowns, `u` and `u_angle`, so the following conditions should be met:
```
mgovers, line 63: i think we need to be explicit here.

```suggestion
unknowns) or an exactly determined system (when the number of measurements is equal to the number of unknowns). For each node, there are two unknowns, `u` and `u_angle`, so the following conditions should be met:
```

alternatively, something along the lines of `at least not an under-determined system`

docs/user_manual/calculations.md
```diff
@@ -72,10 +72,12 @@ Where
 
 $$
     \begin{eqnarray}
-        n_{unknowns}    & = & 2 & \cdot & n_{nodes}
+        n_{unknowns}    & = & 2 & \cdot & n_{nodes} - 1
```
Jerry-Jinfeng-Guo, line 75: Shouldn't it be:
```suggestion
        n_{measurements} >= n_{unknowns}    & = & 2 & \cdot & n_{nodes} - 1
```

docs/user_manual/calculations.md
Jerry-Jinfeng-Guo, line 63: I'd suggest mention here already the important property of `u_angle` mentioned below.
```suggestion
... For each node, there are two unknowns, `u` and `u_angle`. Due to the relative nature of `u_angle` (relevant only in systems with at least two nodes), in total the following conditions should be met:
```

docs/user_manual/calculations.md
```diff
@@ -108,6 +110,16 @@ Based on the requirements of observability mentioned above, user needs to satisf
 - Any sensor on a `Branch` for all branches: Parallel branches with either side of measurements count as one.
 - All `Branch3` sensors.
 
+##### Sufficient observability condition
+
+The condition check above only checks the necessary condition for observability. When the measurements are not independent enough, the system may still be unobservable even if the necessary condition is met.
```
mgovers, line 115: i think in theory it is also possible that measurements are such that they coincidentally result in an underdetermined system of equations (extremely unlikely in real life, but potentially may occur for generated data). or do the matrix equation by construction guarantee that is never the case?

docs/user_manual/calculations.md
```diff
@@ -72,10 +72,12 @@ Where
 
 $$
     \begin{eqnarray}
-        n_{unknowns}    & = & 2 & \cdot & n_{nodes}
+        n_{unknowns}    & = & 2 & \cdot & n_{nodes} - 1
     \end{eqnarray}
 $$
 
+Note that the phase angle is a relative value. So the number of unknowns is reduced by one.
+
```
Jerry-Jinfeng-Guo, line 80: If mentioned above as suggested, this line can be removed.

docs/user_manual/calculations.md
```diff
@@ -96,7 +98,7 @@ In observable systems this helps better outputting correct results. On the other
 
 ##### Necessary observability condition
 
-Based on the requirements of observability mentioned above, user needs to satisfy  at least the following conditions for state estimation calculation in power-grid-model.
+Based on the requirements of observability mentioned above, user needs to satisfy  at least the following conditions for state estimation calculation in `power-grid-model`.
```
Jerry-Jinfeng-Guo, line 101:
```suggestion
Based on the requirements of observability mentioned above, users need to satisfy at least the following conditions for state estimation calculation in `power-grid-model`.
```

docs/user_manual/calculations.md
```diff
@@ -108,6 +110,16 @@ Based on the requirements of observability mentioned above, user needs to satisf
...
 
+##### Sufficient observability condition
+
+The condition check above only checks the necessary condition for observability. When the measurements are not independent enough, the system may still be unobservable even if the necessary condition is met.
+It is rather complicated to do a full sufficient and necessary observability check in generic cases. However, `power-grid-model` performs the sufficient condition check when the following conditions are met:
+
+1. The system is a radial network.
+2. The system does not have voltage phasor measurements.
+
+In this case, the validation of the independent measurements is rather straightforward. If the system is not observable, the calculation will raise a not observable error instead of sparse matrix error.
```
Jerry-Jinfeng-Guo, line 121:
```suggestion
In this case, the validation of the independent measurements is rather straightforward. If the system is not observable, the calculation will raise a `NotObservableError` instead of `SparseMatrixError`.
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,114 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
     return std::make_tuple(n_voltage_sensor, n_voltage_phasor_sensor);
 }
 
-} // namespace detail
+// count flow sensors into ybus structure like
```
Jerry-Jinfeng-Guo, line 31: This sentence is incomplete?

docs/user_manual/calculations.md
```diff
@@ -57,10 +57,10 @@ Output:
 - Power flow through branches
 - Deviation between measurement values and estimated state
 
-In order to perform a state estimation, the system should be observable. If the system is not observable, the calculation will 
-raise a sparse matrix error; the matrix reprensentation of the system is too sparse to solve and particular so when it is
-singular. In short, meeting the requirement of observability indicates that the system is either an overdetermined system (when the number of measurements is larger than the number of
-unknowns) or a balanced system (when the number of measurements is equal to the number of unknowns). For each node, there are two unknowns, `u` and `u_angle`, so the following conditions should be met:
+In order to perform a state estimation, the system should be observable. If the system is not observable, the calculation will either raise a not observable error or 
+a sparse matrix error.
```
mgovers, line 61: we most likely should update the default error message of the sparse matrix error (also check whether the python regex needs to be updated). it's not up to date anymore after this change.

mgovers (commented):
due to other priorities, i have not yet done a full in-depth review but here are some remarks

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,114 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
...
+// lower triangle part is always zero
+// for diagonal part, it will be one if there is bus injection
+// for upper triangle part, it will be one if there is branch flow sensor and the branch is fully connected
 template <symmetry_tag sym>
-inline void necessary_observability_check(MeasuredValues<sym> const& measured_values,
-                                          std::shared_ptr<MathModelTopology const> const& topo) {
+std::vector<int8_t> count_flow_sensors(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
+                                       YBusStructure const& y_bus_structure) {
+    Idx const n_bus{topo.n_bus()};
+    std::vector<int8_t> flow_sensors(y_bus_structure.row_indptr.back(), 0); // initialize all to zero
```
Jerry-Jinfeng-Guo, line 39: Since `flow_sensors` will be 1's and 0's, why not make it `std::vector<bool>`. Since this state for later use is not accumulable, booleans seems perfect.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,114 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
...
+            if (flow_sensors[bus_entry_current] == 1) {
+                // try to steal from current bus
+                flow_sensors[bus_entry_current] = 0;
+                flow_sensors[branch_entry_upstream] = 1;
+            } else if (flow_sensors[bus_entry_upstream] == 1) {
+                // if not possible, steal from upstream bus
+                flow_sensors[bus_entry_upstream] = 0;
+                flow_sensors[branch_entry_upstream] = 1;
+            }
+        }
```
Jerry-Jinfeng-Guo, line 97:
```suggestion
        if (flow_sensors[branch_entry_upstream] == 0) {
            if (flow_sensors[branch_entry_upstream] == 0) {
                if (flow_sensors[bus_entry_current] == 1 || flow_sensors[bus_entry_upstream] == 1) {
                    // try to steal from current bus or upstream bus
                    flow_sensors[bus_entry_current] = 0;
                    flow_sensors[bus_entry_upstream] = 0;
                    flow_sensors[branch_entry_upstream] = 1;
                }
            }
        }
```

docs/user_manual/calculations.md
TonyXiang8787: What do you mean here? If the measurements result in an underdetermined system of equations, they are not independent enough.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: If we prefer to have `int_8`, we could make it one bigger, and use the final entry to count along the way. This avoids the `O(n)` check for `n_flow_sensors`.

docs/user_manual/calculations.md
mgovers: ohhh ok sorry i mis-read this

docs/user_manual/calculations.md
mgovers: the word `usually` may be misleading and/or confusing, though...

```
"Sparse matrix error, possibly singular matrix!\n"
If you get this error from state estimation, it usually means the system is not fully observable, i.e. not enough measurements.
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,114 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
...
+                 element_index != y_bus_structure.y_bus_entry_indptr[ybus_index + 1]; ++element_index) {
+                YBusElement const& element = y_bus_structure.y_bus_element[element_index];
+                // shunt should not be considered
+                // if the branch is fully connected and measured, we consider it as a valid flow sensor
+                // we only need one flow sensor, so the loop will break
+                if (element.element_type != YBusElementType::shunt) {
+                    Idx const branch = element.idx;
+                    if ((measured_values.has_branch_from(branch) || measured_values.has_branch_to(branch)) &&
+                        topo.branch_bus_idx[branch][0] != -1 && topo.branch_bus_idx[branch][1] != -1) {
+                        flow_sensors[ybus_index] = 1;
```
Jerry-Jinfeng-Guo, line 59: If we opt for `int_8` vector but one bigger,
```suggestion
                        flow_sensors[ybus_index] = 1;
                        flow_sensors.back()++;
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,114 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
...
+
+    // for radial grid without phasor measurement, try to assign injection sensor to branch sensor
+    // we can then check sufficient condition for observability
+    if (topo.is_radial && n_voltage_phasor_sensor == 0) {
+        detail::assign_injection_sensor(y_bus_structure, flow_sensors);
+        // count flow sensors again
+        Idx const n_flow_sensor_new = std::reduce(flow_sensors.cbegin(), flow_sensors.cend(), Idx{}, std::plus<Idx>{});
+        if (n_flow_sensor_new < n_bus - 1) {
+            throw NotObservableError{"The number of power sensors seems to be sufficient, but they are not independent "
+                                     "enough. The system is still not observable.\n"};
```
Jerry-Jinfeng-Guo, line 136: "seems to be" is rather vague, in this case we do know the number is indeed enough. Perhaps only:
```suggestion
            throw NotObservableError{"The number of power sensors appears sufficient, but they are not independent "
                                     "enough. The system is still not observable.\n"};
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
TonyXiang8787: I do not understand the proposed change. What's the meaning of double `if` with the same condition?

Jerry-Jinfeng-Guo: They are logically equivalent, just one branch less

TonyXiang8787: You have two nested if with the same condition, why is that?

```
        if (flow_sensors[branch_entry_upstream] == 0) {
            if (flow_sensors[branch_entry_upstream] == 0) {
```

Jerry-Jinfeng-Guo: > You have two nested if with the same condition, why is that?
> 
> ```
>         if (flow_sensors[branch_entry_upstream] == 0) {
>             if (flow_sensors[branch_entry_upstream] == 0) {
> ```

Extra sure ;)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
TonyXiang8787: `std::vector<bool>` is not STL container. So I do not use them.

Count along the way will not help for the sufficient condition check. In the second step we need to move things around and re-count.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
TonyXiang8787: Okay, the proposed code is logically not valid anymore. Now you always set `bus_entry_upstream` to zero. We only want want to set `bus_entry_upstream` to zero if current bus needs to steal the upstream injection into the branch.

If we don't need to do that, the `bus_entry_upstream` should remain `1`. So when it is the turn of that bus, it can be used to enable its own upstream branch.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
TonyXiang8787: `int8` will overflow very quickly if we have more than 128 nodes. 

We can make a separate counter in this process. It can avoid one count for the necessary condition check. For the sufficient condition check, we still need to re-count.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: Apart from not continuous in memory, `std::vector<bool>` is the same the way a container is used here?

docs/user_manual/calculations.md
Jerry-Jinfeng-Guo: Github changes ;)

docs/user_manual/calculations.md
TonyXiang8787: Please re-apply the suggestion. I close this one. The GitHub cannot apply this anymore.

docs/user_manual/calculations.md
```diff
@@ -57,10 +57,10 @@ Output:
...
-raise a sparse matrix error; the matrix reprensentation of the system is too sparse to solve and particular so when it is
-singular. In short, meeting the requirement of observability indicates that the system is either an overdetermined system (when the number of measurements is larger than the number of
-unknowns) or a balanced system (when the number of measurements is equal to the number of unknowns). For each node, there are two unknowns, `u` and `u_angle`, so the following conditions should be met:
+In order to perform a state estimation, the system should be observable. If the system is not observable, the calculation will raise either a `NotObservableError` or 
+a `SparseMatrixError`.
+In short, meeting the requirement of observability indicates that the system is either an overdetermined system (when the number of measurements is larger than the number of
+... For each node, there are two unknowns, `u` and `u_angle`. Due to the relative nature of `u_angle` (relevant only in systems with at least two nodes), in total the following conditions should be met:
```
Jerry-Jinfeng-Guo, line 63:
```suggestion
In short, meeting the requirement of observability indicates that the system is either an overdetermined system (when the number of measurements is larger than the number of unknowns.
For each node, there are two unknowns, `u` and `u_angle`. Due to the relative nature of `u_angle` (relevant only in systems with at least two nodes), in total the following conditions should be met:
```

docs/user_manual/calculations.md
TonyXiang8787: Changed error message.

Jerry-Jinfeng-Guo (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,114 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
...
-                                          std::shared_ptr<MathModelTopology const> const& topo) {
+std::vector<int8_t> count_flow_sensors(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
+                                       YBusStructure const& y_bus_structure) {
+    Idx const n_bus{topo.n_bus()};
+    std::vector<int8_t> flow_sensors(y_bus_structure.row_indptr.back(), 0); // initialize all to zero
+    for (Idx row = 0; row != n_bus; ++row) {
+        // lower triangle is ignored and kept as zero
+        // diagonal for bus injection measurement
+        if (measured_values.has_bus_injection(row)) {
+            flow_sensors[y_bus_structure.bus_entry[row]] = 1;
```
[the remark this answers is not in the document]
TonyXiang8787: I am hesitating on that idea. The numeric number has its own meaning already: number of independent sensors in this bus/branch.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
TonyXiang8787: this is not actually a bool. See my other comments: it is a meaningful number: number of independent sensors.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -86,11 +89,21 @@ TEST_CASE("Necessary observability check") {
...
     }
+    SUBCASE("Not independent") {
+        // set branch sensor to bus_1 <-branch_1-> bus_2
+        // it is not independent with injection sensor of bus_2
+        topo.power_sensors_per_branch_from = {from_sparse, {0, 0, 1, 1}};
+        // set non phasor measurement
+        se_input.measured_voltage = {{{1.0, nan}, 1.0}};
+        // this will throw NotObservableError
+        check_not_observable(topo, param, se_input);
+    }
```
mgovers, line 106: can we add another case with a meshed grid or is that overkill?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,115 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
...
+        // if the upstream branch is not measured
+        if (flow_sensors[branch_entry_upstream] == 0) {
+            if (flow_sensors[bus_entry_current] == 1) {
+                // try to steal from current bus
+                flow_sensors[bus_entry_current] = 0;
+                flow_sensors[branch_entry_upstream] = 1;
+            } else if (flow_sensors[bus_entry_upstream] == 1) {
+                // if not possible, steal from upstream bus
+                flow_sensors[bus_entry_upstream] = 0;
+                flow_sensors[branch_entry_upstream] = 1;
```
mgovers, line 96: does it make sense to use `swap`?
```suggestion
                // try to steal from current bus
                std::swap(flow_sensors[branch_entry_upstream], flow_sensors[bus_entry_current]);
            } else if (flow_sensors[bus_entry_upstream] == 1) {
                // if not possible, steal from upstream bus
                std::swap(flow_sensors[branch_entry_upstream], flow_sensors[bus_entry_upstream]);
```

tests/cpp_unit_tests/test_observability.cpp
TonyXiang8787: I don't think it is necessary here.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -58,29 +28,114 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
...
+        Idx const branch_entry_upstream = bus_entry_current + 1;
+        // there should be only one upstream branch in the upper diagonal
+        // so the next of branch_entry_upstream is already the next row
+        // because the grid is radial
+        assert(y_bus_structure.row_indptr[current_bus + 1] == branch_entry_upstream + 1);
+        Idx const upstream_bus = y_bus_structure.col_indices[branch_entry_upstream];
+        Idx const bus_entry_upstream = y_bus_structure.bus_entry[upstream_bus];
+        // if the upstream branch is not measured
+        if (flow_sensors[branch_entry_upstream] == 0) {
+            if (flow_sensors[bus_entry_current] == 1) {
```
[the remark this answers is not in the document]
TonyXiang8787: See my earlier comments. What do you think?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: In favor of `1` and `0` as well. K.I.S.S.

Jerry-Jinfeng-Guo (approved):
Since @mgovers has signed off. I hereby resolve the comments and approve the PR.

PR #860, opened 2024-12-18
Resolve difference between input and update

mgovers:
It turns out that input data and update data are not treated exactly the same way in power sensors.

When rounding errors caused by this propagate, a difference may occur between single calculations run with measurement data provided in input data, or (batch or single) calculations with power sensor measurements provided in (batch or permanent) update data.

NOTE: similar rounding errors introduced when calculating the inverse update may also propagate to other scenarios, so the result may even differ between different orderings of batch scenarios or threading levels

* [x] power sensor
  * [x] this fix works as is for the input vs update of power sensor
  * [x] ~but not yet for the inverse update~ EDIT: as mentioned below: we do not apply the propagation of inverse update calculation in this PR. It may not be as big of an issue as one would expect.
* [x] voltage sensor
  * [x] ~from a quick investigation, it turns out that the voltage sensor may also be prone to this, but no combination of values that reproduces this bug has been found yet~ EDIT: a combination of values was found and added to the unit tests.
* [x] other component types
  * [x] it seems that other component types are not prone to this bug, because their input/update values are directly stored without modifications.
  * [x] this suggests that doing so for the sensors as well provides a sustainable solution

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/power_sensor.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/power_sensor.hpp
@@ -67,6 +67,8 @@ class GenericPowerSensor : public Sensor {
 };
 
 template <symmetry_tag power_sensor_symmetry_> class PowerSensor : public GenericPowerSensor {
+    static constexpr double inv_base_power = 1.0 / base_power<power_sensor_symmetry_>;
+
   public:
     using power_sensor_symmetry = power_sensor_symmetry_;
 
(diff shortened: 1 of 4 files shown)


TonyXiang8787:
@mgovers what exactly is the drifting here? We do not have addition/subtraction drift right? Multiple/division usually does not cause large drift.

power_grid_model_c/power_grid_model/include/power_grid_model/component/power_sensor.hpp
```diff
@@ -67,6 +67,8 @@ class GenericPowerSensor : public Sensor {
 };
 
 template <symmetry_tag power_sensor_symmetry_> class PowerSensor : public GenericPowerSensor {
+    static constexpr double inv_base_power = 1.0 / base_power<power_sensor_symmetry_>;
```
Jerry-Jinfeng-Guo, line 70: So, in the end it's always fixed by some suggestions existed since last century...

mgovers:
@TonyXiang8787 I added you as a reviewer because I would like you to be the one to sign off on this change.

TonyXiang8787 (approved):

Issue #547, opened 2024-03-26
[FEATURE] Support current measurement in state estimation


TonyXiang8787:
# Background

Currently the state estimation supports two types of sensors: `voltage_sensor` and `power_sensor`. Their meaning is self explanatory. In distribution grids, we have also many current measurements. Supporting current magnitude measurement is difficult since we do not know the direction of active/reactive power. To decouple the current, we need the voltage as a reference, with which we can calculate power. Therefore, we only support power measurements as this moment.

# Current measurement with relative phase angle to the local voltage

Sometimes we have both voltage and current measurements in practice. However, certain voltage magnitude measurement has prohibitable high error margin which makes the calculated power value useless.  Meanwhile, the phase angle between the voltage and current is actually accurately measured. In this case, we have current magnitude + relative phase angle measurement.

# Current measurement using PMU with global phase angle

If we install PMU current measurement in the grid, we can measure the current magnitude and the global phase angle with regarding the reference angle in the HV/MV substation. In this way we measure the current phasor.

# Feature request

This issue proposes to support decoupled current measurement into PGM state estimation. Concretely, the following tasks:

* [x] Define new component `sym_current_sensor` and `asym_current_sensor`.
  * [x] They should have the attributes `i_measured` and `i_angle_measured`, representing the current magnitude and angle.
  * [x] They should have the attributes `angle_measurement_type` which is an enumeration of `local` and `global`.
    * [x] In case of `local`, this is a relative angle to the local voltage.
    * [x] In case of `global`, this is a PMU current phasor with the global angle to the reference angle in the HV/MV substation.
  * [x] The reference direction should be the same as the measured object, similar to `power_sensor`.
  * [x] Also define relevant sigmas.
  * [x] They can only measure `branch` and `branch3`
* [x] In the observability check, treat the current sensor the same as power sensor. They both contribute to observability.
* [x] In the measured values, treat the current sensor similarly as power sensor, but accumulate separately.
  * [x] If there is no voltage angle measurement but there are global angle current sensors in the connected component of the grid, throw an error
  * [x] If there is a mixture of local and global angle current sensors on the same terminal (i.e., the same branch_from or branch_to of the same component), throw an error.
    * [x] NOTE: it is supported to have a mixture as long as they are on different terminals, e.g., a local angle current sensor on the branch from side and a global angle current sensor on the branch to side of the same branch is OK.
[...]

PR #861, opened 2024-12-24
current sensor skeleton

petersalemink95:
Creates a skeleton for #547.


code_generation/data/attribute_classes/input.json
```diff
@@ -494,6 +494,40 @@
...
+                {
+                    "data_type": "AngleMeasurementType",
+                    "names": "angle_measurement_type",
+                    "description": "type of angle measurement"
+                },
+                {
+                    "data_type": "double",
+                    "names": [
+                        "i_sigma", 
+                        "i_angle_sigma"
```
petersalemink95, line 511: TODO: determine the right way to formulate sigma's for a current sensor.
Multiple options to investigate:
- just `i_sigma` - similar as for voltage sensor
- `i_sigma` and `i_angle_sigma` as `double` - similar as `power_sigma`
- `i_sigma` and `i_angle_sigma` as `RealValue<sym>` - similar as `p_sigma` and `q_sigma`

pyproject.toml
```diff
@@ -103,7 +103,7 @@ ignore_missing_imports = true
 show_column_numbers = true
 non_interactive = true
 install_types = true
-
+disable_error_code = ["call-overload"]  # mypy 1.14.0 does not like PGM's data_types
```
petersalemink95, line 106: FYI: mypy's 1.14.0 version - released Friday before the Holidays 😕  - does not like PGM's data_types. We need to look into this after the Holidays. For now, I'm ignoring the error.

code_generation/data/attribute_classes/input.json
```diff
@@ -494,6 +494,42 @@
...
+                    "description": "type of measured terminal"
+                },
+                {
+                    "data_type": "AngleMeasurementType",
+                    "names": "angle_measurement_type",
+                    "description": "type of angle measurement"
+                },
+                {
+                    "data_type": "double",
+                    "names": "i_sigma",
```
petersalemink95, line 514: TODO: determine the right way to formulate sigma's for a current sensor.
Multiple options to investigate:

- just i_sigma - similar as for voltage sensor - chosen now
- i_sigma and i_angle_sigma as double - similar as power_sigma
- i_sigma and i_angle_sigma as RealValue<sym> - similar as p_sigma and q_sigma

petersalemink95: After discussing with @TonyXiang8787: `i_sigma` and `i_angle_sigma` as a double, where `i_angle_sigma` is Optional.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -108,6 +108,20 @@ template <symmetry_tag sym_type> struct PowerSensorCalcParam {
...
+//   * for branches, the direction is node -> branch
+template <symmetry_tag sym_type> struct CurrentSensorCalcParam {
+    using sym = sym_type;
+
+    static constexpr bool symmetric{is_symmetric_v<sym>};
+
+    ComplexValue<sym> value{};
+    RealValue<sym> i_variance{};       // variance (sigma^2) of the error range of the current, in p.u.
+    RealValue<sym> i_angle_variance{}; // variance (sigma^2) of the error range of the current angle, in p.u.
+};
```
TonyXiang8787, line 123: Since the `i_sigma` and `i_angle_sigma` are always single values (not three phase) as decided. The variance should also be always `double`. Just as voltage sensor.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -108,6 +108,20 @@ template <symmetry_tag sym_type> struct PowerSensorCalcParam {
...
+// current sensor calculation parameters for state estimation
+// The value is the complex current
+//   * for appliances, it is always in injection direction
+//   * for branches, the direction is node -> branch
+template <symmetry_tag sym_type> struct CurrentSensorCalcParam {
+    using sym = sym_type;
+
+    static constexpr bool symmetric{is_symmetric_v<sym>};
+
+    ComplexValue<sym> value{};
```
TonyXiang8787, line 120: Angle measurement type should be part of the calculation parameter. We need this to handle differently in the mathematical core.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
petersalemink95: done

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
petersalemink95: done

tests/cpp_unit_tests/test_current_sensor.cpp
```diff
@@ -0,0 +1,227 @@
...
+        CAPTURE(i);
+        check_nan_preserving_equality(actual(i), expected(i));
+    }
+}
+} // namespace
+
+TEST_CASE("Test current sensor") {
+    SUBCASE("Symmetric Current Sensor - generator, branch_from, branch_to, source") {
+        for (auto const terminal_type : {MeasuredTerminalType::generator, MeasuredTerminalType::branch_from,
+                                         MeasuredTerminalType::branch_to, MeasuredTerminalType::source}) {
```
figueroa1395, line 31: Can't the terminal type also be of a `load`, `node`, etc?

tests/cpp_unit_tests/test_current_sensor.cpp
```diff
@@ -0,0 +1,227 @@
...
+        CHECK(actual == doctest::Approx(expected));
+    }
+}
+void check_nan_preserving_equality(RealValue<asymmetric_t> const& actual, RealValue<asymmetric_t> const& expected) {
+    for (auto i : {0, 1, 2}) {
+        CAPTURE(i);
+        check_nan_preserving_equality(actual(i), expected(i));
+    }
+}
+} // namespace
```
figueroa1395, line 26: Maybe out of scope, but I see that these are used across different unit tests, so perhaps it is a good idea to place this in a common utility file for tests, or something like that.

tests/cpp_unit_tests/test_current_sensor.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+TEST_CASE("Test current sensor") {
+    SUBCASE("Symmetric Current Sensor - generator, branch_from, branch_to, source") {
```
figueroa1395, line 29: Aren't we missing asymmetric current sensor?

tests/cpp_unit_tests/test_current_sensor.cpp
```diff
@@ -0,0 +1,227 @@
...
+            CHECK(is_nan(sym_sensor_output.id));
+            CHECK(is_nan(sym_sensor_output.energized));
+            CHECK(is_nan(sym_sensor_output.i_residual));
+            CHECK(is_nan(sym_sensor_output.i_angle_residual));
+
+            // Check symmetric sensor output for asymmetric parameters
+            CHECK(asym_sensor_param.i_variance == doctest::Approx(0.0));
+            CHECK(asym_sensor_param.i_angle_variance == doctest::Approx(0.0));
+            CHECK(real(asym_sensor_param.value[0]) == doctest::Approx(0.0));
+            CHECK(imag(asym_sensor_param.value[1]) == doctest::Approx(0.0));
```
figueroa1395, line 75: Input and output are checked, aren't we missing update?

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+        update_real_value<current_sensor_symmetry>(update_data.i_measured, i_measured_, scalar);
+        update_real_value<current_sensor_symmetry>(update_data.i_angle_measured, i_angle_measured_, 1.0);
+        return {false, false};
+    }
+
+    CurrentSensorUpdate<current_sensor_symmetry> inverse(CurrentSensorUpdate<current_sensor_symmetry> update_data,
+                                                         double const& u_rated) const {
+        assert(update_data.id == this->id() || is_nan(update_data.id));
+
+        double const base_current = base_power_3p / u_rated / sqrt3;
```
nitbharambe, line 108: Can be stored in base_i_ since it gets calculated multiple times

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    explicit CurrentSensor(CurrentSensorInput<current_sensor_symmetry> const& current_sensor_input, double u_rated)
+        : GenericCurrentSensor{current_sensor_input},
+          i_angle_measured_{current_sensor_input.i_angle_measured},
+          i_angle_sigma_{current_sensor_input.i_angle_sigma} {
+        set_current(current_sensor_input, u_rated);
+    };
+
+    UpdateChange update(CurrentSensorUpdate<current_sensor_symmetry> const& update_data, double const& u_rated) {
+        double const base_current = base_power_3p / u_rated / sqrt3;
```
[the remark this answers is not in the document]
mgovers: `inv_sqrt3` is actually defined in `common.hpp` (directly included from `std::numbers`)

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -108,6 +108,21 @@ template <symmetry_tag sym_type> struct PowerSensorCalcParam {
...
+//   * for branches, the direction is node -> branch
+template <symmetry_tag sym_type> struct CurrentSensorCalcParam {
+    using sym = sym_type;
+
+    static constexpr bool symmetric{is_symmetric_v<sym>};
+
+    AngleMeasurementType angle_measurement_type{};
+    ComplexValue<sym> value{};
+    double i_variance{};       // variance (sigma^2) of the error range of the current, in p.u.
+    double i_angle_variance{}; // variance (sigma^2) of the error range of the current angle, in p.u.
```
mgovers, line 123: it feels a bit odd for me that the value is a complex value but then the variances are 2 doubles with a different meaning.

For power sensors, the first represents variance along the real axis while the second represents variance along the imaginary axis.

however, in this case, the first represents magnitude while the second represents variance along the complex phase. Since this is an entirely different representation, I'm leaning towards treating `CurrentSensorCalcParam` as a combination of 2 independent `UniformRandomVariable`s, e.g.

```cpp
// current sensor calculation parameters for state estimation
// The value is the complex current
//   * for appliances, it is always in injection direction
//   * for branches, the direction is node -> branch
template <symmetry_tag sym_type> struct CurrentSensorCalcParam {
    using sym = sym_type;

    static constexpr bool symmetric{is_symmetric_v<sym>};

    AngleMeasurementType angle_measurement_type{};
    UniformRealRandomVariable<sym> i_magnitude{};
    UniformRealRandomVariable<sym> i_angle{};
```

what are your thoughts on that?

code_generation/data/attribute_classes/output.json
```diff
@@ -293,6 +293,21 @@
             "base": "BaseOutput",
             "is_template": false,
             "attributes": []
+        },
+        {
+            "name": "CurrentSensorOutput",
+            "base": "BaseOutput",
```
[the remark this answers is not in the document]
mgovers: cause the base class would be empty. only the input type contains new attributes

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+    virtual CurrentSensorCalcParam<symmetric_t> sym_calc_param() const = 0;
+    virtual CurrentSensorCalcParam<asymmetric_t> asym_calc_param() const = 0;
+
+    virtual CurrentSensorOutput<symmetric_t> get_sym_output(ComplexValue<symmetric_t> const& i) const = 0;
+    virtual CurrentSensorOutput<asymmetric_t> get_asym_output(ComplexValue<asymmetric_t> const& i) const = 0;
+};
+
+template <symmetry_tag current_sensor_symmetry_> class CurrentSensor : public GenericCurrentSensor {
+  public:
+    using current_sensor_symmetry = current_sensor_symmetry_;
```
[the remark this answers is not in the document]
mgovers: that's c++ being meh. they need to be different symbols, which is truly terrible

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+    UpdateChange update(CurrentSensorUpdate<current_sensor_symmetry> const& update_data, double const& u_rated) {
+        double const base_current = base_power_3p / u_rated / sqrt3;
+        double const scalar = convert_direction() / base_current;
+
+        if (!is_nan(update_data.i_sigma)) {
+            i_sigma_ = update_data.i_sigma / base_current;
+        }
+        if (!is_nan(update_data.i_angle_sigma)) {
+            i_angle_sigma_ = update_data.i_angle_sigma;
+        }
```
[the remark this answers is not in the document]
mgovers: yes, that's how our update logic works: either we set it (if value) else we leave it be (nan)

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+        update_real_value<current_sensor_symmetry>(update_data.i_measured, i_measured_, scalar);
+        update_real_value<current_sensor_symmetry>(update_data.i_angle_measured, i_angle_measured_, 1.0);
+        return {false, false};
+    }
+
+    CurrentSensorUpdate<current_sensor_symmetry> inverse(CurrentSensorUpdate<current_sensor_symmetry> update_data,
```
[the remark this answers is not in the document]
mgovers: the way our batch logic works is:
```cpp
auto const cache = cache_state();
setup_scenario(update_data);
calculate();
restore_scenario(cache);
```

but we actually do it more memory-efficiently by only caching the deltas:

```cpp
auto const inverse_update_data = inverse(update_data);
update(update_data)
calculate();
update(inverse_update);
```

where supposedly `inverse_update` is the update data such that if applied to an updated element, it restores the original one.

notice that we take a shortcut in our implementation where it's not an exactly inverse update for some component types, but rather an approximate one, and that's the drift we have been talking about.

code_generation/data/attribute_classes/output.json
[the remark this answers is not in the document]
petersalemink95: Similar as for other components

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    explicit GenericCurrentSensor(GenericCurrentSensorInput const& generic_current_sensor_input)
+        : Sensor{generic_current_sensor_input},
+          terminal_type_{generic_current_sensor_input.measured_terminal_type},
+          angle_measurement_type_{generic_current_sensor_input.angle_measurement_type} {}
+
+    MeasuredTerminalType get_terminal_type() const { return terminal_type_; }
+    AngleMeasurementType get_angle_measurement_type() const { return angle_measurement_type_; }
+
+    template <symmetry_tag sym> CurrentSensorOutput<sym> get_output(ComplexValue<sym> const& i) const {
```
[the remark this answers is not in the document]
petersalemink95: Yes

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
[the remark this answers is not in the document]
petersalemink95: Not something introduced here, is done everywhere, because of what Martijn explained

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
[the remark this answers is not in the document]
petersalemink95: Done @Jerry-Jinfeng-Guo

tests/cpp_unit_tests/test_current_sensor.cpp
petersalemink95: I think there was a reason why @mgovers put these in the tests in a nameless namespace. Don't remember exactly why

tests/cpp_unit_tests/test_current_sensor.cpp
petersalemink95: Yes, we are missing many things - the logic is not implemented, so it does not make sense to add all tests. If you scroll down a bit you see we check if everything is `0.0` or `nan`. Writing full tests should be done when filling the functions with the right logic

tests/cpp_unit_tests/test_current_sensor.cpp
petersalemink95: yes, same here, only way to check update is through running `calc_param` and `get_output`, which are not filled yet

tests/cpp_unit_tests/test_current_sensor.cpp
mgovers: anonymous namespace has only 1 reason: to explicitly tell the compiler that this function is restricted to this translation unit. it is like how you would always have to write `inline` if it is declared in a header file. If you don't, you get name conflicts and linking errors.

It is a consequence of the choice of putting the functions locally in this file, not a reason for it.

I do think moving it to a separate header file makes sense at this point, and indeed, it should not be in an anonymous namespace but marked `inline` at that point.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
TonyXiang8787: I do not get your intention. We want a complex number for current instead of two.

tests/cpp_unit_tests/test_current_sensor.cpp
petersalemink95: Let's do that move in a separate PR ;)

tests/cpp_unit_tests/test_current_sensor.cpp
petersalemink95: No, it behaves similar as a power sensor -> for load and shunt the direction is flipped. However, this question resulted in remembering that we only implement current sensor for branches in this stage, so I adjusted the code. Thanks

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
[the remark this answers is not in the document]
petersalemink95: No, variances come from the `double` sigmas

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
petersalemink95: Done

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: i understand that and i agree that having a complex value makes sense. what i do not completely know for sure is how the 'weird' standard deviation transforms the data and would like to think about it just a little bit more before resolving this.

* on one end, there are 2 independent random variables $|I|$ and `theta_I`
* on another, there is one complex variable $I$ with standard deviations $\sigma_{|I|}$ and $\sigma_{\theta}$ along different axes than the real and imaginary axes.
* however, in complex random number analysis, there is, AFAIK, no standardized way of representing this. Maybe there is, but I would like to follow those conventions if possible

TonyXiang8787: To think about how things will be used in the core, we can do this:

```c++
    ComplexValue<sym> value{};
    double i_real_variance{};       // variance (sigma^2) of the error range of real part of the current, in p.u.
    double i_imag_variance{}; // variance (sigma^2) of the error range of imaginary part of the current, in p.u.
```

For NRSE, we treat the real and imaginary part as separate and independent real random variables with separate `sigma` in NRSE, as we do for power sensor. 

For ILSE, we treat the whole complex current as one complex random variable, we add these two values together to have a `i_variance` which is the variance of the complex number.

In the component class we need to translate `i_angle_sigma` and `i_sigma` to real and imaginary part by (TBD)

mgovers: ok, that makes sense; i agree.

One more question to think about in the long term: how do we want to calculate the real and imaginary variances? do we want to do so by simulation or by linear approximation?

feel free to resolve this comment

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+            throw InvalidMeasuredTerminalType{current_sensor_input.measured_terminal_type, "Current sensor"};
+        }
+    };
+
+    UpdateChange update(CurrentSensorUpdate<current_sensor_symmetry> const& update_data) {
+        double const scalar = 1.0 / base_current_;
+
+        if (!is_nan(update_data.i_sigma)) {
+            i_sigma_ = update_data.i_sigma / base_current_;
+        }
```
Jerry-Jinfeng-Guo, line 99:
```suggestion
        double const scalar = 1.0 / base_current_;

        if (!is_nan(update_data.i_sigma)) {
            i_sigma_ = update_data.i_sigma * scalar;
        }
```

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+        return update_data;
+    }
+
+  private:
+    RealValue<current_sensor_symmetry> i_measured_{};
+    RealValue<current_sensor_symmetry> i_angle_measured_{};
+    double i_sigma_{};
+    double i_angle_sigma_{};
+    double base_current_{};
```
Jerry-Jinfeng-Guo, line 125: Perhaps it's worth caching `base_current_inv_` since it's constructed once and used the same amount as `base_current_` in the `set_current` function and the `update` function.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
petersalemink95: Calculation parameters are updated - calculation of real/imag variances is part of follow up PR when `calc_param` will be filled @mgovers @TonyXiang8787

petersalemink95: @Jerry-Jinfeng-Guo the naming is similar to the rest of the codebase - deviating from that just for the current sensor does not make sense to me ;)

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
petersalemink95: Good catch, done

power_grid_model_c/power_grid_model/include/power_grid_model/component/current_sensor.hpp
petersalemink95: Done

Jerry-Jinfeng-Guo (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: After going over the mathematics one more time, it turns out that the above formula is incorrect.
Instead, it should be:
```math
Var(\Re\{I e^{i\theta}\}) = Var(I \cos \theta) \approx Var(I)\cos^2 \theta + Var(\theta)I^2 \sin^2 \theta
```
```math
Var(\Im\{I e^{i\theta}\}) = Var(I \sin \theta) \approx Var(I)\sin^2 \theta + Var(\theta)I^2 \cos^2 \theta
```
```math
\sigma_{\Re} = \sqrt{Var(\Re\{I e^{i\theta}\})}
```
```math
\sigma_{\Im} = \sqrt{Var(\Im\{I e^{i\theta}\})}
```
This amounts to a box draw that weighs the extrapolated error bars to obtain the corners (see below images)
image
image

However, there are 3 things that need to be considered:
1. The resultant error bars are underestimated in one direction and over-estimated in the other (the supposedly more correct points are shown by arrows in the drawing point at improved corner points). This is especially visible in the 2nd image where $$\theta = \frac{\pi}{2}$$.
2. The estimator of the mean is biased because the center of gravity of the resultant box does not weigh the fact that the prior distribution of the measurement is non-homogeneous. I don't think this is an issue but we should think about this properly
3. What should the error be when the angle is not measured? In that case, there is no prior knowledge about $$\theta$$. Worst case, if we assume the angle to be either $$0$$ or $$\frac{\pi}{2}$$, then either the real or the imaginary component is dominated entirely by the variance of the angle, which is not known.

TonyXiang8787: Agree with the new formula.

1. The over/under estimation is acceptable.
2. This should be an issue.
3. Angle for current sensor is required attribute, unlike voltage sensor.

PR #862, opened 2024-12-30
Using pivot perturbation for ill-conditioned state estimation

### LINKED ISSUE #853: [BUG] *descriptive_name*

TonyXiang8787:
Closes #853 

# Mathematical approach

This PR attempts to fix the problem by using pivot perturbation. This is an alternative method than partial/full pivoting method in the LU factorization. In this way, the structure of the sparse matrix is not changed. The methods are described in [[Li99]](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=A+Scalable+Sparse+Direct+Solver+Using+Static+Pivoting&btnG=) and [[Schenk04]](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=On+Fast+Factorization+Pivoting+Methods+for+Sparse+Symmetric+Indefinite+Systems&btnG=).

This PR applies the methods with following aspects:

1. The perturbation threshold is set to $\epsilon = 10^{-13}$, same as used in [MKL PARDISO](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-0/pardiso-iparm-parameter.html#IPARM9). This applied for both perturbed pivot and iterative refinement.
2. The infinite-norm of the matrix is calculated block-wise. First the infinite norm of each block is calculated, then the norm of the whole matrix by treating each block as a single value. When calculating the whole matrix, the diagonal blocks are ignored. In Power System State Estimation, we sometimes have huge diagonal values. We do not want those values to influence the norm of the matrix.
3. The backward error is calculated by capping the denominators to a minimum to avoid scenarios where both $|x|$ and $|b|$ are zero.

$$
D_{\max} = \max_i {(|A||x| + |b|)_i}
$$

```math
berr = \max_{ i } {\frac{|r|_i}{\max ( (|A||x| + |b|)_i,  \epsilon_{berr} D_{\max}) }}
```

```math
\epsilon_{berr} = 10^{-4}
```

5. The iterative refinement is capped at 5 times. The convergence criteria is $berr < \epsilon$.

# Implementation

We adjust `LUSparseSolver` to allow optionally pivot perturbation. To do this, the original matrix has to be cached to do the iterative refinement. Optional member data are defined in the class to store the cache. This perturbation is an opt-in feature which should only be enabled if we suspect ill-conditioned scenario. Currently, it is only enabled in State Estimation (both NR and IL) when the following two conditions are met:

1. The system is sufficiently observable. We need this to make sure that we do not create fake observability by perturbation.
2. The system could be possibly ill-conditioned. That means the following conditions are met:
    1. The bus does not have (zero) injection measurement.
    2. All upstream branches, i.e. branches between this bus and the buses with higher bus numbers, do not have branch measurement.
    3. The bus does not have a phasor voltage measurement.

# Check list

- [x] Implement the core
- [x] Unit test
- [x] Extensive validation test
- [x] Documentation
- [x] Run Nightly


mgovers:
can you explain to me why:
* you use a different value for epsilon from PARDISO (see below)
* you use a different stop criterion

I like the more stringent choice of parameters, but I would like to understand why certain decisions were made.

for reference, PARDISO uses `epsilon=1e-8` as a default value for symmetric indefinite matrices taken from (https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-0/pardiso-iparm-parameter.html#IPARM9)
image

TonyXiang8787:
> can you explain to me why:
> 
> * you use a different value for epsilon from PARDISO (see below)
> * you use a different stop criterion

For Q1, the PGM version 1.3.x and earlier, we use PARDISO sparse solver with the option `mtype=1` or `mtype=3`, see below. This will apply $\epsilon = 10^{-13}$ for perturbation. NOTE: structurally symmetric (but value not) is treated as not symmetric matrices, thus the same as `mtype=11` or `mtype=13`

https://github.com/PowerGridModel/power-grid-model/blob/abc71c5dd91e9dd2b086785595b242efe3d1df41/include/power_grid_model/math_solver/mkl_pardiso_solver.hpp#L360-L363

For Q2, in SE equations you often have some $x_i=0$ and $b_i=0$, which makes the denominator $|A||x|+|b|$ extremely small but not zero, which results large backward error. Therefore, we cannot perform the piecewise division and then maximum. We have to first select maximum of numerator and denominator, then divide.

mgovers:
> > can you explain to me why:
> > 
> > * you use a different value for epsilon from PARDISO (see below)
> > * you use a different stop criterion
> 
> For Q1, the PGM version 1.3.x and earlier, we use PARDISO sparse solver with the option `mtype=1` or `mtype=3`, see below. This will apply ϵ = 10 − 13 for perturbation. NOTE: structurally symmetric (but value not) is treated as not symmetric matrices, thus the same as `mtype=11` or `mtype=13`
> 
> https://github.com/PowerGridModel/power-grid-model/blob/abc71c5dd91e9dd2b086785595b242efe3d1df41/include/power_grid_model/math_solver/mkl_pardiso_solver.hpp#L360-L363
> 
> For Q2, in SE equations you often have some x i = 0 and b i = 0 , which makes the denominator | A | | x | + | b | extremely small but not zero, which results large backward error. Therefore, we cannot perform the piecewise division and then maximum. We have to first select maximum of numerator and denominator, then divide.

thank you, that explains a lot!

Jonasvdbo (approved):
Thank you for adding the notice to the code file. No further edits are required regarding the notice. I have added @mgovers for a full review of the change.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -57,11 +59,18 @@ std::vector<int8_t> count_flow_sensors(MeasuredValues<sym> const& measured_value
...
                 }
             }
         }
+        // check voltage sensor
+        if (measured_values.has_voltage(row) && measured_values.has_angle_measurement(row)) {
+            has_at_least_one_sensor = true;
+        }
+        // the system could be ill-conditioned if there is no flow sensor for one bus, except the last bus
+        possibly_ill_conditioned = possibly_ill_conditioned || (!has_at_least_one_sensor && row != n_bus - 1);
     }
```
Jerry-Jinfeng-Guo, line 74:
```suggestion
        if (possibly_ill_conditioned) {
            continue;
        }
        // check voltage sensor
        if (measured_values.has_voltage(row) && measured_values.has_angle_measurement(row)) {
            has_at_least_one_sensor = true;
        }
        // the system could be ill-conditioned if there is no flow sensor for one bus, except the last bus
        possibly_ill_conditioned = !has_at_least_one_sensor && row != n_bus - 1;
    }
```
Since `flow_sensors` is not touched and early out is possible.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+constexpr double epsilon_sqrt = 1.49011745e-8; // sqrt(epsilon), sqrt does not have constexpr
+
+// perturb pivot if needed
+// return updated abs value
+template <scalar_value Scalar>
+inline double perturb_pivot(Scalar& value, double perturb_threshold, bool& has_pivot_perturbation) {
+    double const abs_value = cabs(value);
+    if (abs_value < perturb_threshold) {
+        Scalar const scale = (abs_value == 0.0) ? Scalar{1.0} : (value / abs_value);
+        value = scale * perturb_threshold;
```
Jerry-Jinfeng-Guo, line 28:
```suggestion
        Scalar const sign = (abs_value == 0.0) ? Scalar{1.0} : (value / abs_value);
        value = sign * perturb_threshold;
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +404,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+            RealValueType const numerator = cabs(residual[row]);
+            backward_error_max_demoniator = std::max(backward_error_max_demoniator, max_val(denominator));
+            backward_error_max_numerator = std::max(backward_error_max_numerator, max_val(numerator));
+            // iterate x
+            x[row] += dx[row];
+        }
+        // then |r|_max / (|rhs| + |A| * |x|)_max
+        return backward_error_max_numerator / backward_error_max_demoniator;
+    }
+
```
Jerry-Jinfeng-Guo, line 505: In the original literature, the backward error is the outcome of an element/component wise operation. It is also taken as the maximum of the division where the subscript is identical. That being said, the question here is whether it is correct that the maximum numerator divided by maximum denominator results in the maximum of the division?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +404,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+            calculate_residual(x);
+            // solve with residual
+            solve(data, block_perm_array, residual_.value(), dx_.value());
+            // calculate backward error and then iterate x
+            backward_error = iterate_and_backward_error(x);
+        }
+        // reset refinement cache
+        reset_refinement_cache();
+    }
+
```
Jerry-Jinfeng-Guo, line 439: In case of letting this function arrange the `has_pivot_perturbation` logic:
```suggestion
    void solve_with_refinement(std::vector<Tensor> const& data,        // pre-factoirzed data, const ref
                               BlockPermArray const& block_perm_array, // pre-calculated permutation, const ref
                               std::vector<RHSVector> const& rhs, std::vector<XVector>& x,
                               bool has_pivot_perturbation) {
        // initialize refinement
        initialize_refinement(rhs, x);
        double backward_error{10.0};
        solve(data, block_perm_array, rhs_.value(), x);
        if (has_pivot_perturbation) {
            return;
        }
        Idx num_iter{};
        // iterate until convergence
        while (backward_error > epsilon_perturbation) {
            if (num_iter++ == max_iterative_refinement) {
                throw SparseMatrixError{};
            }
            // calculate residual
            calculate_residual(x);
            // solve with residual
            solve(data, block_perm_array, residual_.value(), dx_.value());
            // calculate backward error and then iterate x
            backward_error = iterate_and_backward_error(x);
        }
        // reset refinement cache
        reset_refinement_cache();
    }

```

Or if we can change some function name:

```suggestion
        // ...
        solve_once(data, block_perm_array, rhs_.value(), x);
        if (has_pivot_perturbation) {
            return;
        }
        Idx num_iter{};
        // iterate until convergence
        while (backward_error > epsilon_perturbation) {
            // ...
            // solve iteration with residual
            solve_once(data, block_perm_array, residual_.value(), dx_.value());
            // ...

```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +404,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+                } else {
+                    row_norm += cabs(data[idx]);
+                }
+            }
+            matrix_norm_ = std::max(matrix_norm_, row_norm);
+        }
+    }
+
+    void reset_matrix_cache() {
+        has_pivot_perturbation_ = false;
```
Jerry-Jinfeng-Guo, line 536: This line is redundant as in all entries to this function, `has_pivot_perturbation` is already `false`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -320,6 +392,10 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
             // iterate column position for the pivot
             ++col_position_idx[pivot_row_col];
         }
+        // if no pivot perturbation needed, reset cache
```
Jerry-Jinfeng-Guo, line 395:
```suggestion
        // if no pivot perturbation happened, reset cache
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+        for (int8_t pivot = 0; pivot != size; ++pivot) {
+            block_perm.q.applyTranspositionOnTheRight(pivot, col_transpositions[pivot]);
+        }
+
+        // throw SparseMatrixError if the matrix is ill-conditioned
+        // only check condition number if pivot perturbation is not used
+        double const pivot_threshold = has_pivot_perturbation ? 0.0 : epsilon * max_pivot;
+        for (int8_t pivot = 0; pivot != size; ++pivot) {
+            if (cabs(matrix(pivot, pivot)) < pivot_threshold || !is_normal(matrix(pivot, pivot))) {
+                throw SparseMatrixError{};
```
Jerry-Jinfeng-Guo, line 136:
```suggestion
                throw SparseMatrixError{"Matrix is ill-conditioned"};
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+                }
+                // perturbation used, continue
+            }
+
+            // perturb pivot if needed
+            double const abs_pivot = use_pivot_perturbation ? perturb_pivot(matrix(row_biggest, col_biggest),
+                                                                            perturb_threshold, has_pivot_perturbation)
+                                                            : cabs(matrix(row_biggest, col_biggest));
+            max_pivot = std::max(max_pivot, abs_pivot);
+
```
Jerry-Jinfeng-Guo, line 100: This is signed maximum pivot, right? 'cause `perturb_pivot` does give a signed value instead of an `abs_`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+    struct BlockPerm {
+        PermutationType p;
+        PermutationType q;
+    };
+
+    // factorize in place
+    // put permutation in block_perm in place
+    // put pivot perturbation in has_pivot_perturbation in place
+    template <class Derived>
+    static void factorize_in_place(Eigen::MatrixBase<Derived>&& matrix, BlockPerm& block_perm, double perturb_threshold,
```
Jerry-Jinfeng-Guo, line 60: Just to be more precise, shall we call this function `factorize_block_in_place`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: This returns a signed amount, right?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+            double const biggest_score = matrix.bottomRightCorner(size - pivot, size - pivot)
+                                             .cwiseAbs2()
+                                             .maxCoeff(&row_biggest_eigen, &col_biggest_eigen);
+            // offset with pivot
+            auto const row_biggest = static_cast<int8_t>(row_biggest_eigen + pivot);
+            auto const col_biggest = static_cast<int8_t>(col_biggest_eigen + pivot);
+
+            // check absolute singular matrix
+            if (biggest_score == 0.0) {
+                // no pivot perturbation, cannot proceed
```
Jerry-Jinfeng-Guo, line 83:
```suggestion
                // pivot perturbation not possible, cannot proceed
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+                // set identity permutation and break the loop
+                if (!use_pivot_perturbation) {
+                    for (int8_t remaining_rows_cols = pivot; remaining_rows_cols != size; ++remaining_rows_cols) {
+                        row_transpositions[remaining_rows_cols] = remaining_rows_cols;
+                        col_transpositions[remaining_rows_cols] = remaining_rows_cols;
+                    }
+                    break;
+                }
+                // perturbation used, continue
+            }
```
Jerry-Jinfeng-Guo, line 93:
```suggestion
                }
            }
```

Jerry-Jinfeng-Guo (commented):
First round on the apparent level.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -97,71 +222,10 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
-        // restore permutation for block matrix
-        if constexpr (is_block) {
-            for (Idx row = 0; row != size_; ++row) {
-                x[row] = (block_perm_array[row].q * x[row].matrix()).array();
-            }
+        if (has_pivot_perturbation_) {
+            solve_with_refinement(data, block_perm_array, rhs, x);
+        } else {
+            solve(data, block_perm_array, rhs, x);
         }
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Not really necessary after discussion with @mgovers

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: After discussion, only renaming to `solve_once` might be necessary.

mgovers (commented):
gone over everything except the sparse-LU solver itself

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -103,17 +112,26 @@ inline void assign_injection_sensor_radial(YBusStructure const& y_bus_structure,
...
+};
+
 template <symmetry_tag sym>
-inline void necessary_observability_check(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
-                                          YBusStructure const& y_bus_structure) {
+inline ObservabilityResult necessary_observability_check(MeasuredValues<sym> const& measured_values,
+                                                         MathModelTopology const& topo,
+                                                         YBusStructure const& y_bus_structure) {
     Idx const n_bus{topo.n_bus()};
+    ObservabilityResult result{};
```
mgovers, line 126: i heard someone mentioning a reset but i don't see it in the code. what is that about exactly?

tests/cpp_unit_tests/test_sparse_lu_solver.cpp
Jerry-Jinfeng-Guo, line 1: Shall we also add a test case with extreme small diagonals?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: I still have to investigate more to be sure. (in other words: not sure yet atm)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: It can be a complex number. We want to keep the angle of the original pivot complex number, and scale it into the minimum absolute value.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: See my comment here

https://github.com/PowerGridModel/power-grid-model/pull/862#issuecomment-2573760076

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: This change does not make sense. Is it not like "not possible" but user disables pivot perturbation.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -545,10 +547,10 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
         RealValue<sym> delta_theta{};
         if (measured_values.has_angle_measurement(bus)) {
             delta_theta = RealValue<sym>{arg(measured_values.voltage(bus))} - RealValue<sym>{x_[bus].theta()};
-            w_theta = RealTensor<sym>{1.0};
+            w_theta = RealTensor<sym>{w_v};
```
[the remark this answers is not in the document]
TonyXiang8787: Yes, that's a bug

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -57,11 +59,18 @@ std::vector<int8_t> count_flow_sensors(MeasuredValues<sym> const& measured_value
...
                     }
                 }
             }
         }
+        // check voltage sensor
+        if (measured_values.has_voltage(row) && measured_values.has_angle_measurement(row)) {
+            has_at_least_one_sensor = true;
+        }
+        // the system could be ill-conditioned if there is no flow sensor for one bus, except the last bus
+        possibly_ill_conditioned = possibly_ill_conditioned || (!has_at_least_one_sensor && row != n_bus - 1);
```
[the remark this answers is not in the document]
TonyXiang8787: Creating a class is over engineering, I think. If you think this chain is error-prone, you can use an explicit `if` statement.

```c++
if (!has_at_least_one_sensor && row != n_bus - 1) {
    possibly_ill_conditioned = true;
}
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: This is not possible given the error code/types.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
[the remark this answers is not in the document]
TonyXiang8787: We should also modify the comment in the function.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +405,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+    std::optional<std::vector<XVector>> dx_;
+    std::optional<std::vector<RHSVector>> residual_;
+    std::optional<std::vector<RHSVector>> rhs_;
+
+    void solve_with_refinement(std::vector<Tensor> const& data,        // pre-factoirzed data, const ref
+                               BlockPermArray const& block_perm_array, // pre-calculated permutation, const ref
+                               std::vector<RHSVector> const& rhs, std::vector<XVector>& x) {
+        // initialize refinement
+        initialize_refinement(rhs, x);
+        double backward_error{10.0};
```
mgovers, line 422: i prefer to have no arbitrary bounds
```suggestion
        double backward_error{std::numeric_limits<double>::infinity()};
```
or
```suggestion
        double backward_error{std::numeric_limits<double>::max()};
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +405,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+            if (num_iter++ == max_iterative_refinement) {
+                throw SparseMatrixError{};
+            }
+            // calculate residual
+            calculate_residual(x);
+            // solve with residual
+            solve_once(data, block_perm_array, residual_.value(), dx_.value());
+            // calculate backward error and then iterate x
+            backward_error = iterate_and_backward_error(x);
+        }
```
mgovers, line 436: you don't need the first solve, right? the refinement algorithm already contains the first calculation using `residual_ = rhs_` and `dx_ = x` and the backward error may already be small enough after the first solve.

```suggestion
        double backward_error{std::numeric_limits::infinity()};
        Idx num_iter{};
        // iterate until convergence
        while (backward_error > epsilon_perturbation) {
            if (num_iter++ == max_iterative_refinement) {
                throw SparseMatrixError{};
            }
            // solve with residual
            solve_once(data, block_perm_array, residual_.value(), dx_.value());
            // calculate backward error and then iterate x
            backward_error = iterate_and_backward_error(x);
            // calculate residual
            calculate_residual(x);
        }
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +405,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+    void solve_with_refinement(std::vector<Tensor> const& data,        // pre-factoirzed data, const ref
+                               BlockPermArray const& block_perm_array, // pre-calculated permutation, const ref
+                               std::vector<RHSVector> const& rhs, std::vector<XVector>& x) {
+        // initialize refinement
+        initialize_refinement(rhs, x);
+        double backward_error{10.0};
+        solve_once(data, block_perm_array, rhs_.value(), x);
+        Idx num_iter{};
+        // iterate until convergence
+        while (backward_error > epsilon_perturbation) {
```
mgovers, line 426: this criterion still misses the "still worth it" check that

$$\text{berr} < \frac{1}{2}\cdot \text{lastberr}$$

was that intentional?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +405,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+        auto const& residual = residual_.value();
+        auto const& dx = dx_.value();
+        using RealValueType = std::conditional_t<is_block, Eigen::Array<double, block_size, 1>, double>;
+        // calculate backward error and then iterate x
+        for (Idx row = 0; row != size_; ++row) {
+            // error denominator by |rhs|
+            RealValueType denominator = cabs(rhs[row]);
+            // then append |A| * |x|
+            for (Idx idx = row_indptr[row]; idx != row_indptr[row + 1]; ++idx) {
+                denominator += dot(cabs(original_matrix[idx]), cabs(x[col_indices[idx]]));
```
mgovers, line 495: just to double-check is this indeed correct?

$$
(|A|\cdot |x|)_i := \sum_j |a_{ij}| |x_j|
$$

cause the absolute value of a matrix or a vector aren't technically defined in this paper.

EDIT: I looked up the original reference they cited (https://www3.math.tu-berlin.de/Vorlesungen/SS14/MatricesGraphsPDEs/paper_for_students/SIMAX-ardedu-0610013.pdf) and, indeed, there, they do properly define the symbolic notation $$|A|$$ to be the matrix consisting of entries $$a_{ij}$$, and idem for vectors, so this is correct

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +405,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+            // loop all columns
+            for (Idx idx = row_indptr[row]; idx != row_indptr[row + 1]; ++idx) {
+                // subtract
+                residual[row] -= dot(original_matrix[idx], x[col_indices[idx]]);
+            }
+        }
+    }
+
+    double iterate_and_backward_error(std::vector<XVector>& x) {
+        double backward_error_max_demoniator = 0.0;
```
mgovers, line 480: typo
```suggestion
        double backward_error_max_denominator = 0.0;
```
unless we're summoning demons here, in which case: let's keep it in 😄

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
mgovers: quoting that comment:

> For Q2, in SE equations you often have some $$x_i = 0$$ and $$b_i = 0$$, which makes the denominator $$| A | | x | + | b |$$ extremely small but not zero, which results large backward error. Therefore, we cannot perform the piecewise division and then maximum. We have to first select maximum of numerator and denominator, then divide.

dividing by the maximum of the backward denominators sounds like a dangerous criterion, because there may be a row `k` with a large residual and a low value for the denominator and another row `l` with a small residual but very high value for $$|A| |x| + |b|$$. In that case, the original criterion would fail on `k` (as it should). But in the current implementation, it would use the numerator from `k` and the denominator for `l`, and therefore the criterion will succeed incorrectly.

wouldn't it be better to use the current-row denominator when it is non-zero and only use the a fall-back when needed? In particular, the theorem 2 described in https://www3.math.tu-berlin.de/Vorlesungen/SS14/MatricesGraphsPDEs/paper_for_students/SIMAX-ardedu-0610013.pdf claims that such errors are suppressed if I understand it correctly.

either way, since this function deviates from literature, so we should at the very least add the quote from https://github.com/PowerGridModel/power-grid-model/pull/862#issuecomment-2573760076 as in-line documentation

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +405,210 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+    }
+
+    void initialize_pivot_perturbation(std::vector<Tensor> const& data) {
+        // save a copy of original matrix
+        original_matrix_ = data;
+        // calculate the block-wise non-diagonal infinite norm of the matrix
+        // that is:
+        // 1. calculate the infinite norm of each individual block
+        // 2. sum all norms of the blocks per row, except the diagonal block
+        // 3. take the maximum of all the sums
```
mgovers, line 514: i guess it's approximately good enough, but this is not an exact infinity norm:

take the following example:

$$
\begin{pmatrix}
\begin{pmatrix}
1 && 0 \\
0 && 3
\end{pmatrix} && \begin{pmatrix}
3 && 0 \\
0 && 0
\end{pmatrix}
\end{pmatrix}
$$

the infinity norm should be the abs sum of the 1st row: `1 + 3 = 4`, but the infinity norm of the block-wise infinity norms says that it's `3 + 3 = 6`, which is larger.

should we stick to the existing solution or change it to the `rowwise().sum().maxCoeff()` of the sums of the absolute values of the blocks?

mgovers (commented):
i've done a thorough review of all the code and all the maths. most of it looks good but i still have some open questions regarding the mathematics

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+                 (Derived::RowsAtCompileTime == size) && (Derived::ColsAtCompileTime == size))
+    {
+        TranspositionVector row_transpositions{};
+        TranspositionVector col_transpositions{};
+        double max_pivot{};
+
+        // main loop
+        for (int8_t pivot = 0; pivot != size; ++pivot) {
+            int row_biggest_eigen{};
+            int col_biggest_eigen{};
```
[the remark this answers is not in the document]
TonyXiang8787: EIGEN uses internally `int` for compile-time size of the matrix.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+            block_perm.q.applyTranspositionOnTheRight(pivot, col_transpositions[pivot]);
+        }
+
+        // throw SparseMatrixError if the matrix is ill-conditioned
+        // only check condition number if pivot perturbation is not used
+        double const pivot_threshold = has_pivot_perturbation ? 0.0 : epsilon * max_pivot;
+        for (int8_t pivot = 0; pivot != size; ++pivot) {
+            if (cabs(matrix(pivot, pivot)) < pivot_threshold || !is_normal(matrix(pivot, pivot))) {
+                throw SparseMatrixError{}; // can not specify error code
+            }
```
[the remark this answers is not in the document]
TonyXiang8787: We can throw earlier. I started this by completely mimicking the EIGEN full Pivot solver. But keep like this does not hurt. Who knows if we are going to use the null-space later for some reason.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -191,16 +262,18 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
             // return reference to pivot permutation
             BlockPerm const& block_perm = [&]() -> std::conditional_t<is_block, BlockPerm const&, BlockPerm> {
                 if constexpr (is_block) {
-                    LUFactor lu_factor(lu_matrix[pivot_idx]);
-                    // set a low threshold, because state estimation can have large differences in eigen values
-                    lu_factor.setThreshold(1e-100);
```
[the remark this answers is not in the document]
TonyXiang8787: Yes that's intended. We check against now machine precision. It leads to some failures in MV routes where we find new data issues.

The prevrious `1e-100` is just too sloppy.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -191,16 +262,18 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
-                    block_perm_array[pivot_row_col] = {lu_factor.permutationP(), lu_factor.permutationQ()};
+                    LUFactor::factorize_block_in_place(lu_matrix[pivot_idx].matrix(), block_perm_array[pivot_row_col],
+                                                       perturb_threshold, use_pivot_perturbation,
+                                                       has_pivot_perturbation_);
                     return block_perm_array[pivot_row_col];
                 } else {
+                    if (use_pivot_perturbation) {
+                        // set threshold to machine precision
+                        // record pivot perturbation
+                        perturb_pivot(lu_matrix[pivot_idx], perturb_threshold, has_pivot_perturbation_);
```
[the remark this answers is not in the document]
TonyXiang8787: The LU solver has a scalar mode and block mode.

In the implementation above, that is the block mode, you call Dense LU Factorization with full pivoting and perturbation (if enabled).

Here is the branch for scalar mode: like iterative current power flow.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: I have tried this way before. The program crashed in many places. I could have spent more time to cover some corner cases.

But I decided to revert back to the original.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: Other papers did mention the meaning of this, as you said.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: *not possible without doing what you mentioned and that is not ideal, is it?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: Makes sense

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: Agreed.

tests/cpp_unit_tests/test_sparse_lu_solver.cpp
mgovers: I'd like that

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+                    }
+                    break;
+                }
+            }
+
+            // perturb pivot if needed
+            double const abs_pivot = use_pivot_perturbation ? perturb_pivot(matrix(row_biggest, col_biggest),
+                                                                            perturb_threshold, has_pivot_perturbation)
+                                                            : cabs(matrix(row_biggest, col_biggest));
+            max_pivot = std::max(max_pivot, abs_pivot);
```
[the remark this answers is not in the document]
mgovers: You're right. My bad

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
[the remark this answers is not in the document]
mgovers: Ok

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
+            // check absolute singular matrix
+            if (biggest_score == 0.0) {
+                // pivot perturbation not possible, cannot proceed
+                // set identity permutation and break the loop
+                if (!use_pivot_perturbation) {
+                    for (int8_t remaining_rows_cols = pivot; remaining_rows_cols != size; ++remaining_rows_cols) {
+                        row_transpositions[remaining_rows_cols] = remaining_rows_cols;
+                        col_transpositions[remaining_rows_cols] = remaining_rows_cols;
+                    }
+                    break;
```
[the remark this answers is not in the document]
mgovers: As per the other discussion: not worth it now

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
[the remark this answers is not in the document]
mgovers: Ok, good call, thanks for explaining

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -191,16 +262,18 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
             // return reference to pivot permutation
             BlockPerm const& block_perm = [&]() -> std::conditional_t<is_block, BlockPerm const&, BlockPerm> {
                 if constexpr (is_block) {
-                    LUFactor lu_factor(lu_matrix[pivot_idx]);
-                    // set a low threshold, because state estimation can have large differences in eigen values
-                    lu_factor.setThreshold(1e-100);
-                    if (lu_factor.rank() < block_size) {
-                        throw SparseMatrixError{};
-                    }
+                    // set threshold to machine precision
```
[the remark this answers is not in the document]
mgovers: Cfr. Other comment: please update to "use machine precision by default"

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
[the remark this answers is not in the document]
mgovers: I think machine precision can be deduced from `numeric_limits` and/or other STL functionality

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: no time

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: This is not correct. `has_pivot_perturbation_` is not garanteed to be `false` when we reach here. Please revert the change.

TonyXiang8787: If you run factorization once where `has_pivot_perturbation_` is set to `true`. 

In the next calculation we run factorization again, it calls `reset_matrix_cache` with  `has_pivot_perturbation_` still being `true`. We need to force set it to `false`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: That what I call it block-wise off-diagonal infinite norm.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Jerry-Jinfeng-Guo: Great, resolved then.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: @mgovers , you have a valid point. But sometimes you have extremely small $|x|$ and $|b|$ which make the comparison not logical. My proposal is like this:

First get largest denominator.

$$
D_{\max} = \max_i {(|A||x| + |b|)}
$$

The get piecewise backward error and select the max among those where the denominator is significant to the comparison.

```math
berr = \max_{ i \in \{ i: 0 \le i < N, (|A||x| + |b|)_i \ge \epsilon D_{\max}  \} } {\frac{|r|}{|A||x| + |b|}}
```

Let me know what do you think.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: Yes. We want to the iteration go on. If it reaches maximum iteration (5), we throw Sparse Matrix Error.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
mgovers: this sounds like a good solution. Selecting on something like machine precision filters out rows where rounding errors potentially dominate, while maintaining the original criterion for all other rows.

Let's properly document this in code.

TonyXiang8787: @mgovers , I have one more adjustment. So put the minimum value of the denominator. So, if $|r|$ is super big for that row, it can still be caught.

$$
D_{\max} = \max_i {(|A||x| + |b|)_i}
$$

The get piecewise backward error by setting a minimum denominator, and select max.

```math
berr = \max_{ i } {\frac{|r|_i}{\max ( (|A||x| + |b|)_i,  \epsilon D_{\max}) }}
```

TonyXiang8787: No, I mean `max`. To avoid the denominator become too small.

tests/cpp_unit_tests/test_sparse_lu_solver.cpp
TonyXiang8787: We can add this later with the documentation of LU solver.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: Will not be relevant when implementing the new approach

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: @TonyXiang8787 @Jerry-Jinfeng-Guo @mgovers This is no longer relevant. I initially made the comment and it came from an overlook from my side, but after checking the observability in the correct place, I see it working as intended. I will resolve this.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: New method pushed

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -34,14 +34,16 @@ std::tuple<Idx, Idx> count_voltage_sensors(const Idx n_bus, const MeasuredValues
 // for upper triangle part, it will be one if there is branch flow sensor and the branch is fully connected
 template <symmetry_tag sym>
 std::vector<int8_t> count_flow_sensors(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
-                                       YBusStructure const& y_bus_structure) {
+                                       YBusStructure const& y_bus_structure, bool& possibly_ill_conditioned) {
```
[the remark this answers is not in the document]
TonyXiang8787: What do you mean a potential change? This is a private function which is only used in the main check observability function.

The reference which is passed in the main function is only a local variable which does not have side effect.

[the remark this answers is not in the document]
mgovers: From an outside user perspective, yes. However, if I were to change the code in the future, I might accidentally do such a change that breaks the logic. It's easy to make mistakes on code like this.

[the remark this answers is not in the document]
TonyXiang8787: solved.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -10,10 +10,136 @@
...
 
-template <class Tensor, class RHSVector, class XVector, class = void> struct sparse_lu_entry_trait;
+constexpr double epsilon = std::numeric_limits<double>::epsilon();
+constexpr double epsilon_perturbation = 1e-13; // perturbation threshold
+constexpr double epsilon_sqrt = 1.49011745e-8; // sqrt(epsilon), sqrt does not have constexpr
+
+// perturb pivot if needed
+// return updated abs value
+template <scalar_value Scalar>
+inline double perturb_pivot(Scalar& value, double perturb_threshold, bool& has_pivot_perturbation) {
```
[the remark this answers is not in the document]
TonyXiang8787: resolved.

figueroa1395:
Since this is both a big and a very important PR, this is a friendly reminder to run the Nightly build manually once it is "ready" to be merged.

mgovers:
For future reference purposes only

* important changes to mathematics after the initially proposed PR:
  * back error calculation: https://github.com/PowerGridModel/power-grid-model/pull/862#discussion_r1923397601
* Other discussions worth of reference:
  * absence of early-out found in literature https://github.com/PowerGridModel/power-grid-model/pull/862#discussion_r1920023840
  * element-wise absolute value of matrix: https://github.com/PowerGridModel/power-grid-model/pull/862#discussion_r1920047418
  * block-wise off-diagonal infinite norm: https://github.com/PowerGridModel/power-grid-model/pull/862#discussion_r1920074538
  * different implementation of iteration: https://github.com/PowerGridModel/power-grid-model/pull/862#discussion_r1920008902

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -328,6 +410,230 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+            }
+            // calculate residual
+            calculate_residual(x);
+            // solve with residual
+            solve_once(data, block_perm_array, residual_.value(), dx_.value());
+            // calculate backward error and then iterate x
+            backward_error = iterate_and_backward_error(x);
+        }
+        // reset refinement cache
+        reset_refinement_cache();
```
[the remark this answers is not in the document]
TonyXiang8787: The order is now wired, looking at the back error formula:

$$
D_{\max} = \max_i {(|A||x| + |b|)_i}
$$

```math
berr = \max_{ i } {\frac{|r|_i}{\max ( (|A||x| + |b|)_i,  \epsilon D_{\max}) }}
```

The backward error should be based on the residual and x of the same iteration. In your suggestion you are calculating the backward error with the x of current iteration and the residual of the last iteration.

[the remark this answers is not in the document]
TonyXiang8787: It's actually the same iteration, but wrapped around. The proposal should work.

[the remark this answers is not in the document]
TonyXiang8787: I have made the changes.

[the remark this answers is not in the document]
TonyXiang8787: @mgovers it is not working in some CI build. Maybe still need to revert?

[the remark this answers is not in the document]
TonyXiang8787: I found the bug. It was in the eigen initialization. This is resolved.

TonyXiang8787:
Nightly build in https://github.com/PowerGridModel/power-grid-model/actions/runs/12924710668

mgovers:
Running the benchmark yields comparatively similar results. I found no significant regressions. In fact, the new implementation often appears to give slightly better results, in particular for ClangCL, but it is most most likely within run differences on my machine and therefore not significant.

[output.zip](https://github.com/user-attachments/files/18517496/output.zip)

CMakeLists.txt
```diff
@@ -29,7 +29,11 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 include(GNUInstallDirs)
 
-cmake_policy(SET CMP0167 OLD)  # libboost-headers packaged by conda does not come with BoostConfig.cmake
+if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.30.0") 
+#  libboost-headers packaged by conda does not come with BoostConfig.cmake
+cmake_policy(SET CMP0167 OLD)  
+endif()
```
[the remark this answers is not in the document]
mgovers: whoops this is wrong. I will recreate

CMakeLists.txt
```diff
@@ -29,7 +29,11 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
...
+
 find_package(Boost REQUIRED)
```
mgovers, line 37:
```suggestion
cmake_policy(PUSH)
if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.30.0") 
  #  libboost-headers packaged by conda does not come with BoostConfig.cmake
  cmake_policy(SET CMP0167 OLD)  
endif()
find_package(Boost REQUIRED)
cmake_policy(POP)
```

TonyXiang8787: I don't think we need such a complicated stuff. Setting `CMP0167` permanently (for this project) would not hurt.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: The test on production data shows this adjustment does not work. We might need to revert.

TonyXiang8787: Commit in question: https://github.com/PowerGridModel/power-grid-model/pull/862/commits/cc270c0747c9bb873774407738bc8cb7323c268a

TonyXiang8787: After discussion the decision:

Change the denominator cap to $D_{\max}$ without $\epsilon$, which is essentially the same as the original proposal. 

Later we try to find an optimum scaling.

mgovers:
## Profiling results

using the Visual Studio 2022 Profiler on `msvc-release` on `power_grid_model_benchmark_cpp`

Very similar results between `main` and this branch up to the `SparseLUSolver` (Root in the profiles below). After that, the implementation changes a lot and therefore, the implementation is also different.

Note: the threaded version has almost identical behavior, as expected; see included `.diagsession` files

### `main`
image

### this branch
image
s/assets/61aeb8e5-e172-4d3c-bc42-b8d03dd72ff8)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: Final decision: set the cap to $10^{-4}D_{\max}$.

TonyXiang8787:
Full run in https://github.com/PowerGridModel/power-grid-model/actions/runs/12945646002

mgovers (approved):
I hereby approve the current set of changes

Jerry-Jinfeng-Guo (approved):

PR #863, opened 2024-12-30
disable mypy call-overload

petersalemink95:
Mypy 1.14.0 cannot handle PGM's data_types.

For now, let's ignore this. We need to investigate after the holidays

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ ignore_missing_imports = true
 show_column_numbers = true
 non_interactive = true
 install_types = true
-
+disable_error_code = ["call-overload"]  # mypy 1.14.0 does not like PGM's data_types
 
 # CI build options
 [tool.cibuildwheel]


TonyXiang8787 (approved):

mgovers:
After investigating, seems to be a NumPy typing regression and not a problem on our end. Basically, there are a lot of typing regressions in numpy 2.2.0.

We will probably want to track this overview issue: https://github.com/numpy/numpy/issues/28076
Note that mypy is also taking action in https://github.com/python/mypy/issues/18343

Minimal repro case for our use case below. When that passes, we should be able to revert this PR.

```py
import numpy as np

# our use case
STRUCTURED_ARR = np.array([(1, 2), (3, 4)], dtype=np.dtype([('a', int), ('b', int)]))
STRUCTURED_ARR['a'] = 10  # mypy error

# even smaller taken from https://github.com/numpy/numpy/issues/28076
ARR = np.arange(2)
ARR = x + 1  # mypy error
```

PR #871, opened 2025-01-14
Bugfix for quadratic blow-up of batch validation errors

mgovers:
batch validation errors were appended to the same list for every scenario

This meant that if there were 2 scenarios, then the same error would be reported 2x2=4 times instead of only 2 times (once for each scenario). For 4 scenarios, that would become 16 times, etc.

Note that this may result in high costs when trying to report an error.

This was introduced in #793 (release https://github.com/PowerGridModel/power-grid-model/releases/tag/v1.9.87) in commit https://github.com/PowerGridModel/power-grid-model/commit/3b0c6b60bf35a07cf739167e423a51ae021e0ffc

Credits to @BartSchuurmans for finding this

### DIFF
--- a/tests/unit/validation/test_batch_validation.py
+++ b/tests/unit/validation/test_batch_validation.py
@@ -6,19 +6,19 @@
 import numpy as np
 import pytest
 
-from power_grid_model import DatasetType, LoadGenType, initialize_array
+from power_grid_model import ComponentType, DatasetType, LoadGenType, initialize_array
 from power_grid_model._utils import compatibility_convert_row_columnar_dataset, is_columnar
 from power_grid_model.enum import ComponentAttributeFilterOptions
 from power_grid_model.validation import validate_batch_data
-from power_grid_model.validation.errors import MultiComponentNotUniqueError, NotBooleanError
+from power_grid_model.validation.errors import MultiComponentNotUniqueError, NotBetweenOrAtError, NotBooleanError
 
 
 @pytest.fixture
 def original_input_data() -> dict[str, np.ndarray]:
-    node = initialize_array("input", "node", 4)
+    node = initialize_array(DatasetType.input, ComponentType.node, 4)
     node["id"] = [1, 2, 3, 4]
     node["u_rated"] = 10.5e3
-    line = initialize_array("input", "line", 4)
+    line = initialize_array(DatasetType.input, ComponentType.line, 4)
 
     line["id"] = [5, 6, 7, 8]
     line["from_node"] = [1, 2, 3, 1]
@@ -31,7 +31,7 @@ def original_input_data() -> dict[str, np.ndarray]:
     line["tan1"] = 4.0
     line["i_n"] = 5.0
 
-    asym_load = initialize_array("input", "asym_load", 2)
+    asym_load = initialize_array(DatasetType.input, ComponentType.asym_load, 2)
     asym_load["id"] = [9, 10]
     asym_load["node"] = [1, 2]
     asym_load["status"] = [1, 1]
@@ -39,13 +39,19 @@ def original_input_data() -> dict[str, np.ndarray]:
     asym_load["p_specified"] = [[11e6, 12e6, 13e6], [21e6, 22e6, 23e6]]
     asym_load["q_specified"] = [[11e5, 12e5, 13e5], [21e5, 22e5, 23e5]]
 
-    return {"node": node, "line": line, "asym_load": asym_load}
+    return {
+        ComponentType.node: node,
+        ComponentType.line: line,
+        ComponentType.asym_load: asym_load,
+    }
 
 
 @pytest.fixture
 def original_input_data_columnar_all(original_input_data):
     return compatibility_convert_row_columnar_dataset(
-        original_input_data, ComponentAttributeFilterOptions.everything, DatasetType.input
+        original_input_data,
+        ComponentAttributeFilterOptions.everything,
+        DatasetType.input,
     )
 
 
@@ -57,7 +63,11 @@ def original_input_data_columnar_relevant(original_input_data):
 
 
 @pytest.fixture(
-    params=["original_input_data", "original_input_data_columnar_all", "original_input_data_columnar_relevant"]
+    params=[
+        "original_input_data",
+        "original_input_data_columnar_all",
+        "original_input_data_columnar_relevant",
+    ]
 )
 def input_data(request):
     return request.getfixturevalue(request.param)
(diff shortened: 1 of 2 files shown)


tests/unit/validation/test_batch_validation.py
```diff
@@ -119,7 +137,49 @@ def test_validate_batch_data_input_error(input_data, batch_data):
...
+    transformer_input["tap_size"] = 250
+    test_input_data = {
+        ComponentType.node: node_input,
+        ComponentType.transformer: transformer_input,
+    }
+    test_update_data = {ComponentType.sym_load: initialize_array(DatasetType.update, ComponentType.sym_load, (2, 0))}
+    result = validate_batch_data(test_input_data, test_update_data)
+    assert result is not None
+    assert len(result) == test_update_data[ComponentType.sym_load].shape[0]
+    assert len(result[0]) == test_input_data[ComponentType.transformer].shape[0]
```
figueroa1395, line 182: Given the specific bug, maybe it is a good idea to also add:
`assert len(result[1]) == test_input_data[ComponentType.transformer].shape[0]`
As the error was propagating through scenarios.

figueroa1395 (approved):

tests/unit/validation/test_batch_validation.py
```diff
@@ -79,19 +89,27 @@ def original_batch_data() -> dict[str, np.ndarray]:
...
+        ComponentAttributeFilterOptions.relevant,
+        DatasetType.update,
     )
 
 
 @pytest.fixture(
-    params=["original_batch_data", "original_batch_data_columnar_all", "original_batch_data_columnar_relevant"]
+    params=[
+        "original_batch_data",
+        "original_batch_data_columnar_all",
```
Jerry-Jinfeng-Guo, line 110: This should actually be updated to "original_batch_data_columnar_everything"

figueroa1395 (approved):

tests/unit/validation/test_batch_validation.py
mgovers: I'd consider that out of scope of this PR. Please create a new one with that small change (or add it to another PR)

mgovers:
> How does this change prevents from appending? The original code was:
> 
> ```
>         batch_errors = input_errors + id_errors
> 
>         if not id_errors:
>             batch_errors = input_errors
>             # ...
> ```
> 
> The new one is:
> 
> ```
>         batch_errors = input_errors + id_errors
> 
>         if not id_errors:
>             # ...
> ```

In Python, everything is a reference, except if it's not

* `batch_errors = input_errors` makes `batch_errors` a reference to `input_errors`
* `batch_errors = input_errors + id_errors` makes a new list with the contents of `input_errors`, followed by the contents of `id_errors`

that means that if you append to `batch_errors`, then in the former case, you actually append to `input_errors`.

If you then create a reference when adding the current scenario errors to the full list, then in the former case, every scenario will point to `input_errors`. In the latter, you basically move the contents of `batch_errors` (new list) into the full error list (technically, it's creating a reference, updating the reference count of the former, and then when it goes out of scope, the reference counter is set back to `0`, which is equivalent to moving the object)

Jerry-Jinfeng-Guo:
> How does this change prevents from appending? The original code was:
> 
> ```
>         batch_errors = input_errors + id_errors
> 
>         if not id_errors:
>             batch_errors = input_errors
>             # ...
> ```
> 
> The new one is:
> 
> ```
>         batch_errors = input_errors + id_errors
> 
>         if not id_errors:
>             # ...
> ```

OK, I got it now, this line duplicates the input error for all [batch] entry

Jerry-Jinfeng-Guo:
When you do complexity analysis, it's important not to confuse N. In this case, the complexity is `N log(N)`.

mgovers:
> When you do complexity analysis, it's important not to confuse N. In this case, the complexity is `N log(N)`, suppose one of them is at log scale of the other one, n=LogN, i.e., number_of_batches * sum_of_errors.

Actually no because each scenario updates the same object it references, and also a reference to the original object is used. That means that at scenario k, you also still update the errors referenced for scenario 0, and they all grow identically the same (interestingly, as a side effect, there is no additional memory overhead (linear in the amount of scenarios), but each error is referenced N times, resulting in the quadratic overhead.

Even if that were not the case, and each scenario would only contain a sum of the former, you would end up in triangle numbers, which is also quadratic: 1st scenatio references 1 error, 2nd scenario references 2 errors, ..., k references k errors. Total = $$\sum_{k=1}^n k = \frac{n(n-1)}{2}$$

mgovers:
> > When you do complexity analysis, it's important not to confuse N. In this case, the complexity is `N log(N)`, suppose one of them is at log scale of the other one, n=LogN, i.e., number_of_batches * sum_of_errors.
> 
> Actually no because each scenario updates the same object it references, and also a reference to the original object is used. That means that at scenario k, you also still update the errors referenced for scenario 0, and they all grow identically the same (interestingly, as a side effect, there is no additional memory overhead (linear in the amount of scenarios), but each error is referenced N times, resulting in the quadratic overhead.

Note that this also was verified experimentally by @bartschuurmans

Jerry-Jinfeng-Guo:
> > When you do complexity analysis, it's important not to confuse N. In this case, the complexity is `N log(N)`, suppose one of them is at log scale of the other one, n=LogN, i.e., number_of_batches * sum_of_errors.
> 
> Actually no because each scenario updates the same object it references, and also a reference to the original object is used. That means that at scenario k, you also still update the errors referenced for scenario 0, and they all grow identically the same (interestingly, as a side effect, there is no additional memory overhead (linear in the amount of scenarios), but each error is referenced N times, resulting in the quadratic overhead.
> 
> Even if that were not the case, and each scenario would only contain a sum of the former, you would end up in triangle numbers, which is also quadratic: 1st scenatio references 1 error, 2nd scenario references 2 errors, ..., k references k errors. Total = ∑ k = 1 n k = n ( n − 1 ) 2

When looking at such problems the total amount of (non duplicate) errors is dominant, you don't (need to) look at the batch size. I am not saying NlogN is any better than quadratic, on the contrary, it might even lead to close to cubic complexity.

Jerry-Jinfeng-Guo:
> > > When you do complexity analysis, it's important not to confuse N. In this case, the complexity is `N log(N)`, suppose one of them is at log scale of the other one, n=LogN, i.e., number_of_batches * sum_of_errors.
> > 
> > 
> > Actually no because each scenario updates the same object it references, and also a reference to the original object is used. That means that at scenario k, you also still update the errors referenced for scenario 0, and they all grow identically the same (interestingly, as a side effect, there is no additional memory overhead (linear in the amount of scenarios), but each error is referenced N times, resulting in the quadratic overhead.
> 
> Note that this also was verified experimentally by @BartSchuurmans

Like I said, anything between nlogn (lower case) and cubic is to be expected.

mgovers:
> > > > When you do complexity analysis, it's important not to confuse N. In this case, the complexity is `N log(N)`, suppose one of them is at log scale of the other one, n=LogN, i.e., number_of_batches * sum_of_errors.
> > > 
> > > 
> > > Actually no because each scenario updates the same object it references, and also a reference to the original object is used. That means that at scenario k, you also still update the errors referenced for scenario 0, and they all grow identically the same (interestingly, as a side effect, there is no additional memory overhead (linear in the amount of scenarios), but each error is referenced N times, resulting in the quadratic overhead.
> > 
> > 
> > Note that this also was verified experimentally by @BartSchuurmans
> 
> Like I said, anything between nlogn (lower case) and cubic is to be expected.

How did you arrive at `n = log N`? I expect 2 types of data validation errors to happen in production: all scenarios suffer from the same issue (erroneous input data) or some scenarios suffer from some issue (erroneous update data). The former has a constant contribution to each scenario, and the latter is statistically distributed across the scenarios, which is also constant, but with a different prefactor.

That said: in the issue here, the amount of errors per scenario only contributes to the prefactor, not to the scaling in terms of the amount of scenarios, which is the issue described here.

Jerry-Jinfeng-Guo:
> How did you arrive at n = log N?

That's the only assumption we could make. I see no need to further this discussion as it does not impact the fix you did here.

PR #873, opened 2025-01-15
disable cmake policy that removes FindBoost.cmake

mgovers:
Workaround for compatibility with CMake >= 3.30 due to https://cmake.org/cmake/help/latest/policy/CMP0167.html#policy:CMP0167

Depending on the outcome of https://github.com/conda-forge/boost-feedstock/issues/226 and when CMake actually removes the deprecated functionality, we may need to move to `libboost-devel` for Windows instead in the future.


mgovers:
discussion for sustainable fix in https://github.com/conda-forge/boost-feedstock/issues/226

Jerry-Jinfeng-Guo (approved):

TonyXiang8787 (approved):

PR #874, opened 2025-01-17
Use Linux arm64 instance to build Python linux aarch64

TonyXiang8787:
Linux ARM64 instances are now available in GitHub Actions for open source projects.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -188,7 +188,7 @@ jobs:
             os: ubuntu-24.04
             cibw_build: "*_x86_64"
           - platform: linux-aarch64
-            os: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


TonyXiang8787:
Now the GCC release build is the real bottleneck.

mgovers (commented):
great change, but I can't find whether they are free; only that the runner images are in public beta. https://github.com/actions/partner-runner-images/blob/main/README.md

I will approve after I know for sure that it's free for us.

mgovers (approved):
> @mgovers look at here
> 
> https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

great, thank you!

PR #877, opened 2025-01-20
use link time optimization for CI release build

TonyXiang8787:
Use Link Time Optimization for CI release build for all targets. This is achieved by setting `CMAKE_INTERPROCEDURAL_OPTIMIZATION` in the preset.

In the user build, the C-API always enables LTO.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -254,7 +254,10 @@
     },
     {
       "name": "ci-clang-release",
-      "inherits": "clang-release"
+      "inherits": "clang-release",
+      "cacheVariables": {
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "TRUE"
+      }
     },
     {
       "name": "ci-gcc-debug",
@@ -270,7 +273,10 @@
       "name": "ci-gcc-release",
       "inherits": [
         "gcc-release"
-      ]
+      ],
+      "cacheVariables": {
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "TRUE"
+      }
     },
     {
       "name": "ci-sonar",
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -129,11 +129,7 @@
         "CMAKE_C_CLANG_TIDY": "clang-tidy",
         "CMAKE_CXX_CLANG_TIDY": "clang-tidy",
         "CMAKE_C_COMPILER": "clang",
-        "CMAKE_CXX_COMPILER": "clang++",
-        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "TRUE"
-      },
-      "environment": {
-        "ASAN_OPTIONS": "detect_odr_violation=0"
+        "CMAKE_CXX_COMPILER": "clang++"
       },
       "inherits": "unix-base",
       "hidden": true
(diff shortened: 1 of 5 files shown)


CMakePresets.json
```diff
@@ -270,7 +273,13 @@
       "name": "ci-gcc-release",
       "inherits": [
         "gcc-release"
-      ]
+      ],
+      "cacheVariables": {
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "TRUE"
+      },
+      "environment": {
+        "ASAN_OPTIONS": "detect_odr_violation=0"
```
mgovers, line 281: do we want to re-enable this since we found the `static constexpr lambda` thing? or is it a completely separate issue?

CMakePresets.json
```diff
@@ -254,7 +254,10 @@
     },
     {
       "name": "ci-clang-release",
-      "inherits": "clang-release"
+      "inherits": "clang-release",
+      "cacheVariables": {
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "TRUE"
+      }
```
mgovers, line 260: can you please also add this to `ci-clang-tidy-release` and `ci-clang-tidy-debug`? I wonder if that speeds up clang-tidy

CMakePresets.json
TonyXiang8787: This is a separate issue.

CMakePresets.json
TonyXiang8787: Why do we want that? Clang-tidy is not the bottleneck in the CI.

mgovers: we disabled part of the clang-tidy build in CI due to the time it takes to run all checks. If the reason it takes so long is indeed the optimization, and if we can solve it using this flag, then we might enable it in the full build in CI again.

TonyXiang8787: I seriously doubt it, but let me try.

TonyXiang8787: @mgovers  full build in https://github.com/PowerGridModel/power-grid-model/actions/runs/12888364732

TonyXiang8787: It is not faster. I will revert the change.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -91,4 +91,10 @@ using IntSVector = std::vector<IntS>;
 template <class T, class... Ts>
 concept is_in_list_c = (std::same_as<std::remove_const_t<T>, Ts> || ...);
 
+// functor to include all
+struct IncludeAll {
+    constexpr bool operator()(Idx /*ignored*/) const { return true; }
+};
+constexpr IncludeAll include_all{};
+
```
mgovers, line 99: is this the correct location for this? sounds too specific to be in the most root-level file. I'm OK with keeping it in `common/` but maybe in its own file?

If you prefer to keep it here, you can also just resolve my comment. I have no hard feelings

mgovers (dismissed):

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
TonyXiang8787: I think it makes sense to be here. This functor gets used in different places (maybe even more in the future).

I can change it to more generic.

TonyXiang8787: Please have a look again. @mgovers

mgovers (approved):

PR #882, opened 2025-01-27
Feature/clang tidy

mgovers:
Ensures `clang-tidy` runs on Visual Studio 2022 latest version.

Relates to #267 .

**Note:** I had to go from `std::ranges::copy(foo, std::back_inserter(bar));` to `bar.insert(bar.end(), foo.begin(), foo.end());` due to https://github.com/llvm/llvm-project/issues/120151 . I prefer to keep using the former but I first wanted to do a full build

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
@@ -44,7 +44,7 @@ using nlohmann::json;
 // visitor for json conversion
 struct JsonMapArrayData {
     size_t size{};
-    msgpack::sbuffer buffer{};
+    msgpack::sbuffer buffer;
 };
 
 struct JsonSAXVisitor {
@@ -130,8 +130,8 @@ struct JsonSAXVisitor {
         throw SerializationError{ss.str()};
     }
 
-    std::stack<JsonMapArrayData> data_buffers{};
-    msgpack::sbuffer root_buffer{};
+    std::stack<JsonMapArrayData> data_buffers;
+    msgpack::sbuffer root_buffer;
 };
 
 // visitors for parsing
@@ -192,8 +192,15 @@ struct MapArrayVisitor : DefaultErrorVisitor<MapArrayVisitor<map_array>> {
         std::same_as<map_array, visit_map_t> || std::same_as<map_array, visit_map_array_t>;
     static constexpr bool enable_array =
         std::same_as<map_array, visit_array_t> || std::same_as<map_array, visit_map_array_t>;
-    static constexpr std::string_view static_err_msg =
-        enable_map ? (enable_array ? "Map or an array expected." : "Map expected.") : "Array expected.";
+    static constexpr std::string_view static_err_msg = [] {
+        if constexpr (enable_map && enable_array) {
+            return "Map or an array expected.";
+        } else if constexpr (enable_map) {
+            return "Map expected.";
+        } else {
+            return "Array expected.";
+        }
+    }();
 
     Idx size{};
     bool is_map{};
@@ -228,7 +235,7 @@ struct MapArrayVisitor : DefaultErrorVisitor<MapArrayVisitor<map_array>> {
 struct StringVisitor : DefaultErrorVisitor<StringVisitor> {
     static constexpr std::string_view static_err_msg = "String expected.";
 
-    std::string_view str{};
+    std::string_view str;
     bool visit_str(const char* v, uint32_t size) {
         str = {v, size};
         return true;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
@@ -78,8 +78,8 @@ struct JsonConverter : msgpack::null_visitor {
 
     Idx indent;
     Idx max_indent_level;
-    std::stringstream ss{};
-    std::stack<MapArray> map_array{};
+    std::stringstream ss;
+    std::stack<MapArray> map_array;
 
     void print_indent() {
         if (indent < 0) {
@@ -293,7 +293,7 @@ class Serializer {
     std::vector<ComponentBuffer> component_buffers_; // list of components, then all scenario flatten
 
     // msgpack pakcer
-    msgpack::sbuffer msgpack_buffer_{};
+    msgpack::sbuffer msgpack_buffer_;
     msgpack::packer<msgpack::sbuffer> packer_;
     bool use_compact_list_{};
     std::map<MetaComponent const*, std::vector<MetaAttribute const*>> attributes_;
@@ -606,9 +606,8 @@ class Serializer {
         });
     }
     void pack_attribute(AttributeBuffer<void const> const& attribute_buffer, Idx idx) {
-        return ctype_func_selector(attribute_buffer.meta_attribute->ctype, [&]<class T> {
-            packer_.pack(*(reinterpret_cast<T const*>(attribute_buffer.data) + idx));
-        });
+        ctype_func_selector(attribute_buffer.meta_attribute->ctype,
+                            [&]<class T> { packer_.pack(*(reinterpret_cast<T const*>(attribute_buffer.data) + idx)); });
     }
 
     static constexpr BufferView advance(BufferView buffer_view, Idx offset) {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
@@ -93,7 +93,10 @@ concept is_in_list_c = (std::same_as<std::remove_const_t<T>, Ts> || ...);
 
 // functor to include all
 struct IncludeAll {
-    template <class... T> constexpr bool operator()(T&&... /*ignored*/) const { return true; }
+    template <class... T> constexpr bool operator()(T&&... ignored) const {
+        ((void)std::forward<T>(ignored), ...);
+        return true;
+    }
 };
 constexpr IncludeAll include_all{};
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
@@ -281,7 +281,7 @@ inline void add_diag(Eigen::ArrayBase<DerivedA>& x, Eigen::ArrayBase<DerivedB> c
 }
 template <rk2_tensor DerivedA, column_vector DerivedB>
 inline void add_diag(Eigen::ArrayBase<DerivedA>&& x, Eigen::ArrayBase<DerivedB> const& y) {
-    x.matrix().diagonal() += y.matrix();
+    std::move(x).matrix().diagonal() += y.matrix();
 }
 
 // zero tensor
@@ -351,13 +351,14 @@ template <symmetry_tag sym, class Proxy>
 inline void update_real_value(RealValue<sym> const& new_value, Proxy&& current_value, double scalar) {
     if constexpr (is_symmetric_v<sym>) {
         if (!is_nan(new_value)) {
-            current_value = scalar * new_value;
+            std::forward<Proxy>(current_value) = scalar * new_value;
         }
     } else {
         for (size_t i = 0; i != 3; ++i) {
             if (!is_nan(new_value(i))) {
-                current_value(i) = scalar * new_value(i);
+                current_value(i) = scalar * new_value(i); // can't forward due to runtime element access
             }
+            (void)std::forward<Proxy>(current_value);
         }
     }
 }
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
@@ -87,7 +87,7 @@ template <typename Component, typename IndexType, class ComponentContainer, type
                                  decltype(*comp_base_sequence_cbegin<Component>(MainModelState<ComponentContainer>{}))>
 constexpr ResIt produce_output(MainModelState<ComponentContainer> const& state, ResIt res_it, ResFunc&& func) {
     return std::transform(get_component_citer<Component>(state).begin(), get_component_citer<Component>(state).end(),
-                          comp_base_sequence_cbegin<Component>(state), res_it, func);
+                          comp_base_sequence_cbegin<Component>(state), res_it, std::forward<ResFunc>(func));
 }
 
 } // namespace detail
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
@@ -29,6 +29,7 @@ inline void iterate_component_sequence(Func&& func, ForwardIterator begin, Forwa
         // get component directly using sequence id
         func(*it, sequence_idx[seq]);
     }
+    (void)std::forward<Func>(func);
 }
 
 template <typename T> bool check_id_na(T const& obj) {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
@@ -191,6 +191,7 @@ inline void calculate_pf_result(YBus<sym> const& y_bus, PowerFlowInput<sym> cons
         calculate_load_gen_result<sym>(load_gens, bus_number, input, output, load_gen_func);
         calculate_source_result<sym>(sources, bus_number, y_bus, input, output, load_gens);
     }
+    (void)std::forward<LoadGenFunc>(load_gen_func);
 }
 
 template <symmetry_tag sym>
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
@@ -167,6 +167,7 @@ inline void calculate_load_gen_result(IdxRange const& load_gens, Idx bus_number,
         }
         output.load_gen[load_gen].i = conj(output.load_gen[load_gen].s / output.u[bus_number]);
     }
+    (void)std::forward<LoadGenFunc>(load_gen_func);
 }
 
 template <symmetry_tag sym, typename LoadGenFunc>
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
@@ -138,6 +138,7 @@ template <rk2_tensor Matrix> class DenseLUFactor {
                 throw SparseMatrixError{}; // can not specify error code
             }
         }
+        (void)std::move(matrix);
     }
 };
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
@@ -329,7 +329,7 @@ class Topology {
             permuted_node_indices[reordered[idx]] = n_non_cyclic_nodes + idx;
         }
 
-        std::ranges::copy(reordered, std::back_inserter(dfs_node));
+        dfs_node.insert(dfs_node.end(), reordered.begin(), reordered.end());
         for (auto [from, to] : fills) {
             auto from_reordered = permuted_node_indices[from];
             auto to_reordered = permuted_node_indices[to];
@@ -530,6 +530,7 @@ class Topology {
                 coupling[topo_comp_i] = Idx2D{topo_idx, new_math_comp_i};
             }
         }
+        (void)std::forward<GetMathTopoComponent>(get_component_topo);
     }
 
     void couple_all_appliance() {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -53,7 +53,7 @@ template <symmetry_tag sym_type> struct BranchShortCircuitSolverOutput {
 
 // fault math calculation parameters and math output
 struct FaultCalcParam {
-    DoubleComplex y_fault{};
+    DoubleComplex y_fault;
     FaultType fault_type{};
     FaultPhase fault_phase{};
 };
@@ -136,7 +136,7 @@ static_assert(sensor_calc_param_type<PowerSensorCalcParam<asymmetric_t>>);
 struct TransformerTapRegulatorCalcParam {
     double u_set{};
     double u_band{};
-    DoubleComplex z_compensation{};
+    DoubleComplex z_compensation;
     IntS status{};
 };
 
@@ -203,7 +203,7 @@ struct SourceCalcParam {
     DoubleComplex y1;
     DoubleComplex y0;
 
-    template <symmetry_tag sym> inline ComplexTensor<sym> y_ref() const {
+    template <symmetry_tag sym> ComplexTensor<sym> y_ref() const {
         if constexpr (is_symmetric_v<sym>) {
             return y1;
         } else {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
@@ -161,7 +161,7 @@ class DenseGroupedIdxVector {
 
         IdxVector const* dense_vector_{};
         Idx group_{};
-        std::pair<group_iterator, group_iterator> group_range_{};
+        std::pair<group_iterator, group_iterator> group_range_;
 
         friend class boost::iterator_core_access;
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_like_concepts.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_like_concepts.hpp
@@ -16,14 +16,14 @@ concept iterator_like = requires(T const t) {
 
 template <typename T, typename ElementType>
 concept forward_iterator_like = std::regular<T> && iterator_like<T, ElementType> && requires(T t) {
-    { t++ } -> std::same_as<T>;
-    { ++t } -> std::same_as<T&>;
+    { t++ } -> std::same_as<T>;  // NOLINT(bugprone-inc-dec-in-conditions)
+    { ++t } -> std::same_as<T&>; // NOLINT(bugprone-inc-dec-in-conditions)
 };
 
 template <typename T, typename ElementType>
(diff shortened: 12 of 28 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -93,7 +93,10 @@ concept is_in_list_c = (std::same_as<std::remove_const_t<T>, Ts> || ...);
 
 // functor to include all
 struct IncludeAll {
-    template <class... T> constexpr bool operator()(T&&... /*ignored*/) const { return true; }
+    template <class... T> constexpr bool operator()(T&&... ignored) const {
+        ((void)std::forward<T>(ignored), ...);
```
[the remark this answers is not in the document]
TonyXiang8787: Why are you doing this instead of the original? Have a unnamed argument will not trigger an unused argumeng warning.

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
```diff
@@ -530,6 +530,7 @@ class Topology {
                 coupling[topo_comp_i] = Idx2D{topo_idx, new_math_comp_i};
             }
         }
+        (void)std::forward<GetMathTopoComponent>(get_component_topo);
```
TonyXiang8787, line 533: why is this happening?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -138,6 +138,7 @@ template <rk2_tensor Matrix> class DenseLUFactor {
                 throw SparseMatrixError{}; // can not specify error code
             }
         }
+        (void)std::move(matrix);
```
TonyXiang8787, line 141: Why here?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -167,6 +167,7 @@ inline void calculate_load_gen_result(IdxRange const& load_gens, Idx bus_number,
         }
         output.load_gen[load_gen].i = conj(output.load_gen[load_gen].s / output.u[bus_number]);
     }
+    (void)std::forward<LoadGenFunc>(load_gen_func);
```
TonyXiang8787, line 170: why this?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
```diff
@@ -191,6 +192,7 @@ inline void calculate_pf_result(YBus<sym> const& y_bus, PowerFlowInput<sym> cons
         calculate_load_gen_result<sym>(load_gens, bus_number, input, output, load_gen_func);
         calculate_source_result<sym>(sources, bus_number, y_bus, input, output, load_gens);
     }
+    (void)std::forward<LoadGenFunc>(load_gen_func);
```
TonyXiang8787, line 195: why this?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -29,6 +29,7 @@ inline void iterate_component_sequence(Func&& func, ForwardIterator begin, Forwa
         // get component directly using sequence id
         func(*it, sequence_idx[seq]);
     }
+    (void)std::forward<Func>(func);
```
TonyXiang8787, line 32: why this?

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -281,7 +281,7 @@ inline void add_diag(Eigen::ArrayBase<DerivedA>& x, Eigen::ArrayBase<DerivedB> c
 }
 template <rk2_tensor DerivedA, column_vector DerivedB>
 inline void add_diag(Eigen::ArrayBase<DerivedA>&& x, Eigen::ArrayBase<DerivedB> const& y) {
-    x.matrix().diagonal() += y.matrix();
+    std::move(x).matrix().diagonal() += y.matrix();
```
TonyXiang8787, line 284: this is very strange

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -351,13 +351,14 @@ template <symmetry_tag sym, class Proxy>
 inline void update_real_value(RealValue<sym> const& new_value, Proxy&& current_value, double scalar) {
     if constexpr (is_symmetric_v<sym>) {
         if (!is_nan(new_value)) {
-            current_value = scalar * new_value;
+            std::forward<Proxy>(current_value) = scalar * new_value;
```
TonyXiang8787, line 354: also this seems to be very strange.

TonyXiang8787 (approved):
@mgovers please look very critically for artificial `void`, `move`, and `forward`.

I strongly object the idea to use those artificially just to get `clang-tidy` happy. If we have a considered choice to ignore the warning, we should either silence it globally or locally, or use an unnamed argument.

TonyXiang8787 (changes requested):
@mgovers please look very critically for artificial void, move, and forward.

I strongly object the idea to use those artificially just to get clang-tidy happy. If we have a considered choice to ignore the warning, we should either silence it globally or locally, or use an unnamed argument.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
mgovers: this is how a perfectly forwarded proxy works: the `current_value` is either taken by `&` or by `&&` or by value depending on what the type in caller is, and then here it is called as

```cpp
std::forward<Proxy>.operator=(scalar * new_value);
```
which can be rewritten as `std::forward<Proxy>(current_value) = scalar * new_value;`

This deliberately consumes the `current_value` in the case of forwarding on a prvalue.

If we do not want to do perfect forwarding, can you please explain why we went for this solution in the first place here? I do understand that this implementation creates a modifiable view of the 3-phase tensor by value, while taking mutable reference for other objects, but it this really the way that `Eigen` wants us to do it?

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
mgovers: this is not a perfect forwarding because it contains `Eigen::ArrayBase<DerivedA>` (a deduced type) instead of `template A` directly. Therefore, it creates a view by value, which is then passed into this function as an rvalue and therefore, it has to be moved.

Alternatively, we can decide to not capture by `&&` but by value, but I don't know whether that is how `Eigen` works, I would need to dig into it.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: this function captures `func` by perfect forwarding. It is not allowed by the standard to not forward a forwarded argument, or move a moved argument.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
mgovers: this function captures `func` by perfect forwarding. It is not allowed by the standard to not forward a forwarded argument, or move a moved argument.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
mgovers: this function captures `func` by prvalue. It is not allowed by the standard to not move a moved argument, or forward a forwarded argument.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
mgovers: this function captures `func` by perfect forwarding. It is not allowed by the standard to not forward a forwarded argument, or move a moved argument.

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
mgovers: this function captures `func` by perfect forwarding. It is not allowed by the standard to not forward a forwarded argument, or move a moved argument.

mgovers:
> @mgovers please look very critically for artificial void, move, and forward.
> 
> I strongly object the idea to use those artificially just to get clang-tidy happy. If we have a considered choice to ignore the warning, we should either silence it globally or locally, or use an unnamed argument.

on the eigen things, we should really have a look whether this is the correct way to capture views on Eigen matrices. On the function versions, the standard tells us that we should use the `&` version of calling a function on all except the last call. in the last call, the function should be perfectly forwarded into its next call. it looks ugly, I know, but it's the only correct usage.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: What do you mean not allowed by standard? I do not think standard forbid you to capture a `rvalue` but not to forward or move them. It is just a (possibly false) warning from `clang-tidy` right?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
TonyXiang8787: This and the other for file `common_solver_functions.hpp`, the root cause is we define the argument for the lambda as `LoadGenFunc&& load_gen_func`, which is not correct. 

Lambdas in function argument should be passed by value unless there is a strong reason of not to do.

If we change that to `LoadGenFunc load_gen_func`. The stuff will go away.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
TonyXiang8787: See https://github.com/PowerGridModel/power-grid-model/pull/882#discussion_r1938948404

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
TonyXiang8787: See https://github.com/PowerGridModel/power-grid-model/pull/882#discussion_r1938948404.

The argument `get_component_topo` should be passed by value as lambdas.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
TonyXiang8787: Same here for https://github.com/PowerGridModel/power-grid-model/pull/882#discussion_r1938948404.

`func` should be defined as pass by value in the function argument.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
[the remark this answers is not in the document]
TonyXiang8787: After re-thinking about it, the original implementation is not correct. We do not do anything with the argument so it does not make sense to make it universal reference. Just define as unnamed const lvalue reference and throw them away.

```suggestion
    template <class... T> constexpr bool operator()(T const&... /*ignored*/) const { return true; }
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
TonyXiang8787: Now this (and other places) is indeed a nasty one. `EIGEN` uses expression template extensively, which is basically a (temporary) proxy object.

For this function, we have a `rvalue` overload because the caller can do something like:

```c++
add_diag(real(complex_tensor), diag_value);
```

Where `real(complex_tensor)` returns a temporary `rvalue` of `EIGEN` expression template, so the other `lvalue` overload will not work. This is what I call the third use-case of `rvalue` reference in the function argument: `move`, `forward`, and capture temporary mutable proxy object. The `clang-tidy` does not correctly recognize the third use-case of `rvalue` reference and raises a false warning.

We can silence the warning locally. We can also choose to define a named variable in all the caller side, so

```c++
auto real_value_proxy = real(complex_tensor);
add_diag(real_value_proxy , diag_value);
```

Then we just need the `lvalue` overload for the function. This also applies to other parts.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -351,13 +351,14 @@ template <symmetry_tag sym, class Proxy>
...
         }
     } else {
         for (size_t i = 0; i != 3; ++i) {
             if (!is_nan(new_value(i))) {
-                current_value(i) = scalar * new_value(i);
+                current_value(i) = scalar * new_value(i); // can't forward due to runtime element access
             }
+            (void)std::forward<Proxy>(current_value);
```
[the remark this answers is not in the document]
TonyXiang8787: see https://github.com/PowerGridModel/power-grid-model/pull/882#discussion_r1938977795

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
TonyXiang8787: see https://github.com/PowerGridModel/power-grid-model/pull/882#discussion_r1938977795

TonyXiang8787:
> > @mgovers please look very critically for artificial void, move, and forward.
> > I strongly object the idea to use those artificially just to get clang-tidy happy. If we have a considered choice to ignore the warning, we should either silence it globally or locally, or use an unnamed argument.
> 
> on the eigen things, we should really have a look whether this is the correct way to capture views on Eigen matrices. On the function versions, the standard tells us that we should use the `&` version of calling a function on all except the last call. in the last call, the function should be perfectly forwarded into its next call. it looks ugly, I know, but it's the only correct usage.
> 
> if we really need to we should silence the warning locally, but i'm very hesitant to disable it on a global level because it usually really **is** a bug to forget a forward or move

To summarize, there are two major things here:

1. Lambda argument wrongly defined as `rvalue`. They should be passed by value. The problems go away.
2. `EIGEN` expression stuff. We can either silence the warning locally, or define a named variable at caller side.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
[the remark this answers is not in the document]
mgovers: after digging into it, perfect forwarding is correct and there should be a NOLINT here

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp
mgovers: good point. i'll double-check wherever this is used

mgovers: https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/common_solver_functions.hpp#L175-L191 `calculate_load_gen_result` is in the hot-loop. should be captured by reference there

TonyXiang8787: Pass by value is the standard practice for lambdas as a function argument. The lambda object itself is not so big.  Note that the internal state of the lambda is already captured by reference.

Also, in Release build the compiler might inline all the functions in one go including the lambda self.

mgovers: while standard practice, what's the reason you so strongly object preventing passing by perfect forwarding? as the person writing the function, you by design don't know whether the person calling the function at a later state will capture by reference, or do a single expensive copy. perfect forwarding handles at the function-creation-side, not as a user responsibility. This makes it much more fail-safe on both ends.

TonyXiang8787: Same goes with all STL algorithms, they take lambda as value. STL authors just assume that the user of STL algorithm will create cheap-to-copy lambda.

mgovers: done

TonyXiang8787 (approved):

figueroa1395 (commented):
Even though this has been merged, I believe there is some discussion worth having over some of the changes.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -192,8 +192,15 @@ struct MapArrayVisitor : DefaultErrorVisitor<MapArrayVisitor<map_array>> {
...
-        enable_map ? (enable_array ? "Map or an array expected." : "Map expected.") : "Array expected.";
+    static constexpr std::string_view static_err_msg = [] {
+        if constexpr (enable_map && enable_array) {
+            return "Map or an array expected.";
+        } else if constexpr (enable_map) {
+            return "Map expected.";
+        } else {
+            return "Array expected.";
+        }
+    }();
```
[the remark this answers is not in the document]
mgovers: 👍 you can share it indeed

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -606,7 +604,7 @@ class Serializer {
         });
     }
     void pack_attribute(AttributeBuffer<void const> const& attribute_buffer, Idx idx) {
-        return ctype_func_selector(attribute_buffer.meta_attribute->ctype, [&]<class T> {
+        ctype_func_selector(attribute_buffer.meta_attribute->ctype, [this, &attribute_buffer, idx]<class T> {
```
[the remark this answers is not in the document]
mgovers: no reason to 😬 it's not a spurious `return` statement, just an unnecessary one (since `ctype_func_selector` has return type `auto`

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -91,9 +91,20 @@ using IntSVector = std::vector<IntS>;
 template <class T, class... Ts>
 concept is_in_list_c = (std::same_as<std::remove_const_t<T>, Ts> || ...);
 
+namespace capturing {
+// perfect forward into void
+template <class... T>
+constexpr void into_the_void(T&&... /*ignored*/) { // NOLINT(cppcoreguidelines-missing-std-forward)
+    // do nothing; the constexpr allows all compilers to optimize this away
+}
+} // namespace capturing
```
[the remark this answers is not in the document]
mgovers: might as well share it in the knowledge sharing session if you're showing `std::forward` in the first place, under the pretext that "so what happens if you really do not want to forward but do want to capture? well, then you just forward/move it directly `into_the_void`

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
mgovers: well, the think is that even though we don't move any underlying data, we still `move` the proxy. as it is not known (in principle) how big or small this Eigen expression proxy object is, it should still be moved (it's not a trivial object, per se)

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -348,17 +348,22 @@ inline auto all_zero(RealValue<asymmetric_t> const& value) { return (value == Re
...
+            capturing::into_the_void<Proxy>(std::forward<Proxy>(current_value));
         }
     } else {
         for (size_t i = 0; i != 3; ++i) {
             if (!is_nan(new_value(i))) {
-                current_value(i) = scalar * new_value(i);
+                current_value(i) = scalar * new_value(i); // can't forward due to runtime element access
             }
         }
+        capturing::into_the_void<Proxy>(std::forward<Proxy>(current_value));
```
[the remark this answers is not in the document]
mgovers: it forwards as `Proxy`. no underlying data is used. However, a `Proxy` object may still be quite big. Imagine, e.g., an Eigen expression that contains `(A + (B * C) - ((B + D)/E))`: it contains a quite complicated sequence of objects, which all take up data. Therefore, we should not capture by value.

Since we don't know here whether we have a value that can be converted to an lvalue reference, we cannot capture by `const&` alone. Instead we either need to have 2 overloads: 1 for `const&` and one for `&&`.

OR we can can use the short-hand, which is perfect forwarding. However, it is considered bad practice to capture something by move and then not move it (as it would literally throw data down the drain). Idem for forwarding (as it's a shorthand for "either by lvalue reference or by rvalue").

Now finally, since it's a common pattern, I did not want to just add NOLINTs everywhere in the code - people will forget. Instead, I do believe that `into_the_void` is clear enough that from a function reader perspective, there's a `forward` happening, but the reader also knows that it's going no-where. I'm open for input, but let's take that offline

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -87,7 +87,7 @@ template <typename Component, typename IndexType, class ComponentContainer, type
                                  decltype(*comp_base_sequence_cbegin<Component>(MainModelState<ComponentContainer>{}))>
 constexpr ResIt produce_output(MainModelState<ComponentContainer> const& state, ResIt res_it, ResFunc&& func) {
     return std::transform(get_component_citer<Component>(state).begin(), get_component_citer<Component>(state).end(),
-                          comp_base_sequence_cbegin<Component>(state), res_it, func);
+                          comp_base_sequence_cbegin<Component>(state), res_it, std::forward<ResFunc>(func));
```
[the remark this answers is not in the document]
mgovers: > Why do we perfect forward a lambda here, shouldn't we pass it by copy? From eyeballing it looks like such lambdas don't contain heavy data, but are rather an intermediate step for outputing. This would involve changing the signature of the parent function as well.

i mean, yes, but there's also no real reason not to forward, right? might as well use best practices (namely: capture by `const&` if you can, or by rvalue if you need)

> Also, what does ResFunc stand for?

`ResFunc` = "the function that produces the result output for a component"

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -138,6 +138,7 @@ template <rk2_tensor Matrix> class DenseLUFactor {
                 throw SparseMatrixError{}; // can not specify error code
             }
         }
+        capturing::into_the_void(std::move(matrix));
```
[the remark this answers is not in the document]
mgovers: cause you can only move once. While `operator[]` on arrays can actually be moved on a per-index basis, that is not true for `operator()` (because you may not receive the actual element but a proxy). Therefore, you can only `std::move` the entire matrix, and not just the individual elements.

The function can be symbolically represented as

```cpp
void f(auto&& matrix) {
    for (int i = 0; i < matrix.size(); ++i) {
        // g(std::move(matrix(i))); // error: cannot move from this matrix
        // g(std::move(matrix)(i)); // error: syntactically correct, but moving already moved object is ill-formed (no diagnostic required)
        g(matrix(i));
    }
    // missing move
}
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp
```diff
@@ -76,10 +76,10 @@ struct MapArray {
...
     static constexpr char sep_char = ' ';
 
-    Idx indent;
-    Idx max_indent_level;
-    std::stringstream ss{};
-    std::stack<MapArray> map_array{};
+    Idx indent{};
+    Idx max_indent_level{};
+    std::stringstream ss{};           // NOLINT(readability-redundant-member-init)
+    std::stack<MapArray> map_array{}; // NOLINT(readability-redundant-member-init)
```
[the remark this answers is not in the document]
figueroa1395: Ugh... Noted.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
figueroa1395: Then I will try to explain this one the best I can in a knowledge sharing as it can come across very surprising. However, just to be sure, making the proxy an lvalue on the caller side first would have been a valid option as well, right? You just went with the option you thought was best, correcT?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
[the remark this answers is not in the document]
figueroa1395: My takeaway then is: Always perfect forward unless there is a strong reason not to. Which I find very reasonable. However, this is a change, albeit in a good direction, should be taken in consideration with the team, as now we probably have a bit of a mixed "standard" (unless you managed to perfect forward all the lambdas that were needed to, which is probably hard to scan for unless pointed out by clang tidy).

In the knowledge session, I will try to showcase perfect forwarding and come with these examples where it is a best practice and never hurts, but also show strong reasons not to as in the eigen example above. Let's see what can I come up with.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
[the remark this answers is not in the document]
figueroa1395: Maybe my question was miss-understood. I intended to ask, why do we even need to move the matrix at this point? I thought clang tidy was complaining about some "no use" thingy, which seemed incorrect for me. This read to me as, ok we are done factorizing the block, it is of no use anymore, let's throw it away.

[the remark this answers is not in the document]
mgovers: we do exactly that but with the proxy (which is captured by value)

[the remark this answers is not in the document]
figueroa1395: I actually got confused in my own head. I meant to ask, why do we need to "send" `matrix` `into_the_void` after all is done here? The `move` I get just fine. In any case, we can take this offline is prefered.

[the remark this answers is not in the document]
figueroa1395: After offline discussion:

Not moving the matrix results in `NOLINT(cppcoreguidelines-missing-std-forward)` warning by clang tidy indicating that we didn't perfect forward the `matrix` which is "captured" by `&&` as an argument. However, since the `matrix` is an `Eigen` template expression or a `proxy`, forwarding it before doing operations on it may results in unnecesary copies of a possibly large matrix. Hence, the forward is done after all the operations are performed and since the `matrix` is no longer needed, it is sent `into_the_void`. This avoids the manual `NOLINT` and shows intentionallity by moving the matrix.

PR #887, opened 2025-01-27
validation test for automatic tap

petersalemink95:
Validation test for #885 

Test is a copy of auto-tap-changer-meshed-any, with adjusted u_band for transformer regulators in the second rank -> u_band too small to find a solution.


tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-any-max-iter/input.json
mgovers, line 1: do you think it makes sense to try to come up with an even smaller repro case? if not, it's good as is, but it is a little bit long

mgovers, line 1: can you please double-check whether this validation case reproduces similar behavior on the new implementation for the step up transformer?

petersalemink95: The network looks like this, with trafos in 1-3, 5-7, 2-4, 6-8, 9-10:
1 - 3 - 5 - 7
 |                |
0                ----
 |                       |
2 - 4 - 6 - 8 - 9 - 10

I tried the following things to reduce:
- remove 10
- remove line between 7-9, making it radial
- only 0-1-3-5-7

This did not result in the right solution, so I think we should keep it as is.

petersalemink95: Did the following:

```
git checkout feature/step-up-transformer-tap-changer-support
git pull
git checkout -b tmp-branch
git merge test/auto-tap-test
build && test
```

Test passes

mgovers (approved):
resolved all comments with latest input + reviewed once more. LGTM. Glad we found this

PR #868, opened 2025-01-08
Step-up transformer tap changer support: additional tests

figueroa1395:
Add missing tests and refactor inappropriate ones for (#826):
- [x] Add [validation test case](https://github.com/PowerGridModel/power-grid-model/pull/868/commits/52184a9018e84a1c90623cf7aafd8a3e6a6f72eb) for transformer control side on the HV side. This test is marked as failing.
- [x] Add [validation test case](https://github.com/PowerGridModel/power-grid-model/pull/868/commits/e81df62c3ba70714c6abca35fe1ddb3c2e87abc5) for a step up transformer. This test is marked as failing.
- [x] Add [new test grid](https://github.com/PowerGridModel/power-grid-model/pull/868/commits/d947149daa534c4a5b9389e6d9a8817311385e27), as a unit test, for transformer ranking that works with the new logic.

PR #892, opened 2025-02-12
Step-up transformer tap changer support: unit test for new control logic

Jerry-Jinfeng-Guo:
Unit test


tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -324,15 +324,64 @@ TEST_CASE("Test Transformer ranking") {
...
+        main_core::add_component<Node>(state, nodes.begin(), nodes.end(), 50.0);
+
+        std::vector<TransformerInput> transformers{
+            get_transformer(11, 0, 1, BranchSide::from), get_transformer(12, 0, 1, BranchSide::from),
+            get_transformer(13, 2, 3, BranchSide::from), get_transformer(14, 6, 7, BranchSide::from),
+            get_transformer(15, 5, 8, BranchSide::from), get_transformer(16, 9, 10, BranchSide::from)};
+        main_core::add_component<Transformer>(state, transformers.begin(), transformers.end(), 50.0);
+
+        std::vector<ThreeWindingTransformerInput> transformers3w{
+            get_transformer3w(17, 0, 4, 5, Branch3Side::side_2, 0)};
```
mgovers, line 356: is it correct that the tap side is `side_2` here? should we add multiple test cases for it?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -324,15 +324,64 @@ TEST_CASE("Test Transformer ranking") {
...
+            get_regulator(30, 17, ControlSide::side_2)};
+        main_core::add_component<TransformerTapRegulator>(state, regulators.begin(), regulators.end(), 50.0);
+
+        state.components.set_construction_complete();
+
+        // Subcases
+        SUBCASE("Ranking complete the graph") {
+            pgm_tap::RankedTransformerGroups order = pgm_tap::rank_transformers(state);
+            pgm_tap::RankedTransformerGroups const ref_order{
+                {{Idx2D{3, 0}, Idx2D{3, 1}, Idx2D{4, 0}, Idx2D{3, 4}}, {Idx2D{3, 2}, Idx2D{3, 3}, Idx2D{3, 5}}}};
```
mgovers, line 381: i probably made a mistake in my workout but i believe that this is not correct according to the new transformer ranking workout:

undirected graph with tap and control side and without links:

```
      0
     tt \-----
  11//   \  17 \ 
   //12   \     5(tc)
  c||c     4--/   t
  1/2      |      |15
    t      6      c
    |      t      8/9
   c|     /      /
   3/7---/14   /
    |  c     /
     \     /
       10
```
following the dirctions in the mathematical update on the one-note, i get (vertex ranks in (), edge rank contributions in plain):

```
     0---------\1
  1//1 \0       v
   ||    \      (1)
   vv   (0)-----^  \
  (1)     |   1     |1
   |1    (0)        v
   v    /          (2)
  (1)<-/1          /
     \            /1
       \--(1)<---/
```

That would mean that transformer 13 and 16 are not reachable. I probably missed something but maybe you can elaborate

mgovers (commented):
will continue my review on monday

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -878,9 +997,10 @@ TEST_CASE("Test Tap position optimizer") {
...
                 state_b.u_pu = [&state_b, &regulator_b](ControlSide side) {
-                    CHECK(side == regulator_b.control_side());
-
-                    // tap pos closer to tap_max at tap side <=> lower voltage at control side
+                    if (state_b.tap_side == regulator_b.control_side()) {
+                        return static_cast<DoubleComplex>(
+                            test::normalized_lerp(state_b.tap_pos, state_b.tap_min, state_b.tap_max));
+                    }
                     return static_cast<DoubleComplex>(
                         test::normalized_lerp(state_b.tap_pos, state_b.tap_max, state_b.tap_min));
```
mgovers, line 1005: since this code repeats fairly often, maybe you can move it outside this function?

I also see that you use
```cpp
                    bool const control_at_tap = side == regulator_b.control_side();
                    auto const value = test::normalized_lerp(state_b.tap_pos, state_b.tap_min, state_b.tap_max);

                    auto const value = control_at_tap
                                           ? test::normalized_lerp(state_b.tap_pos, state_b.tap_max, state_b.tap_min)
                                           : test::normalized_lerp(state_b.tap_pos, state_b.tap_min, state_b.tap_max);
                    return DoubleComplex{value, value};
```
somewhere else, which kind of amounts to the same thing. That can be homogenized to reduce cognitive complexity.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -917,122 +1041,133 @@ TEST_CASE("Test Tap position optimizer") {
...
                 }
                 SUBCASE("positive reactance") {
                     update_data.line_drop_compensation_x = 0.125 / base_power_3p;
-                    check_b = test::check_exact_per_strategy(3, 5, 2);
+                    check_b = test::check_compensated_exact_per_strategy(3, 3, 4, 2, 4, 1);
                 }
                 SUBCASE("negative reactance") {
                     update_data.line_drop_compensation_x = -0.5 / base_power_3p;
-                    check_b = test::check_exact_per_strategy(3, 5, 3);
+                    check_b = test::check_compensated_exact_per_strategy(3, 3, 3, 3, 4, 1);
```
mgovers, line 1052: this has a little bit too many arguments to my liking: what's the order? what you can do is either:

* make a struct that has named arguments that you can use as input for the `test_compensated_exact_per_strategy`
  * e.g., `test::check_compensated_exact_per_strategy({.tap_pos = 3, .tap_pos_comp = 3, .tap_min = 3, .tap_min_comp = 3, .tap_max = 4, .tap_max_comp = 1})` (actually, i probably already messed up the order here, which kind of proves my point haha)
* or group them by pairs,
  * e.g. `test::check_compensated_exact_per_strategy({.value = 3, .comp_value = 3}, {.value = 3, .comp_value = 3}, {.value = 4, .comp_value = 1})`
* or group them by uncomp and comp
  * e.g. `test::check_compensated_exact_per_strategy({3, 3, 4}, {3, 3, 1})`

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -917,122 +1041,133 @@ TEST_CASE("Test Tap position optimizer") {
...
-                    // u_2a = (u_1a * n_1) / (1.0 + relative_tap_pos_a)
-                    // consider u_1a = n_1 = 1.0
-                    // For a tap_size of 0.1 and tap_nom of 0, tap_pos_relative_a = 0.1 * (tap_pos_a - 0)
-                    auto const relative_tap_a = static_cast<double>(state_a.tap_pos) * 0.1;
-                    return static_cast<DoubleComplex>(1.0 / (1.0 + relative_tap_a));
-                };
-
-                state_b.u_pu = [&state_a, &regulator_a, &state_b, &regulator_b](ControlSide side) {
-                    CHECK(side == regulator_b.control_side());
+                SUBCASE("multiple transformers with control function based on ranking") {
```
mgovers, line 1057: did you deliberately move this subcase into the negative reactance subcase`?
```suggestion
            }
            SUBCASE("multiple transformers with control function based on ranking") {
```

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -917,122 +1041,133 @@ TEST_CASE("Test Tap position optimizer") {
...
-                    auto const relative_tap_b = static_cast<double>(state_b.tap_pos) * 0.1;
-                    return state_a.u_pu(regulator_a.control_side()) / (1.0 + relative_tap_b);
-                };
+                    state_b.u_pu = [&state_a, &regulator_a, &state_b, &regulator_b](ControlSide side) {
+                        CHECK(side == regulator_b.control_side());
+
+                        // u_2b = f(tap_pos_a, tap_pos_b) when rank is 1
+                        // u_2b = (u_1b * n_2) / (1.0 + relative_tap_pos_b)
+                        // consider n_2 = 1. Also u_1a = u_2b
+                        // For a tap_size of 0.1 and tap_nom of 0, tap_pos_relative_b = 0.1 * (tap_pos_b - 0)
```
mgovers, line 1085: comment is outdated: you have now added support for tap side == control side

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -917,122 +1041,133 @@ TEST_CASE("Test Tap position optimizer") {
...
+                        auto const relative_tap_b = static_cast<double>(state_b.tap_pos) * 0.1;
+                        if (state_b.tap_side == regulator_b.control_side()) {
+                            return (1.0 + relative_tap_b) / state_a.u_pu(regulator_a.control_side());
```
mgovers, line 1088: i don't think this is correct unit-wise. I think it should just be
```suggestion
                            return (1.0 + relative_tap_b) * state_a.u_pu(regulator_a.control_side());
```

that may also make the rest of the logic simpler

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -917,122 +1041,133 @@ TEST_CASE("Test Tap position optimizer") {
...
+                    }
+                    SUBCASE("Situation 2") {
+                        regulator_a.update({.id = 3, .u_set = 1.1111, .u_band = 0.01});
+                        regulator_b.update({.id = 4, .u_set = 1.5873, .u_band = 0.01});
+                        check_a = test::check_exact(-1);
+                        check_b = test::check_compensated_exact_per_strategy(-3, 5, 5, -3, -3, 5);
+                    }
+                    SUBCASE("Situation 3") {
+                        regulator_a.update({.id = 3, .u_set = 1.0, .u_band = 0.01});
+                        regulator_b.update({.id = 4, .u_set = 1.0, .u_band = 0.01});
```
mgovers, line 1107: was it really necessary to change these values? i think the new values can be additions to the original ones, but i think the original ones should remain valid even with the new control logic

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -917,122 +1041,133 @@ TEST_CASE("Test Tap position optimizer") {
...
-                    check_a = test::check_exact_per_strategy(0, 2, 0);
-                    check_b = test::check_exact_per_strategy(2, 0, 2);
-                }
-                SUBCASE("Rank a == Rank b") {
-                    state_a.rank = 0;
-                    state_b.rank = 0;
-                    check_a = test::check_exact(1);
-                    check_b = test::check_exact(1);
-                }
+                //// The following test is no longer making sense
```
mgovers, line 1113: can you explain why?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -917,122 +1041,133 @@ TEST_CASE("Test Tap position optimizer") {
...
+                // }
             }
-
-            auto const initial_a{transformer_a.tap_pos()};
-            auto const initial_b{transformer_b.tap_pos()};
+            auto const initial_tap_pos_a{transformer_a.tap_pos()};
+            auto const initial_tap_pos_b{transformer_b.tap_pos()};
 
             for (auto strategy_search_side : test::strategy_search_and_sides) {
+                // for (auto strategy_search_side : test::strategy_search_and_sides_xs) {
```
mgovers, line 1170: what do you mean `xs`?

mgovers (commented):
For some parts, i'm wondering whether it would be easier to just make 2 completely separate test cases: one for control side == tap side, and one for the originally present test case.

i would expect the test logic for control side == tap side to be easier than for the control side != tap side, not more difficult

tests/cpp_unit_tests/test_optimizer.hpp
```diff
@@ -146,7 +146,8 @@ static_assert(std::convertible_to<decltype(stub_const_dataset_update), ConstData
 
 constexpr auto strategies = [] {
     using enum OptimizerStrategy;
-    return std::array{any, global_minimum, global_maximum, local_minimum, local_maximum};
+    // return std::array{any, global_minimum, global_maximum, local_minimum, local_maximum};
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Let's just say that it's philosophical

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -576,26 +625,84 @@ template <main_core::main_model_state_c State> class MockTransformerRanker {
...
+            }
             break;
         case local_minimum:
         case global_minimum:
-            CHECK(value == tap_pos_min);
+            if (control_at_tap_side) {
+                CHECK(value == tap_range_max_comp);
+            } else {
+                CHECK(value == tap_range_min_comp);
+            }
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: This is one of the things to investigate on the workout level: the updated control logic yield such results which may make sense since this is not exactly a symmetric scenario.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -825,26 +934,36 @@ TEST_CASE("Test Tap position optimizer") {
...
+                        if (control_at_tap_side) {
+                            CHECK(value == state_b.tap_min); // Min tap pos for min voltage
+                        } else {
+                            CHECK(value == state_b.tap_max); // Max tap pos for min voltage
+                        }
                         break;
                     default:
                         FAIL("unreachable");
                     }
                 };
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: That could work

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: This is the result of aggregating test cases for combinatorial configurations. The second suggestion does make things clearer.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: When debugging, some brackets have to be relocated ;)

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: Suggestion?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
figueroa1395: I will participate in the comments involving this test case as I believe I created it in #868. 

Please correct me if the scope has changed (@Jerry-Jinfeng-Guo), but these grid and test, were created only with the purpose of testing the new ranking for an arbitratry grid to which the new algorithm actually applied to (see `SUBCASE("Ranking complete the graph")` for `Full grid 1`). So indeed `tap side` was set to be `side_2` intentionally by construction, as that is the grid I came up with.

In addition, I don't think multiple cases for `tap side` should be added here, as others may or may not classify for the new ranking algorithm. This is just one possible artificial grid to test it.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -324,15 +324,64 @@ TEST_CASE("Test Transformer ranking") {
...
+        std::vector<LinkInput> links{{20, 1, 2, 1, 1}, {21, 3, 7, 1, 1}, {22, 8, 9, 1, 1}};
+        main_core::add_component<Link>(state, links.begin(), links.end(), 50.0);
+
+        std::vector<SourceInput> sources{{23, 0, 1, 1.0, 0, nan, nan, nan}};
+        main_core::add_component<Source>(state, sources.begin(), sources.end(), 50.0);
+
+        std::vector<TransformerTapRegulatorInput> regulators{
+            get_regulator(24, 11, ControlSide::to),    get_regulator(25, 12, ControlSide::to),
+            get_regulator(26, 13, ControlSide::to),    get_regulator(27, 14, ControlSide::to),
+            get_regulator(28, 15, ControlSide::to),    get_regulator(29, 16, ControlSide::to),
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: > ... `control_side` is an updateable parameter, ...

How so? PowerGridModel/power-grid-model/blob/455d29b5c161777e8f94f6301f849ca1a894a901/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/update.hpp#L195-L210

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -324,15 +324,64 @@ TEST_CASE("Test Transformer ranking") {
...
+
+        std::vector<ThreeWindingTransformerInput> transformers3w{
+            get_transformer3w(17, 0, 4, 5, Branch3Side::side_2, 0)};
+        main_core::add_component<ThreeWindingTransformer>(state, transformers3w.begin(), transformers3w.end(), 50.0);
+
+        std::vector<LineInput> lines{get_line_input(18, 4, 6), get_line_input(19, 3, 10)};
+        main_core::add_component<Line>(state, lines.begin(), lines.end(), 50.0);
+
+        std::vector<LinkInput> links{{20, 1, 2, 1, 1}, {21, 3, 7, 1, 1}, {22, 8, 9, 1, 1}};
+        main_core::add_component<Link>(state, links.begin(), links.end(), 50.0);
```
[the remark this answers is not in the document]
figueroa1395: For the purpose of this specific test, the use of `link`s or `line`s should be interchangable, as this only concerns the ranking. Hence, I don't see why there would be added difficulty because of them here. I would just leave them such that we can always check that for the ranking it doesn't really matter which one we use.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: The combinations of old `u_set` and `u_band` values will lead to max iteration reached in certain cases.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: In this particular case, the `u_pu` of both transformers are set to mimic what could have been. However, this two way check (with `tap_a_sign` and `tap_b_sign`) is no longer enough since the control side of regulator a is at `side_2`. In short, we would need to have `3 * 3` sets of values to check, instead of the current max `3 * 2`, which is already not nice.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -324,15 +324,64 @@ TEST_CASE("Test Transformer ranking") {
         }
 
         SUBCASE("Ranking complete the graph") {
-            // (TODO: jguo) existing demo grid is not compatible with the updated ranking
-            // The test grid needs to be updated here to match the new ranking logic
-            // pgm_tap::RankedTransformerGroups order = pgm_tap::rank_transformers(state);
-            // pgm_tap::RankedTransformerGroups const ref_order{
-            //     {{Idx2D{3, 0}, Idx2D{3, 1}, Idx2D{4, 0}}, {Idx2D{3, 3}, Idx2D{3, 2}, Idx2D{3, 4}}}};
-            // CHECK(order == ref_order);
+            // The test grid 1 is not compatible with the updated logic for step up transformers
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Closing this thread as these couple commits came from https://github.com/PowerGridModel/power-grid-model/pull/868/commits/d947149daa534c4a5b9389e6d9a8817311385e27

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Closing this thread as these couple commits came from https://github.com/PowerGridModel/power-grid-model/pull/868/commits/d947149daa534c4a5b9389e6d9a8817311385e27

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: Closing this thread as these couple commits came from https://github.com/PowerGridModel/power-grid-model/pull/868/commits/d947149daa534c4a5b9389e6d9a8817311385e27

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: Closing this thread as these couple commits came from https://github.com/PowerGridModel/power-grid-model/pull/868/commits/d947149daa534c4a5b9389e6d9a8817311385e27

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -576,26 +625,84 @@ template <main_core::main_model_state_c State> class MockTransformerRanker {
...
+    };
 };
-auto check_exact_per_strategy(IntS tap_pos_any, IntS tap_pos_min, IntS tap_pos_max) -> TapPositionCheckFunc {
-    return [tap_pos_any, tap_pos_min, tap_pos_max](IntS value, OptimizerStrategy strategy) {
+auto check_compensated_exact(IntS tap_pos, IntS tap_pos_comp) -> TapPositionCheckFunc {
+    return [tap_pos, tap_pos_comp](IntS value, OptimizerStrategy /*strategy*/, bool control_at_tap_side) {
+        if (control_at_tap_side) {
+            CHECK(value == tap_pos_comp);
+        } else {
+            CHECK(value == tap_pos);
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: There is a whole dedicated unit test for line drop compensation: https://github.com/PowerGridModel/power-grid-model/blob/baa3ce2aea94439760ba42b6167cbfbcee034bc1/tests/cpp_unit_tests/test_tap_position_optimizer.cpp#L1016-L1017

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: Closing this thread as these couple commits came from https://github.com/PowerGridModel/power-grid-model/pull/868/commits/d947149daa534c4a5b9389e6d9a8817311385e27

figueroa1395: @mgovers I see that you have placed the tap and the control sides on the 3w transformer on the 5th node, but according to me they should both be on the 4th node, changing the ranking completely. Let's discuss offline.

figueroa1395:
The validation cases from https://github.com/PowerGridModel/power-grid-model/pull/868 that were marked as failing, should now pass, am I right? If so, don't forget to "add" them here.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: maybe you can mention the 2 different functions? e.g. 1 saying `1/(1 + rel_tap_pos)`, the other saying `* (1 + rel_tap_pos)`

Jerry-Jinfeng-Guo:
Only the unit test file needs review, the rest have been covered in https://github.com/PowerGridModel/power-grid-model/pull/894 and https://github.com/PowerGridModel/power-grid-model/pull/893

Jerry-Jinfeng-Guo:
> The validation cases from #868 that were marked as failing, should now pass, am I right? If so, don't forget to "add" them here.

This PR is about the unit test but I'll see if that is manageable given the amount of work left in this PR (sonar).

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo, line 1: This file has been reviewed in https://github.com/PowerGridModel/power-grid-model/pull/893 and https://github.com/PowerGridModel/power-grid-model/pull/894. No need to review

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo, line 1: The test case "Test Transformer ranking" was introduced in another PR, no need to review here.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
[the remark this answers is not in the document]
figueroa1395: Offline discussion result: I will change the tap/control configuration of one of the parallel trafos such that they are both on the same side. The ranking should remain the same.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: `side_2` is 1-based instead of 0-based, so it is correct as is. my bad

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: as discussed: 1-based index vs 0-based index. my bad

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: please fix this before merging

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: to me it sounds like we should test this *more* rather than less if it is getting more complicated. you mention that it's even more control logic now, so we should test that control logic and not just assume it works.

mgovers:
FYI: new control logic implementation is verified to work with the old tests, and the old control logic implementation is verified to work with the new tests. I followed the following steps:

Checkout, build and run with the old unit tests
```sh
git checkout feature/tap-changer-unit-test-for-new-control-logic
git checkout origin/feature/step-up-transformer-tap-changer-support -- .\tests\
cmake --build --preset clang-cl-debug
ctest --preset clang-cl-debug
```
To the failing tests, add the following logic to ignore newly supported step-up transformer logic:
```cpp
if (state_a.tap_side == regulator_a.control_side() || state_b.tap_side == regulator_b.control_side()) {
    continue;
}
```
Rebuild and run
```sh
cmake --build --preset clang-cl-debug
ctest --preset clang-cl-debug
```

vice versa (new test logic with old control logic), the same steps can be followed but with `git checkout origin/feature/step-up-transformer-tap-changer-support -- .\power_grid_model_c\`

Therefore, from a functional perspective, this PR is in a good state. I will now proceed with code review

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -576,26 +625,151 @@ template <main_core::main_model_state_c State> class MockTransformerRanker {
...
         case global_minimum:
-            CHECK(value == tap_pos_min);
+            CHECK(value == comp_result.get_min(control_at_tap_side));
+            break;
+        default:
+            FAIL("Unreachable");
+        }
+    };
+}
+struct GenericResultPerStratety {
```
mgovers, line 699: typo
```suggestion
struct GenericResultPerStrategy {
```

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: new test logic looks good

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: new test logic looks good

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: new test logic looks good

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: new logic looks good

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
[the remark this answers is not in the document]
mgovers: new logic LGTM

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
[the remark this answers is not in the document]
mgovers: :+1:

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: feel free to resolve this. I think it looks good as is. More is always possible, but probably not necessary so I leave that up to you.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -426,6 +434,9 @@ template <transformer_c... TransformerTypes> class TransformerWrapper {
     IntS tap_max() const {
         return apply([](auto const& t) { return t.tap_max(); });
     }
+    IntS tap_side() const {
+        return std::visit([](auto const& t) { return static_cast<IntS>(t.get().tap_side()); }, transformer_);
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: added 
```
return apply([](auto const& t) { return static_cast<IntS>(t.tap_side()); });
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -1150,11 +1173,19 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
     auto pilot_run(std::vector<std::vector<RegulatedTransformer>> const& regulator_order) {
         using namespace std::string_literals;
 
-        constexpr auto max_voltage_pos = [](transformer_c auto const& transformer) -> IntS {
+        constexpr auto max_voltage_pos = [](transformer_c auto const& transformer, bool control_at_tap_side) -> IntS {
+            if (control_at_tap_side) {
+                // min voltage at control side => max voltage at tap side => max tap pos
```
figueroa1395, line 1178: Even though the logic here is fine, the comment seems confusing, as it reads to me as if to get the minimum voltage at the control side (when it is on the same side as the tap), we need the max voltage at the tap side, but the side is the same. Could you modify the comment a bit please.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -1150,11 +1173,19 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
+                return transformer.tap_max();
+            }
             // max voltage at control side => min voltage at tap side => min tap pos
             return transformer.tap_min();
         };
-        constexpr auto min_voltage_pos = [](transformer_c auto const& transformer) -> IntS {
+        constexpr auto min_voltage_pos = [](transformer_c auto const& transformer, bool control_at_tap_side) -> IntS {
+            if (control_at_tap_side) {
+                // max voltage at control side => min voltage at tap side => min tap pos
```
figueroa1395, line 1186: Same as above.

figueroa1395 (commented):
Looks good to me, just a minor comment.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: Addressed.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: Addressed.

figueroa1395 (approved):

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -576,26 +625,151 @@ template <main_core::main_model_state_c State> class MockTransformerRanker {
...
+            case local_minimum:
+            case global_minimum:
+                CHECK(value == (control_at_tap_side ? tap_range_max : tap_range_min));
+                break;
+            default:
+                FAIL("Unreachable");
+            }
+        };
+}
+struct CompensatedResultPerStratety {
```
nitbharambe, line 664: Spell error
```suggestion
struct CompensatedResultPerStrategy {
```

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -746,9 +920,9 @@ TEST_CASE("Test Tap position optimizer") {
 
     SUBCASE("optimization") {
         main_core::emplace_component<test::MockTransformer>(
-            state, 1, MockTransformerState{.id = 1, .math_id = {.group = 0, .pos = 0}});
+            state, 1, MockTransformerState{.id = 1, .tap_side = ControlSide::from, .math_id = {.group = 0, .pos = 0}});
         main_core::emplace_component<test::MockTransformer>(
-            state, 2, MockTransformerState{.id = 2, .math_id = {.group = 0, .pos = 1}});
+            state, 2, MockTransformerState{.id = 2, .tap_side = ControlSide::from, .math_id = {.group = 0, .pos = 1}});
```
nitbharambe, line 925:
```suggestion
            state, 1, MockTransformerState{.id = 1, .tap_side = TapSide::from, .math_id = {.group = 0, .pos = 0}});
        main_core::emplace_component<test::MockTransformer>(
            state, 2, MockTransformerState{.id = 2, .tap_side = TapSide::from, .math_id = {.group = 0, .pos = 1}});
```

Jerry-Jinfeng-Guo: This change would require subsequent casting along the lines and is not in line with other places where the `tap side` property was set.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -1061,31 +1218,185 @@ TEST_CASE("Test Tap position optimizer") {
...
                 if (state_a.rank != MockTransformerState::unregulated) {
-                    check_a(get_output_tap_pos(state_a.id), strategy);
+                    check_a(get_output_tap_pos(state_a.id), strategy, control_at_tap_side_a);
                 }
                 if (state_b.rank != MockTransformerState::unregulated) {
-                    check_b(get_output_tap_pos(state_b.id), strategy);
+                    check_b(get_output_tap_pos(state_b.id), strategy, control_at_tap_side_b);
                 }
 
                 // reset
```
nitbharambe, line 1234: What does this reset?
Or is it to verify if the tap_pos are not modified. If so, why is that specially needed?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -746,9 +920,9 @@ TEST_CASE("Test Tap position optimizer") {
 
     SUBCASE("optimization") {
         main_core::emplace_component<test::MockTransformer>(
-            state, 1, MockTransformerState{.id = 1, .math_id = {.group = 0, .pos = 0}});
+            state, 1, MockTransformerState{.id = 1, .tap_side = ControlSide::from, .math_id = {.group = 0, .pos = 0}});
```
nitbharambe, line 923: Do the unit tests follow all combinations of `tap_at_control_side` for a and b transformers? Since we have conditions added for handling them at multiple places

Jerry-Jinfeng-Guo: The extensive combinatorial for both `a` and `b` are done in a dedicated generic test, here it is not needed that we vary `a` as well.

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
Jerry-Jinfeng-Guo: I believe this is from the original test

Jerry-Jinfeng-Guo:
@nitbharambe is there anything preventing this PR from merging?

nitbharambe (approved):

PR #894, opened 2025-02-12
Step-up transformer tap changer support: binary search control logic

Jerry-Jinfeng-Guo:
Updated control logic for binary search


power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -719,7 +722,7 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
         }
 
         void propose_new_pos(bool strategy_max, bool above_range) {
-            bool const is_down = above_range == tap_reverse_;
+            bool const is_down = (above_range == tap_reverse_) ^ control_at_tap_side_;
```
mgovers, line 725: please use `!=` instead of bitwise XOR for boolean values.

Rationale: the exact implementation of `bool` is not not specified in the C/C++ standard. As a result, bitwise operations on `bool`s are undefined behavior.

Example: if you have a `bool` variable represented by an `int8_t` and its binary representation is `00000001` another `bool` with a binary representation `00000010`, then their boolean values are both `1`. The boolean XOR should therefore be `0`. However, the bitwise XOR is `00000011`, which represents `1`, not `0`.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -777,25 +781,27 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
             }
         }
 
-        IntS search(bool prefer_higher = true) const {
+        IntS search(bool prefer_higher_) const {
             // This logic is used to determin which of the middle points could be of interest
```
mgovers, line 785:
```suggestion
            // This logic is used to determine which of the middle points could be of interest
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -777,25 +781,27 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
             // given strategy used in optimization:
             // Since in BinarySearch we insist on absolute upper and lower bounds, we only need to
             // find the corresponding mid point. std::midpoint returns always the lower mid point
             // if the range is of even length. This is why we need to adjust bounds accordingly.
             // Not because upper bound and lower bound might be reversed, which is not possible.
+            bool const prefer_higher = control_at_tap_side_ ? !prefer_higher_ : prefer_higher_;
```
mgovers, line 791: optional different implementation:

```suggestion
            bool const prefer_higher = control_at_tap_side_ != prefer_higher_;
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -777,25 +781,27 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
             auto const primary_bound = prefer_higher ? upper_bound_ : lower_bound_;
             auto const secondary_bound = prefer_higher ? lower_bound_ : upper_bound_;
```
mgovers, line 793: this logic was originally introduced to swap upper and lower bound. Maybe it's a good idea to make the `control_at_tap_side` the new default, e.g.

```suggestion
            bool const prefer_lower = control_at_tap_side_ == prefer_higher_;
            auto const primary_bound = prefer_lower ? lower_bound_ : upper_bound_ ;
            auto const secondary_bound = prefer_lower ? upper_bound_ : lower_bound_ ;
```

or otherwise explicitly mention

```suggestion
            bool const inverted_bounds = control_at_tap_side_ != prefer_higher_;
            auto const primary_bound = inverted_bounds ? upper_bound_ : lower_bound_;
            auto const secondary_bound = inverted_bounds ? lower_bound_ : upper_bound_;
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -777,25 +781,27 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
-        bool tap_reverse_{false};    // tap range normal or reversed
-        bool inevitable_run_{false}; // inevitable run
+        IntS lower_bound_{};              // tap position lower bound
+        IntS upper_bound_{};              // tap position upper bound
+        IntS current_{0};                 // current tap position
+        bool last_down_{false};           // last direction
+        bool last_check_{false};          // last run checked
+        bool tap_reverse_{false};         // tap range normal or reversed
+        bool inevitable_run_{false};      // inevitable run
+        bool control_at_tap_side_{false}; // regulator control side is at tap side
```
mgovers, line 804: rather than `control_at_tap_side`, maybe you can combine this with the logic surrounding `tap_reverse`?

mgovers:
linux build fails

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: That is actually an orthogonal degree that is indeed effectively combinable but would result in un-readable code.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: This suggestion is making less sense to me and appears to be a matter of opinion.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: I am merging this to the other PR without resolving this thread, I think this discussion is still worth having and if we decide to combine we will do it in the base.base PR

Jerry-Jinfeng-Guo:
Merging this PR to PowerGridModel/power-grid-model/pull/893 since the only remaining topic can be carried over to the base PR

PR #895, opened 2025-02-13
Improve memory footprint of columnar deserialization using relevant filter

TonyXiang8787:
# Background

When using PGM deserialization, the user can select a relevant filter for columnar data format. The de-serializer will return only relevant attributes for that component.

Internally, the Python routine still make buffers of all attributes. Then the attributes with only `nan` are discarded. This is un-wanted memory footprint.

# Improvement

In general, we cannot know which attributes are relevant before parse the whole dataset. However, if the condense attribute list is given in the dataset header, and there are no map-like elements in the data, we know for sure that the relevant attributes can only be those attributes specified in the header. In this way, we can ask the Python routine to only create those buffers. This should save memory footprint of the de-serializer.

# Implementation

* [x] C++ core
* [x] C API
* [x] Python API
* [x] C++ unit test
* [x] C API test
* [x] Python unit test (not needed, no Python behavior is changed)


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -685,7 +707,14 @@ class Deserializer {
             elements_per_scenario < 0 ? std::reduce(counter.cbegin(), counter.cend()) : // aggregation
                 elements_per_scenario * batch_size;                                     // multiply
         handler.add_component_info(component_key_, elements_per_scenario, total_elements);
-        msg_data_offsets_.push_back(component_byte_meta);
+        // check if all scenarios does not have any map
+        bool const has_attribute_indications = std::none_of(component_byte_meta.cbegin(), component_byte_meta.cend(),
+                                                            [](auto const& x) { return x.has_map; });
```
mgovers, line 712: maybe more elaborate for the reader
```suggestion
        bool const only_values_in_data = std::none_of(component_byte_meta.cbegin(), component_byte_meta.cend(),
                                                      [](auto const& x) { return x.has_map; });
        bool const only_attribute_indications = !only_values_in_data;
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -754,6 +783,16 @@ class Deserializer {
...
+        // for columnar buffer
+        // if there is no intersection between the pre-defined attributes and the user provied buffer
+        // and the whole component does not have map
+        // skip the whole component for all scenarios and all elements
+        if constexpr (std::same_as<row_or_column_t, columnar_t>) {
+            if (info.has_attribute_indications && reordered_attribute_buffers.empty()) {
+                component_key_ = "";
+                return;
+            }
+        }
```
mgovers, line 795: i personally prefer not to have any early returns in the middle of the function (at start are OK of course), but i understand why it's there.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -83,6 +83,47 @@ PGM_API PGM_Idx PGM_dataset_info_elements_per_scenario(PGM_Handle* handle, PGM_D
  */
 PGM_API PGM_Idx PGM_dataset_info_total_elements(PGM_Handle* handle, PGM_DatasetInfo const* info, PGM_Idx component_idx);
 
+/**
+ * @brief Return if a component has attribute indications.
```
mgovers, line 87:
```suggestion
 * @brief Return whether a component has attribute indications.
```

otherwise, it may suggest that it never returns if not 😂

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -83,6 +83,47 @@ PGM_API PGM_Idx PGM_dataset_info_elements_per_scenario(PGM_Handle* handle, PGM_D
...
+ * Attribute indications are used to indicate the presence of meaningful attributes for a certain component in the
+ * dataset. It returns 1 for a component if all the elements of all the scenarios for this component are list-like in
+ * the serialized data.
+ *
+ * @param handle
+ * @param info A pointer to the info object.
+ * @param component_idx The index number of the component.
+ * @return 1 if the component has attribute indications, 0 if it does not.
+ */
+PGM_API PGM_Idx PGM_dataset_info_has_attribute_indications(PGM_Handle* handle, PGM_DatasetInfo const* info,
```
mgovers, line 98: here but also elsewhere: `has_attribute_indications` feels ambiguous, because it suggests that there are indications in the header, but it does not necessarily suggest that there are no in-body attributes.

Maybe it's good to rename this to something like `*has_only_attribute_indications` or `*has_packed_data` or `*is_packed` or `*only_preamble_attribute_indications` or something similar?

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -83,6 +83,47 @@ PGM_API PGM_Idx PGM_dataset_info_elements_per_scenario(PGM_Handle* handle, PGM_D
...
+                                                           PGM_Idx component_idx);
+
+/**
+ * @brief Return the number of attribute indications for a component.s
```
mgovers, line 102:
```suggestion
 * @brief Return the number of attribute indications for a component.
```

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -83,6 +83,47 @@ PGM_API PGM_Idx PGM_dataset_info_elements_per_scenario(PGM_Handle* handle, PGM_D
...
+ *
+ * @param handle
+ * @param info A pointer to the info object.
+ * @param component_idx The index number of the component.
+ * @return The number of attribute indications for the component.
+ * It is UB if PGM_dataset_info_has_attribute_indications() returns zero.
```
mgovers, line 108: why is this necessary? it may just return `0`, right?

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -279,14 +279,34 @@ TEST_CASE("Deserializer") {
...
-            CHECK(info.get_component_info("source").total_elements == 3);
-            CHECK(info.get_component_info("sym_load").elements_per_scenario == 4);
-            CHECK(info.get_component_info("sym_load").total_elements == 4);
+
+            auto const& node_info = info.get_component_info("node");
+            CHECK(node_info.elements_per_scenario == 3);
+            CHECK(node_info.total_elements == 3);
+            CHECK(node_info.has_attribute_indications == true);
+            CHECK(node_info.attribute_indications[0]->name == "id"s);
+            CHECK(node_info.attribute_indications[1]->name == "u_rated"s);
```
mgovers, line 288: please use `REQUIRE` to check before doing unchecked element access (`REQUIRE` is fatal while `CHECK` is not)
```suggestion
            CHECK(node_info.has_attribute_indications == true);
            REQUIRE(node_info.attribute_indications.size() == 2);
            CHECK(node_info.attribute_indications[0]->name == "id"s);
            CHECK(node_info.attribute_indications[1]->name == "u_rated"s);
```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -279,14 +279,34 @@ TEST_CASE("Deserializer") {
...
+
+            auto const& line_info = info.get_component_info("line");
+            CHECK(line_info.elements_per_scenario == 2);
+            CHECK(line_info.total_elements == 2);
+            CHECK(line_info.has_attribute_indications == false);
```
mgovers, line 293:
```suggestion
            CHECK(line_info.has_attribute_indications == false);
            CHECK(line_info.attribute_indications.empty());
```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -279,14 +279,34 @@ TEST_CASE("Deserializer") {
...
+
+            auto const& source_info = info.get_component_info("source");
+            CHECK(source_info.elements_per_scenario == 3);
+            CHECK(source_info.total_elements == 3);
+            CHECK(source_info.has_attribute_indications == false);
```
mgovers, line 298:
```suggestion
            CHECK(source_info.has_attribute_indications == false);
            CHECK(source_info.attribute_indications.empty());
```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -279,14 +279,34 @@ TEST_CASE("Deserializer") {
...
+            auto const& sym_load_info = info.get_component_info("sym_load");
+            CHECK(sym_load_info.elements_per_scenario == 4);
+            CHECK(sym_load_info.total_elements == 4);
+            CHECK(sym_load_info.has_attribute_indications == true);
+            CHECK(sym_load_info.attribute_indications[0]->name == "id"s);
+            CHECK(sym_load_info.attribute_indications[1]->name == "node"s);
+            CHECK(sym_load_info.attribute_indications[2]->name == "status"s);
+            CHECK(sym_load_info.attribute_indications[3]->name == "type"s);
+            CHECK(sym_load_info.attribute_indications[4]->name == "p_specified"s);
+            CHECK(sym_load_info.attribute_indications[5]->name == "q_specified"s);
```
mgovers, line 309:
```suggestion
            CHECK(sym_load_info.has_attribute_indications == true);
            REQUIRE(sym_load_info.attribute_indications.size() == 6);
            CHECK(sym_load_info.attribute_indications[0]->name == "id"s);
            CHECK(sym_load_info.attribute_indications[1]->name == "node"s);
            CHECK(sym_load_info.attribute_indications[2]->name == "status"s);
            CHECK(sym_load_info.attribute_indications[3]->name == "type"s);
            CHECK(sym_load_info.attribute_indications[4]->name == "p_specified"s);
            CHECK(sym_load_info.attribute_indications[5]->name == "q_specified"s);
```

tests/cpp_unit_tests/test_deserializer.cpp
```diff
@@ -431,8 +451,10 @@ TEST_CASE("Deserializer") {
             auto const& info = deserializer.get_dataset_info();
             CHECK(info.get_component_info("sym_load").elements_per_scenario == -1);
             CHECK(info.get_component_info("sym_load").total_elements == 4);
+            CHECK(info.get_component_info("sym_load").has_attribute_indications == false);
             CHECK(info.get_component_info("asym_load").elements_per_scenario == 1);
             CHECK(info.get_component_info("asym_load").total_elements == 4);
+            CHECK(info.get_component_info("asym_load").has_attribute_indications == false);
```
mgovers, line 457:
```suggestion
            CHECK(info.get_component_info("sym_load").has_attribute_indications == false);
            CHECK(info.get_component_info("sym_load").attribute_indications.empty());
            CHECK(info.get_component_info("asym_load").elements_per_scenario == 1);
            CHECK(info.get_component_info("asym_load").total_elements == 4);
            CHECK(info.get_component_info("asym_load").has_attribute_indications == false);
            CHECK(info.get_component_info("asym_load").attribute_indications.empty());
```

tests/native_api_tests/test_api_serialization.cpp
```diff
@@ -176,6 +176,12 @@ TEST_CASE("API Serialization and Deserialization") {
             CHECK(info.n_components() == n_components);
             CHECK(info.component_name(0) == "node"s);
             CHECK(info.component_name(1) == "source"s);
+            CHECK(info.has_attribute_indications(0) == true);
+            CHECK(info.has_attribute_indications(1) == false);
+            auto const node_attributes = info.attribute_indications(0);
+            CHECK(node_attributes.size() == 2);
```
mgovers, line 182:
```suggestion
            REQUIRE(node_attributes.size() == 2);
```

tests/native_api_tests/test_api_serialization.cpp
```diff
@@ -176,6 +176,12 @@ TEST_CASE("API Serialization and Deserialization") {
             CHECK(info.n_components() == n_components);
             CHECK(info.component_name(0) == "node"s);
             CHECK(info.component_name(1) == "source"s);
+            CHECK(info.has_attribute_indications(0) == true);
+            CHECK(info.has_attribute_indications(1) == false);
```
mgovers, line 180:
```suggestion
            CHECK(info.has_attribute_indications(0));
            CHECK_FALSE(info.has_attribute_indications(1));
```

mgovers (commented):
real nice PR!

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: I do not get your intention here.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787: The name makes sense from user perspective. `has_attribute_indications` means the user can rely on the `attribute_indications` to create buffers and not missing any meaningful data.

If `has_attribute_indications == false`. The user does not have such indications so that the user has to create buffers for all attributes or use row-based buffer.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: The logic also seems to be inverted.

The name is as we said indicated in the https://github.com/PowerGridModel/power-grid-model/pull/895#discussion_r1991520332

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
mgovers: My issue is the following edge case:

```json
{
    "version": "1.0",
    "type": "input",
    "is_batch": false,
    "attributes": {
        "node": [
            "id",
            "u_rated"
        ]
    },
    "data": {
        "node": [
            [
                5,
                10500
            ],
            {
                "id": 6,
                "u_rated": 10500
            }
        ]
    }
}
```

As a user, i now expect `has_attribute_indications` to be `true` (because `node` is present in the header), but the logic says it's `false` (because there are maps in the node input).

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: My intention was that I, as a reader of the code, do not see the direct relationship between `none_of(x, has_map)` and `has_attribute_indications`. You're right in that I made a typo, but I still had to read this code multiple times before I understood what the intention is.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
mgovers: Why is the UB necessary?

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787: The point is, we should not focus on a particular serialized format. The idea of `has_attribute_indications` is from the information perspective, regardless of the actual serialization format and data.

It answers the following questions for the user: if I only create attribute buffers as in `attribute_indications`, am I guaranteed to retain all the relevant information from the dataset?

That's the reason I give this name. Currently, this condition just happens to be satisfied by "no-map in the data section in the msgpack/json format". But it does not stay that way. We can, for example, in the future to scan the actual map-like elements and still give the `has_attribute_indications` with `attribute_indications` of all the meaningful attributes.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787: It has been changed.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: The first line you only change the name of the variable. What is the intension for your second line?

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -83,6 +83,47 @@ PGM_API PGM_Idx PGM_dataset_info_elements_per_scenario(PGM_Handle* handle, PGM_D
  */
 PGM_API PGM_Idx PGM_dataset_info_total_elements(PGM_Handle* handle, PGM_DatasetInfo const* info, PGM_Idx component_idx);
 
+/**
+ * @brief Return whether a component has attribute indications.
+ *
+ * Attribute indications are used to indicate the presence of meaningful attributes for a certain component in the
+ * dataset. It returns 1 for a component if all the elements of all the scenarios for this component are list-like in
+ * the serialized data.
```
mgovers, line 91:
```suggestion
 * dataset. If it returns 1 for that component, you can create only the columnar buffers with the indicated attributes. In this case, you are guaranteed that all information will be retained. Otherwise, no such guarantees are made.
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -40,6 +40,11 @@ struct ComponentInfo {
     // for non-uniform component, this is -1, we use indptr to describe the elements per scenario
     Idx elements_per_scenario{};
     Idx total_elements{};
+    // does the dataset contain only a subset of attributes with meaningful value
```
mgovers, line 43:
```suggestion
    // whether there is a subset of meaningful attributes that was deduced
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -602,6 +619,10 @@ class Deserializer {
                 attributes_per_component.push_back(&component->get_attribute(single_component.second[element_number_]));
             }
             attributes_[component] = std::move(attributes_per_component);
+            // set attribute intidation if enabled
```
mgovers, line 622:
```suggestion
            // set attribute indication if enabled
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
mgovers: resolved cfr. discussion

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
mgovers: resolved cfr. discussion

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -754,6 +783,16 @@ class Deserializer {
...
+        // for columnar buffer
+        // if there is no intersection between the pre-defined attributes and the user provided buffer
+        // and the whole component does not have map
+        // skip the whole component for all scenarios and all elements
+        if constexpr (std::same_as<row_or_column_t, columnar_t>) {
+            if (info.has_attribute_indications && reordered_attribute_buffers.empty()) {
+                component_key_ = "";
+                return;
+            }
+        }
```
figueroa1395, line 795: What if we only calculate `reordered_attribute_buffers` if `info.has_attribute_indications` is `true`, and if not just initialize it as an "empty". Could this potentially bring a performance boost for deserialization of large datasets where some attributes are present but are to be ignored by the user?

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/dataset.hpp
```diff
@@ -63,6 +63,20 @@ class DatasetInfo {
...
+
+    std::vector<std::string> attribute_indications(Idx component_idx) const {
+        Idx const n_attributes = handle_.call_with(PGM_dataset_info_n_attribute_indications, info_, component_idx);
+        std::vector<std::string> attributes;
+        attributes.reserve(n_attributes);
+        for (Idx idx = 0; idx < n_attributes; ++idx) {
+            attributes.emplace_back(handle_.call_with(PGM_dataset_info_attribute_name, info_, component_idx, idx));
+        }
+        return attributes;
+    }
```
figueroa1395, line 78: For consistency and testability, let's split this function into two: `n_attribute_indications` and `attribute_indications`, where a call to the first is to be used in the implementation of the second.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
TonyXiang8787: Can you elaborate more what performance boost are you proposing?

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -83,6 +83,47 @@ PGM_API PGM_Idx PGM_dataset_info_elements_per_scenario(PGM_Handle* handle, PGM_D
  */
 PGM_API PGM_Idx PGM_dataset_info_total_elements(PGM_Handle* handle, PGM_DatasetInfo const* info, PGM_Idx component_idx);
 
+/**
+ * @brief Return whether a component has attribute indications.
+ *
+ * Attribute indications are used to indicate the presence of meaningful attributes for a certain component in the
+ * dataset. If it returns 1 for that component, you can create columnar data with only the indicated attributes. In this
+ * case, you are guaranteed that all information will be retained. Otherwise, no such guarantees are made.
```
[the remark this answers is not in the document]
TonyXiang8787: It is intended to not to go specifics. We want to explain from information perspective. Can you explain why this is not clear to you?

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/dataset.hpp
TonyXiang8787: I personally prefer a single `vector` as output without a separate `n_attribute_indications`. As the vector has size inside.

But the CPP API is just experimental now. I proposal we keep like that until we decide to make CPP API official. Then we can walk through some of the API design choices.

mgovers: i think postponing it makes sense. overloading would be non-breaking (because it's C++, not C), e.g., and even if we don't want to overload, naming should not be an issue
```cpp
std::vector<std::string> attribute_indications(Idx component_idx) const;
Idx n_attribute_indications(Idx component_idx) const;
std::string attribute_indication(Idx component_idx, Idx attribute_idx) const;
```

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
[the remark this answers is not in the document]
TonyXiang8787: Can you check if the following text makes it more clear to you?

```
Attribute indications are used to indicate the presence of meaningful attributes for a certain component in the
dataset. 

If it returns 1 for that component, you can create columnar data with only the indicated attributes. In this
case, you are guaranteed that all information from the original serialized data will be retained for this component,
i.e., you do not lose any meaningful information during the deserialization process.

Otherwise, no such guarantees are made.
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
figueroa1395: Nevermind my comment, I didn't notice that `reordered_attribute_buffers` is needed in spite of the result of this PR. I am resolving this comment.

TonyXiang8787:
@figueroa1395 all comments are resolved. If you agree with the PR, please approve it.

mgovers (approved):

PR #896, opened 2025-02-13
Measurement statistics module

nitbharambe:
Implements variance related calculations mentioned in # https://github.com/PowerGridModel/power-grid-model/issues/547

A generic module for handling different types of variances in state estimation.


power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,167 @@
...
+    using sym = sym_type;
+
+    static constexpr bool symmetric{is_symmetric_v<sym>};
+
+    ComplexValue<sym> value{};
+    RealValue<sym> variance{}; // variance (sigma^2) of the error range, in p.u.
+
+    explicit operator UniformComplexRDV<sym>() const {
+        return UniformComplexRDV<sym>{.value = value, .variance = sum_val(variance)};
+    }
```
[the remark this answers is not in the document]
mgovers: so do you want me to review as is? i would be reviewing my own code though 😬

mgovers (commented):
from a very quick look it seems OK to me, but maybe it's good to let someone else review, as i've written a significant part of the code

nitbharambe:
> from a very quick look it seems OK to me, but maybe it's good to let someone else review, as i've written a significant part of the code

True, asked @Jerry-Jinfeng-Guo .

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+#pragma once
+
+#include "common.hpp"
+#include "three_phase_tensor.hpp"
+
+/**
+ * @file statistics.hpp
+ * @brief This file contains various structures and functions for handling statistical representations of
+ * randomly distributed variables(RDV) used in the Power Grid Model, like in the State Estimation algorithms to
```
Jerry-Jinfeng-Guo, line 13: `randomly distributed variable` is quite a strange expression. For this specific context, 'random variable' is accurate and comcise enough. 
```suggestion
 * random variables (RandVar) used in the Power Grid Model, like in the State Estimation algorithms to
```
or
```suggestion
 * random variables (RV) used in the Power Grid Model, like in the State Estimation algorithms to
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
...
+ * handle measurements.
+ *
+ * The structures provided in this file are used to represent measured values of sensors
+ * with different types of variances. These structures support both symmetric and asymmetric representations and
+ * provide conversion operators to transform between these representations.
+ *
+ * A Randomly distributed variable in PGM can have following characteristics:
```
Jerry-Jinfeng-Guo, line 20:
```suggestion
 * A random variable in PGM can have following characteristics:
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
...
+ *  - Uniform: Single Variance for all phases
+ *  - Independent: Unique Variance for each phase
+ *  - Real: The Real value without direction, eg. real axis: RealValue (* 1), imaginary axis: RealValue (* 1i).
+ *  - Complex: A combined complex value in `a + bi` notation.
```
Jerry-Jinfeng-Guo, line 24:
```suggestion
 *  - Unified: total variance for all phases
 *  - Independent: all phases are independent from each other
 *  - Scalar: a scalar value `RealValue`, eg. real axis: RealValue (* 1), imaginary axis: RealValue (* i).
 *  - Complex: a complex value with real and imaginary parts.
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * Based on these, we use combine variables in Polar/Decomposed forms:
+ * - Decomposed: Treat RDV individually as in cartesian co-ordinates with a separate variance for both real and
+ * complex component.
+ * - Polar: RDV is in polar co-ordinates, with magnitude and angle.
```
Jerry-Jinfeng-Guo, line 29: This part is a bit hard to flow from the above, perhaps some inbetween liaison?
```suggestion
 * [liaison]
 * Based on these, we use combined variables in Decomposed/Polar forms:
 * - Decomposed: treat random variables individually as in cartesian co-ordinates with separated variances for both real and
 * imaginary part.
 * - Polar: random variables are in polar co-ordinates, with magnitudes and angles.
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
...
+    explicit operator UniformRealRDV<asymmetric_t>() const
+        requires(is_symmetric_v<sym>)
+    {
+        return {.value = RealValue<asymmetric_t>{std::piecewise_construct, value}, .variance = variance};
+    }
+    explicit operator UniformRealRDV<symmetric_t>() const
+        requires(is_asymmetric_v<sym>)
+    {
+        return {.value = mean_val(value), .variance = variance / 3.0};
+    }
```
Jerry-Jinfeng-Guo, line 51: Is the division by `3` enough? If I convert one sym rand var to asym rand var and back, would I still get the same result?

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
...
+        auto const real_component = magnitude.value * cos_theta;
+        auto const imag_component = magnitude.value * sin_theta;
+        return DecomposedComplexRDV<sym>{
+            .real_component = {.value = real_component,
+                               .variance = magnitude.variance * cos_theta * cos_theta +
+                                           imag_component * imag_component * angle.variance},
+            .imag_component = {.value = imag_component,
+                               .variance = magnitude.variance * sin_theta * sin_theta +
+                                           real_component * real_component * angle.variance}};
+    }
```
Jerry-Jinfeng-Guo, line 179: Should we add the exact equations in, especially here, to make it easier to follow?

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
...
+    explicit operator IndependentRealRDV<symmetric_t>() const
+        requires(is_asymmetric_v<sym>)
+    {
+        return {.value = mean_val(value), .variance = mean_val(variance) / 3.0};
+    }
+};
+
+// Complex measured value of a sensor in p.u. with a uniform variance across all phases and axes of the complex plane
+// (rotationally symmetric)
+template <symmetry_tag sym_type> struct UniformComplexRDV {
```
Jerry-Jinfeng-Guo, line 82:
```suggestion
template <symmetry_tag sym_type> struct UnifiedComplexRandVar {
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,181 @@
+// SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+//
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+#include "common.hpp"
+#include "three_phase_tensor.hpp"
```
Jerry-Jinfeng-Guo, line 8: this one again ...

Jerry-Jinfeng-Guo (commented):
Couple comments on the file `statistics.hpp` in general before proceeding to the test file:
- Please rename `RDV` to `RandVar`
- Please rename `Uniform` to `Unified`
- Is the `symmetric` member variable actually useful?
- Shall we add some simple equations as reference to the conversions between the forms?

tests/cpp_unit_tests/test_statistics.cpp
```diff
@@ -0,0 +1,820 @@
...
+#include <power_grid_model/common/statistics.hpp>
+
+#include <doctest/doctest.h>
+
+namespace power_grid_model {
+namespace {
+using AsymRealValue = RealValue<asymmetric_t>;
+
+using std::numbers::sqrt2;
+constexpr auto inv_sqrt2 = sqrt2 / 2;
```
Jerry-Jinfeng-Guo, line 14:
```suggestion
constexpr auto inv_sqrt2 = std::numbers::sqrt2 * 0.5;
constexpr auto sqrt3_2 = std::numbers::sqrt3 * 0.5;
```

tests/cpp_unit_tests/test_statistics.cpp
```diff
@@ -0,0 +1,820 @@
...
+                    CHECK(decomposed.real_component.value == doctest::Approx(polar.magnitude.value));
+                    CHECK(decomposed.imag_component.value == doctest::Approx(0.0));
+                    CHECK(decomposed.real_component.variance == doctest::Approx(polar.magnitude.variance));
+                    CHECK(decomposed.imag_component.variance ==
+                          doctest::Approx(magnitude * magnitude * polar.angle.variance));
+                    CHECK(real(decomposed.value()) == doctest::Approx(real(polar.value())));
+                    CHECK(imag(decomposed.value()) == doctest::Approx(imag(polar.value())));
+                }
+
+                SUBCASE("Perpendicular phase shift") {
```
Jerry-Jinfeng-Guo, line 443: `Perpendicular` would be matching a rotation.
```suggestion
                SUBCASE("90deg phase shift") {
```

Jerry-Jinfeng-Guo (commented):
No major comments, except:
- Please rename `RDV` to `RandVar`
- Please rename `U(u)niform` to `U(u)nified`
- Is it possible to add additional test for the conversion back and forth? I.e., `original = Convert_to_a(Convert_to_b(original))`
- I understant that most (first level) subcases are now adopting the function/struct name directly, is it desirable? I.e., `PolarComplexRDV<asymmetric_t>` or `Polar complex random variable - asymmetric`

nitbharambe:
> No major comments, except:
> 
> * Please rename `RDV` to `RandVar`
> * Please rename `U(u)niform` to `U(u)nified`
> * Is it possible to add additional test for the conversion back and forth? I.e., `original = Convert_to_a(Convert_to_b(original))`
> * I understant that most (first level) subcases are now adopting the function/struct name directly, is it desirable? I.e., `PolarComplexRDV<asymmetric_t>` or `Polar complex random variable - asymmetric`

- Done.
- Uniform kind of stands for "uniform variance over all three phases".
- A round trip would not be possible. We do not implement the complete cycle as there is no requirement for it. 
- I think so. The subcase names depict the respective functions one to one.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
nitbharambe: We dont have round trip conversions as that was not required. Division should be enough I suppose.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+};
+
+// Complex measured value of a sensor modeled as separate real and imaginary components with independent
+// variances (rotationally symmetric)
+template <symmetry_tag sym_type> struct DecomposedComplexRandVar {
+    using sym = sym_type;
+
+    IndependentRealRandVar<sym> real_component;
+    IndependentRealRandVar<sym> imag_component;
```
figueroa1395, line 114: Maybe there is no use case and that is why is ommited, but aren't we missing conversion operators from/to symmetric to/from assymetric for `DecomposedComplexRandVar` struct itself? If the members are explicitly converted, I don't think in means that the struct itself implicitly will.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+};
+
+// Complex measured value of a sensor in polar coordinates (magnitude and angle)
+// (rotationally symmetric)
+template <symmetry_tag sym_type> struct PolarComplexRandVar {
+    using sym = sym_type;
+
+    UniformRealRandVar<sym> magnitude;
+    UniformRealRandVar<sym> angle;
```
figueroa1395, line 133: Same as above?

figueroa1395 (commented):
Pretty cool module. Nice job. Also, very thorough testing!

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
Jerry-Jinfeng-Guo: Hmm, I am actually very curious, since the two steps themselves are possible. If we have A->B, and B->A, in theory it should be equivalent?

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+        requires(is_symmetric_v<sym>)
+    {
+        return {.value = RealValue<asymmetric_t>{std::piecewise_construct, value}, .variance = variance};
+    }
+    explicit operator UniformRealRandVar<symmetric_t>() const
+        requires(is_asymmetric_v<sym>)
+    {
+        return {.value = mean_val(value), .variance = variance / 3.0};
+    }
+};
```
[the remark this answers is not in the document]
mgovers: that was actually my original implementation. However, there's a good reason as to why that is not the case. You sample from 2 different things:

* When going from asym -> sym, The original measurements are sampled from a prior distribution with fixed mean and variance. Then, you basically make an estimation of the mean of the measurement (modulo the shift). The variance of the estimator of the mean is inversely proportional to the amount of measurements.
* When going from sym -> asym, you treat the sym measurement + variance as a "known" distribution and "sample" from that distribution, so the new asym measurement has the same variance as the original sym one.

This asymmetry can be summarized as going from prior distro -> sample ===>>> estimator of new distribution -> new sample. The ===>>> is the exact moment the equivalence does not hold anymore as a direct consequence of the Mean Value Theorem

Tricky, I know, but it makes sense

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
mgovers: See https://github.com/PowerGridModel/power-grid-model/pull/896#discussion_r1973491898

mgovers: It's probably good to document that knowledge

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+template <symmetry_tag sym_type> struct IndependentRealRandVar {
+    using sym = sym_type;
+
+    RealValue<sym> value{};
+    RealValue<sym> variance{}; // variance (sigma^2) of the error range
+
+    explicit operator UniformRealRandVar<symmetric_t>() const {
+        constexpr auto scale = is_asymmetric_v<sym> ? 3.0 : 1.0;
+        return {.value = mean_val(value), .variance = mean_val(variance) / scale};
+    }
```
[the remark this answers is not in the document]
mgovers: The mean value of the variance is an estimator of the variance of the original distribution. However, the result returned here is actually a new distribution with the estimator of the mean and its variance as variables (it's more accurate because it is obtained by sampling 3 separate measurements, by Mean Value Theorem). Therefore, the variance of the estimator of the mean is the estimator of the variance of the original distribution, divided by the amount of measurements.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+    explicit operator IndependentRealRandVar<asymmetric_t>() const
+        requires(is_symmetric_v<sym>)
+    {
+        return {.value = RealValue<asymmetric_t>{std::piecewise_construct, value},
+                .variance = RealValue<asymmetric_t>{std::piecewise_construct, variance}};
+    }
+    explicit operator IndependentRealRandVar<symmetric_t>() const
+        requires(is_asymmetric_v<sym>)
+    {
+        return {.value = mean_val(value), .variance = mean_val(variance) / 3.0};
```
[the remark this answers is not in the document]
mgovers: See https://github.com/PowerGridModel/power-grid-model/pull/896#discussion_r1973504108

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+};
+
+// Complex measured value of a sensor with a uniform variance across all phases and axes of the complex plane
+// (rotationally symmetric)
+template <symmetry_tag sym_type> struct UniformComplexRandVar {
+    using sym = sym_type;
+
+    ComplexValue<sym> value{};
+    double variance{}; // variance (sigma^2) of the error range
+};
```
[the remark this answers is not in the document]
mgovers: we didn't add these yet because there was no reason to - we don't use this conversion yet, except for going via `pos_seq()` and `three_phase()`. The reason why we opted for separate functions is the following.

In the most generic situation, you don't know whether an asymmetric complex variable is phase shifted or not. E.g.: while a measurement of the current **is** phase shifted, a measurement of the complex (active and reactive) power is not. So the way to go from the complex measurement to the real one depends on what type of measurement it is:

* the magnitude of the current shall be obtained from the `pos_seq()`
* the apparent power shall be obtained by a direct summation (NOT a `mean_val`!!!)

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+    using sym = sym_type;
+
+    IndependentRealRandVar<sym> real_component;
+    IndependentRealRandVar<sym> imag_component;
+
+    ComplexValue<sym> value() const { return {real_component.value, imag_component.value}; }
+
+    explicit operator UniformComplexRandVar<sym>() const {
+        return static_cast<UniformComplexRandVar<sym>>(static_cast<IndependentComplexRandVar<sym>>(*this));
+    }
```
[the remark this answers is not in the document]
mgovers: IKR, and IMO the best thing is: if desired, it can be improved in the future - as long as the tests hold, it's OK

mgovers:
it's a good sign that there are as many questions about the actual contents of this change as there are. It proves that making a separate statistics module was really really needed

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
nitbharambe: > See [#896 (comment)](https://github.com/PowerGridModel/power-grid-model/pull/896#discussion_r1973491898)

Apart from variance, the fact is that the `value` itself goes from zero, +ve, -ve sequence in asym to only pos sequence in sym makes it such that we can't do the conversions back and forth.

mgovers: well, `pos_seq` and `three_phase` are supposedly inverse operators (if one doesn't look at the variance)

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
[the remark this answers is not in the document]
figueroa1395: Now it makes sense, because why would the variance increase threefold when going sym -> asym... Whereas it makes sense if we go asym -> sym to divide by three because the uniform representation of the variance holds the total of the three independent phases of the asym, hence we are basically getting the mean. This should definitely be commented in code.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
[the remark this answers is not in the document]
figueroa1395: But the mean of the variance already represents the average among all the three phases, hence equivalent to the variance of a symmetric measurement. This way reads to me as if the variance of each phase was represented as a uniform variance individually. Maybe we can continue offline and post the results of the discussion here.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
figueroa1395: So this is a case of, if we don't need it (expect it), don't add it. If we were to need it, then we can add it. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
figueroa1395: https://github.com/PowerGridModel/power-grid-model/pull/896#discussion_r1973727680

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+// Complex measured value of a sensor with separate variances per phase (but rotationally symmetric in the
+// complex plane)
+template <symmetry_tag sym_type> struct IndependentComplexRandVar {
+    using sym = sym_type;
+
+    ComplexValue<sym> value{};
+    RealValue<sym> variance{}; // variance (sigma^2) of the error range
+
+    explicit operator UniformComplexRandVar<sym>() const {
```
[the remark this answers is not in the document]
figueroa1395: https://github.com/PowerGridModel/power-grid-model/pull/896#discussion_r1973727680

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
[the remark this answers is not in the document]
figueroa1395: https://github.com/PowerGridModel/power-grid-model/pull/896#discussion_r1973504108 explains it. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
Jerry-Jinfeng-Guo: Ok, it is physical then? If this is confirmed to be true, this thread can be resolved.

figueroa1395 (approved):

Jerry-Jinfeng-Guo (approved):

Jerry-Jinfeng-Guo (approved):

PR #897, opened 2025-02-14
Refactor for easier testing

petersalemink95:
This PR will not change behavior: only includes a refactor + additional tests


power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +583,27 @@ template <symmetry_tag sym> struct NodeState {
     }
 };
 
+class RankIterator {
+  public:
+    RankIterator() = default;
+
```
mgovers, line 589: default constructor is not necessary

```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +583,27 @@ template <symmetry_tag sym> struct NodeState {
...
+    void iterate_ranks(auto const& regulator_order, auto adjust_transformer_in_rank, bool& tap_changed,
+                       std::vector<IntS>& iterations_per_rank, Idx& rank_index) {
```
mgovers, line 591: do you think it makes more sense to make `iterations_per_rank` and `rank_index` member variables (private + public `constexpr` getters for testability)? The `adjust_transformer_in_rank` can then be provided into the constructor, and `tap_changed` can be an output parameter (just like it is for `adjust_transformer_in_rank`). That way, you can also really test per iteration

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +583,27 @@ template <symmetry_tag sym> struct NodeState {
     }
 };
 
+class RankIterator {
+  public:
+    RankIterator() = default;
+
+    void iterate_ranks(auto const& regulator_order, auto adjust_transformer_in_rank, bool& tap_changed,
```
mgovers, line 590: since this logic is actually more generic than regulated transformers, maybe the names here can also be more generic, e.g.:
* `regulator_order` -> `ranked_order`
* `adjust_transformer_in_rank` -> `apply_single_object_in_rank`
* `tap_changed` -> `adjusted` or `additive_result`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -890,21 +911,18 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
...
+                tap_changed = adjust_transformer(regulator, state, result, update_data, search, options) || tap_changed;
+                return tap_changed;
+            };
+
+            RankIterator rank_iterator;
+            rank_iterator.iterate_ranks(regulator_order, adjust_transformer_in_rank, tap_changed, iterations_per_rank,
+                                        rank_index);
+
             if (tap_changed) {
                 if (static_cast<uint64_t>(iterations_per_rank[rank_index]) > 2 * max_tap_ranges_per_rank[rank_index]) {
```
mgovers, line 927: i think it's good to do the outer `while` loop + stop criterion in the `RankIterator` as well if possible

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +583,27 @@ template <symmetry_tag sym> struct NodeState {
     }
 };
 
+class RankIterator {
```
mgovers, line 586: `Iterator` is typically used as an object to iterate over. `RankIterator` may cause confusion. Maybe you can rename it to something like `RankedIteration`

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
petersalemink95: Don't think I agree. The fact that while `tap_changed == true` there needs to happen something feels like behavior of `TapPositionOptimizerImpl`. The iterations itself can happen in the `RankIteration`.

Next to that I think moving the while loop + stop criteria would make thing overly complicated.

mgovers: ok

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -1167,6 +1167,35 @@ TEST_CASE("Test tap position optmizer I/O") {
...
+    bool tap_changed{false};
+    std::vector<IntS> iterations_per_rank = {0, 3, 6};
+    Idx rank_index{0};
+    bool update{false};
+
+    auto const mock_lambda = [&](Idx const& /*rank_idx*/, Idx const& /*transformer_idx*/,
+                                 std::vector<IntS> const& /*same_rank_regulators*/) { return update; };
+    SUBCASE("Test tap not changed") {
+        // std::vector<IntS> iterations_per_rank_original = iterations_per_rank;
+        rank_iterator.iterate_ranks(regulator_order, mock_lambda, tap_changed, iterations_per_rank, rank_index);
```
mgovers, line 1182: since setting `tap_changed` is a side-effect of the `RankIteration`, please also check it

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
petersalemink95: I don't get why we would do that

petersalemink95: Addressed most of these points - as discussed offline

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +583,39 @@ template <symmetry_tag sym> struct NodeState {
     }
 };
 
+class RankIteration {
+  public:
+    RankIteration(std::vector<IntS> iterations_per_rank, Idx rank_index)
+        : iterations_per_rank_{std::move(iterations_per_rank)}, rank_index_{rank_index} {}
+
+    // Getters
+    constexpr auto iterations_per_rank() const { return iterations_per_rank_; }
```
mgovers, line 592:
```suggestion
    constexpr auto const& iterations_per_rank() const { return iterations_per_rank_; }
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +583,39 @@ template <symmetry_tag sym> struct NodeState {
...
+    constexpr auto rank_index() const { return rank_index_; }
+
+    // Setters
+    void set_rank_index(Idx rank_index) { rank_index_ = rank_index; }
+
+    bool iterate_ranks(auto const& ranked_order, auto apply_single_object_in_rank, bool adjusted) {
+        for (Idx i = 0; i < static_cast<Idx>(ranked_order.size()); ++i) {
```
mgovers, line 599: not strictly needed but looks nicer
```suggestion
        for (Idx i = 0; i < ranked_order.ssize(); ++i) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -583,6 +583,39 @@ template <symmetry_tag sym> struct NodeState {
...
+            auto const& same_rank_regulators = ranked_order[i];
+            for (Idx j = 0; j < static_cast<Idx>(same_rank_regulators.size()); ++j) {
```
mgovers, line 601: idem
```suggestion
            for (Idx j = 0; j < same_rank_regulators.ssize(); ++j) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: this should also fix the readthedocs compilation error

mgovers (approved):

PR #901, opened 2025-02-17
Generic branch: small fixes to the documentation

scud-soptim:
Fix for Documentation Issue: https://github.com/PowerGridModel/power-grid-model/pull/900

### Changes proposed in this PR:  
- Removed German comments  
- Reformatted Python code examples  
- Fixed Markdown formatting

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -13,6 +13,7 @@
 		"ms-toolsai.jupyter-renderers",
 		"ms-toolsai.vscode-jupyter-cell-tags",
 		"ms-toolsai.vscode-jupyter-slideshow",
+		"ms-toolsai.datawrangler",
 		"ms-vscode.cmake-tools",
 		"ms-vscode.cpptools",
 		"ms-vscode.cpptools-extension-pack",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,19 @@
     "cmake.configureSettings": {
         "POWER_GRID_MODEL_SANITIZER": 1
     },
-    "testMate.cpp.test.executables": "${command:cmake.buildDirectory}/bin/power_grid_model_*_tests",
+    "testMate.cpp.test.executables": "${command:cmake.buildDirectory}/bin/power_grid_model_*_tests*",
+    "testMate.cpp.debug.configTemplate": {
+        "type": "cppvsdbg",
+        "linux": { "type": "cppdbg", "MIMode": "gdb" },
+        "darwin": { "type": "cppdbg", "MIMode": "lldb" },
+        "win32": { "type": "cppvsdbg" },
+        "program": "${exec}",
+        "args": "${argsArray}",
+        "cwd": "${cwd}",
+        "env": "${envObj}",
+        "environment": "${envObjArray}",
+        "sourceFileMap": "${sourceFileMapObj}",
+    },
     "doxdocgen.generic.filteredKeywords": ["PGM_API"],
     "doxdocgen.generic.includeTypeAtReturn": false,
     "python.testing.pytestArgs": [


docs/examples/Generic Branch Example.ipynb
```diff
@@ -18,7 +18,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 7,
```
mgovers, line 21: quite minor but one of the things we do to guarantee that the example is correct, is to restart the kernel, clear all outputs and then re-running the notebook. A quick check to see whether that has happened is by checking the `execution_count`, which should start at `0` and increment by `1` for each cell.

Can you please do that?

mgovers (commented):
Hi @scud-soptim , Nice findings. Before we accidentally merge too quickly again, can you please go over everything again and let us know when it's ready (or pull the PR back to draft if there is still work required). That way we can reduce the amount of follow-up PRs

docs/examples/Generic Branch Example.ipynb
```diff
@@ -18,7 +18,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 3,
```
mgovers, line 21: can you please restart kernel, clear outputs and rerun?

I'd expect this number to be `0`

docs/examples/Generic Branch Example.ipynb
```diff
@@ -796,7 +779,7 @@
     "This chapter provides a guide on how to use the Generic Branch to import data from CIM/CGMES network datasets and integrate it into a PGM model.\n",
```
mgovers, line 779: maybe out of scope for this PR but it feels to me like this section should go to PGM-IO instead. See e.g. our example on importing data from Arrow in https://power-grid-model-io.readthedocs.io/en/stable/examples/arrow_example.html , which we technically do not have an off-the-shelf conversion function for, but we do have some documentation on how to handle it.

That said, this is really useful documentation! Thank you for contributing this!

docs/examples/Generic Branch Example.ipynb
```diff
@@ -1056,7 +1039,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.12"
+   "version": "3.11.0rc1"
```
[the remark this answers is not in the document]
mgovers: ok in that case, i can also try to do it if you are OK with that

docs/examples/Generic Branch Example.ipynb
mgovers: that's odd, which interpreter are you using? jupyter notebook or raw ipykernel (e.g. using the VS Code plugin for notebooks)

docs/examples/Generic Branch Example.ipynb
mgovers: definitely not! only this section on CIM/CGMES conversion probably belongs in PGM-IO. The rest of the document that shows how the Generic Branch component works is definitely in the right place as it is.

mgovers: we can also decide to move it later

mgovers: it is of high interest 😉 we are extremely happy with this and @petersalemink95 already forwarded it internally. it's mainly the purpose of PGM vs the purpose of PGM-IO that makes this a little bit tricky. We'll discuss it offline and do a follow-up PR if necessary. Again, thanks for the input!

docs/examples/Generic Branch Example.ipynb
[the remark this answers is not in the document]
mgovers: done in [3cf022a](https://github.com/PowerGridModel/power-grid-model/pull/901/commits/3cf022aacc8525adac922b1f499353e286fd7577)

mgovers (approved):

PR #906, opened 2025-02-19
Drop python 3.10: fix documentation

Jerry-Jinfeng-Guo:
The drop python 3.10 saga [PowerGridModel/power-grid-model/pull/879)] missed the documentation build guide.


mgovers (approved):
Whoops, good catch. If you like, we can also immediately bump to 3.13 since we want people to always use the latest greatest but i leave that up to you

petersalemink95 (approved):

PR #907, opened 2025-02-20
Bugfix: rule in none missing yield false positives

Jerry-Jinfeng-Guo:
Closes PowerGridModel/power-grid-model/issues/902


mgovers:
i think this is ready for review but would like to double-check with you first cause of the additional validation thing with asym sensor + mixed p/q-sigma instantiation

src/power_grid_model/validation/rules.py
```diff
@@ -777,23 +772,21 @@ def none_missing(
...
+            invalid = np.isnan(data[component][field])
         else:
-            invalid = np.equal(data[component][field][index], nan)
+            invalid = np.equal(data[component][field], nan)
 
         if invalid.any():
-            if isinstance(invalid, np.ndarray):
-                invalid = np.any(invalid)
+            # handle asymmetric values
+            invalid = np.any(invalid, axis=tuple(range(1, invalid.ndim)))
```
figueroa1395, line 783: Nitpick: I would suggest commenting a bit more why this fix handles asymmetric values, so later on is evident at first glance. Important because this is the bug source.

src/power_grid_model/validation/rules.py
```diff
@@ -812,16 +805,18 @@ def valid_p_q_sigma(data: SingleDataset, component: ComponentType) -> list[Multi
     q_nan = np.isnan(q_sigma)
```
figueroa1395, line 805: Why isn't this handled like above with `_nan_type`? Is it because we know for certain this specific fields' `nan`s are not custom?

src/power_grid_model/validation/validation.py
```diff
@@ -247,40 +246,26 @@ def validate_ids(update_data: SingleDataset, input_data: SingleDataset) -> list[
...
-        if _power_sigma.ndim < _mask.ndim:
-            _mask = np.any(_mask, axis=tuple(range(_power_sigma.ndim, _mask.ndim)))
-
-        for sublist, should_remove in zip(_sensor_mask, _mask):
-            if should_remove and "power_sigma" in sublist:
-                sublist = cast(list[str], sublist)
-                sublist.remove("power_sigma")
-
-    if _check_sensor_in_data(data, sensor):
+    if sensor in data:
```
figueroa1395, line 256: Since `sensor` can theoretically be any `ComponentType`, shouldn't we add a fallback for when, by mistake, we use a "wrong" component as an argument (which wouldn't have the field `power_sigma`)? For for easier debuging.

I am assuming (didn't check in detail), that `_check_sensor_in_data()` use to do something along those lines (?)

src/power_grid_model/validation/validation.py
```diff
@@ -247,40 +246,26 @@ def validate_ids(update_data: SingleDataset, input_data: SingleDataset) -> list[
...
+        q_sigma = sensor_data["q_sigma"]
+
+        # virtual patch to handle missing power_sigma
+        asym_axes = tuple(range(sensor_data.ndim, p_sigma.ndim))
+        mask = np.logical_and(np.isnan(power_sigma), np.any(np.logical_not(np.isnan(p_sigma)), axis=asym_axes))
+        power_sigma[mask] = np.nansum(p_sigma[mask], axis=asym_axes)
 
-            _process_power_sigma_in_list(sensor_mask, power_sigma, p_sigma, q_sigma)
+        mask = np.logical_and(np.isnan(power_sigma), np.any(np.logical_not(np.isnan(q_sigma)), axis=asym_axes))
+        power_sigma[mask] = np.nansum(q_sigma[mask], axis=asym_axes)
```
figueroa1395, line 268: Is this logic correct? Shouldn't the `nan` values of `power_sigma` be overriten by a combination of `p_sigma`and `q_sigma`? Here it seems to me like ones may be overriten by one or the other, or the `q_sigma` values could potentially be ignored alltogether.

src/power_grid_model/validation/errors.py
```diff
@@ -472,6 +472,14 @@ class FaultPhaseError(MultiFieldValidationError):
     _message = "The fault phase is not applicable to the corresponding fault type for {n} {objects}."
 
 
+class PQSigmaPairError(MultiFieldValidationError):
+    """
+    The combination of p_sigma and q_sigma is not valid. They should be both present or both absent.
+    """
+
+    _message = "The combination of p_sigma and q_sigma is not valid for {n} {objects}."
+
```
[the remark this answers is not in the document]
mgovers: that should already be handled by the `ValidationError` and `MultiFieldValidationError` base classes

src/power_grid_model/validation/rules.py
mgovers: `q` should always be floating point, so there's no doubt about whether it's an integer or not

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/blob/812fe84435970154d2dde078b485e7f46b9a3cff/src/power_grid_model/validation/validation.py#L387-L388: they are only of type `ComponentType.sym_power_sensor` and `ComponentType.asym_power_sensor`. For private functions, a comment should suffice.

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: We decided to have overwriting here as this is only happening on the proxy data.

src/power_grid_model/validation/validation.py
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model/blob/eda52acdf7313c6de9a307357745c51d2a072888/src/power_grid_model/validation/validation.py#L256-L258

src/power_grid_model/validation/validation.py
figueroa1395: What do you mean? Isn't this to handle in general the cases when `power_sigma` is `nan` and `p_sigma` and `q_sigma` are provided? Shouldn't  then the "total" sigma value be calculated from both contributions?

mgovers: you may be correct. let's double-check.

figueroa1395: I understand the decision to not override when `power_sigma` is defined, but when both `p_sigma` and `q_sigma` are provided, one should take the total value including both, and that doesn't happen here. Let's take the discussion offline and post the conclusion here.

Jerry-Jinfeng-Guo: > but when both `p_sigma` and `q_sigma` are provided, one should take the total value including both

Can you elaborate on this? I do not quite follow yet

figueroa1395: Conclusion of offline discussion: My thought was that if `power_sigma` was `nan`, it should be calculated as the norm with components `p_sigma`, `q_sigma`, which is not reflected here. However, even though my statement is true, this is not the test for it and that was my confusion. Here, we only check that if `power_sigma` is `nan`, either of the other two should be present and its value is overwritten, but the validity of that overwritten value is checked elsewhere. It is about separation of concerns and reducing complexity. This logic is fine as is. Resolving.

figueroa1395 (approved):

PR #912, opened 2025-03-03
Step-up transformer tap changer support: Patch tests

figueroa1395:
The tests in https://github.com/PowerGridModel/power-grid-model/pull/868 need to be addressed:

- [ ] One 2w transformer in the transformer ranking unit test needs to be modified such that the tap and the control side are on the same side away from the source.
- [ ]  The validation test cases for a transformer with tap and control side on the same side need to be modified such that the tap and the control are on the side farther away from the source.
- [ ] Once the control logic is merged in the main branch, the tests should not be marked as failing anymore.

Note: When cleaning the commit history in https://github.com/PowerGridModel/power-grid-model/pull/909 , the tests introduced in https://github.com/PowerGridModel/power-grid-model/pull/868 for the feature branch https://github.com/PowerGridModel/power-grid-model/pull/826  were removed - for some unknown reason to me - from the feature branch itself. In this PR I cherry pick the commits that were deleted and add them back while also doing the changes proposed in this PR.


figueroa1395:
@nitbharambe Can you please take a look at [01ceabf](https://github.com/PowerGridModel/power-grid-model/pull/912/commits/01ceabf53f390cbb3ea9fb40c6efdd24dc5d7a79). Make sure it makes sense. This is the amended test given that the control side should be far from the source, together with the tap side.

nitbharambe (approved):

figueroa1395:
@nitbharambe Let's wait for #892 to be merged into #826 before merging this one. That way we can actually test the tests and fix where necessary.

figueroa1395:
The validation tests pass on the C++ side, but fail on the Python side. This is because in C++ we don't check (yet) for the `u_angle` results, which we do on the Python side. This happened because I used C++ to create the tests, so it never warned me about the `u_angle` results being wrong (as they aren't checked), so they were never right in the first place. I will update and push again.

Jerry-Jinfeng-Guo:
> However, I would like to give extra attention to one particular test. Given my by-hand calculations, I was expecting the test trafo-control-tap-same-side-any-valid-tap to give the same resulting tap position as the fast_any strategy, but even though the result is within the u_band, it is the same of the min_voltage_tap one. 

There is no guarantee that ‘fast_any’ would yield the exactly same as ‘any’. As the strategy name suggests, it might give anything within the range.

Jerry-Jinfeng-Guo:
When debugging through the steps (C++), I noticed that the `step-up-transformer-` cases are not added to `all_cases`. Maybe it's because none of step-up transformer cases contain `update.json` while it is set to add cases in batch mode. @figueroa1395 could you take a look?

Jerry-Jinfeng-Guo:
Great work! Based on the outcome of all passing tests and that this has been reviewed in https://github.com/PowerGridModel/power-grid-model/pull/868, I here by approve.

Jerry-Jinfeng-Guo (approved):

PR #915, opened 2025-03-10
Use cmake to build the DLL for conda package

TonyXiang8787:
Use `cmake` to build the DLL/SO for conda package. In runtime, if the `power_grid_core` detects that it is in conda environment, load the DLL/SO from `PATH`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -268,7 +268,7 @@ jobs:
           cmake --install build/
 
       - name: Build and install cmake target for Linux
-        if: matrix.os == 'windows'
+        if: matrix.os == 'linux'
         run: |
           cmake -GNinja -DCMAKE_BUILD_TYPE=Release -B build/ -S . -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DCMAKE_PREFIX_PATH=$CONDA_PREFIX
           cmake --build build/ --verbose -j1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -235,10 +235,15 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu", "windows"] # We do not test conda for MacOS
+        include:
+          - os: ubuntu
+            shell: bash -el {0}
+          - os: windows
+            shell: powershell
   
-    # defaults:
-    #   run:
-    #     shell: bash -el {0}
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -235,10 +235,15 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu", "windows"] # We do not test conda for MacOS
+        include:
+          - os: ubuntu
+            shell: bash -el {0}
+          - os: windows
+            shell: powershell
```
mgovers, line 242: shouldn't we also include macos since you added it later on?

src/power_grid_model/_core/power_grid_core.py
```diff
@@ -125,11 +126,26 @@ def _load_core() -> CDLL:
...
+            dll_file = "power_grid_model_c.dll"
+        elif platform.system() == "Darwin":
+            dll_file = "libpower_grid_model_c.dylib"
+        elif platform.system() == "Linux":
+            dll_file = "libpower_grid_model_c.so"
+        else:
+            raise NotImplementedError(f"Unsupported platform: {platform.system()}")
+        # the dll will be found through conda environment
+        dll_path = dll_file
     else:
```
mgovers, line 141: this is potentially breaking for users who use conda as their base environment (e.g. because they installed python via conda) and run both conda and venv next to eachother. i think we should first look for a local file of PGM and then change to the global version.

Especially people installing PGM themselves may accidentally load the PGM version from outside the current working directory

.github/workflows/main.yml
TonyXiang8787: We disable macos because conda support is corrupted in GA.

mgovers (approved):
a question more out of personal interest: what's the reason the `dylib` is only used for MacOS for conda and not for PGM? is that something we need to fix?

TonyXiang8787:
> a question more out of personal interest: what's the reason the `dylib` is only used for MacOS for conda and not for PGM? is that something we need to fix?

@mgovers `cmake` in `macOS` produces `.dylib` for shared library. `setuptools` in `macOS` produces `.so` in compiled extension. We are just following the convention.

mgovers:
> > a question more out of personal interest: what's the reason the `dylib` is only used for MacOS for conda and not for PGM? is that something we need to fix?
> 
> @mgovers `cmake` in `macOS` produces `.dylib` for shared library. `setuptools` in `macOS` produces `.so` in compiled extension. We are just following the convention.

clear, thanks!

PR #916, opened 2025-03-14
Use math solver dispatch to separate math solver from main model

TonyXiang8787:
This PR uses `MathSolverDispatcher` to separate the real implementation of math solver in a different translation unit. The handle of the dispatcher is an input argument for the constructor of the `MainModel`. In this way, the math solver function pointers are set in run-time.

The core library is still header-only but with clear boundaries. In C-API separate translation units can compile the `MainModel` and `MathSolver`.

This enables the `conda-forge` again without hitting the RAM limit. See https://github.com/conda-forge/power-grid-model-feedstock/pull/328


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
```diff
@@ -0,0 +1,156 @@
...
+template <template <symmetry_tag> class MathSolverType> struct math_solver_tag {};
+
+struct MathSolverDispatcher {
+    template <symmetry_tag sym> struct MathSolverDispatcherConfig {
+        template <template <symmetry_tag> class MathSolverType>
+        constexpr MathSolverDispatcherConfig(math_solver_tag<MathSolverType>)
+            : create{[](std::shared_ptr<MathModelTopology const> const& topo_ptr) -> void* {
+                  return new MathSolverType<sym>{topo_ptr};
+              }},
+              destroy{[](void const* solver) { delete reinterpret_cast<MathSolverType<sym> const*>(solver); }},
```
mgovers, line 34: here and below: is there a way so that we do not have to use `void const*` + `reinterpret_cast`?

if not, maybe you can do something like:
1. an alias like `using AnySolver = void const*`, as well as
2. create a wrapper for `reinterpret_cast` (can be a `private` template function in the `MathSolverDispatcher`), e.g.
   ```cpp
   template <template <symmetry_tag> class MathSolverType, typename VoidType>
       requires std::same_as<std::remove_const_t<VoidType>, void>
   auto& load(math_solver_tag<MathSolverType> /*tag*/, VoidType* solver) {
       using MaybeConstSolverType = std::conditional_t<std::is_const_v<VoidType>, MathSolverType<sym> const, MathSolverType<sym>>;
       assert(solver != nullptr);
       return *reinterpret_cast<MaybeConstSolverType*>(solver)
   }
   ```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+    template <symmetry_tag sym> MathSolverDispatcherConfig<sym> const& get_dispather_config() const {
+        if constexpr (is_symmetric_v<sym>) {
+            return sym_config;
+        } else {
+            return asym_config;
+        }
+    }
+    MathSolverDispatcherConfig<symmetric_t> sym_config;
+    MathSolverDispatcherConfig<asymmetric_t> asym_config;
```
mgovers, line 88: probably better to make these `private` since you have a `get_dispatcher_config()` entrypoint anyway
```suggestion

  private:
    MathSolverDispatcherConfig<symmetric_t> sym_config;
    MathSolverDispatcherConfig<asymmetric_t> asym_config;
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
```diff
@@ -0,0 +1,156 @@
...
+            run_short_circuit;
+        std::add_pointer_t<void(void*)> clear_solver;
+        std::add_pointer_t<void(void*, bool)> parameters_changed;
+    };
+
+    template <template <symmetry_tag> class MathSolverType>
+    constexpr MathSolverDispatcher(math_solver_tag<MathSolverType> /* unused */)
+        : sym_config{math_solver_tag<MathSolverType>{}}, asym_config{math_solver_tag<MathSolverType>{}} {}
+
+    template <symmetry_tag sym> MathSolverDispatcherConfig<sym> const& get_dispather_config() const {
```
mgovers, line 80:
```suggestion
    template <symmetry_tag sym> MathSolverDispatcherConfig<sym> const& get_dispatcher_config() const {
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+namespace math_solver {
+
+// forward declare YBus
+template <symmetry_tag sym> class YBus;
+
+template <template <symmetry_tag> class MathSolverType> struct math_solver_tag {};
+
+struct MathSolverDispatcher {
+    template <symmetry_tag sym> struct MathSolverDispatcherConfig {
```
mgovers, line 28: optionally, you can rename this to just `Config` since it's a member struct anyways

mgovers (commented):
really nice!

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
```diff
@@ -0,0 +1,156 @@
...
+        std::add_pointer_t<ShortCircuitSolverOutput<sym>(void*, ShortCircuitInput const&, CalculationInfo&,
+                                                         CalculationMethod, YBus<sym> const&)>
+            run_short_circuit;
+        std::add_pointer_t<void(void*)> clear_solver;
+        std::add_pointer_t<void(void*, bool)> parameters_changed;
+    };
+
+    template <template <symmetry_tag> class MathSolverType>
+    constexpr MathSolverDispatcher(math_solver_tag<MathSolverType> /* unused */)
+        : sym_config{math_solver_tag<MathSolverType>{}}, asym_config{math_solver_tag<MathSolverType>{}} {}
```
TonyXiang8787, line 78: @mgovers can you have a look at the failure? I really do not understand why.

https://github.com/PowerGridModel/power-grid-model/actions/runs/13853372262/job/38764829184?pr=916

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+// forward declare YBus
+template <symmetry_tag sym> class YBus;
+
+template <template <symmetry_tag> class MathSolverType> struct math_solver_tag {};
+
+struct MathSolverDispatcher {
+    template <symmetry_tag sym> struct MathSolverDispatcherConfig {
+        template <template <symmetry_tag> class MathSolverType>
+        constexpr MathSolverDispatcherConfig(math_solver_tag<MathSolverType> /* unused */)
```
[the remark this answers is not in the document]
TonyXiang8787: You are right. The `math_solver_tag` is a tag type to capture the actual type. In constructor template of the class, you can only rely on implicit deduction.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -22,10 +22,12 @@ class MainModel {
...
 
-    explicit MainModel(double system_frequency, ConstDataset const& input_data, Idx pos = 0)
-        : impl_{std::make_unique<Impl>(system_frequency, input_data, pos)} {}
-    explicit MainModel(double system_frequency, meta_data::MetaData const& meta_data)
-        : impl_{std::make_unique<Impl>(system_frequency, meta_data)} {};
+    explicit MainModel(double system_frequency, ConstDataset const& input_data,
+                       MathSolverDispatcher const* math_solver_dispatcher, Idx pos = 0)
+        : impl_{std::make_unique<Impl>(system_frequency, input_data, math_solver_dispatcher, pos)} {}
+    explicit MainModel(double system_frequency, meta_data::MetaData const& meta_data,
+                       MathSolverDispatcher const* math_solver_dispatcher)
```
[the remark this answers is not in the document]
TonyXiang8787: Using `const&` indeed makes sense.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
TonyXiang8787: @figueroa1395 @mgovers using `void*` is the only way in the function pointer definition because we are doing type erasure here. That means:

1. In the dispatcher function pointer types, there should not be concrete types in the signature.
2. The lambdas receives `void*` and needs to `reinteprete_cast` to the concrete type. This happens in a separate compilation units as intended.
3. The lambdas should be stateless otherwise it will not decay to a function pointer. So it cannot call a member function of any class (like `Config` class).
4. We can theoretically make a templated `static` function in the `Config` class for the `reinteprete_cast`, and the lambas can call that function. But I highly hesitate the added value of such a design. It does not provide more value then directly using `reinteprete_cast` in lambdas.
5. I also do not see the added value of justing defining a type alias for `void*`.

TonyXiang8787: @mgovers @figueroa1395, after some thinking, there is another way here to avoid type erasure: define pure virtual base class for `MathSolver`. In the dispatcher we define a function pointer to create a concrete solver which returns the pointer to base class.

This should avoid all `reintepret_cast`.

TonyXiang8787: This method is maybe better because it asks less boilerplate.

TonyXiang8787: @mgovers @figueroa1395  see the new implementation. It is much less boilerplate and without `reinteprete_cast`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/math_solver_dispatch.hpp
```diff
@@ -0,0 +1,117 @@
...
+            : create{[](std::shared_ptr<MathModelTopology const> const& topo_ptr) -> MathSolverBase<sym>* {
+                  return new MathSolverType<sym>{topo_ptr};
+              }} {}
+
+        std::add_pointer_t<MathSolverBase<sym>*(std::shared_ptr<MathModelTopology const> const&)> create;
+    };
+
+    template <template <class> class MathSolverType>
+    constexpr MathSolverDispatcher(math_solver_tag<MathSolverType> /* unused */)
+        : sym_config_{math_solver_tag<MathSolverType>{}}, asym_config_{math_solver_tag<MathSolverType>{}} {}
```
mgovers, line 70:
```suggestion
    constexpr MathSolverDispatcher(math_solver_tag<MathSolverType> tag)
        : sym_config_{tag}, asym_config_{tag} {}
```

mgovers (approved):
One nitpick, otherwise looks great

TonyXiang8787:
@mgovers  please approve again. I will merge this after trying conda build again.

mgovers (approved):

PR #918, opened 2025-03-14
Automatic tap changer step up transformer support: documentation

Jerry-Jinfeng-Guo:


docs/user_manual/calculations.md
```diff
@@ -667,7 +667,7 @@ This assumption is reflected in the requirements mentioned in {hoverxreftooltip}
 ```
 
 ```{note}
-The control logic assumes that the (compensated) control voltage decreases when the tap position increases.
+The control logic assumes that the (compensated) control voltage decreases when the tap position changes (increases when the control side of the regulator and the tap side of the transformer are at different sides, or decreases they are at the same side).
```
mgovers, line 670: this entire line can go
```suggestion
```

docs/user_manual/components.md
```diff
@@ -814,7 +814,6 @@ The actual grid state is not changed after calculations are done.
 
 The following additional requirements exist on the input parameters.
```
mgovers, line 815: Should we add another requirement or note that the regulator should be on the far side (with a reference to the explanation in `calculations.md`?

mgovers (commented):
let's wait with merging this PR until we get the OK on the new functionality from the users who requested this

docs/user_manual/components.md
Jerry-Jinfeng-Guo: This new requirement is indeed introduced with the new work-out.

mgovers (approved):

PR #921, opened 2025-03-17
Drop manylinux_2_24 and MacOS < 13.3

### LINKED ISSUE #717: [FEATURE] Drop support for manylinux_2_24
Currently the pre-built binary wheels in the PyPI supports `manylinux_2_24`. We would like to upgrade this to `manylinux_2_28`. The main issue with this upgrade is that `amazonlinux 2` will not be supported anymore.

`amazonlinux 2` will be EoL on 30 June 2025. We plan also to drop the support of `manylinux_2_24` on that day.

TonyXiang8787:
Closes #717


mgovers (commented):
maybe it's good as a test to see if `std::views` now actually works

mgovers (approved):

TonyXiang8787:
> maybe it's good as a test to see if `std::views` now actually works

Go ahead to try.

mgovers (approved):

PR #925, opened 2025-03-19
Try to fix macOS version bump

TonyXiang8787:


mgovers (dismissed):

mgovers (approved):

mgovers (approved):

mgovers (approved):

mgovers (dismissed):

setup.py
```diff
@@ -18,6 +18,10 @@
     if_win = True
 elif platform.system() in ["Linux", "Darwin"]:
     if_win = False
+    if platform.system() == "Darwin" and (
+        "MACOSX_DEPLOYMENT_TARGET" not in os.environ or float(os.environ["MACOSX_DEPLOYMENT_TARGET"]) < 13.3
+    ):
+        os.environ["MACOSX_DEPLOYMENT_TARGET"] = "13.3"
```
mgovers, line 24:
```suggestion
    os.environ["MACOS_DEPLOYMENT_TARGET"] |= "13.3"
```

We're dealing with 2 potential edge cases:
1. we version bump again and the user has an old MACOS_DEPLOYMENT_TARGET that we suddenly don't support anymore. This should raise an error
2. the user desires a newer minimum version than our minimum requirement.

we have to consider 3 possibilities:
* [ ] this solution (conditional overwriting)
  1. tacitly bumps minimum version => may break user's runtime: Not OK
  2. uses user-picked latest version: OK
* [ ] always overwrite (`os.environ["MACOS_DEPLOYMENT_TARGET"] = "13.3"`)
  1. tacitly bumps minimum version => may break user's runtime: Not OK
  2. builds for older version => OK-ish but not great either
* [ ] only set when not set by parent environment (`os.environ["MACOS_DEPLOYMENT_TARGET"] |= "13.3"`)
  1. throws error (minimum version not correct): OK
  2. uses user-picked latest version: OK

figueroa1395: `os.environ["MACOS_DEPLOYMENT_TARGET"] |= "13.3"` did not seem to work locally. See [0c1979c](https://github.com/PowerGridModel/power-grid-model/pull/925/commits/0c1979c56c4c6860b93dacbf51b7037b5a0f727a) for a different fix.

figueroa1395:
> @figueroa1395 found an error when building locally. to be investigated before merging

The problem found was that even though `cmake` compiled the project without any issues, `pip` defaults to an old version of the environment variable "MACOSX_DEPLOYMENT_TARGET" for compatibility reasons. The fix was to manually overwrite this variable.

PR #929, opened 2025-03-20
Unreachable regulators condition is moved

nitbharambe:
We reintroduced this bug mentioned in #702 (PR: #703) for unreachable regulators which would be the reason for more failing cases

Move the condition for unreachable regulators


figueroa1395:
Should we add a minimum reproducible example, as a validation test case, that fails before the fix and passes after the order correction?

nitbharambe:
> Should we add a minimum reproducible example, as a validation test case, that fails before the fix and passes after the order correction?

Indeed, adding test

Jerry-Jinfeng-Guo:
The bug that originally removed the isolated transformers is introduced in https://github.com/PowerGridModel/power-grid-model/commit/b642c970b5cf3e19f655c989fde8b2203614695d

Jerry-Jinfeng-Guo (approved):

PR #930, opened 2025-03-21
The public CPP API from PGM should rely on c++17 standard

TonyXiang8787:
The public CPP API of PGM should rely on  c++17 standard as decided by the TSC. However, there are c++20 features used. In this PR we force the package test to build in c++17. And the error will show up.

### DIFF
--- a/power_grid_model_c/power_grid_model_cpp/CMakeLists.txt
+++ b/power_grid_model_c/power_grid_model_cpp/CMakeLists.txt
@@ -30,6 +30,7 @@ target_sources(power_grid_model_cpp PUBLIC
 set_target_properties(power_grid_model_cpp PROPERTIES
   VERSION ${PGM_VERSION}
   SOVERSION ${PGM_VERSION}
+  CXX_STANDARD 17
 )
 
 
--- a/power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp
+++ b/power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp
@@ -71,8 +71,8 @@ class Handle {
             auto const* const failed_scenario_seqs = PGM_failed_scenarios(handle_ptr);
             auto const* const failed_scenario_messages = PGM_batch_errors(handle_ptr);
             for (Idx i = 0; i < n_failed_scenarios; ++i) {
-                failed_scenarios[i] = PowerGridBatchError::FailedScenario{.scenario = failed_scenario_seqs[i],
-                                                                          .error_message = failed_scenario_messages[i]};
+                failed_scenarios[i] =
+                    PowerGridBatchError::FailedScenario{failed_scenario_seqs[i], failed_scenario_messages[i]};
             }
             clear_error();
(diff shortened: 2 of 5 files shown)


TonyXiang8787:
@mgovers this is the commit which bumped the version
https://github.com/PowerGridModel/power-grid-model/commit/c825471fb50782d0e73e58e0d0c9c8a34d7b73e8

power_grid_model_c/power_grid_model_cpp/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ target_sources(power_grid_model_cpp PUBLIC
 set_target_properties(power_grid_model_cpp PROPERTIES
   VERSION ${PGM_VERSION}
   SOVERSION ${PGM_VERSION}
+  CXX_STANDARD 17
```
TonyXiang8787, line 33: Will this work as intended? The idea is that the C++ API self is C++17 compliant. But it should not require the user to use C++17 only. User (like validation test) should be able to use any C++ version which is equal or higher than C++17.

mgovers: it will work cause it's the target property, not the overload `set(CMAKE_CXX_STANDARD)`. See also how the other libraries still compile, even if they use C++20 features. It does not enforce upstream/downstream dependencies.

In fact, setting the `CXX_STANDARD` property on the library still doesn't enforce the compiler to only use that interface library with C++17 features only. It's more for exposure than for actually doing something

TonyXiang8787: Okay good to know.

TonyXiang8787:
@mgovers if that is the only non-compliant code. Feel free to approve and merge the PR.

power_grid_model_c/power_grid_model_cpp/CMakeLists.txt
mgovers: ok interestingly, for MSVC, it does make a difference, but not for clang-cl. welp, clang-tidy-19 seems to use the `CXX_STANDARD` target property as well so it's all good now

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp
```diff
@@ -71,8 +71,8 @@ class Handle {
             auto const* const failed_scenario_seqs = PGM_failed_scenarios(handle_ptr);
             auto const* const failed_scenario_messages = PGM_batch_errors(handle_ptr);
             for (Idx i = 0; i < n_failed_scenarios; ++i) {
-                failed_scenarios[i] = PowerGridBatchError::FailedScenario{.scenario = failed_scenario_seqs[i],
-                                                                          .error_message = failed_scenario_messages[i]};
+                failed_scenarios[i] =
+                    PowerGridBatchError::FailedScenario{failed_scenario_seqs[i], failed_scenario_messages[i]};
```
mgovers, line 75: designated initializers are a C++20 feature. Clang-tidy-19 on Windows is OK with this change

mgovers (approved):

PR #762, opened 2024-10-04
Implementation of feature 720 asym_line

leovsch:
**Implement feature**: `# 720`

### Changes proposed in this PR include:

Implementation of [feature 720](https://github.com/PowerGridModel/power-grid-model/issues/720).
Add a new component to support asym_line specification with R, X and C matrix values

### Could you please pay extra attention to the points below when reviewing the PR:

I'm a new contributor so pay attention to the coding standards applicable to this project along with achitectural rules etc.

### Check list

- [x] Check if the available input/update/output data suffices (check `power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/` `input.hpp` / `update.hpp` / `output.hpp` or in the documentation directly)
- [x] If not, add the new data format to `code_generation/data/attribute_classes/` `input.json` / `update.json` / `output.json`  + run `code_generation/code_gen.py`
- [x] Create a new component in a new `power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/component/[component].hpp` file that at least inherits from Base, but in this case `GenericBranch` should inherit from `Branch`
- [x] If necessary: add new enums or exceptions
- [x] Create the necessary unit tests in `power-grid-model/tests/cpp_unit_tests/test_[component].cpp`
- [x] Add the `test_[component].cpp` to `power-grid-model/tests/cpp_unit_tests/CMakeLists.txt`
- [x] Add component to `power_grid_model_c/power_grid_model/include/power_grid_model/all_components.hpp`
- [x] Not necessary for this component (If necessary update `main_core/topology.hpp` / `input.hpp` / `output.hpp` / `update.hpp`)
- [x] Add component to `code_generation/data/dataset_class_maps/dataset_definitions.json` + re-run `code_generation/code_gen.py`
- [x] Add validation test cases to `tests/data`
- [x] Update input/update data validator for the new component: `src/power_grid_model/validation/validation.py` + add corresponding tests

Issue #720, opened 2024-09-12
[FEATURE] support asymmetric line


leovsch:
# Background
Currently the `line` component only supports symmetric components as input. Namely, `r1` `x1` `r0` `x0`. 

## Feature request
Opendss also supports the full specification of Rmatrix and Xmatrix see [the opendss manual](https://opendss.epri.com/Line.html) section 
'Matrix Impedance Definition Properties'. We would like PGM to also support this by having an new component called `asym_line` so that we can reproduce the same result as OpenDSS.

petersalemink95:
Hi @leovsch,

Thank you for your request; this would be a nice addition to Power Grid Model.
I heard from @TonyXiang8787 that you're interested in picking-up the feature yourself. 
That would be awesome.
I'll add the steps that need to be taken to add a new component to Power Grid Model below.
Know that the maintainers are there to support you, if you need help.

You can either request help through the issue or get our contacts directly through @TonyXiang8787.

- [ ] Check if the available input/update/output data suffices (check `power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/` `input.hpp` / `update.hpp` / `output.hpp` or in the documentation directly)
- [ ] If not, add the new data format to `code_generation/data/attribute_classes/` `input.json` / `update.json` / `output.json`  + run `code_generation/code_gen.py`
- [ ] Create a new component in a new `power-grid-model/power_grid_model_c/power_grid_model/include/power_grid_model/component/[component].hpp` file that at least inherits from Base, but in this case `GenericBranch` should inherit from `Branch`
- [ ] If necessary: add new enums or exceptions
- [ ] Create the necessary unit tests in `power-grid-model/tests/cpp_unit_tests/test_[component].cpp`
- [ ] Add the `test_[component].cpp` to `power-grid-model/tests/cpp_unit_tests/CMakeLists.txt`
- [ ] Add component to `power_grid_model_c/power_grid_model/include/power_grid_model/all_components.hpp`
- [ ] Not necessary for this component (If necessary update `main_core/topology.hpp` / `input.hpp` / `output.hpp` / `update.hpp`)
- [ ] Add component to `code_generation/data/dataset_class_maps/dataset_definitions.json` + re-run `code_generation/code_gen.py`
- [ ] Add validation test cases to `tests/data`
- [ ] Update input/update data validator for the new component: `src/power_grid_model/validation/validation.py` + add corresponding tests

Note: the order is recommended, but not necessary

[...]

PR #941, opened 2025-03-31
Full clang-tidy

Laurynas-Jagutis:
For now, this PR is for experimentation with the changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,10 +53,32 @@ jobs:
   reuse-compliance:
     uses: "./.github/workflows/reuse-compliance.yml"
 
+  obtain-clang-tidy-target:
+    runs-on: ubuntu-24.04
+    outputs:
+      target: ${{ steps.set-target.outputs.target }}
+    steps:
+      - name: set target depending on event
+        id: set-target
+        run: |
+          echo "target=power_grid_model_api_tests" >> "$GITHUB_OUTPUT"
+          if [ ${{ github.event_name }} = "pull_request" ]; then
+            echo "target=power_grid_model_api_tests" >> "$GITHUB_OUTPUT"
+          fi
+
+  debug-clang-tidy-target:
+    needs: obtain-clang-tidy-target
+    runs-on: ubuntu-24.04
+    steps:
+      - name: debug target
+        run: |
+          echo ${{ needs.obtain-clang-tidy-target.outputs.target }}
+
   clang-tidy:
+    needs: obtain-clang-tidy-target
     uses: "./.github/workflows/clang-tidy.yml"
     with:
-      target: "power_grid_model_api_tests"
+      target: ${{ needs.obtain-clang-tidy-target.outputs.target }}
 
   ci-passed:
     runs-on: ubuntu-latest
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -78,4 +78,4 @@ jobs:
       - name: Build
         run: |
           cmake --preset ${{ env.PRESET }}
-          cmake --build --preset ${{ env.PRESET }} --target ${{ env.TARGET }}
+          cmake --build --preset ${{ env.PRESET }} --target ${{ env.TARGET }} -- -k 0


.github/workflows/ci.yml
```diff
@@ -53,10 +53,32 @@ jobs:
   reuse-compliance:
     uses: "./.github/workflows/reuse-compliance.yml"
 
+  obtain-clang-tidy-target:
+    runs-on: ubuntu-24.04
```
figueroa1395, line 57: Please change to `ubuntu-latest`

.github/workflows/ci.yml
```diff
@@ -53,10 +53,32 @@ jobs:
...
+    outputs:
+      target: ${{ steps.set-target.outputs.target }}
+    steps:
+      - name: set target depending on event
+        id: set-target
+        run: |
+          echo "target=power_grid_model_api_tests" >> "$GITHUB_OUTPUT"
+          if [ ${{ github.event_name }} = "pull_request" ]; then
+            echo "target=all" >> "$GITHUB_OUTPUT"
+          fi
```
figueroa1395, line 67: Is this desired? I think I understand the thought process behind it, but I don't think it fits our use case very well. We create PRs when the code is very immature in many cases, so why run a full clang tidy run only when opening the PRs? I don't think that would really make use of clang tidy and may lead to problems appearing later.

.github/workflows/clang-tidy.yml
```diff
@@ -78,4 +73,4 @@ jobs:
       - name: Build
         run: |
           cmake --preset ${{ env.PRESET }}
-          cmake --build --preset ${{ env.PRESET }} --target ${{ env.TARGET }}
+          cmake --build --preset ${{ env.PRESET }} --target ${{ env.TARGET }} -- -k 0
```
figueroa1395, line 76: This is equivalent to `ninja`s default `-k 1`, right? If so, why the change? I am just curious.

.github/workflows/ci.yml
mgovers: i think we should just do the full clang-tidy build always. it doesn't matter if the code is still immature, but we don't want to keep breaking the nightly on a weekly basis.

Another important thing: `power_grid_model_benchmark` is marked `EXCLUDE_FROM_ALL` so we should explicitly add it here (as is the case in the `Nightly` build)

.github/workflows/clang-tidy.yml
mgovers: it's equivalent to `ninja`'s `-k 0`. It's the amount of failing targets after which it errors out, i.e., `-k 0` means `always continue`, while `-k 1` means `stop after the first failing target`

figueroa1395: Ohh so I miss-read the documentation then. Now it makes sense: we want to see the whole failing log and for it to continue until it literally can't. Feel free to resolve.

.github/workflows/ci.yml
figueroa1395: I agree with you, maybe I miss-read the code then? To me it reads, when opening a PR do a full clang tidy, if not, just run it on the API. And we want to always do a full run.

.github/workflows/ci.yml
Laurynas-Jagutis: This was removed

.github/workflows/ci.yml
Laurynas-Jagutis: Yes, you did not miss-read the code :). I think that I just got a little confused with the wording in my ticket: "enable full clang-tidy build again on every PR".
But I have now changed it to how you described it.

figueroa1395:
The PR looks good do me. However, before approval, if I take a look at the `clang-tidy` workflow in

> Proof of 2nd bullet point, with a purposefully created error: https://github.com/PowerGridModel/power-grid-model/actions/runs/14214214087/job/39827059570?pr=941

I see that it is failing due to a minor thing (I wonder why it now shows here but it didn't before in the nightly). Could you fix it? I believe it is false positive (so can be probably marked as such), but please take a look at it. Here should be fine since I think the fix would be a one-liner.

Laurynas-Jagutis:
> The PR looks good do me. However, before approval, if I take a look at the `clang-tidy` workflow in
> 
> > Proof of 2nd bullet point, with a purposefully created error: https://github.com/PowerGridModel/power-grid-model/actions/runs/14214214087/job/39827059570?pr=941
> 
> I see that it is failing due to a minor thing (I wonder why it now shows here but it didn't before in the nightly). Could you fix it? I believe it is false positive (so can be probably marked as such), but please take a look at it. Here should be fine since I think the fix would be a one-liner.

This error was created on purpose by me in that single commit. I wanted to test if clang tidy is going to continue with other build objects even if one of them fails. In the link that I attached it shows that indeed it did continue even with the error that I purposefully included, which is the desired behavior. 
After testing it out I removed the error (which I myself included on purpose), thus all the checks pass

figueroa1395:
> > The PR looks good do me. However, before approval, if I take a look at the `clang-tidy` workflow in
> > > Proof of 2nd bullet point, with a purposefully created error: https://github.com/PowerGridModel/power-grid-model/actions/runs/14214214087/job/39827059570?pr=941
> > 
> > 
> > I see that it is failing due to a minor thing (I wonder why it now shows here but it didn't before in the nightly). Could you fix it? I believe it is false positive (so can be probably marked as such), but please take a look at it. Here should be fine since I think the fix would be a one-liner.
> 
> This error was created on purpose by me in that single commit. I wanted to test if clang tidy is going to continue with other build objects even if one of them fails. In the link that I attached it shows that indeed it did continue even with the error that I purposefully included, which is the desired behavior. After testing it out I removed the error (which I myself included on purpose), thus all the checks pass

Are you sure? Because the error points towards the `read` function in the `validation.cpp` file, which I think is unrelated to your "on purpose" error. Just making sure :)

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -37,7 +37,7 @@ class UnsupportedValidationCase : public PowerGridError {
 
 using nlohmann::json;
 
-auto read_file(std::filesystem::path const& path) {
+auto read_file(std::filesystem::path const& path) { // NOLINT(clang-diagnostic-unused-function)
```
mgovers, line 40: is this really unused? if so, please delete, but i expect this to be a false positive. can you find the underlying reason?

figueroa1395 (approved):

mgovers:
@Laurynas-Jagutis i also think it's ready. Can you please press the merge button if you think so as well?

.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
   clang-tidy:
     uses: "./.github/workflows/clang-tidy.yml"
     with:
-      target: "power_grid_model_api_tests"
+      target: "all power_grid_model_benchmark_cpp"
```
TonyXiang8787, line 59: Do we want to include benchmark target? I think this should still be out of the scope until benchmark is refactored to route through C-API

figueroa1395: Makes sense. Then we can keep it in the nightly only.

Jerry-Jinfeng-Guo:
@Laurynas-Jagutis Can we have a short demo session regarding the implications to the rest of the team before merging? Like an extended version of workflow-wise implication of the Tech Review session.

.github/workflows/ci.yml
mgovers: @TonyXiang8787 are you OK with keeping PGM-benchmark in this PR so that this can supersede the nightly entirely?

if not, @Laurynas-Jagutis can you exclude it here?

TonyXiang8787: What do you mean supersede?

TonyXiang8787:
@mgovers @Laurynas-Jagutis 

My proposal is to wait this PR until #762 is merged. So it does not create more additional CI fails there. After #762 has been merged, we might get build failures in this PR because of the clang-tidy stuff for the test code of `asym_line`.

.github/workflows/ci.yml
TonyXiang8787: I think we should keep excluding benchmark target until we route it via C-API

mgovers: > What do you mean supersede?

i meant that, currently, the nightly will fail every time we make a change in one of the headers that causes clang-tidy errors in the tests and PGM benchmark (which are not triggered in regular CI at the moment). This PR enables this, so that we don't accidentally break the Nightly in one of our PRs, effectively changing the purpose of the Nightly from a "catch all things we accidentally break ourselves" to only "catch things that external dependencies may break", which is important.

* Removing PGM-benchmark from this PR means accepting the fact that we will keep continuing to break the PGM-benchmark clang-tidy, and therefore the Nightly on a regular basis.
* If we keep PGM-benchmark, the added overhead is small and it will prevent us from accidentally messing up. The fixes are usually small but currently require a separate PR and also are missed on a regular basis, because we ourselves have to check whether the Nightly still passes manually instead of automatically.

What is your objection against adding the PGM benchmark here?

TonyXiang8787: I propose to even exclude benchmark target in the nightly build. The benchmark target is not in a good shape, and we should park it until we figure out what to do with it, instead of trying to fix clang-tidy or whatever issues come with it in its current form.

It makes less sense to spend effort to detect and fix potential break in clang-tidy in benchmark target, if the whole target is going to be extensively re-designed and re-written.

mgovers: ok, that's fair. @Laurynas-Jagutis can you do that please?

mgovers (approved):
#720 was merged, so this is now good to go. All looks good

mgovers:
@Laurynas-Jagutis quick reminder to double-check that the Nightly works as intended tomorrow

PR #942, opened 2025-04-01
Fix json serialization precision

TonyXiang8787:
The `json` serializer does not provide enough precision for `double` to be able to round trip.

This PR reproduces the problem by creating an edge case. Then it fixes the issue.


tests/native_api_tests/test_api_serialization.cpp
```diff
@@ -240,4 +241,26 @@ TEST_CASE("API Serialization and Deserialization") {
...
+TEST_CASE("API Serialization and Deserialization with float precision") {
+    // dataset with one double value
+    std::vector<double> const u_rated_ref{1.8014398509481982e+16};
+    DatasetConst dataset{"input", true, 1};
+    dataset.add_buffer("node", 1, 1, nullptr, nullptr);
+    dataset.add_attribute_buffer("node", "u_rated", u_rated_ref.data());
+
+    // serialize
+    Serializer json_serializer{dataset, PGM_json};
+    std::string const json_result = json_serializer.get_to_zero_terminated_string(0, -1);
```
mgovers, line 253: do you think it makes sense to add a `CHECK` for the string as well?

mgovers (approved):

PR #944, opened 2025-04-04
Add trusted publisher management in PyPI

figueroa1395:
This PR is used as an experiment to see if we can switch to trusted publisher management as the method to publish in PyPI.

Testing 1 - Make sure behavior works as intended:
- [x] Remove the use of PyPI's secrets and the **_Publish wheels_** sub-workflow in `test-build-release.yml` - [e46d201](https://github.com/PowerGridModel/power-grid-model/pull/944/commits/e46d2010322ac8021a7e508ab648c956c975595d)
- [x] Create a [dedicated workflow](https://github.com/PowerGridModel/power-grid-model/blob/02d5412d8fb498c7b7510c7714a90bd74e1b68f9/.github/workflows/publish-pypi.yml) for publishing in PyPI.
- [x] Add trusted publisher first in [TestPyPI](https://test.pypi.org/manage/project/power-grid-model/settings/publishing/).
- [x] Publish first to [TestPyPI](https://test.pypi.org/manage/project/power-grid-model/release/1.10.81a112406191486/). Make sure it is correct.

Testing 2 - Make sure this works from an independent workflow.
- [x] [Manual CI build](https://github.com/PowerGridModel/power-grid-model/actions/runs/14403952882/job/40395790160) with pre-release.
- [x] [TestPyPI](https://test.pypi.org/manage/project/power-grid-model/release/1.10.82a1400959606/).
- [x] Clean `test-workflow.yml` - [aca7184](https://github.com/PowerGridModel/power-grid-model/pull/944/commits/aca7184f5caf35bf6159870ab15da27ea95fe429)

Note: To test a new workflow that requires a `pre-release` as it was done here. First create a `pre-release` via manual trigger of `build-test-release.yml` with the `pre-release` checkbox; if done via `ci.yml`, it will fail because of `code-quality-check.yml`. Finally, manually trigger `test-workflow.yml` with the new action. Don't forget to clean it up before merging.


figueroa1395:
Manual workflow trigger: https://github.com/PowerGridModel/power-grid-model/actions/runs/14266201056

TonyXiang8787:
@mgovers @figueroa1395 

My proposal to migrate this:

1. Remove the PyPI publish step in the current workflow.
2. Create a new workflow, triggered by `released published`. In the workflow, it downloads the artifacts from GitHub release assets, and upload them to PyPI.

This also creates a separation of release and publish. Release means we make a tag and upload the assets to GitHub Release. Publish means we take the assets and publish them to a package manager.

figueroa1395:
@TonyXiang8787 I'm on it, let me also explore other possibilities and I'll get back to you later if I find something else.

figueroa1395:
@TonyXiang8787 I went with something along the lines of what you proposed. However, from this point, I will move the newly created workflow `publish_pypi.yml` into `ci.yml`, such that it can be tested already. I will undo these changes if/when I have shown that it was correctly published to TestPyPI.

figueroa1395:
General question: Should we "clean up" the release candidates generated during the testing of this PR? i.e. https://github.com/PowerGridModel/power-grid-model/releases

.github/workflows/build-test-release.yml
```diff
@@ -324,19 +317,14 @@ jobs:
...
-        run: |
-          pip install twine
-          echo "Publish to PyPI..."
-          twine upload --verbose wheelhouse/*
-
-      - name: Release
+      - name: Create GitHub release
         uses: softprops/action-gh-release@v2
         if: (inputs.create_release)
         with:
```
mgovers, line 330: probably best to create the GH release in the same action as the PyPI publishing

.github/workflows/ci.yml
```diff
@@ -64,3 +61,43 @@ jobs:
...
+
+      - name: List downloaded assets
+        run: ls -la assets-to-publish
+
+      # Rename to "Publish to PyPI" when testing is finished
+      - name: Publish to TestPyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          # remove when testing is finished
+          repository-url: https://test.pypi.org/legacy/
```
mgovers, line 99: probably best to have a step and/or workflow input parameter that determines to which site to publish. My guess is that most of us will not remember that there is such a thing as a test PyPI half a year from now.

.github/workflows/ci.yml
```diff
@@ -64,3 +61,43 @@ jobs:
 
     steps:
       - run: echo "ci passed"
+
+  publish-to-pypi:
+    name: Publish to PyPI
+    runs-on: ubuntu-latest
+    needs: [ci-passed]
+    if: ${{ github.event_name == 'release' && github.event.action == 'published' }}
```
mgovers, line 69: I think this thing causes the PyPI upload to be skipped, right? https://github.com/PowerGridModel/power-grid-model/actions/runs/14312315151/job/40111394049 did not publish

image

.github/workflows/ci.yml
figueroa1395: The problem with automatic the TestPyPI step, is that in TestPyPI, the repos are cleaned up constantly. So it is very likely that if we automate this, we would still manually have to create again a dummy repo there for testing, so we must remember about PyPI because we need to set it up with trusted publisher again.

On the other hand, if I automate it in such a way that we can check a box to send to TestPyPI when manually triggering the workflow, maybe if the TestPyPI repo isn't there, a nice error would be raised reminding the dev to do so. 

I won't resolve and leave for the end as a possible nice to have. I will also add some comments.

.github/workflows/build-test-release.yml
figueroa1395: I don't think so actually. The GitHub release is based on successful "local" build and testing and the "publishing" is done locally as well (local as in within a single GitHub environment). On the contrary, the publishing to PyPI involves fully third party publishing and is basically done only once everything else has succeeded.

.github/workflows/ci.yml
figueroa1395: I know, I was trying to "replicate" more or less the actual behavior of the action when in its own workflow, but I guess it is more important right now to figure out whether the publishing can be done or not. So I'll change that to make sure we can actually publish.

figueroa1395:
[5e797e2](https://github.com/PowerGridModel/power-grid-model/pull/944/commits/5e797e2a57b44508716ee3cd90edf32099719f1d) proved that this experiment works. The [manual CI run](https://github.com/PowerGridModel/power-grid-model/actions/runs/14327276205) was successful and the project was published to [TestPyPI](https://test.pypi.org/project/power-grid-model/1.10.81a112406191486/) with trusted publisher.

What is left is to separate the workflow back again to its own dedicated one and resolve comments regarding implementation. But the investigation is concluded and the conclusion is that we can easily migrate.

I will continue with this and remove from draft after discussions with the team on the next/final step of this implementation.

.github/workflows/build-test-release.yml
mgovers: but on the other hand, every github release is usually associated with a PyPI release

figueroa1395:
Testing via [independent workflow](https://github.com/PowerGridModel/power-grid-model/actions/runs/14403952882/job/40395790160) triggered manually was successful. The pre-release (made for testing purposes), was uploaded to [TestPyPI](https://test.pypi.org/manage/project/power-grid-model/release/1.10.82a1400959606/). Now I will remove from `test-workflow.yml` and add to `publish-pypi.yml`.

.github/workflows/build-test-release.yml
TonyXiang8787: @mgovers 

You can't in the same action. The PyPI publishing has to be in an independent workflow, instead of re-usable workflow.

figueroa1395: I still don't think both actions should be put together. Maybe splitting up `build-test-release.yml` could be something to further discuss. In any case, if you feel strongly about it, we can have an offline discussion on Monday about this.

TonyXiang8787: What do you mean should be put together? It literally cannot. That's the requirement of the PyPI publish action.

figueroa1395:
If it is relevant, I added a cleaner/shorter `pre-release` tag as well, that includes RC (release candidate) and the commit hash. See [02d5412](https://github.com/PowerGridModel/power-grid-model/pull/944/commits/02d5412d8fb498c7b7510c7714a90bd74e1b68f9)

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,55 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
```
TonyXiang8787, line 10: You can't do that. If you initiate the workflow by dispatch, there will be no `github.event.release.tag_name`. Therefore the steps below will fail.

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,55 @@
...
+  publish:
+    name: Publish to PyPI
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Download assets from GitHub release
+        uses: robinraju/release-downloader@v1
+        with:
+          repository: ${{ github.repository }}
+          latest: true
```
TonyXiang8787, line 26: How can you both download the latest release and a specific tag? These two options are conflict right?

.github/workflows/build-test-release.yml
figueroa1395: The PyPI publish action needs to be a workflow on its own, yes, but it can have other jobs that do other stuff in that same workflow. My understanding of what @mgovers wants to achieve is to have the github release job as part of the PyPI publishing workflow. That should be doable, but I don't think we should.

.github/workflows/build-test-release.yml
```diff
@@ -324,19 +317,18 @@ jobs:
...
       - name: Get tag
         id: tag
-        run: echo "tag=v$(cat PYPI_VERSION)" >> $GITHUB_OUTPUT
+        # pre-release tag is "RC-commit-hash" (release candidate)
+        run: |
+          if [ "${GITHUB_REF}" != 'refs/heads/main' ]; then
+            echo "tag=RC-$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+          else
+            echo "tag=v$(cat PYPI_VERSION)" >> $GITHUB_OUTPUT
+          fi
```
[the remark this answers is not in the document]
figueroa1395: We can do it if we need some testing (for instance what I did in this PR) if we manually trigger `ci.yml` and check the box for pre-release. So it is useful for testing phases. And I took the liberty to change this tag because it is more readable, as the default is very long and doesn't give much information. In any case, this can easily be reverted if it is not desirable.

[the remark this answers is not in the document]
TonyXiang8787: I do not agree to create a GitHub release and PyPI upload for testing purpose at all. So also remove the `pre-release` option in the `ci.yml`.

If we want the CI to build something for testing purpose, we trigger the CI but no GitHub release or PyPI upload. We just download the CI built artifact (or ask our beta user to download from there) to test.

.github/workflows/publish-pypi.yml
figueroa1395: Since I define `latest: true` earlier, `tag` is then optional and the latest tag is taken. This way, we can use this action via `workflow_dispatch` which leaves `tag` empty, which is how I was able to publish [it](https://test.pypi.org/manage/project/power-grid-model/release/1.10.82a1400959606/).

Then, having `tag: ${{ github.event.release.tag_name }}` is probably not necesary and we can just leave `latest: true`.

.github/workflows/publish-pypi.yml
figueroa1395: See comment above.

But to add, not really, they can live together as `tag` overrides `latest`, but when `tag` can't be obtained, then `latest` is used (such as during `workflow_dispatch`).

.github/workflows/publish-pypi.yml
figueroa1395: I removed `tag` and only kept `latest: release`, this results in the intended behaviour and we don't run into this problem when triggering a workflow manually.

figueroa1395:
Pre-releases created while testing were cleaned up: https://github.com/PowerGridModel/power-grid-model/releases

.github/workflows/build-test-release.yml
mgovers: another benefit of moving the release logic to the parent workflow reason would be that we then can remove the `write` permissions from this workflow file, but i get your point.

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,54 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+name: Publish to PyPI
+
+on:
+  release:
```
mgovers, line 8: ohhh i understand. ok this makes sense

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,54 @@
...
+    types: [published]
+  workflow_dispatch:
+
+permissions:
+  contents: write
```
mgovers, line 13: does it really need to have `contents: write` permissions?

mgovers (commented):
i think it looks pretty good

.github/workflows/publish-pypi.yml
figueroa1395: Yes, it is explicitly mentioned in PyPI [docs](https://docs.pypi.org/trusted-publishers/using-a-publisher/). It is imperative that this permission is enabled.

.github/workflows/build-test-release.yml
figueroa1395: Since I don't think there are any strong opinions left, I am resolving.

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Download assets from GitHub release
+        uses: robinraju/release-downloader@v1
+        with:
+          repository: ${{ github.repository }}
+          latest: true
+          # also download assets from pre-releases
+          preRelease: true
```
TonyXiang8787, line 28: we should not download pre-release right?

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,54 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
```
mgovers, line 10: wouldn't it make sense to give the opportunity to `workflow_dispatch` with some kind of tag/branch/release (and make sure the action fails if the package was already uploaded)

pyproject.toml
```diff
@@ -20,15 +20,14 @@ authors = [
...
-license = { text = "MPL-2.0" }
+license = "MPL-2.0"
 classifiers = [
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: C++",
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
     "Intended Audience :: Science/Research",
-    "License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)",
```
mgovers, line 31: why?

.github/workflows/build-test-release.yml
```diff
@@ -44,12 +39,13 @@ jobs:
           python-version: "3.12"
 
       - name: Set PyPI Version
-        run: |
-          pip install requests build
-          python set_pypi_version.py
+        uses: PowerGridModel/pgm-version-bump@main
```
mgovers, line 42: it's probably a good idea to just create a release and call it `@v1` cause `main` is always tricky

pyproject.toml
TonyXiang8787: Deprecation of setuptools. This is the new way of specifying license.

.github/workflows/build-test-release.yml
figueroa1395: How so? Specially if the version bump repo would be rarely touched?

.github/workflows/publish-pypi.yml
figueroa1395: I wouldn't remove it yet, as we still have the pre-release option for the `ci-build` and `build-test-release` actions. I would leave this there until we remove the pre-release option.

.github/workflows/publish-pypi.yml
figueroa1395: I believe that the publishing actions fails by design if the package to be uploaded is a duplicate, so that is already covered. Regarding this:

> wouldn't it make sense to give the opportunity to `workflow_dispatch` with some kind of tag/branch/release

It is not clear to me what you are suggesting.

.github/workflows/build-test-release.yml
mgovers: yes, but IF it happens, we want to be there for it.

.github/workflows/publish-pypi.yml
mgovers: it would be nice to have the RC versions to also be increasing monotonically in some kind of way. i may create a branch `release/v1.11.0-rc` and then in its current version, i have no control over the specific naming of the release candidates. it would be nice to do something like `v1.11.0-rc1-HASH`

.github/workflows/build-test-release.yml
TonyXiang8787: I doubt very much the benefit here. Using main branch is the standard way for small composite actions. Just have a look at other small action examples. 

Basically you are living at head. If we want to change the the version bump action, we can play with it in the feature branch.

I do not see the risk of using main here.

.github/workflows/publish-pypi.yml
TonyXiang8787: The combination is the problem. Now you always retrieve the latest release including pre-release.

.github/workflows/publish-pypi.yml
TonyXiang8787: The version bump for feature/release branch includes the build number in the version, then the commit hash. So it is already increasing.

.github/workflows/publish-pypi.yml
figueroa1395: Why is that a problem? Usually we don't have pre-releases and if we do is for a very specific reason in which we probably would want to retrieve that one.

figueroa1395: Convinced.

.github/workflows/publish-pypi.yml
figueroa1395: Resolving since I don't think there is anything else to do here.

.github/workflows/build-test-release.yml
figueroa1395: Given that we don't have any automated action inside the action repo itself that may update something automatically and possibly break the action, I'd say in this case using @main is fine. However, if we were to change that, then we should either @v1 or even @commit-sha if we want an even simpler version.

@mgovers Is it okay to resolve this, or do you have a strong opinion? Creating the tag on the other repo isn't too much work either, so we might as well do it if it is that way.

TonyXiang8787: Creating a tag in the other repo means we also need to retarget the tag when we have new changes. This adds extra step.

mgovers: OK with resolving this

figueroa1395: Let's then leave as is for now.

PR #951, opened 2025-04-10
Iterative Linear State Estimation: power sensor accuracy

mgovers:
The Iterative Linear State Estimation algorithm assumes that linearized rescaling power to current keeps the variances the same. This assumption is true when node voltages are close to `1 p.u.`. However, when voltages may vary significantly, this assumption becomes increasingly bad, and may lead to incorrect weighing of different measurements, and may even lead to incorrect answers.

This becomes especially important when looking at grids with both power and current sensors.

TBD (do-not-merge label added):
* [ ] the new implementation causes the prefactorization to happen on a per-iteration level, instead of on a global level.
  * [ ] What is the computational impact of that? (benchmark)
  * [ ] Is this change desirable?
* [ ] the new implementation may lead to slightly different answers from the original implementation
  * [ ] What is the user impact of that?


power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -204,4 +204,51 @@ template <symmetry_tag sym_type> struct PolarComplexRandVar {
...
+template <symmetry_tag sym, template <symmetry_tag> typename RandVarType>
+    requires is_in_list_c<RandVarType<sym>, UniformComplexRandVar<sym>, IndependentComplexRandVar<sym>>
+inline auto scale(RandVarType<sym> const& var, DoubleComplex const& scale_factor) {
+    return RandVarType<sym>{.value = var.value * scale_factor, .variance = var.variance * abs2(scale_factor)};
+}
+
+inline auto scale(IndependentComplexRandVar<asymmetric_t> const& var, ComplexValue<asymmetric_t> const& scale_factor) {
+    return IndependentComplexRandVar<asymmetric_t>{.value = var.value * scale_factor,
+                                                   .variance = var.variance * abs2(scale_factor)};
+}
```
mgovers, line 231: TODO: try to combine into single function

mgovers: not possible without weird long `requires` clauses. The thing is that the following table needs to be captured in a single requires clause, which is highly non-trivial:

| `RandVarType`\`decltype(scale_factor)` | `RealValue<symmetric_t>` | `RealValue<asymmetric_t>` | ComplexValue<symmetric_t> | ComplexValue<asymmetric_t> |
| --- | --- | --- | --- | --- |
| UniformRealRandVar<symmetric_t> | ✅ | ❌ | ❌ | ❌ |
| UniformRealRandVar<asymmetric_t> | ✅ | ✅ | ❌ | ❌ |
| IndependentRealRandVar<symmetric_t> | ✅ | ❌ | ❌ | ❌ |
| IndependentRealRandVar<asymmetric_t> | ✅ | ✅ | ❌ | ❌ |
| UniformComplexRandVar<symmetric_t> | ✅ | ❌ | ✅ | ❌ |
| UniformComplexRandVar<asymmetric_t> | ✅ | ✅ | ✅ | ✅ |
| IndependentComplexRandVar<symmetric_t> | ✅ | ❌ | ✅ | ❌ |
| IndependentComplexRandVar<asymmetric_t> | ✅ | ✅ | ✅ | ✅ |

figueroa1395:
This one can now be closed after offline conclusions, right? And then add the documentation needed to the issue #547, correct?

TonyXiang8787:
@mgovers based on the decision. We should close this PR, right?

mgovers:
> This one can now be closed after offline conclusions, right? And then add the documentation needed to the issue #547, correct?

> @mgovers based on the decision. We should close this PR, right?

Indeed. To summarize the discussions:

* [x] The variances are indeed not one-to-one scaled.
* [x] The power sensors used in ILSE are an approximation.
  * [x] Uncertainties/sigmas of the measurements are typically best-effort values.
  * [x] Linearization of the measurements and error margins are merely an approximation in itself.
  * [x] Interpretation of power measurements as current measurements is also an assumption.
* [x] It is probably not a huge issue that the power sensor variances are not correctly handled within the domain in which linearization is valid and in which we are operating.
* [x] The inner-loop factorization of the matrix would come at a significant computational cost.

Based on these arguments, we decided to go with the following:

* [x] Present ILSE as an approximate method and properly document the assumptions that are made.

The following options were also considered but rejected in favor of the chosen approach.

~* [ ] Do the change and present it as an improvement to the ILSE~
~* [ ] Do the change and present it as a bugfix~
~* [ ] Add the change as a separate calculation method to the original ILSE~

In conclusion, this PR will be closed and superseded by a follow-up documentation PR.

PR #956, opened 2025-04-16
Fix MacOS build for Xcode 16.3

figueroa1395:
As mentioned in [Xcode 16.3 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-16_3-release-notes): _The base template for std::char_traits has been removed. If you are using std::char_traits with types other than char, wchar_t, char8_t, char16_t, char32_t or a custom character type for which you specialized std::char_traits, your code will stop working. The Standard does not mandate that a base template is provided, and such a base template is bound to be incorrect for some types, which could previously cause unexpected behavior while going undetected._

This affects us indirectly as we use [nlohmann/json](https://github.com/nlohmann/json/tree/v3.12.0) and they still use `std::char_traits`, resulting in failed compilation attempt (only for MacOS devs with Xcode16.3) due to the usage of `std::vector<std::byte>` as an argument of `nlohmann::ordered_json::from_msgpack()` in our tests.

In this PR, we modify our tests such that this is not an issue anymore.

Note: There was a recent release of the [library](https://github.com/nlohmann/json/releases/tag/v3.12.0), but it didn't address this issue.

### DIFF
--- a/tests/native_api_tests/test_api_serialization.cpp
+++ b/tests/native_api_tests/test_api_serialization.cpp
@@ -81,9 +81,11 @@ TEST_CASE("API Serialization and Deserialization") {
             Serializer msgpack_serializer{dataset, 1};
 
             SUBCASE("Round trip") {
-                std::vector<char> msgpack_data{};
+                std::vector<std::byte> msgpack_data{};
+                std::span<std::underlying_type_t<std::byte>> const msgpack_data_view{
+                    reinterpret_cast<std::underlying_type_t<std::byte>*>(msgpack_data.data()), msgpack_data.size()};
                 msgpack_serializer.get_to_binary_buffer(0, msgpack_data);
-                auto const json_document = nlohmann::ordered_json::from_msgpack(msgpack_data);
+                auto const json_document = nlohmann::ordered_json::from_msgpack(msgpack_data_view);
                 auto const json_result = json_document.dump(-1);
                 CHECK(json_result == json_data);
             }
--- a/tests/native_api_tests/test_api_serialization.cpp
+++ b/tests/native_api_tests/test_api_serialization.cpp
@@ -82,9 +82,9 @@ TEST_CASE("API Serialization and Deserialization") {
 
             SUBCASE("Round trip") {
                 std::vector<std::byte> msgpack_data{};
+                msgpack_serializer.get_to_binary_buffer(0, msgpack_data);
                 std::span<std::underlying_type_t<std::byte>> const msgpack_data_view{
                     reinterpret_cast<std::underlying_type_t<std::byte>*>(msgpack_data.data()), msgpack_data.size()};
-                msgpack_serializer.get_to_binary_buffer(0, msgpack_data);
                 auto const json_document = nlohmann::ordered_json::from_msgpack(msgpack_data_view);
                 auto const json_result = json_document.dump(-1);
                 CHECK(json_result == json_data);
--- a/tests/native_api_tests/test_api_serialization.cpp
+++ b/tests/native_api_tests/test_api_serialization.cpp
@@ -81,7 +81,7 @@ TEST_CASE("API Serialization and Deserialization") {
             Serializer msgpack_serializer{dataset, 1};
 
             SUBCASE("Round trip") {
-                std::vector<std::byte> msgpack_data{};
+                std::vector<char> msgpack_data{};
                 msgpack_serializer.get_to_binary_buffer(0, msgpack_data);
                 auto const json_document = nlohmann::ordered_json::from_msgpack(msgpack_data);
                 auto const json_result = json_document.dump(-1);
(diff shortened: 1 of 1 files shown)


tests/native_api_tests/test_api_serialization.cpp
```diff
@@ -82,8 +82,10 @@ TEST_CASE("API Serialization and Deserialization") {
 
             SUBCASE("Round trip") {
                 std::vector<std::byte> msgpack_data{};
+                std::span<std::underlying_type_t<std::byte>> const msgpack_data_view{
+                    reinterpret_cast<std::underlying_type_t<std::byte>*>(msgpack_data.data()), msgpack_data.size()};
                 msgpack_serializer.get_to_binary_buffer(0, msgpack_data);
-                auto const json_document = nlohmann::ordered_json::from_msgpack(msgpack_data);
+                auto const json_document = nlohmann::ordered_json::from_msgpack(msgpack_data_view);
                 auto const json_result = json_document.dump(-1);
```
mgovers, line 89: can you please double-check that the latest `nlohmann_json` version is used?

mgovers (approved):

TonyXiang8787:
@figueroa1395 did you raise issue at `nlohmann/json` board?

figueroa1395:
> @figueroa1395 did you raise issue at `nlohmann/json` board?

I haven't, but I will. I will post here when I do.

mgovers:
Converted back to Draft because not all issues have been resolved yet. Sustainable solution will take a bit longer.

figueroa1395:
Issue created in https://github.com/nlohmann/json/issues/4756

TonyXiang8787:
@mgovers @figueroa1395 

If I understand correctly, we just need to change from `std::vector<std::byte>` to `std::vector<char>` in the following two line, the problem is going away right?

https://github.com/PowerGridModel/power-grid-model/blob/fc123b369cc7de95094da4da7f8c132ee8979a6d/tests/native_api_tests/test_api_serialization.cpp#L84

https://github.com/PowerGridModel/power-grid-model/blob/fc123b369cc7de95094da4da7f8c132ee8979a6d/tests/native_api_tests/test_api_serialization.cpp#L71

figueroa1395:
> @mgovers @figueroa1395
> 
> If I understand correctly, we just need to change from `std::vector<std::byte>` to `std::vector<char>` in the following line, the problem is going away right?
> 
> https://github.com/PowerGridModel/power-grid-model/blob/fc123b369cc7de95094da4da7f8c132ee8979a6d/tests/native_api_tests/test_api_serialization.cpp#L84

@TonyXiang8787 Yes, and that was my first quick and naive solution. However, if we just do that, I think we would be taking a step back. At the CPP API level we should be able to use `std::byte` when handling raw data, not just go back to `char` which can also be used for characters or even numbers.

Apparently there is someone interested in implementing a PR to solve my issue in nlohmman/json, so we can just wait a bit more there. In any case, I'll be giving my input as well. I just didn't have the time to do the PR myself.

TonyXiang8787:
> > @mgovers @figueroa1395
> > If I understand correctly, we just need to change from `std::vector<std::byte>` to `std::vector<char>` in the following line, the problem is going away right?
> > https://github.com/PowerGridModel/power-grid-model/blob/fc123b369cc7de95094da4da7f8c132ee8979a6d/tests/native_api_tests/test_api_serialization.cpp#L84
> 
> @TonyXiang8787 Yes, and that was my first quick and naive solution. However, if we just do that, I think we would be taking a step back. At the CPP API level we should be able to use `std::byte` when handling raw data, not just go back to `char` which can also be used for characters or even numbers.
> 
> Apparently there is someone interested in implementing a PR to solve my issue in nlohmman/json, so we can just wait a bit more there. In any case, I'll be giving my input as well. I just didn't have the time to do the PR myself.

@figueroa1395 @mgovers 
We do not have to change the CPP API. But just cast the pointer to `char const*`. It will then solve the problem. So change the lines:

https://github.com/PowerGridModel/power-grid-model/blob/edc1018f372234373a647e4693b6a2c835a8cbe3/tests/native_api_tests/test_api_serialization.cpp#L84-L86

To:

```c++
                std::vector<std::byte> msgpack_data{};
                msgpack_serializer.get_to_binary_buffer(0, msgpack_data);
                auto const char_start = reintepret_cast<char const*>(msgpack_data.data());
                auto const char_end = char_start + msgpack_data.size();
                auto const json_document = nlohmann::ordered_json::from_msgpack(char_start, char_end);
```

TonyXiang8787 (approved):

figueroa1395:
@mgovers @TonyXiang8787 FYI: The [issue](https://2025.dddeurope.com) I created relevant to this has already been closed as completed via https://github.com/nlohmann/json/pull/4760. Now we have to wait until they make their next release. Since that may take very long, should I create a simple issue here to keep track of it for once it gets released on nlohmann/json's side?

PR #965, opened 2025-04-23
patch automatic publish PyPI

mgovers:
Very unfortunate patch to #944 

This seems to be caused by the fact that GitHub explicitly excludes releases created by GitHub Actions workflows from triggering `on: release` events. See also https://github.com/orgs/community/discussions/25281

NOTE: this is probably not a sustainable solution.


.github/workflows/ci.yml
```diff
@@ -61,3 +61,45 @@ jobs:
 
     steps:
       - run: echo "ci passed"
+
+  publish:
+    name: Publish to PyPI
+    runs-on: ubuntu-latest
+    needs: build-test-release
```
figueroa1395, line 68:
```suggestion
    needs: ci-passed
```

figueroa1395 (commented):
So basically, the original failed because `on-release` isn't actually `on-release` if called from an another workflow? Hence making it part of the "main" workflow fixes it? That unless we want to go with the official way to fix this (which we shouldn't) which is to [use personal tokens to access](https://github.com/orgs/community/discussions/25281#discussioncomment-3247262)? Doesn't make any sense to be honest, and worse that it isn't documented beyond threads were people complain... Besides the small suggestion, I am on board with this.

.github/workflows/ci.yml
mgovers: i had this officially, but then i realized that the original version depended on ONLY the `Create GitHub Release` job, which is part of the `build-test-release` workflow. Fine with changing, but it would change behavior

figueroa1395: I'm on board with this chance since we are already in this workflow.

mgovers:
Added do-not-merge label as input is desired

mgovers:
> @mgovers is this the only way? Can we still put the upload-pypi in a separate workflow. And in the CI-passed job of the main workflow, we trigger the other workflow by some means.

That would indeed be the alternative i mentioned in the offline thread. The way to do that would be to add a `repository_dispatch` trigger in the `publish-pypi.yml` combined with adding something like this to the `build-test-release.yml`

```yml
      - name: Trigger PyPI publish event
        if: (inputs.create_release)
        run: |
          gh api \
            --method POST \
            -H "Accept: application/vnd.github+json" \
            -H "X-GitHub-Api-Version: 2022-11-28" \
            /repos/PowerGridModel/power-grid-model/dispatches \
            -f "event_type=pypi-publish" -F "client_payload[unit]=false" -F "client_payload[integration]=true"
```

However, again, we cannot test this unless we do a 2-stage merge so that the `repository_dispatch` trigger is in `main`

TonyXiang8787:
> > @mgovers is this the only way? Can we still put the upload-pypi in a separate workflow. And in the CI-passed job of the main workflow, we trigger the other workflow by some means.
> 
> That would indeed be the alternative i mentioned in the offline thread. The way to do that would be to add a `repository_dispatch` trigger in the `publish-pypi.yml` combined with adding something like this to the `build-test-release.yml`
> 
> ```yaml
>       - name: Trigger PyPI publish event
>         if: (inputs.create_release)
>         run: |
>           gh api \
>             --method POST \
>             -H "Accept: application/vnd.github+json" \
>             -H "X-GitHub-Api-Version: 2022-11-28" \
>             /repos/PowerGridModel/power-grid-model/dispatches \
>             -f "event_type=pypi-publish" -F "client_payload[unit]=false" -F "client_payload[integration]=true"
> ```
> 
> However, again, we cannot test this unless we do a 2-stage merge so that the `repository_dispatch` trigger is in `main`

Do we have an existing action to do that in the market place?

figueroa1395:
> @mgovers is this the only way? Can we still put the upload-pypi in a separate workflow. And in the CI-passed job of the main workflow, we trigger the other workflow by some means.

The problem is that the publishing action to PyPI can't be triggered from another workflow. So if we want this publishing workflow to be standalone, we would have to be able to trigger it within the workflow itself with something different than `on: release`. However, maybe we can do something like:
```
name: Publish to PyPI
on:
  workflow_run:
    workflows: ["Wathever"]  # name of the CI finished workflow for instance (we need to add a name to it)
    types:
      - completed
  workflow_dispatch:
  

jobs:
  publish:
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    ...
```

would something like this work instead?

mgovers:
> > > @mgovers is this the only way? Can we still put the upload-pypi in a separate workflow. And in the CI-passed job of the main workflow, we trigger the other workflow by some means.
> > 
> > 
> > That would indeed be the alternative i mentioned in the offline thread. The way to do that would be to add a `repository_dispatch` trigger in the `publish-pypi.yml` combined with adding something like this to the `build-test-release.yml`
> > ```yaml
> >       - name: Trigger PyPI publish event
> >         if: (inputs.create_release)
> >         run: |
> >           gh api \
> >             --method POST \
> >             -H "Accept: application/vnd.github+json" \
> >             -H "X-GitHub-Api-Version: 2022-11-28" \
> >             /repos/PowerGridModel/power-grid-model/dispatches \
> >             -f "event_type=pypi-publish" -F "client_payload[unit]=false" -F "client_payload[integration]=true"
> > ```
> >   
> > However, again, we cannot test this unless we do a 2-stage merge so that the `repository_dispatch` trigger is in `main`
> 
> Do we have an existing action to do that in the market place?

maybe, i haven't been able to investigate this yet due to other priorities, but i didn't find anything other than this (github official way) in my quick 1min search.

TonyXiang8787:
@mgovers @figueroa1395 

I think the solution would be, as similar to other organizations such as `conda-forge`, to create service bot account. We can then create personal access token for the service account and use in the release step of main CI. In that way the release is not created by `GITHUB_TOKEN` but by our own token. The publish PyPI can then be triggerd. This is also suggested in official GitHub Actions documentation:

softprops/action-gh-release

figueroa1395:
> @mgovers @figueroa1395
> 
> I think the solution would be, as similar to other organizations such as `conda-forge`, to create service bot account. We can then create personal access token for the service account and use in the release step of main CI. In that way the release is not created by `GITHUB_TOKEN` but by our own token. The publish PyPI can then be triggerd. This is also suggested in official GitHub Actions documentation:
> 
> https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow
> 
> It also mentioned to use GitHub App to generate a token.

In summary, create an account to be used as a bot, make it a maintainer of the pgm projects, and then use its personal GitHub access token to run the action instead of the workflow token? It doesn't look like it is too hard, but probably needs a bit of investigation and some trial an error.

.github/workflows/ci.yml
figueroa1395: @mgovers since we don't want to break anything, let's just keep it as it was and scratch this suggestion.

.github/workflows/ci.yml
```diff
@@ -61,3 +61,45 @@ jobs:
...
+    if: (github.event_name == 'workflow_dispatch') || github.event_name == 'push'
+
+    steps:
+      - name: Download assets from GitHub release
+        uses: robinraju/release-downloader@v1
+        with:
+          repository: ${{ github.repository }}
+          # download the latest release
+          latest: true
+          # don't donwload pre-releases
```
mgovers, line 78:
```suggestion
          # don't download pre-releases
```

mgovers:
> Added do-not-merge label as input is desired

do-not-merge label removed cfr. offline discussions: skip publishing the last 2 patch releases (no important changes to them) and proceed with the lastest

.github/workflows/ci.yml
```diff
@@ -61,3 +61,45 @@ jobs:
...
+
+    steps:
+      - name: Download assets from GitHub release
+        uses: robinraju/release-downloader@v1
+        with:
+          repository: ${{ github.repository }}
+          # download the latest release
+          latest: true
+          # don't download pre-releases
+          preRelease: false
```
TonyXiang8787, line 79: We decided to be able to upload a alpha version right? So this has to be changed back to true?

mgovers: i'd consider that out of scope of this PR. currently we are unable to upload to PyPI at all (not even `main`). Let's first fix that and then we can do follow-ups

TonyXiang8787 (approved):

PR #962, opened 2025-04-22
add current sensor + power sensor variance diff to ILSE docs

mgovers:
* [x] Add current sensor to component docs
  * [x] Add experimental feature note
* [x] Add current sensor to observability docs
* [x] Add current sensor to ILSE docs
* [x] Explicitly exclude current sensor from supported sensors in NRSE docs
* [x] Document decision made in #951 .

Relates to #547 .
Part of #764 .
Do not merge before #964 is merged into `main`

PR #975, opened 2025-04-29
voltage sensor angle residual mod 2pi

mgovers:
voltage angle residuals currently can be in the range `[-2pi, 2pi]`. This PR restricts it to the range `(-pi, pi]`, effectively doing `mod 2pi`. This is also cfr. #962

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/component/voltage_sensor.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/component/voltage_sensor.hpp
@@ -151,7 +151,7 @@ template <symmetry_tag sym> class VoltageSensor : public GenericVoltageSensor {
         } else {
             value.u_residual = (real(u1_measured) - cabs(u)) * u_rated_;
         }
-        value.u_angle_residual = arg(u1_measured) - arg(u);
+        value.u_angle_residual = arg(ComplexValue<symmetric_t>{exp(1.0i * (arg(u1_measured) - arg(u)))});
         return value;
     }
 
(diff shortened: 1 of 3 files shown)


figueroa1395 (approved):
Only a small related-ish comment. Rest looks good.

> 
> NOTE: this PR changes the PGM behavior slightly, but it should only be an improvement to the output for users.

Also, I see all the validation tests remained intact, so this should be okay. If anything, users would now face corrections.

power_grid_model_c/power_grid_model/include/power_grid_model/component/voltage_sensor.hpp
```diff
@@ -151,7 +151,7 @@ template <symmetry_tag sym> class VoltageSensor : public GenericVoltageSensor {
         } else {
             value.u_residual = (real(u1_measured) - cabs(u)) * u_rated_;
         }
-        value.u_angle_residual = arg(u1_measured) - arg(u);
+        value.u_angle_residual = arg(ComplexValue<symmetric_t>{exp(1.0i * (arg(u1_measured) - arg(u)))});
```
figueroa1395, line 154: Just thought about something else that I had mentioned before. Do we want to make this an actual function (maybe `bounded_arg` or something like that) that lives in `common.hpp` and we just test the actual function and re-use it here and in the current sensor implementation?

power_grid_model_c/power_grid_model/include/power_grid_model/component/voltage_sensor.hpp
```diff
@@ -160,7 +160,7 @@ template <symmetry_tag sym> class VoltageSensor : public GenericVoltageSensor {
         value.id = id();
         value.energized = 1;
         value.u_residual = (u_measured_ - cabs(u)) * u_rated_ / sqrt3;
```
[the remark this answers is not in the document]
mgovers: the `has_angle` is a significant difference though. that's a behavioral change

power_grid_model_c/power_grid_model/include/power_grid_model/component/voltage_sensor.hpp
mgovers: not sure... it's not possible to create it efficiently (requires exponentiation + logarithm), so we shouldn't use it without thinking about it properly

mgovers: done

TonyXiang8787 (approved):

Issue #977, opened 2025-04-30
[Feature] PGM Meet-up 2025-05-21 Hackatons


mgovers:
This is the parent issue for the Hackatons planned for the [2025-05-21 PGM meet-up](https://community.linuxfoundation.org/events/details/lfhq-lf-energy-presents-6th-power-grid-model-meetup-2025/).

Please register via the [event page](https://community.linuxfoundation.org/events/details/lfhq-lf-energy-presents-6th-power-grid-model-meetup-2025/) if you are interested in joining. While most of the hackaton topics will be in-person only, we will facilitate some topics that can be attended remotely.

petersalemink95:
Closing, since the meet-up has been finished

PR #980, opened 2025-04-30
meet-up hackaton: update build guide

mgovers:
Relates to #977

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ include("cmake/pgm_version.cmake")
 
 project (power_grid_model VERSION ${PGM_VERSION})
 
-option(PGM_ENABLE_TESTS "Enable tests" OFF)
+option(PGM_ENABLE_DEV_BUILD "Enable developer build (e.g.: tests)" OFF)
 
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
--- a/docs/advanced_documentation/build-guide.md
+++ b/docs/advanced_documentation/build-guide.md
@@ -175,7 +175,7 @@ You can use any IDE to develop this project. As a popular cross-platform IDE, th
 
 ```{note}
 VSCode (as well as some other IDEs) does not set its own build environment itself. For optimal usage, open the folder
-using `cmake <project_dir>` from a terminal that has the environment set up. See above section for tips.
+using `code <project_dir>` from a terminal that has the environment set up. See above section for tips.
 ```
 
 ## Build Script for Linux/macOS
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ include("cmake/pgm_version.cmake")
 
 project (power_grid_model VERSION ${PGM_VERSION})
 
-option(PGM_ENABLE_TESTS "Enable tests" OFF)
+option(PGM_ENABLE_DEV_BUILD "Enable developer build (e.g.: tests)" OFF)
```
mgovers, line 14: this is a bugfix

note to the reviewer: please double-check that this is indeed the option added to the `CMakePresets.json` and that it is the one used to enable tests later in this file

docs/advanced_documentation/build-guide.md
```diff
@@ -374,9 +374,9 @@ The IDE should be able to automatically detect the Visual Studio cmake configura
   Prompt`).
 ```
 
-## Example Setup for macOS (Big Sur)
```
mgovers, line 377: we sent out a notice when upgrading macos that we officially don't support Big Sur anymore

docs/advanced_documentation/build-guide.md
```diff
@@ -427,7 +427,7 @@ As an example, go to the root folder of repo. Use the following command to build
 ./build.sh -p <preset>
 ```
 
-To list the available presets, run `./build.sh -h`.
+To list the available presets, run `cmake --list-presets`.
```
mgovers, line 430: this is a fix

docs/advanced_documentation/build-guide.md
```diff
@@ -175,7 +175,7 @@ You can use any IDE to develop this project. As a popular cross-platform IDE, th
 
 ```{note}
 VSCode (as well as some other IDEs) does not set its own build environment itself. For optimal usage, open the folder
-using `cmake <project_dir>` from a terminal that has the environment set up. See above section for tips.
+using `code <project_dir>` from a terminal that has the environment set up. See above section for tips.
```
mgovers, line 178: this is a fix

CMakeLists.txt
TonyXiang8787: This option is not used at all in any places. So just remove this line.

mgovers: It is used by CMakePresets.json to enable dev build. Without this, only the C API is built by default, not the tests. With it enabled, the tests are also built. This was introduced to allow a better default experience for people who want to build from source with a minimal set of features enabled.

This bugfix fixes the mismatch between the CMake option and the actual CMake variable used in CMakeLists.txt to enable dev build

TonyXiang8787: I mean, this particular `option` line is not used. Without this, it still works as intended. When user or preset defines `PGM_ENABLE_DEV_BUILD`, developer build is enabled, otherwise only the C-API.

mgovers: still, it is probably a good idea to tell users that and how it is possible to enable the DEV build without digging through the code base. `option` is designed specifically for that. From a best-practice point of view, I do think it's a good idea to keep it. What is your reasoning for explicitly removing it?

TonyXiang8787: You are correct. I did not know the meaning of `option` in `cmake`.

PR #988, opened 2025-05-08
Re add isnan internal import for PGM-IO

nitbharambe:
**Relates to**: https://github.com/PowerGridModel/power-grid-model-io/issues/297

### DIFF
--- a/src/_utils.py
+++ b/src/_utils.py
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+"""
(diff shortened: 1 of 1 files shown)


src/_utils.py
mgovers, line 1: Do you think it makes sense to add a deprecation warning? See the way we do such warnings we also used for some of the old data import functionality. It's not strictly necessary because it concerns a private module import so it's optional

mgovers (approved):

src/_utils.py
petersalemink95: Indeed, conclusion was users should not use it in the first place - apparently we did ourselves in pgm-io. Nevertheless, deprecation warning would not be necessary

petersalemink95 (approved):

PR #996, opened 2025-05-15
Current sensor: validation

zhen0427:
Update the data validator so it checks current sensors. Similar checks as for power sensor

### Checks try to be implemented

- [x] MissingValueError: the following attributes are required
  - [x] id
  - [x] measured_object
  - [x] measured_terminal_type
  - [x] angle_measurement_type
  - [x] i_sigma
  - [x] i_angle_sigma
  - [x] i_measured
  - [x] i_angle_measured
- [x] All attributes are bigger than zero
- [x] All valid enum values
- [x] Current sensor can only measure branch and branch3 type of component.
  - [x] MeasuredTerminalType should therefore be branch_from, branch_to, branch3_1, branch3_2, branch3_3
  - [x] MeasuredTerminalType should only have `branch3_*` for 3w transformer and `branch_*` for regular branches
- [x] current sensor on link is not supported
- [x] angle measurement type should be local_angle or global_angle
- [x] the local_angle and global_angle cannot be simultaneously measured on the same terminal
  - [x] implementation
  - [x] thoroughly test the implementation
- [x] if global angle provided, then there needs to be a voltage angle measurement in the dataset
- [x] It is not possible to add both a [power sensor](https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html#generic-current-sensor) and a [current sensor](https://power-grid-model.readthedocs.io/en/stable/user_manual/calculations.html#./components.mdgeneric-current-sensor) to the same terminal of the same component.

Check the logic of validation is full-rounded


src/power_grid_model/validation/_validation.py
```diff
@@ -962,6 +982,63 @@ def validate_generic_power_sensor(data: SingleDataset, component: ComponentType)
     return errors
 
 
+def validate_generic_current_sensor(data: SingleDataset, component: ComponentType) -> list[ValidationError]:
+
+    errors = validate_base(data, component)
+    errors += _all_greater_than_zero(data, component, "i_sigma")
+    errors += _all_greater_than_zero(data, component, "i_angle_sigma")
+    errors += _all_valid_enum_values(data, component, "measured_terminal_type", MeasuredTerminalType)
```
nitbharambe, line 990: We would need a separate error InvalidEnumSubsetError? because we dont have current sensor on all terminal types.

tests/unit/validation/test_input_validation.py
```diff
@@ -558,6 +587,78 @@ def test_validate_input_data_sym_calculation(input_data):
...
+                "line",
+                "asym_line",
+                "generic_branch",
+                "transformer",
+                "three_winding_transformer",
+            ],
+        )
+        in validation_errors
+    )
+    assert NotGreaterThanError("sym_current_sensor", "i_sigma", [9, 10], 0) in validation_errors
```
figueroa1395, line 608: Nitpick and probably out of scope, but this and many others can be replaced with `ComponentType.component`.

tests/unit/validation/test_input_validation.py
```diff
@@ -558,6 +587,78 @@ def test_validate_input_data_sym_calculation(input_data):
...
+            "measured_object",
+            [7, 10],
+            ["line", "asym_line", "generic_branch", "transformer"],
+            {"measured_terminal_type": MeasuredTerminalType.branch_to},
+        )
+        in validation_errors
+    )
+    assert (
+        InvalidEnumValueError("sym_current_sensor", "measured_terminal_type", [9], MeasuredTerminalType)
+        in validation_errors
```
figueroa1395, line 622: Related to this one, maybe we want to make one of the `measured_terminal_type`s to be within `MeasuredTerminalType` but be different to `0` and `1` to directly test here for `UnsupportedMeasuredTerminalType`. So add a small new test to compliment this one.

tests/unit/validation/test_input_validation.py
```diff
@@ -296,6 +297,22 @@ def original_data() -> dict[ComponentType, np.ndarray]:
     asym_power_sensor["power_sigma"] = [1.0, np.nan, 0.0, -1.0]
     asym_power_sensor["measured_terminal_type"] = [1, 1, 10, 1]
 
+    sym_current_sensor = initialize_array(DatasetType.input, ComponentType.sym_current_sensor, 4)
+    sym_current_sensor["id"] = [7, 8, 9, 10]
+    sym_current_sensor["measured_object"] = [12, 3, 13, 200]
+    sym_current_sensor["i_sigma"] = [1.0, np.nan, 0.0, -1.0]
+    sym_current_sensor["i_angle_sigma"] = [1.0, np.nan, 0.0, -1.0]
+    sym_current_sensor["measured_terminal_type"] = [1, 1, 10, 1]
```
figueroa1395, line 308: Note: These four should break when the check for _It is not possible to add both a [power sensor](https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html#generic-current-sensor) and a [current sensor](https://power-grid-model.readthedocs.io/en/stable/user_manual/calculations.html#./components.mdgeneric-current-sensor) to the same terminal of the same component._ is added, as it "overlaps" with the power sensors.

tests/unit/validation/test_validation_functions.py
```diff
@@ -718,6 +756,144 @@ def test_validate_generic_power_sensor__terminal_types(
...
+        ANY,
+        field=ANY,
+        ref_components=ANY,
+        measured_terminal_type=measured_terminal_type,
+    )
+
+
+@pytest.mark.parametrize("current_sensor_type", [ComponentType.sym_current_sensor, ComponentType.asym_current_sensor])
+@pytest.mark.parametrize(
+    "measured_terminal_type,supported",
```
figueroa1395, line 793: Nitpick:
```suggestion
    "measured_terminal_type, supported",
```

tests/unit/validation/test_validation_functions.py
```diff
@@ -718,6 +756,144 @@ def test_validate_generic_power_sensor__terminal_types(
...
+    validate_generic_current_sensor(data={}, component="")  # type: ignore
+
+    # Assert
+    _all_valid_ids.assert_any_call(
+        ANY,
+        ANY,
+        field=ANY,
+        ref_components=ANY,
+        measured_terminal_type=measured_terminal_type,
+    )
```
figueroa1395, line 788: Isn't this test redundant if we consider the one starting at line 843?

tests/unit/validation/test_input_validation.py
mgovers: changed the 4th one to `2` (source)

tests/unit/validation/test_validation_functions.py
mgovers: i wouldn't consider so: the check whether it is even a valid `MeasuredTerminalType` in the first place is less restrictive than the check whether it's a supported one.

Sure, the user will get both error messages, but the meaning they carry is different.

mgovers (commented):
comments resolved

src/power_grid_model/validation/errors.py
```diff
@@ -518,3 +542,36 @@ def enum_str(self) -> str:
...
+
+
+class MixedPowerCurrentSensorError(MultiComponentValidationError):
+    """
+    Mixed power and current sensor error.
+    """
+
+    _message = (
+        "Mixture of power and current sensors on the same terminal for {n} {objects}. "
+        "If multiple sensors measure the same terminal of the same object, all sensors must measure the same quantity."
```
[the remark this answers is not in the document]
figueroa1395: Let's just keep it as quantity then. That is fine. Resolving.

PR #1000, opened 2025-05-20
Linting and formatting: add Ruff

Laurynas-Jagutis:
Adds Ruff linting and formatting, replaces the old linting and formatting tools.


docs/examples/Power Flow Example.ipynb
```diff
@@ -1372,7 +1384,7 @@
       "\n",
       "Failed scenarios: [3 7]\n",
       "Succeeded scenarios: [0 1 2 4 5 6 8 9]\n",
-      "Error messages: ['The id cannot be found: 1000\\n', 'Iteration failed to converge after 20 iterations! Max deviation: 5.81174e+85, error tolerance: 1e-08.\\n']\n"
+      "Error messages: ['The id cannot be found: 1000\\n', 'Sparse matrix error, possibly singular matrix!\\nIf you get this error from state estimation, it might mean the system is not fully observable, i.e. not enough measurements.\\nIt might also mean that you are running into a corner case where PGM cannot resolve yet.See https://github.com/PowerGridModel/power-grid-model/issues/864.']\n"
```
mgovers, line 1387: 😮 interesting... i did not think we changed behavior... can you please investigate in a separate branch which version caused this difference?... or maybe the change happened when we improved the sparse matrix calculator?

it's at least good to have an idea of what caused this

docs/examples/State Estimation Example.ipynb
```diff
@@ -802,7 +809,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 9,
+   "execution_count": null,
```
mgovers, line 812: please re-run the notebook and make sure that all cells are executed

docs/examples/Validation Examples.ipynb
```diff
@@ -101,24 +101,10 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
```
mgovers, line 104: please re-run the notebook and make sure that all cells are executed

pyproject.toml
```diff
@@ -39,7 +39,7 @@ dynamic = ["version"]
 
 
 [project.optional-dependencies]
-dev = ["msgpack", "pre-commit", "pylint", "pytest", "pytest-cov"]
+dev = ["msgpack", "pre-commit", "pylint", "pytest", "pytest-cov", "ruff"]
```
mgovers, line 42: `pylint` is not necessary anymore, right?

pyproject.toml
```diff
@@ -83,17 +83,65 @@ addopts = [
...
+[tool.ruff]
+# Exclude a variety of commonly ignored directories.
+exclude = [
+    ".bzr",
+    ".direnv",
+    ".eggs",
+    ".git",
+    ".git-rewrite",
+    ".hg",
+    ".ipynb",
```
mgovers, line 95: ? why would you need to exclude notebooks? also, if you did exclude them, then why are there changes to them included in this PR?

pyproject.toml
```diff
@@ -83,17 +83,65 @@ addopts = [
...
+]
+
+# Same as Black.
 line-length = 120
-target-version = ['py311']
+indent-width = 4
+show-fixes = true
+
+# Assume Python 3.12
+target-version = "py312"
```
mgovers, line 124: this should still be `py311` because we haven't dropped python3.11 support yet. that is expected to happen in January (https://power-grid-model.readthedocs.io/en/stable/release_and_support/RELEASE.html#actively-supported-python-versions)

pyproject.toml
```diff
@@ -83,17 +83,65 @@ addopts = [
...
+# Imports that are imported using keyword "as" and are from the same source - are combined.
+combine-as-imports = true
 
-[tool.pylint]
-max-line-length = 120
+[tool.ruff.lint.per-file-ignores]
+# Ignore `F811` (redefinition violations) in all examples notebooks since we use redefinition.
+"docs/examples/*.ipynb" = ["F811"]
+# Ignore 'E402' (import not at the top) in all docs files.
+"docs/*" = ["E402"]
```
mgovers, line 144: :+1:

mgovers, line 144: which other, non-example file in `docs/` has this offending thing? we may want to fix it (not in the notebooks though)

src/power_grid_model/__init__.py
```diff
@@ -27,3 +27,26 @@
     WindingType,
 )
 from power_grid_model.typing import ComponentAttributeMapping
+
+__all__ = [
```
mgovers, line 31: why was this necessary?

src/power_grid_model/data_types.py
```diff
@@ -11,7 +11,7 @@
 
 # Import non class symbols as private to reassign and provide a docstring.
 # Not necessary for class, as sphinx autodoc finds their docstrings
-from power_grid_model._core.data_types import (  # pylint: disable=unused-import
+from power_grid_model._core.data_types import (  # noqa: F401, I001
```
mgovers, line 14: i prefer disabling by message, not by line. does `ruff` not support that? idem below

EDIT: nope, ruff only supports disabling codes

src/power_grid_model/validation/__init__.py
```diff
@@ -12,3 +12,14 @@
...
+__all__ = [
+    "assert_valid_batch_data",
+    "assert_valid_data_structure",
+    "assert_valid_input_data",
+    "errors_to_string",
+    "validate_batch_data",
+    "validate_input_data",
+    "ValidationError",
+    "ValidationException",
+]
```
mgovers, line 25: why was this needed?

pyproject.toml
```diff
@@ -83,17 +83,65 @@ addopts = [
 ]
 xfail_strict = true
 
-[tool.black]
+[tool.ruff]
+# Exclude a variety of commonly ignored directories.
+exclude = [
```
mgovers, line 88: i would expect this to be easier... is this really the only way?

mgovers (commented):
although not too bad, it is quite a lot of files. I propose the following:
create a separate branch off main in which you do the github workflows, precommit and pyproject.toml changes
this branch should not contain any code changes.
instead, you can disable any violating rules in the pyproject.toml in the configuration
then rebase this branch off the other one
note: this may cause some merge/rebase conflicts but after merging, this branch in your local, rebased checkout should be completely the same as the one in the PR ( git diff origin/feature/add-ruff yields an empty diff)
all the configuration disablings in the pyproject.toml should now be "normal" again
then this change is split into 2 stages which helps people migrate, as well as prove that this is indeed the right way to go forward

pyproject.toml
mgovers: please disable in that line  using `# noqa: E402` because it's very specific to that file

src/power_grid_model/validation/__init__.py
Laurynas-Jagutis: Otherwise we get: 

F401 `power_grid_model.validation.utils.errors_to_string` imported but unused; consider removing, adding to `__all__`, or using a redundant alias

The current solution with __all__ was chosen because it already is used in PGM-DS:
https://github.com/PowerGridModel/power-grid-model-ds/blob/main/src/power_grid_model_ds/__init__.py

src/power_grid_model/__init__.py
Laurynas-Jagutis: same as in the other files

src/power_grid_model/validation/__init__.py
mgovers: OK. let's go with `__all__` indeed then, because it explicitly tells users that this is intended to be exposition.

c.c. @emabre this may be interesting to you

pyproject.toml
```diff
@@ -83,17 +83,65 @@ addopts = [
...
+
+[tool.ruff.lint]
+select = [
+    # pycodestyle
+    "E",
+    # Pyflakes
+    "F",
+    # isort
+    "I",
+]
```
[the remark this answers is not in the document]
mgovers: > I have experimented with them a little bit. For example "SIM" seemed to be against if-else blocks, also complained about how we access keys of dictionaries - SIM118 Use key in dict instead of key in dict.keys().

the former, i don't really understand. can you give a concrete example?

the latter is indeed the new way to do it (separate PR!)

>"UP" complained about how we use typing - UP006 [*] Use type instead of Type for type annotation, also it complained about situations like these: if isinstance(ref_components, (str, ComponentType)): UP038 Use X | Y in isinstance call instead of (X, Y)

also the new way to do things (separate PR!)

another one i found:

```
    # Numpy 2.0 downstream : https://numpy.org/devdocs/numpy_2_0_migration_guide.html
    "NPY201",
```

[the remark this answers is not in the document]
Laurynas-Jagutis: > the former, i don't really understand. can you give a concrete example?

in tests\unit\utils.py, line 183.

```
if data_dir is not None:  
        relevant_calculations = [data_dir]  
 else:  
        relevant_calculations = ["power_flow", "state_estimation", "short_circuit"]  
```

We get SIM108 Use ternary operator `relevant_calculations = [data_dir] if data_dir is not None else ["power_flow", "state_estimation", "short_circuit"]` instead of `if`-`else`-block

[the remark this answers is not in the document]
mgovers: also a good one for another separate PR

docs/examples/Power Flow Example.ipynb
nitbharambe: Example was last modified pre LU solver bug. I suppose this power flow gets classified first as SparseMatrix before actually diverging.

pyproject.toml
```diff
@@ -39,7 +39,7 @@ dynamic = ["version"]
 
 
 [project.optional-dependencies]
-dev = ["msgpack", "pre-commit", "pylint", "pytest", "pytest-cov"]
+dev = ["msgpack", "pre-commit", "pytest", "pytest-cov", "ruff"]
```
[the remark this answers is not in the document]
emabre: It added it in  https://github.com/PowerGridModel/power-grid-model/pull/869 because pylint contains pyreverse, which is needed for generating the package and class diagrams.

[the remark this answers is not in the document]
mgovers: we can either:
* keep it, or
* only add `pyreverse` here, or
* only add it to the `docs` optional dependency set

pyproject.toml
```diff
@@ -83,17 +83,34 @@ addopts = [
...
+select = [
+    # pycodestyle
+    "E",
+    # Pyflakes
+    "F",
+    # isort
+    "I",
+    "SIM",
+]
+ignore = ["SIM108", "SIM118", "SIM110", "SIM211"]
```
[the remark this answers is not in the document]
Laurynas-Jagutis: Yes, will do, the initial idea was to do exactly that. First we match the ruff setup as it is in PGM-DS and then we can keep on adding other checks such as "SIM"

pyproject.toml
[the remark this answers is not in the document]
Laurynas-Jagutis: in #1005 I kept it and if it really does not matter to us which option is chosen, then the same will be done here.

docs/examples/Power Flow Example.ipynb
```diff
@@ -1403,16 +1415,16 @@
...
-      " [9.98796126e-001 9.37530046e-001 9.50447962e-001]\n",
-      " [9.98688504e-001 9.29997471e-001 9.44441670e-001]]\n",
+      "[[ 9.99401170e-001  9.92685785e-001  9.94521366e-001]\n",
+      " [ 9.99347687e-001  9.86226389e-001  9.89352855e-001]\n",
+      " [ 9.99288384e-001  9.79654011e-001  9.84095542e-001]\n",
+      " [-2.66881060e+116  2.33997016e-302  6.70346672e-198]\n",
+      " [ 9.99151380e-001  9.66149483e-001  9.73298790e-001]\n",
+      " [ 9.99073166e-001  9.59205860e-001  9.67750710e-001]\n",
+      " [ 9.98988099e-001  9.52126208e-001  9.62096474e-001]\n",
+      " [-2.44756775e+092  5.35663612e-256  1.91838796e-203]\n",
```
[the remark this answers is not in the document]
mgovers: yes: invalid data means rubbish output

mgovers:
> You need to change the recommended vscode extensions to ruff.

this was done

src/power_grid_model/_utils.py
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+"""
+This file contains helper functions for library-internal use only.
+Disclaimer!
+We do not officially support this functionality and may remove features in this library at any given time!
+"""
```
figueroa1395: Why add this here? Not a big deal, just curious.

figueroa1395 (commented):
Only one minor additional thing left from me. There are still some `pylint` comments at `dataset_class_maps.py.jinja`. Remove those as well and don't forget to run the code generator afterwards. Besides that, it looks good to me and locally it is formatted the same as here.

mgovers:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=1000) **Quality Gate passed**
> 
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=1000&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=1000&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=PowerGridModel_power-grid-model&pullRequest=1000&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [96.7% Coverage on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=1000&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=1000&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=1000)

please just add a `nosonar` because it's intended

src/power_grid_model/_utils.py
Laurynas-Jagutis: This file was on main previously, but was removed, so I removed it from this branch as well

PR #1005, opened 2025-05-28
Linting and formatting: setup of Ruff, no code changes

Laurynas-Jagutis:
This PR holds the ruff setup with no linting/formatting changes to the code base.

### DIFF
--- a/tests/unit/validation/test_assertions.py
+++ b/tests/unit/validation/test_assertions.py
@@ -28,10 +28,14 @@ def test_validation_exception(errors_to_string_mock: MagicMock):
 def test_assert_valid_input_data(validate_mock: MagicMock):
     validate_mock.return_value = None
     assert_valid_input_data(
-        input_data={"foo": np.array([1])}, calculation_type=CalculationType.state_estimation, symmetric=False  # type: ignore
+        input_data={"foo": np.array([1])},
+        calculation_type=CalculationType.state_estimation,
+        symmetric=False,  # type: ignore
     )
     validate_mock.assert_called_once_with(
-        input_data={"foo": np.array([1])}, calculation_type=CalculationType.state_estimation, symmetric=False  # type: ignore
+        input_data={"foo": np.array([1])},
+        calculation_type=CalculationType.state_estimation,
+        symmetric=False,  # type: ignore
     )
 
     validate_mock.return_value = [ValidationError()]
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -44,6 +44,7 @@ jobs:
         run: |
           pip install ruff .
           ruff check .
+          ruff format .
           git restore README.md
 
       - name: Install and run clang-format
--- a/docs/examples/Power Flow Example.ipynb
+++ b/docs/examples/Power Flow Example.ipynb
@@ -1388,8 +1388,8 @@
     "except PowerGridBatchError as e:\n",
     "    print(e)\n",
     "    print(f\"Failed scenarios: {e.failed_scenarios}\")\n",
-    "    print(f\"Succeeded scenarios: {e.succeeded_scenarios }\")\n",
-    "    print(f\"Error messages: {e.error_messages }\")"
+    "    print(f\"Succeeded scenarios: {e.succeeded_scenarios}\")\n",
+    "    print(f\"Error messages: {e.error_messages}\")"
    ]
   },
   {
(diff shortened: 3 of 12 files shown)


mgovers (commented):
as discussed before: please ask everyone in the team to upgrade before merging this PR

pyproject.toml
```diff
@@ -83,17 +82,34 @@ addopts = [
...
-line_length = 120
+[tool.ruff.lint.isort]
+# Imports that are imported using keyword "as" and are from the same source - are combined.
+combine-as-imports = true
 
-[tool.pylint]
-max-line-length = 120
+[tool.ruff.lint.per-file-ignores]
+# Ignore `F811` (redefinition violations) in all examples notebooks since we use redefinition.
+"docs/examples/*.ipynb" = ["F811", "E402"]
```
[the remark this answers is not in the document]
Laurynas-Jagutis: This should probably be done when/if the notebooks are adjusted, basically done in a different PR than 1000, right?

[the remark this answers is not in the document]
mgovers: ohhhh right these were the big ones... nvm then. resolving

pyproject.toml
```diff
@@ -52,7 +51,6 @@ doc = [
     "numpydoc",
     "pandas",
     "gitpython",
```
mgovers:
```suggestion
    "gitpython",
    "pylint",
```

otherwise the docs build will be broken: https://app.readthedocs.org/projects/power-grid-model/builds/28366968/

see also github.com/PowerGridModel/power-grid-model/pull/1000#discussion_r2116226974

pyproject.toml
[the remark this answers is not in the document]
Laurynas-Jagutis: Tbh they are not that big, the amount of errors is not huge, it is just that the team would have to decide if we want to make changes to the notebooks or not, then if yes I could proceed

figueroa1395 (commented):
Commenting here due to not having a better place to do so: Please also update the docstring at the begining of .github/workflows/check-code-quality.yml  (lines 5-9).

Besides that, this has been tested and works locally for me. Won't give approval yet as maybe others want to first try locally.

.pre-commit-config.yaml
```diff
@@ -7,14 +7,14 @@ repos:
...
     hooks:
-      - id: isort
-  - repo: https://github.com/psf/black
-    rev: 24.10.0
-    hooks:
-      - id: black-jupyter
+      # Run the linter.
+      - id: ruff-check
+      # Run the formatter.
+      - id: ruff-format
```
mgovers, line 17: let's double-check whether this needs to have the `--fix` arg and why (not)

mgovers (dismissed):
approving. feel free to resolve my comment if you have a good reason to

.pre-commit-config.yaml
Laurynas-Jagutis: We will be adding --fix arguments for ruff-check in #1000

pyproject.toml
```diff
@@ -83,17 +82,34 @@ addopts = [
...
+
+[tool.ruff.lint]
+select = [
+    # pycodestyle
+    "E",
+    # Pyflakes
+    "F",
+    # isort
+    "I",
+    "SIM",
```
mgovers, line 102:
```suggestion
    # suggestions for improvements
    "SIM",
```

tests/unit/validation/test_assertions.py
```diff
@@ -28,10 +28,14 @@ def test_validation_exception(errors_to_string_mock: MagicMock):
...
-        input_data={"foo": np.array([1])}, calculation_type=CalculationType.state_estimation, symmetric=False  # type: ignore
+        input_data={"foo": np.array([1])},
+        calculation_type=CalculationType.state_estimation,
+        symmetric=False,  # type: ignore
     )
     validate_mock.assert_called_once_with(
-        input_data={"foo": np.array([1])}, calculation_type=CalculationType.state_estimation, symmetric=False  # type: ignore
+        input_data={"foo": np.array([1])},
+        calculation_type=CalculationType.state_estimation,
+        symmetric=False,  # type: ignore
```
mgovers, line 38: i think the `type: ignore`s are in the wrong location (i think they should be on `input_data`). 
In addition, `ruff` uses `noqa[CODE]` instead of `type: ignore`. please double-check in #1000 as it's out of scope of this PR.

mgovers (approved):
the added comments will be deferred to #1000

pyproject.toml
mgovers: deferred to #1000

PR #1006, opened 2025-05-30
Documentation: add advanced programming terminology documentation

mgovers:
Prep for design discussion on main model cleanup as well as follow-up of discussions during the meet-up.


docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+
+* The C API is a native data interface. Although the power grid model can check whether the input
+  contain null-pointers (`nullptr`/`NULL`/`0` values where they are not allowed) and can act
+  accordingly, it is not possible to check whether a non-null pointer was actually valid. The
+  behavior of the power grid model in such cases is [undefined](#undefined-behavior-ub-as-a-bug).
+  This is also true for any other native interface that deals with pointers.
+  It is the users' responsibility to provide correct input data. See also [below](#bad-input).
+* If the input data provided to the power grid model is incorrect, the power grid model cannot
+  guarantee that it will be able to find and handle that correctly. The behavior is
+  [undefined](#undefined-behavior-ub-as-a-bug). This is also true for any other native interface.
```
figueroa1395: You mention the phrase "This is also true for any other native interface" several times, and I agree with it. However, since it is such a strong statement, can we cite this? Probably not, but worth doing a quick search.

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+  It is the users' responsibility to provide correct input data. See also [below](#bad-input).
+* The Python wrapper contains human-readable code. Some people consider this a weakness of Python as
+  a programming language. Always make sure to treat your Python code and configuration files as
+  company secrets.
```
figueroa1395: Even though I agree with what is written here, does it belong under the security vulnerability heading?

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+error handling.
+
+Their visibility makes solving them usually easy. If you encounter them, please report them. We will
+usually post announcements regarding bugs of this nature.
+
+#### Undefined behavior bugs
+
+See also [below](#undefined-behavior-ub-as-a-bug).
+
+Although the power grid model takes all the care they can in preventing
```
figueroa1395:
```suggestion
Although the power grid model takes all the care it can in preventing
```

nitbharambe:
Its quite nice that all possible cases are collected and that we are aware of uncertain things present in the library. But I am not sure where this page should lie as the complete list is irrelevant to the user and maybe also for a less involved opensource developer. Nevertheless, quite useful.

What if 
* We split whole of bad input + Incorrect results, Nonsensical results, Incorrect exceptions into one section and rest into other? This section can lie in the user manual.
* Or we have a small section in user manual mentioning these different cases of issues linking to their explanation in this page.

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+[undefined behavior](#undefined-behavior-ub-as-a-bug) like infinite loops or extreme blow-up of the
+data usage, it is still possible that such behavior exists in the code. If you find any such issues,
+please report the issue and we will fix them as soon as possible.
+
+#### Incorrect exceptions
+
+Sometimes, exceptions are raised when they should not be. For example, a `NotObservableError` that
+is raised, even though the scenario actually contains an observable grid, would be a bug. They can
```
figueroa1395: I find the wording a bit weird here. Maybe something along the lines of:

```suggestion
Sometimes, exceptions are raised when they should not be. For example, a `NotObservableError` that
is raised when the current scenario contains an observable grid, would be a bug. They can
```

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+In certain cases, the [data validator](../user_manual/data-validator.md) can catch data issues.
+However, it is not guaranteed to provide an exhaustive check.
+
+#### Unsupported input
+
+The power grid model explicitly forbids certain (combinations of) values. For example, a negative
+`u_rated` on a node is not allowed. Unsupported input will always be documented in the
+power grid model documentation.
+
+Checking every single combination of input parameters would come at a significant performance cost
```
figueroa1395:
```suggestion
Checking every single combination of input parameters would come at a significant performance cost.
```

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+In addition, for most use cases, incorrect input data happens mostly during the development phase of
+a user workflow. In production environments, when performance is usually important, the data is
+usually already sanitized and can be assumed to be correct. Because of this, assuring the
+correctness is the users' responsibility.
```
figueroa1395:
```suggestion
correctness of the input data is the users' responsibility.
```

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### Incorrect input
+
+The power grid model assumes that the input data is correct. Of course, the power grid model cannot
+know whether a user provided the grid correctly. For example, if the user adds a source to the wrong
+node, the power grid model can only obey.
+
+Input correctness is entirely the users' responsibility. The
+[power-grid-model-ds visualiser](https://power-grid-model-ds.readthedocs.io/en/stable/visualizer.html)
+may provide some tools to help with the investigation.
```
figueroa1395: I think this and the above (Incorrect input) can be combined together. Thoughts?

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+  * Short-circuit solutions (spurious)
+  * Divergent domain (spurious)
+  * Differences in amount of iterations before convergence is reached (not spurious)
+
+#### Non-spurious edge cases
+
+In non-spurious edge cases, the current behavior of the power grid model is actually correct.
+
+Many edge cases are inherent to the problem statement, like the amount of iterations it takes before
+a solution is found in an iterative calculation method, are not problematic. Even if you encounter
```
figueroa1395: What did you intend to say with "are not problematic"? Phrasing seems weird.

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+may be extremely difficult to find and/or resolve. Occasionally, even, more research is still needed
+before a satisfactory fix to the edge case can be found. An example of the latter is shown in
+[this paper](https://doi.org/10.48550/arXiv.2504.11650), which was published as collaboration
+between ICT for industy and power grid model.
+
+#### Resolving edge cases
+
+If you encounter any edge case, please double-check your data quality, e.g., by running our
+[data validator](../user_manual/data-validator.md) and verifying that the source of your data is
+up-to-date and correct. Please also try things like increasing the error tolerance to see if that
```
figueroa1395:
```suggestion
up-to-date and correct. Please also try things like increasing the error tolerance or the number of iterations (when applicable) to see if that
```

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+changes the behavior from unspecified behavior to specified exception-throwing.
+
+**NOTE:** PRs changing unspecified behavior may still be marked as a bugfix PR, if developers
+consider the previous behavior unexpected and/or undesirable. Examples are fixes to the behavior of
+a new feature that is still experimental/under development, but for which the developers thought
+that the behavior was already correctly implemented, before they found out that it was not.
+
+##### Undefined behavior (subclass)
+
+Undefined behavior is also a subclass of undefined behavior that is defined as follows:
```
figueroa1395: Is this phrasing intended? UB is a subclass of UB itself?

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+and should therefore be avoided. Triggering undefined behavior should always be conidered a bug,
+regardless of whether it is caused by an incorrect internal implementation of the power grid model
+or a result of incorrect usage by a user.
+
+To help both users and maintainers to find incorrect usage of and bugs in the power grid model, it
+uses an error handling mechanism. It also has extended assertion checks in Debug build configuration
+used in test and development environments to prevent internal bugs.
+
+To prevent, find and fix harder-to-find instances of undefined behavior, the power grid model uses
+tools like AddressSanitizer and Sonar Qube Cloud. We strongly recommend all users to do the same -
```
figueroa1395: Nitpick: I believe this is the correct naming.
```suggestion
tools like AddressSanitizer and Sonar QubeCloud. We strongly recommend all users to do the same -
```

docs/advanced_documentation/terminology.md
```diff
@@ -0,0 +1,329 @@
...
+especially the ones using the C API.
+
+If you find or expect any undefined behavior, please report it and/or
+[contribute](../user_manual/model-validation.md#test-case-creation) the repro case, as doing so
```
figueroa1395:
```suggestion
[contribute](../user_manual/model-validation.md#test-case-creation) with a reproducible case, as doing so
```

figueroa1395:
> Its quite nice that all possible cases are collected and that we are aware of uncertain things present in the library. But I am not sure where this page should lie as the complete list is irrelevant to the user and maybe also for a less involved opensource developer. Nevertheless, quite useful.
> 
> What if
> 
> * We split whole of bad input + Incorrect results, Nonsensical results, Incorrect exceptions into one section and rest into other? This section can lie in the user manual.
> * Or we have a small section in user manual mentioning these different cases of issues linking to their explanation in this page.

I think that as long as we explicitly say that the list is not exhaustive and that this belongs in the advanced documentation, it should be okay to keep here and public. However, this would function sort of a public guideline on how do we tackle and consider edge cases, UB, etc., which is very important. Hence it would be nice if the whole team can take a look and see whether we agree on everything written, or if some things need to be added, changed or removed.

docs/advanced_documentation/terminology.md
mgovers: https://en.wikipedia.org/wiki/Garbage_in,_garbage_out

docs/advanced_documentation/terminology.md
mgovers: removed the section

docs/advanced_documentation/terminology.md
mgovers: i had that originally, but i found that they really are 2 different types: one is detectable, the other is not. it's quite difficult to combine them, especially because we're talking terminology here

docs/advanced_documentation/terminology.md
mgovers: rephrased. better?

docs/advanced_documentation/terminology.md
mgovers: yes it is like that in the C++ standard https://en.cppreference.com/w/cpp/language/ub.html

docs/advanced_documentation/terminology.md
mgovers: nope, if we're going to nitpick stylization, it should be SonarQube Cloud according to their own website https://www.sonarsource.com/products/sonarcloud/
image

updated with the correct stylization

mgovers (commented):
resolved comments

> > Its quite nice that all possible cases are collected and that we are aware of uncertain things present in the library. But I am not sure where this page should lie as the complete list is irrelevant to the user and maybe also for a less involved opensource developer. Nevertheless, quite useful.
> > What if
> > 
> > * We split whole of bad input + Incorrect results, Nonsensical results, Incorrect exceptions into one section and rest into other? This section can lie in the user manual.
> > * Or we have a small section in user manual mentioning these different cases of issues linking to their explanation in this page.
> 
> I think that as long as we explicitly say that the list is not exhaustive and that this belongs in the advanced documentation, it should be okay to keep here and public. However, this would function sort of a public guideline on how do we tackle and consider edge cases, UB, etc., which is very important. Hence it would be nice if the whole team can take a look and see whether we agree on everything written, or if some things need to be added, changed or removed.

I agree with @figueroa1395. That was also my intention: to use for (cross-)reference and as a guideline to what we consider best-effort

PR #1010, opened 2025-06-03
CI: add check no label

Jerry-Jinfeng-Guo:
This PR adds the check for no label for PR's.
The `check-blocking-label` YAML file is renamed to `check-pr-label`.


.github/workflows/check-pr-labels.yml
```diff
@@ -18,14 +18,22 @@ on:
...
-  check-blocking-labels:
+  check-pr-labels:
     runs-on: ubuntu-latest
     steps:
-
+      
+      - name: fail-if-no-label
+        if: github.event_name == 'pull_request'
+        run: |
+          if [ "$(jq '.pull_request.labels | length' "$GITHUB_EVENT_PATH")" -eq 0 ]; then
```
figueroa1395, line 32: Pretty neat! ⚡

figueroa1395:
Everything looks good to me, but can you remove the label temporarily and trigger the CI so it is documented that it works as intended when no label is present?

Jerry-Jinfeng-Guo:
> Everything looks good to me, but can you remove the label temporarily and trigger the CI so it is documented that it works as intended when no label is present?

image

@figueroa1395

figueroa1395:
@Jerry-Jinfeng-Guo If I click on the checks under the commit [74ee648](https://github.com/PowerGridModel/power-grid-model/pull/1010/commits/74ee648d730ac5b77b41ac006a31546c55135b2e) - "empty commit to trigger CI", and go to the specific action status https://github.com/PowerGridModel/power-grid-model/actions/runs/15414358746/job/43373565225, the `do-not-merge` check fails, not the `fail-if-no-label` one. Hence my question.

Jerry-Jinfeng-Guo:
> the `do-not-merge` check fails, not the `fail-if-no-label` one.

@figueroa1395 that's exactly what's supposed to happen, isn't it. No label -> new check fails; do-not-block label -> existing check works, new check passes.

figueroa1395:
@mgovers annoyingly enough, `check-blocking-labels` is stuck forever (renaming in both the workflow name and the .yml file). Best way to bypass this is to just disable it manually in `...` in the [actions](https://github.com/PowerGridModel/power-grid-model/actions/workflows/check-blocking-labels.yml), right? Then this should be "fixed" once the Pr is merged, correct?

mgovers:
> @mgovers annoyingly enough, `check-blocking-labels` is stuck forever (renaming in both the workflow name and the .yml file). Best way to bypass this is to just disable it manually in `...` in the [actions](https://github.com/PowerGridModel/power-grid-model/actions/workflows/check-blocking-labels.yml), right? Then this should be "fixed" once the Pr is merged, correct?

It's not a "bypass" in this case actually: it's an intended change that is required. The new status check should already be discoverable by the required status checks. You can go ahead.

PR #1017, opened 2025-06-12
Test case creation: add helper function to create test case

nitbharambe:

A helper function is added to make adding validation test cases easier.
- Serializes data to file
- Update data can be optionally provided to make batch test case
- It automatically generates license files.

### DIFF
--- a/src/power_grid_model/utils.py
+++ b/src/power_grid_model/utils.py
@@ -407,7 +407,8 @@ def self_test():
 
 
 def _make_test_case(
-    save_path: Path,
+    *,
+    output_path: Path,
     input_data: SingleDataset,
     params: dict,
     output_data: Dataset,
@@ -438,27 +439,27 @@ def _make_test_case(
         Writes JSON files for input, update (if provided), output, and parameters,
         all relevant license files, to save_path.
     """
-    output_file_stem = str(output_dataset_type.name) if isinstance(output_dataset_type, DatasetType) else None
+    output_file_stem = output_dataset_type.name if isinstance(output_dataset_type, DatasetType) else None
     if output_dataset_type in [DatasetType.input, DatasetType.update] or output_file_stem is None:
         raise ValueError(
             f"Invalid output dataset type: {output_dataset_type}. Expected one of: sym_output, asym_output, sc_output."
         )
 
-    save_path.mkdir(parents=True, exist_ok=True)
-    json_serialize_to_file(file_path=save_path / "input.json", data=input_data, dataset_type=DatasetType.input)
-    (save_path / "input.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
+    output_path.mkdir(parents=True, exist_ok=True)
+    json_serialize_to_file(file_path=output_path / "input.json", data=input_data, dataset_type=DatasetType.input)
+    (output_path / "input.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
 
     if update_data is not None:
         json_serialize_to_file(
-            file_path=save_path / "update_batch.json", data=update_data, dataset_type=DatasetType.update
+            file_path=output_path / "update_batch.json", data=update_data, dataset_type=DatasetType.update
         )
-        (save_path / "update_batch.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
+        (output_path / "update_batch.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
         output_file_stem += "_batch"
     json_serialize_to_file(
-        file_path=save_path / f"{output_file_stem}.json", data=output_data, dataset_type=output_dataset_type
+        file_path=output_path / f"{output_file_stem}.json", data=output_data, dataset_type=output_dataset_type
     )
-    (save_path / f"{output_file_stem}.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
+    (output_path / f"{output_file_stem}.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
 
     params_json = json.dumps(params, indent=2)
-    (save_path / "params.json").write_text(data=params_json, encoding="utf-8")
-    (save_path / "params.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
+    (output_path / "params.json").write_text(data=params_json, encoding="utf-8")
+    (output_path / "params.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
--- a/src/power_grid_model/utils.py
+++ b/src/power_grid_model/utils.py
@@ -438,8 +438,8 @@ def _make_test_case(
         Writes JSON files for input, update (if provided), output, and parameters,
         all relevant license files, to save_path.
     """
-    output_file_stem = str(output_dataset_type) if isinstance(output_dataset_type, DatasetType) else None
-    if output_file_stem is None:
+    output_file_stem = str(output_dataset_type.name) if isinstance(output_dataset_type, DatasetType) else None
+    if output_dataset_type in [DatasetType.input, DatasetType.update] or output_file_stem is None:
         raise ValueError(
             f"Invalid output dataset type: {output_dataset_type}. Expected one of: sym_output, asym_output, sc_output."
         )
--- a/src/power_grid_model/utils.py
+++ b/src/power_grid_model/utils.py
@@ -50,7 +50,8 @@ _DEPRECATED_JSON_SERIALIZATION_MSG = f"{_DEPRECATED_FUNCTION_MSG} Please use jso
 
 LICENSE_TEXT = (
     "SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>\n\n"
-    "SPDX-License-Identifier: MPL-2.0\n"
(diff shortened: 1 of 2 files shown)


src/power_grid_model/utils.py
```diff
@@ -398,3 +403,65 @@ def self_test():
...
+    """
+    output_file_stem: str | None = {
+        DatasetType.sym_output: "sym_output",
+        DatasetType.asym_output: "asym_output",
+        DatasetType.sc_output: "sc_output",
+    }.get(output_dataset_type)
+    if output_file_stem is None:
+        raise ValueError(
+            f"Invalid output dataset type: {output_dataset_type}. Expected one of: sym_output, asym_output, sc_output."
+        )
```
mgovers, line 446: it is possible to generate an empty output dataset (e.g. when an error is thrown when it should not). do we need to capture that possibility?

src/power_grid_model/utils.py
```diff
@@ -398,3 +403,65 @@ def self_test():
...
+    Side Effects:
+        Writes JSON files for input, update (if provided), output, and parameters,
+        all relevant license files, to save_path.
+    """
+    output_file_stem: str | None = {
+        DatasetType.sym_output: "sym_output",
+        DatasetType.asym_output: "asym_output",
+        DatasetType.sc_output: "sc_output",
+    }.get(output_dataset_type)
+    if output_file_stem is None:
```
mgovers: this should do the trick
```suggestion
    output_file_stem = str(output_dataset_type) if isinstance(output_dataset_type, DatasetType) else None
    if output_file_stem is None:
```

src/power_grid_model/utils.py
nitbharambe: Yes,it should be possible to give an output dataset which is empty dictionary for such case.

TonyXiang8787 (changes requested):
@nitbharambe @mgovers 
My proposal is to move the validation test function (not the test case finder and parametrize) to the PGM package. 

The validation test function should read a path to a test case, and some meta data (like calculation method, calculation type, etc.). And it should run the validation tests.

In this way, user can easily run more validation test just by install PGM self without going into developer mode.

mgovers:
> @nitbharambe @mgovers My proposal is to move the validation test function (not the test case finder and parametrize) to the PGM package.
> 
> The validation test function should read a path to a test case, and some meta data (like calculation method, calculation type, etc.). And it should run the validation tests.
> 
> In this way, user can easily run more validation test just by install PGM self without going into developer mode.

While indeed useful, I think that's a separate feature that is independent of this PR. In the first place, we want users to be able to provide us with something that is debugable (a test case). If we want to allow them to debug themselves by publishing the test loader, then we should do some serious cleanup work on it first.

We would also need to decide whether to publish the test data as well (or maybe a subset of the data), and a design on how to do that without making the PGM package too big

src/power_grid_model/utils.py
```diff
@@ -398,3 +404,61 @@ def self_test():
             print("Self test finished.")
         except Exception as e:
             raise PowerGridError from e
+
+
+def _make_test_case(
+    save_path: Path,
+    input_data: SingleDataset,
+    params: dict,
```
figueroa1395, line 413: As a follow up, we may want to create a proper `class`or `enum` or something more specific rather than a `dict`for the paramenters. If the goal is to make users' life easier, and ours, this would be a good next step.

tests/unit/test_utils.py
```diff
@@ -164,3 +167,49 @@ def test_msgpack_serialize(serialize_mock: MagicMock, open_mock: MagicMock):
...
+    _make_test_case(
+        save_path=save_path,
+        input_data=input_data,
+        output_data=output_data,
+        params=params,
+        output_dataset_type=output_dataset_type,
+        update_data=update_data,
+    )
+
+    serialize_to_file_mock.assert_any_call(
```
figueroa1395, line 202: We are only testing that the calls are made correctly since the actual serialization is tested elsewhere, right? and `Path.write_text` should also work, correct?

src/power_grid_model/utils.py
mgovers: what about `**kwargs`? e.g. 
```suggestion
    calculation_type: CalculationType,
    calculation_method: CalculationMethod | list[CalculationMethod]
    *,
    rtol: float = 1e-8,
    atol: float = 1e-6,
    experimental_features: _ExperimentalFeatures | None = None,
    extra_params: dict = dict(),
```

figueroa1395: That would do it and follows what we do in the calculation function for instance, but the we should list them all and not have `extra_params` or such. My intention is to easily see them all and make it REALLY easy to use for the user.

src/power_grid_model/utils.py
```diff
@@ -398,3 +404,61 @@ def self_test():
...
+    output_data: Dataset,
+    output_dataset_type: DatasetType,
+    update_data: Dataset | None = None,
```
mgovers: probably good to make everything kwargs for future compatibility
```suggestion
def _make_test_case(
    *,
    save_path: Path,
    input_data: SingleDataset,
    params: dict,
    output_data: Dataset,
    output_dataset_type: DatasetType,
    update_data: Dataset | None = None,
```

src/power_grid_model/utils.py
```diff
@@ -398,3 +404,61 @@ def self_test():
             print("Self test finished.")
         except Exception as e:
             raise PowerGridError from e
+
+
+def _make_test_case(
+    save_path: Path,
```
mgovers: maybe better: `output_path` / `output_dir`

src/power_grid_model/utils.py
mgovers: `extra_params` allow nesting, e.g. to add settings overloads for a certain calculation method. See e.g. https://github.com/PowerGridModel/power-grid-model/blob/main/tests/data/state_estimation/current-sensor/global-current-sensor/params.json#L11-L18

TonyXiang8787:
> > @nitbharambe @mgovers My proposal is to move the validation test function (not the test case finder and parametrize) to the PGM package.
> > The validation test function should read a path to a test case, and some meta data (like calculation method, calculation type, etc.). And it should run the validation tests.
> > In this way, user can easily run more validation test just by install PGM self without going into developer mode.
> 
> While indeed useful, I think that's a separate feature that is independent of this PR. In the first place, we want users to be able to provide us with something that is debugable (a test case). If we want to allow them to debug themselves by publishing the test loader, then we should do some serious cleanup work on it first.
> 
> We would also need to decide whether to publish the test data as well (or maybe a subset of the data), and a design on how to do that without making the PGM package too big

I think this is the same feature from user-story perspective. If user has the function to create a validation dataset, the user wants to be able to run the validation without installing PGM in developer mode or cloning the source code (including the test scripts).

So from user perspective merging only this function alone does not make sense.

src/power_grid_model/utils.py
mgovers: nested classes are possible (using `"ClassName"`) but probably not a good idea

mgovers:
> > > @nitbharambe @mgovers My proposal is to move the validation test function (not the test case finder and parametrize) to the PGM package.
> > > The validation test function should read a path to a test case, and some meta data (like calculation method, calculation type, etc.). And it should run the validation tests.
> > > In this way, user can easily run more validation test just by install PGM self without going into developer mode.
> > 
> > 
> > While indeed useful, I think that's a separate feature that is independent of this PR. In the first place, we want users to be able to provide us with something that is debugable (a test case). If we want to allow them to debug themselves by publishing the test loader, then we should do some serious cleanup work on it first.
> > We would also need to decide whether to publish the test data as well (or maybe a subset of the data), and a design on how to do that without making the PGM package too big
> 
> I think this is the same feature from user-story perspective. If user has the function to create a validation dataset, the user wants to be able to run the validation without installing PGM in developer mode or cloning the source code (including the test scripts).
> 
> So from user perspective merging only this function alone does not make sense.

Still, i would argue that running validation cases is a separate PR at the minimum. I think it's good to follow an incremental improvement approach here. I think cleaning up the test case runner needs to be prioritized with the team and would cause needless delays for merging this PR

src/power_grid_model/utils.py
nitbharambe: Okay an internal dataclass common among the functions using params might be more neat.

tests/unit/test_utils.py
nitbharambe: Yes, thats correct

src/power_grid_model/utils.py
nitbharambe: > what about `**kwargs`? e.g.

The overload of params makes it less ideal to go kwargs way for the `_make_test_case` function of this PR and for the upcoming `validate_test_case` function. I don't think there is a good way of implementing this overload logic.

We can either choose 
* keep params as dict and not have a special support for it.
* Change structure of the params to be a collection of multiple calculation params based on calculation method.

nitbharambe: IMO
1. lets keep params as dict. Same as how it is currently. 
2. Create the new validate_test_case function which would only be about moving functionality.
3. Create params class

TonyXiang8787 (approved):
The test runner into package will be done in a separate PR. I hereby approve this PR.

PR #1018, opened 2025-06-12
Cleanup main model: Job dispatch into separate file

mgovers:
Migrate (job/batch) dispatch logic into separate file

No additional testing was added in this PR yet because the dispatch logic was not really tested in the first place.

NOTE: future improvements can make batch dispatch an input argument into main model rather than the other way around. However, that first requires splitting the logic on how a main model does setup/winddown scenario away from the batch dispatch logic itself. Maybe we can do that by making it static functions that take `main_model` as an input argument?

- [x] Benchmark to confirm no significant performance penalty


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -178,7 +183,6 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
           meta_data_{&meta_data},
           math_solver_dispatcher_{&math_solver_dispatcher} {}
 
-  private:
```
mgovers, line 181: here and below: i don't like this, but i don't know of a good alternative except making static functions for this. Maybe we can provide a `main_model_view` to the batch dispatch logic? thoughts?

we can also do that as a future improvement as just discussed.

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
```diff
@@ -0,0 +1,255 @@
...
+                try {
+                    winddown_(args...);
+                } catch (...) {
+                    recover_from_bad_(args...);
+                }
+            }
+        };
+    }
+
+    static auto scenario_update_restore(
```
mgovers: the contents of this function feel out-of-place here. maybe we need to re-add this function to main model impl again. thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
```diff
@@ -0,0 +1,255 @@
...
+#include "main_model_fwd.hpp"
+
+#include "main_core/calculation_info.hpp"
+#include "main_core/update.hpp"
+
+#include <thread>
+
+namespace power_grid_model {
+
+template <class MainModel, class... ComponentType> class BatchDispatch {
```
mgovers: we probably want to revert this dependency: `BatchDispatch` should be MainModel-agnostic. would be a good future improvement. It probably would help to move `scenario_update_restore` back to main model. thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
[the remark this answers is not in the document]
mgovers: ? this was moved from main model into here.

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
mgovers: another possibility is to make `MainModel` not a template parameter of the class but of each function separately

nitbharambe: What if we eventually would have scenario_update_restore not depend on main_model in the upcoming refactoring? No reason to move it back then, I suppose. 
The logic about copy base data -> apply mutation -> reverse mutation seems to be a job of batch dispatch.

mgovers: idd so the problem is not necessarily the logic but its explicit dependence on `main_model`. however, we would still have to have an interface on "how to update", which is currently a call to the `main_model`

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
```diff
@@ -0,0 +1,256 @@
...
+        if (update_data.empty()) {
+            std::forward<Calculate>(calculation_fn)(model, result_data, 0);
+            return BatchParameter{};
+        }
+
+        // get batch size
+        Idx const n_scenarios = update_data.batch_size();
+
+        // if the batch_size is zero, it is a special case without doing any calculations at all
+        // we consider in this case the batch set is independent and but not topology cachable
```
figueroa1395: Niptick: Can you fix this comment? Specifically `"... and but not ..." is not very readable.

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
```diff
@@ -0,0 +1,256 @@
...
+  private:
+    using SequenceIdxView = std::array<std::span<Idx2D const>, main_core::utils::n_types<ComponentType...>>;
+
+  public:
+    static constexpr Idx ignore_output{-1};
+    static constexpr Idx sequential{-1};
+
+    template <typename Calculate>
+        requires std::invocable<std::remove_cvref_t<Calculate>, MainModel&, MutableDataset const&, Idx>
+    static BatchParameter batch_calculation_(MainModel& model, CalculationInfo& calculation_info,
```
figueroa1395: I would probably want to have this as the only `public` function in the class, or make a simpler entry point. The rest can just be private as are implementation details.

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
figueroa1395: I'd say this is where we need to put most the effort and the "hard" part of taking batch dispatch out of main model. I am all in for completly making `BatchDispatch` `MainModel` agnostic; right now both are very tightly coupled.

I wouldn't like to make `MainModel` a template parameter of each function either, I still think this is halfway there. Instead, I propose we go with either full interface or use dependency injection:
- Full interface: We create either a pure virtual interface or make it the CRTP way (I have preference for the second), and then some interface implementation that uses the main model calls and then we just pass that as an argument to the actual batch dispatch. Pros: Easy to test and mock, boundaries are clear, increases cohesion and decreases coupling and probably easier to read. Cons: Lot's of boilerplate, might be more tedious (take longer) to maintain, and potential overhead if we go virtual instead of CRTP, and CRTP might make it less readable.
- Dependency injection: We could mostly use function and object injection, i.e. create a `struct` in `MainModel` with the necessary calls - or directly implement the calls - that would be passed/forwarded to the batch dispatch logic; then the batch dispatch expects a mechanism to do stuff without really knowing what's up. Pros: Likely simpler to maintain and implement, less boilerplate, and increases cohesion and reduces coupling as well. Cons: Harder to test and mock, but still possible, and probably harder to read.

There may be more pros and cons, just wrote some out of my head; also, there are proabably other alternatives as well, so this is not exhaustive. Furthermore, since we are figuring out what is the best way to go, we could probably use batch dispatch as our first toy model and implement both in separate PRs (as well as other possible solutions) - even if it is just a partial implementation - and then decide the way to go based on that. I have the feeling that we will encounter this issue when trying to separate most things, because we not only want to move to a separate module, but actually separate concerns and invert dependencies.

Thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
figueroa1395: I actually think that this function belongs in this place, it still handles batch dispatch logic. However, we do need to refactor and split apart if possible, this is a pain to read. Also, renaming might be a good idea.

On the other hand, for this and the others, I strongly believe we need to invert the `MainModel` dependency.

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
```diff
@@ -0,0 +1,256 @@
...
+        all_scenarios_sequence = main_core::update::get_all_sequence_idx_map<ComponentType...>(
+            base_model.state(), update_data, 0, components_to_update, update_independence, false);
+
+        return [&base_model, &exceptions, &infos, calculation_fn_ = std::forward<Calculate>(calculation_fn),
+                &result_data, &update_data, &all_scenarios_sequence_ = std::as_const(all_scenarios_sequence),
+                components_to_update, update_independence](Idx start, Idx stride, Idx n_scenarios) {
+            assert(n_scenarios <= narrow_cast<Idx>(exceptions.size()));
+            assert(n_scenarios <= narrow_cast<Idx>(infos.size()));
+
+            Timer const t_total(infos[start], 0000, "Total in thread");
```
figueroa1395: If I see this way of timing, and assuming we invert the dependency on `MainModel`, I actually don't think the way we benchmark would look so bad afterall. As long as the timer module is a floating thing above everything, that should be alright.

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
figueroa1395: Thinking about it a bit more, dependency injection is going to be a very weird implementation which might still be harder to completely decouple. If I have some spare time, I will work on a proof on concept of using an interface, or my idea behind it at least.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: We should try to completely decouple them... maybe this would look better then.

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
mgovers: idd so the problem is not necessarily the logic but its explicit dependence on `main_model`. however, we would still have to have an interface on "how to update", which is currently a call to the `main_model`

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
mgovers: the way we use the `timer` now is actually not by factory pattern or logger pattern (i.e. it's not opt-in but mandatory). in addition, it's not the timer here that's the problem but the `CalculationInfo` that is neither thread-safe nor efficient. there's better ways to do it, but we can go over that another time 😄

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
mgovers:
```suggestion
        // we consider in this case the batch set is independent but not topology cacheable
```

Jerry-Jinfeng-Guo:
Let's take this chance and rename `batch_dispatch` to `job_dispatch`. I like the name `orchestrator`, but not in our current code form. In my opinion, what is now the `main_model` could be named `orchestrator`.

mgovers:
> Let's take this chance and rename `batch_dispatch` to `job_dispatch`. I like the name `orchestrator`, but not in our current code form. In my opinion, what is now the `main_model` could be named `orchestrator`.

ok, sounds good. Are you OK with me doing that in a separate PR to more clearly show the distinction between the move and rename? or do you prefer doing it here instead?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: Resolving: refer to https://github.com/PowerGridModel/power-grid-model/pull/1036

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
figueroa1395: Resolving: refer to https://github.com/PowerGridModel/power-grid-model/pull/1036

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
figueroa1395: Resolving: refer to https://github.com/PowerGridModel/power-grid-model/pull/1036

power_grid_model_c/power_grid_model/include/power_grid_model/batch_dispatch.hpp
figueroa1395: Resolving: refer to https://github.com/PowerGridModel/power-grid-model/pull/1036

Jerry-Jinfeng-Guo:
> > Let's take this chance and rename `batch_dispatch` to `job_dispatch`. I like the name `orchestrator`, but not in our current code form. In my opinion, what is now the `main_model` could be named `orchestrator`.
> 
> ok, sounds good. Are you OK with me doing that in a separate PR to more clearly show the distinction between the move and rename? or do you prefer doing it here instead?

Either is fine by me, although I do not see the particular need for a separate PR. So please proceed whichever way you'd like.

Jerry-Jinfeng-Guo:
Added a `do-not-merge` until the benchmark to-do is checked.

figueroa1395:
I messed up my branches internally and pushed [c8cc106](https://github.com/PowerGridModel/power-grid-model/pull/1018/commits/c8cc1060795eda1cff8c468713c7e09ae5b22009) and [ce332ac](https://github.com/PowerGridModel/power-grid-model/pull/1018/commits/ce332ac6f839350ff2fe72196689c2b6bd61f43f) here by mistake. I'll try to fix this tomorrow morning by removing from here and actually pushing to https://github.com/PowerGridModel/power-grid-model/pull/1036. @mgovers Input to don't make this horribly horrible is appreciated :)

mgovers:
> I messed up my branches internally and pushed [c8cc106](https://github.com/PowerGridModel/power-grid-model/pull/1018/commits/c8cc1060795eda1cff8c468713c7e09ae5b22009) and [ce332ac](https://github.com/PowerGridModel/power-grid-model/pull/1018/commits/ce332ac6f839350ff2fe72196689c2b6bd61f43f) here by mistake. I'll try to fix this tomorrow morning by removing from here and actually pushing to #1036. @mgovers Input to don't make this horribly horrible is appreciated :)

fixed

mgovers:
benchmark OK. Removing `do-not-merge` flag

Jerry-Jinfeng-Guo (approved):

PR #1023, opened 2025-06-16
Linting and Formatting: add a couple more Ruff checks

Laurynas-Jagutis:
Adds and fixes SIM errors
Adds and fixes NPY errors

### DIFF
--- a/src/power_grid_model/_core/utils.py
+++ b/src/power_grid_model/_core/utils.py
@@ -685,9 +685,6 @@ def _extract_row_based_data(
     Returns:
         SingleArray | DenseBatchArray: the contents of row based data
     """
-    if_is_batch = [2] if is_batch else [1]
-    allowed_dims = if_is_batch if is_batch is not None else [1, 2]
-
     if is_batch is None:
         allowed_dims = [1, 2]
     elif is_batch:


setup.py
```diff
@@ -76,10 +76,7 @@ def get_tag(self):
 class MyBuildExt(build_ext):
     def build_extensions(self):
         if not if_win:
-            if "CXX" in os.environ:
-                cxx = os.environ["CXX"]
-            else:
-                cxx = self.compiler.compiler_cxx[0]
+            cxx = os.environ["CXX"] if "CXX" in os.environ else self.compiler.compiler_cxx[0]
```
mgovers: or
```suggestion
            cxx = os.environ.get("CXX", self.compiler.compiler_cxx[0])
```

setup.py
mgovers, line 1: this file is probably not worth a lot of effort due to #1019

src/power_grid_model/_core/utils.py
```diff
@@ -651,10 +651,8 @@ def _extract_columnar_data(
     """
     not_columnar_data_message = "Expected columnar data"
 
-    if is_batch is not None:
-        allowed_dims = [2, 3] if is_batch else [1, 2]
-    else:
-        allowed_dims = [1, 2, 3]
+    if_is_batch = [2, 3] if is_batch else [1, 2]
+    allowed_dims = (if_is_batch) if is_batch is not None else [1, 2, 3]
```
mgovers: although the SIM suggestion is better than the original, this still feels like it can be rewritten in a waaayyyy better way.

E.g. try
```suggestion
    if is_batch is None:
        allowed_dims = [1, 2, 3]
    elif is_batch:
        allowed_dims = [2, 3]
    else:
        allowed_dims = [1, 2]
```
now please run the `SIM` check again after that.

This is a good example of why such a linter is important (we actually get suggestions) and then our next step is t always double-check whether this really is the right path going forward

src/power_grid_model/_core/utils.py
```diff
@@ -683,10 +681,8 @@ def _extract_row_based_data(
     Returns:
         SingleArray | DenseBatchArray: the contents of row based data
     """
-    if is_batch is not None:
-        allowed_dims = [2] if is_batch else [1]
-    else:
-        allowed_dims = [1, 2]
+    if_is_batch = [2] if is_batch else [1]
+    allowed_dims = if_is_batch if is_batch is not None else [1, 2]
```
mgovers: same as before

src/power_grid_model/validation/_rules.py
```diff
@@ -892,10 +892,7 @@ def none_missing(data: SingleDataset, component: ComponentType, fields: str | li
         fields = [fields]
     for field in fields:
         nan = _nan_type(component, field)
-        if np.isnan(nan):
-            invalid = np.isnan(data[component][field])
-        else:
-            invalid = np.equal(data[component][field], nan)
+        invalid = np.isnan(data[component][field]) if np.isnan(nan) else np.equal(data[component][field], nan)
```
mgovers, line 895: there should be a utility for this... out of scope of this PR tho

PR #1024, opened 2025-06-17
Observability check: Phasor voltage sensor

### LINKED ISSUE #914: [FEATURE] Addition of phasor voltage sensors to the necessary and sufficient conditions for observability in radial grids
# Background

The full observability check for any grid (radial or meshed) with any type of sensors is described in #84. In #508, we implemented a necessary condition check for observability, which was later extended in #838 to a necessary and sufficient condition for observability in radial grids. However, the checks implemented in #508 and #838 have not yet been extended to meshed grids. Additionally, they only consider regular voltage sensors and do not account for phasor voltage sensors.

# Proposal

This issue focuses solely on incorporating phasor voltage sensors into the necessary and sufficient condition check for observability in radial grids, extending #838 and serving as an additional intermediate step toward the full observability check described in #84.

# Mathematics

## Build graph

We don't need to build any additional graph. The one used in #858 is enough to perform the observability check. Accounting for phasor voltage sensors doesn't modify the graph, but adds a new tag in the already constructed vertex, similarly to what the injection sensors did in #858. Details are explained in the next section.

## Check observability

To check observability, we traverse the vertices of the above grid in the reverse order of the DFS results. For each vertex, we do the following:

1. If the predecessor edge is `not_measured` .
    1. If the current vertex has `injection_available` or `phasor_voltage_sensor_available`
        1. change the predecessor edge to `measured`.
    1. Else
        1. If the predecessor vertex has `injection_available`
            1. change the predecessor edge to `measured`.
            2. change the predecessor vertex to `injection_unavailable`.
        2. Else if the predecessor vertex has `phasor_voltage_sensor_available`
            1. change the predecessor edge to `measured`.
            2. change the predecessor vertex to `phasor_voltage_sensor_unavailable`.
2. If the current vertex has `phasor_voltage_sensor_available`
    1. set `phasor_voltage_sensor_unmoved` to `true`.
3. Change the current vertex to `injection_unavailable` and `phasor_voltage_sensor_unavailable`, regardless of the original state.

Note: `phasor_voltage_sensor_available` is the new vertex tag introduced by the addition of phasor voltage sensors.

After traversing all the vertices in reverse order, do the following:
1. If all the edges are now `measured`
    1. if at least one voltage sensor is present
        1. the system is observable.
    2. Else if at least one phasor voltage sensor is present
        1. if `phasor_voltage_sensor_unmoved` is `true`
            1. the system is observable.
        2. Else
            1. the system is NOT observable.
2. Else
    1. the system is NOT observable.

This is the full necessary and sufficient condition check for radial grids considering phasor voltage sensors.

Note: The same order layed out in the logic, is not necessarily the most efficient order for the implementation details.

# Implementation

The implementation follows directly from #868. The DFS search information is already contained and ordered in the `YBusStructure`; its upper triangle contains all the information on the predecessor vertex and edge. In addition, the `measured_values` in the already implemented `necessary_observability_check` also contain the information of the phasor voltage sensors. 

In short, the implementation proposal is as follows:
1. Add the new  `phasor_voltage_sensor_available` tag to all the vertexes in the graph. This can probably be done by generalizing/extending the `count_flow_sensors` function to add the new sensors.
2. Since each vertex can potentially have two sensors now, move either one to different branches if needed. This can probably be done by generalizing/extending the `assign_injection_sensor_radial` function to move available phasor voltage sensors to unmeasured branches. 
3. This is enough for the sufficient observability condition, as the necessary condition is already implemented, even for voltage sensors.

Note: Everything can be implemented directly inside `observability.hpp`

figueroa1395:
Closes #914

Checklist:

- [ ] Implementation
- [ ] Unit tests
- [ ] Validation tests
- [ ] Refactor
- [ ] Documentation


figueroa1395:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=1024) **Quality Gate passed**
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=1024&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=PowerGridModel_power-grid-model&pullRequest=1024&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=PowerGridModel_power-grid-model&pullRequest=1024&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [97.8% Coverage on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=1024&metric=new_coverage&view=list)  [1.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=1024&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=1024)

I purposely didn't do what sonar is suggesting as that would involve splitting the `if` statement in two, making it larger and less readable in my opinion. However, if there are strong opinions in favor of this, I will make the change.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -12,38 +12,23 @@
...
-// count flow sensors into integer matrix with ybus structure
-// lower triangle part is always zero
-// for diagonal part, it will be one if there is bus injection
-// for upper triangle part, it will be one if there is branch flow sensor and the branch is fully connected
-// return a pair of
+// count flow and voltage phasor sensors for the observability check
+// use the ybus structure
+// the lower triangular part is always zero
+// the diagonal part will be one if there is bus injection or a voltage phasor sensor
+// the upper triangle part will be one if there is branch flow sensor and the branch is fully connected
```
mgovers:
```suggestion
// the upper triangle part will be one where there is a branch flow sensor and the branch is fully connected
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -12,38 +12,23 @@
...
-
-// count flow sensors into integer matrix with ybus structure
-// lower triangle part is always zero
-// for diagonal part, it will be one if there is bus injection
-// for upper triangle part, it will be one if there is branch flow sensor and the branch is fully connected
-// return a pair of
+// count flow and voltage phasor sensors for the observability check
+// use the ybus structure
+// the lower triangular part is always zero
+// the diagonal part will be one if there is bus injection or a voltage phasor sensor
```
mgovers:
```suggestion
// the diagonal part will be one if there is a complete bus injection or voltage phasor sensor
```

or

```suggestion
// the diagonal part will be one if there is a complete bus injection sensor or a voltage phasor sensor
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -12,38 +12,23 @@
...
+// the upper triangle part will be one if there is branch flow sensor and the branch is fully connected
+// return a tuple of
 //      a vector of flow sensor count
+//      a vector of voltage phasor sensor count
 //      a boolean indicating if the system is possibly ill-conditioned
 template <symmetry_tag sym>
-std::pair<std::vector<int8_t>, bool> count_flow_sensors(MeasuredValues<sym> const& measured_values,
-                                                        MathModelTopology const& topo,
-                                                        YBusStructure const& y_bus_structure) {
+std::tuple<std::vector<int8_t>, std::vector<int8_t>, bool>
```
mgovers: can you make this return type a named object? e.g.
```cpp
struct ObservingSensorCount {
    std::vector<int8_t> flow_sensor_count;
    std::vector<int8_t> voltage_phasor_count;
    bool maybe_ill_conditioned{false};
};
```

or alternatively (but maybe overloads or static accessors are needed for `const` access)

```cpp
struct ObservingSensorCount {
    std::tuple<std::vector<int8_t>, std::vector<int8_t>, bool> data{};
    constexpr decltype(auto) flow_sensor_count() { return std::get<0>(data); }
    constexpr decltype(auto) voltage_phasor_sensor_count() { return std::get<1>(data); }
    constexpr bool maybe_ill_conditioned() { return std::get<2>(data); }
};
```

I'm personally not a fan of unnamed tuples

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -12,38 +12,23 @@
...
+count_observability_sensors(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
+                            YBusStructure const& y_bus_structure) {
     Idx const n_bus{topo.n_bus()};
-    std::pair<std::vector<int8_t>, bool> result{};
-    auto& [flow_sensors, possibly_ill_conditioned] = result;
-    flow_sensors.resize(y_bus_structure.row_indptr.back(), 0); // initialize all to zero
-    possibly_ill_conditioned = false;
+    auto voltage_phasor_sensors = std::vector<int8_t>(n_bus, 0);
+    auto flow_sensors = std::vector<int8_t>(y_bus_structure.row_indptr.back(), 0);
+    auto possibly_ill_conditioned = false;
```
mgovers: can be in-place when using the named return type

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -74,25 +59,26 @@ std::pair<std::vector<int8_t>, bool> count_flow_sensors(MeasuredValues<sym> cons
...
             possibly_ill_conditioned = true;
         }
     }
-    return result;
+    return std::make_tuple(std::move(flow_sensors), std::move(voltage_phasor_sensors), possibly_ill_conditioned);
 }
 
 // re-organize the flow sensor for radial grid without phasor measurement
 // this mutate the flow sensor vector to try to assign injection sensor to branch sensor
 // all the branch should be measured if the system is observable
```
mgovers: docstring is outdated

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -74,25 +59,26 @@ std::pair<std::vector<int8_t>, bool> count_flow_sensors(MeasuredValues<sym> cons
...
 // this is a sufficient condition check
 // if the grid is not radial, the behavior is undefined.
-inline void assign_injection_sensor_radial(YBusStructure const& y_bus_structure, std::vector<int8_t>& flow_sensors) {
-    Idx const n_bus = std::ssize(y_bus_structure.row_indptr) - 1;
+inline void assign_independent_sensors_radial(YBusStructure const& y_bus_structure, std::vector<int8_t>& flow_sensors,
+                                              std::vector<int8_t>& voltage_phasor_sensors, Idx const n_bus) {
```
mgovers: why does `n_bus` need to be an input variable? if you need it as one, please add an assertion to verify no undefined behavior is triggered here.

```suggestion
                                              std::vector<int8_t>& voltage_phasor_sensors, Idx const n_bus) {
    assert(n_bus == std::ssize(y_bus_structure.row_indptr) - 1);
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -101,26 +87,70 @@ inline void assign_injection_sensor_radial(YBusStructure const& y_bus_structure,
...
     flow_sensors[y_bus_structure.bus_entry[n_bus - 1]] = 0;
 }
 
+inline void necessary_observability_condition(Idx const n_voltage_phasor_sensors, Idx const n_flow_sensors,
+                                              Idx const n_bus, bool const has_global_angle_current) {
+    if (n_voltage_phasor_sensors == 0 && n_flow_sensors < n_bus - 1) {
+        throw NotObservableError{};
+    }
+    if (n_voltage_phasor_sensors > 0 && n_flow_sensors + n_voltage_phasor_sensors < n_bus) {
+        throw NotObservableError{};
```
mgovers: maybe good to add these reasons as well to the error messages

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -101,26 +87,70 @@ inline void assign_injection_sensor_radial(YBusStructure const& y_bus_structure,
...
         }
-        // remove the current bus injection regardless of the original state
+        // remove the current bus injection sensors regardless of the original state
         flow_sensors[bus_entry_current] = 0;
     }
     // set last bus injection to zero
     flow_sensors[y_bus_structure.bus_entry[n_bus - 1]] = 0;
 }
 
+inline void necessary_observability_condition(Idx const n_voltage_phasor_sensors, Idx const n_flow_sensors,
```
mgovers: nice

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -90,18 +98,19 @@ TEST_CASE("Necessary observability check") {
...
-        SUBCASE("voltage phasor available condition for unobservable grid") {
+        SUBCASE("Voltage phasor available condition for unobservable grid") {
             check_not_observable(topo, param, se_input);
         }
 
-        SUBCASE("Parallel branch get counted as one sensor") {
-            // Add sensor on branch 3 to side. Hence 2 parallel sensors
+        SUBCASE("Power sensors on parallel branches gets counted as one sensor") {
+            // add sensor on branch 2 to-side
+            // m ove sensor on branch 0 to-side to branch 1 to side
```
mgovers: typo
```suggestion
            // move sensor on branch 0 to-side to branch 1 to side
```

tests/data/state_estimation/not-independent-sensor/not-independent-sensor-1/input.json
mgovers, line 1: `not-independent-sensor-1` and `not-independent-sensor-2` are not really elaborate 😬 is there a better way to describe the difference between the two test cases?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: I know it may look more 'consistent' with the rest, but
```suggestion
    bool possibly_ill_conditioned = false;
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -101,26 +87,70 @@ inline void assign_injection_sensor_radial(YBusStructure const& y_bus_structure,
...
             } else if (flow_sensors[bus_entry_upstream] == 1) {
-                // if not possible, steal from upstream bus
+                // if not possible, try to steal injection sensor from upstream bus
                 std::swap(flow_sensors[branch_entry_upstream], flow_sensors[bus_entry_upstream]);
+            } else if (voltage_phasor_sensors[current_bus] == 1) {
+                // if not possible, try to steal voltage phasor sensor from current bus
+                std::swap(flow_sensors[branch_entry_upstream], voltage_phasor_sensors[current_bus]);
+            } else if (voltage_phasor_sensors[current_bus + 1] == 1) {
+                // if not possible, try to steal voltage phasor sensor from upstream bus
+                std::swap(flow_sensors[branch_entry_upstream], voltage_phasor_sensors[current_bus + 1]);
```
Jerry-Jinfeng-Guo: What is the difference between `current_bus + 1` and `bus_entry_upstream`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -130,51 +160,34 @@ struct ObservabilityResult {
...
+    Idx const n_voltage_phasor_sensors =
+        std::reduce(voltage_phasor_sensors.cbegin(), voltage_phasor_sensors.cend(), Idx{}, std::plus<Idx>{});
 
-    if (measured_values.has_global_angle_current() && n_voltage_phasor_sensor == 0) {
-        throw NotObservableError{
-            "Global angle current sensors require at least one voltage angle measurement as a reference point.\n"};
-    }
+    // check necessary condition for observability
+    detail::necessary_observability_condition(n_voltage_phasor_sensors, n_flow_sensors, n_bus,
+                                              measured_values.has_global_angle_current());
```
Jerry-Jinfeng-Guo: What is the argument behind the design choice here to not include the state of `necessary_observability_condition`  in the `ObservabilityResult` struct like the `sufficient_observability_condition`? This introduces two different approach towards the full observability, which is only true when the network is sufficiently and necessarily observable.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -130,51 +160,34 @@ struct ObservabilityResult {
...
-            throw NotObservableError{
-                "The number of power/current sensors appears sufficient, but they are not independent "
-                "enough. The system is still not observable.\n"};
-        }
-        result.is_sufficiently_observable = true;
+    // check the sufficient condition for observability
+    // the check is currently only implemented for radial grids
+    if (topo.is_radial) {
+        result.is_sufficiently_observable = detail::sufficient_observability_condition(
+            y_bus_structure, flow_sensors, voltage_phasor_sensors, n_voltage_phasor_sensors, n_bus);
```
Jerry-Jinfeng-Guo: I wonder if we should throw or log something for meshed grid. This check here applies: either there was no guard before this step, or this check is redundant.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo, line 1: Are `.value` and `.variance` in measurement arbitrarily set here? Why are they almost of the same scale?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: The suggestion introduces more confusion. I'd prefer the original version with slight extension.
```suggestion
// the upper triangle part will be one when there exist branch flow sensors and the branch is fully connected
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -130,51 +160,34 @@ struct ObservabilityResult {
...
     }
-
     return result;
```
Jerry-Jinfeng-Guo: Just curious, why not also do a RVO-ish direct return here?

Jerry-Jinfeng-Guo (commented):
I am a little bit hesitant to say the current flow of handling is optimal. The reason is inhomogeneous processing of states ins the observability status. Additionally, depending on the electrical context, we might want to differentiate unobservable grids when they are not sufficiently observable and when they are not necessarily observable.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: It doesn't need to be an input variable. I can just calculate it indeed from the `YBusStructure`. Fixing!

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: So this one is tricky and probably I need to come up with an actual name for `current_bus + 1`. The difference is that `bus_entry_upstream` gives the upstream bus index in the sparse matrix context, as the `flow_sensors` vector keeps track of power sensors at both the nodes and the branches themselves. On the other hand, since the voltage phasor sensors can only be in located in the nodes, the `voltage_phasor_sensors` vector only contains information about the diagonal the addmittance matrix.

Let me know if the explanation was clear enough. If so, do you have any suggestions to make it clearer/more readable?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: The necessary condition doesn't guarantee anything, it is only a one way path towards observability. However, the sufficient condition is a two way path. Nevertheless, I decided to split it in this way for clarity and because if the first condition is not satisfied, we just `throw` and stop computation at that point.

That is why onl the sufficient condition is tracked.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: I believe (correct me if I am wrong) that it is well documented that the observability check is only performed for radial grids. For this reason, we can't `throw` when the grid is meshed. In the latter case, if the grid is not observable, the user would probably get a `SparseMatrix` error in the current state.

If you also check current main (keeping in mind that phasor voltage sensors aren't yet introduced in the check), this check has always been in place. If the grid is radial we check for observability, if not, we don't know and just continue to calculate.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Yes, they are completely arbitrary since they aren't relevant for the observability test itself. However, they are still needed for the `YBus` construction.

tests/data/state_estimation/not-independent-sensor/not-independent-sensor-1/input.json
figueroa1395: I will try to come up with better names. I wasn't too inspired when I wrote it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: By definition, sufficient and necessary are both one way and they are the 'opposite' direction. If the condition checked here is indeed two way, it should not be referred to sufficient condition and should simply be referred to as the observable condition. A grid is only observable when both conditions are met.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: So the observability check is performed by default as part of the process. Check.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: In https://github.com/PowerGridModel/power-grid-model/issues/914, it says
```
...
    Else if the current vertex has phasor_voltage_sensor_available
        change the predecessor edge to measured.
        change the current vertex to phasor_voltage_sensor_unavailable.
    Else if the predecessor vertex has phasor_voltage_sensor_available
        change the predecessor edge to measured.
        change the predecessor vertex to phasor_voltage_sensor_unavailable.
...
```
They are appearing to be checking the same predecessor vertex as in the `injection_available` check. Now I am even more confused after the explanation.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Done, went with the first option. Just simpler.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: Then would it be possible to have less uniformity in them? One could question why `sigma = real(*) = img(*)`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Made some changes, let me know what you think.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: You are right. The sufficient and necessary condition is just called `is_sufficiently_observable`. This comes from language abuse as the sufficient condition is only checked if the necessary one has already been satisfied. I will address this.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Done. I made sure they are now more randomized.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: I have changed it to `is_observable`. This reflects the actual state now. Thoughts?

figueroa1395:
> I am a little bit hesitant to say the current flow of handling is optimal. The reason is inhomogeneous processing of states ins the observability status. Additionally, depending on the electrical context, we might want to differentiate unobservable grids when they are not sufficiently observable and when they are not necessarily observable.

Let me know what you think after the changes. Right now we only use the observability check for two reasons:
1. If either the necessary or sufficient conditions fail, then we `throw` a `NotObservableError` with a descriptive message.
2. If both conditions are successful, we then use (newly named) `is_observable` variable to determine whether we should use perturbation or not, in conjunction with the `is_possibly_ill_conditioned` check.

The flow is clear in my opinion. If we were to explicitly need the sufficient and necessary conditions individually in the future, it is trivial to add them in the current state. Also, the flow is kept the same as introduced in https://github.com/PowerGridModel/power-grid-model/pull/858.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Yes, they do. However, the vector that stores `flow_sensors` has a shape like of the `YBusStructure` because power sensors can be placed in both nodes and branches. On the other hand, the vector  for `voltage_phasor_sensors` has a shape like only the diagonal part of the `YBusStructure`. So both are traversed differently. In the first I go over the nodes and branches, in the second only over the nodes.

So `current_bus + 1` is used to traverse `voltage_phasor_sensors` as `current_bus` goes over the diagonal elements as if they were a flat array. On the contrary `bus_entry_upstream` counts the diagonal elements taking into account that the "input" matrix is actually a sparse matrix, not just a diagonal one.

If this is not yet clear, let's continue offline.

tests/data/state_estimation/not-independent-sensor/not-independent-sensor-1/input.json
figueroa1395: Addressed.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: [4c5646e](https://github.com/PowerGridModel/power-grid-model/pull/1024/commits/4c5646e1bbb7724e41bfc060de8ab6c4ca9ef4ec) does this make it a little more clear together with the explanation?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: [4c5646e](https://github.com/PowerGridModel/power-grid-model/pull/1024/commits/4c5646e1bbb7724e41bfc060de8ab6c4ca9ef4ec) should reflect your requested change better.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: Looks great! There is a similar one in `count_observability_sensors`, now that you are already on it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -66,22 +80,28 @@ count_observability_sensors(MeasuredValues<sym> const& measured_values, MathMode
         }
         // the system could be ill-conditioned if there is no flow sensor for one bus, except the last bus
         if (!has_at_least_one_sensor && row != n_bus - 1) {
-            possibly_ill_conditioned = true;
+            is_possibly_ill_conditioned = true;
         }
     }
-    return std::make_tuple(std::move(flow_sensors), std::move(voltage_phasor_sensors), possibly_ill_conditioned);
+    return ObservabilitySensorsResult{flow_sensors, voltage_phasor_sensors, is_possibly_ill_conditioned};
```
mgovers, line 86: i think `std::move` is still necessary here, right? or you need to pre-declare the `ObservabilitySensorResult result;` earlier. Or is RVO that good these days that we don't need it anymore?

Jerry-Jinfeng-Guo: > Or is RVO that good these days that we don't need it anymore?

I think so yes, if we do `std::move` we are sure to have a move. RVO are also applied on temporary construct/object like here.

mgovers: RVO apparently isn't good enough yet. see also https://github.com/PowerGridModel/power-grid-model/pull/1033/files#diff-ab2bbf5516616253c0ee3797ccdca9b8fee7125697e9d4cff9d59ac72ab28f63R36-R38

figueroa1395: My understanding is that in this case it is "trivial" for the compiler to do this and that explicitly moving might even make it worse in some cases. If this is so detail-prone, maybe one of us can dig a bit into it and do some knowledge sharing about it. I would personally like to learn more about the caveats and might be useful for our codebase as well.

figueroa1395: To clarify and close this comment. https://github.com/PowerGridModel/power-grid-model/blob/399e51198f71320273f34d3f38eafeb55df86dfe/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp#L36 was the opted "fix" (out of two). So the problem wasn't that the result wasn't doing RVO, but that by doing so we were copying possibly big vectors (as we didn't `std::move` them into the result); in this fix, named RVO was used instead, where the `std::move` is avoided altogether.

Jerry-Jinfeng-Guo: > To clarify and close this comment.
> 
> https://github.com/PowerGridModel/power-grid-model/blob/399e51198f71320273f34d3f38eafeb55df86dfe/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp#L36
> 
> was the opted "fix" (out of two). So the problem wasn't that the result wasn't doing RVO, but that by doing so we were copying possibly big vectors (as we didn't `std::move` them into the result); in this fix, named RVO was used instead, where the `std::move` is avoided altogether.

I think this is the most reasonable explanation, thanks!

PR #1025, opened 2025-06-18
Linting and Formatting: adding Pylint checks to Ruff

Laurynas-Jagutis:
This implementation is based on a version of the repository before moving to Ruff, which can be accessed here:
https://github.com/PowerGridModel/power-grid-model/tree/c127b2cf42ff179fe985f91688e10a7cf2671a90

As per the previous implementation, Pylint only ran on src files, thus I tried to mimic the behavior:
https://github.com/PowerGridModel/power-grid-model/blob/c127b2cf42ff179fe985f91688e10a7cf2671a90/.pre-commit-config.yaml

Most of the noqa comments alligned with the pylint version:
https://github.com/PowerGridModel/power-grid-model/blob/c127b2cf42ff179fe985f91688e10a7cf2671a90/src/power_grid_model/validation/_rules.py

Most misalignments happened in src/power_grid_model/_core/power_grid_model.py, where the old pylint version did not have comments but now there was a need for them. A possible explanation could be: https://github.com/pylint-dev/pylint/issues/8667

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,8 +113,9 @@ select = [
     "FURB",
     "FLY",
     "SLOT",
+    "PL"
 ]
-ignore = ["SIM108", "SIM118", "SIM110", "SIM211"]
+ignore = ["SIM108", "SIM118", "SIM110", "SIM211", "PLR2004"]
 
 [tool.ruff.lint.isort]
 # Imports that are imported using keyword "as" and are from the same source - are combined.
@@ -123,6 +124,8 @@ combine-as-imports = true
 [tool.ruff.lint.per-file-ignores]
 # Ignore `F811` (redefinition violations) in all examples notebooks since we use redefinition.
 "docs/examples/*.ipynb" = ["F811", "E402"]
+# Pylint was only run in src directory before moving to Ruff
+"tests/*" = ["PL"]
 
 [tool.mypy]
 follow_imports = "silent"
--- a/src/power_grid_model/_core/power_grid_core.py
+++ b/src/power_grid_model/_core/power_grid_core.py
@@ -466,7 +466,7 @@ class PowerGridCore:
         pass  # pragma: no cover
 
     @make_c_binding
-    def dataset_mutable_add_buffer(  # type: ignore[empty-body]
+    def dataset_mutable_add_buffer(  # type: ignore[empty-body]  # noqa: PLR0913
         self,
         dataset: MutableDatasetPtr,
         component: str,
--- a/src/power_grid_model/_core/power_grid_model.py
+++ b/src/power_grid_model/_core/power_grid_model.py
@@ -238,7 +238,7 @@ class PowerGridModel:
             decode_error=decode_error,
         )
 
-    def _calculate_impl(
+    def _calculate_impl(  # noqa: PLR0913
         self,
         calculation_type: CalculationType,
         symmetric: bool,
@@ -303,7 +303,7 @@ class PowerGridModel:
 
         return output_data
 
-    def _calculate_power_flow(
+    def _calculate_power_flow(  # noqa: PLR0913
         self,
         *,
         symmetric: bool = True,
@@ -340,7 +340,7 @@ class PowerGridModel:
             experimental_features=experimental_features,
         )
 
-    def _calculate_state_estimation(
+    def _calculate_state_estimation(  # noqa: PLR0913
         self,
         *,
         symmetric: bool = True,
@@ -375,7 +375,7 @@ class PowerGridModel:
             experimental_features=experimental_features,
         )
 
-    def _calculate_short_circuit(
+    def _calculate_short_circuit(  # noqa: PLR0913
         self,
         *,
         calculation_method: CalculationMethod | str = CalculationMethod.iec60909,
@@ -409,7 +409,7 @@ class PowerGridModel:
             experimental_features=experimental_features,
         )
 
-    def calculate_power_flow(
+    def calculate_power_flow(  # noqa: PLR0913
         self,
         *,
         symmetric: bool = True,
@@ -508,7 +508,7 @@ class PowerGridModel:
             tap_changing_strategy=tap_changing_strategy,
         )
 
-    def calculate_state_estimation(
+    def calculate_state_estimation(  # noqa: PLR0913
         self,
         *,
         symmetric: bool = True,
@@ -602,7 +602,7 @@ class PowerGridModel:
             decode_error=decode_error,
         )
 
-    def calculate_short_circuit(
+    def calculate_short_circuit(  # noqa: PLR0913
         self,
         *,
         calculation_method: CalculationMethod | str = CalculationMethod.iec60909,
--- a/src/power_grid_model/validation/_rules.py
+++ b/src/power_grid_model/validation/_rules.py
@@ -241,7 +241,7 @@ def all_less_or_equal(
     return none_match_comparison(data, component, field, not_less_or_equal, ref_value, NotLessOrEqualError)
 
 
-def all_between(
+def all_between(  # noqa: PLR0913
     data: SingleDataset,
     component: ComponentType,
     field: str,
@@ -281,7 +281,7 @@ def all_between(
     )
 
 
-def all_between_or_at(
+def all_between_or_at(  # noqa: PLR0913
     data: SingleDataset,
     component: ComponentType,
     field: str,
@@ -331,7 +331,7 @@ def all_between_or_at(
     )
 
 
-def none_match_comparison(
+def none_match_comparison(  # noqa: PLR0913
     data: SingleDataset,
     component: ComponentType,
     field: str,
@@ -554,7 +554,7 @@ def all_valid_enum_values(
     return []
 
 
-def all_valid_associated_enum_values(
+def all_valid_associated_enum_values(  # noqa: PLR0913
     data: SingleDataset,
     component: ComponentType,
     field: str,
--- a/src/power_grid_model/validation/_validation.py
+++ b/src/power_grid_model/validation/_validation.py
@@ -279,7 +279,7 @@ def _process_power_sigma_and_p_q_sigma(
         power_sigma[mask] = np.nansum(q_sigma[mask], axis=asym_axes)
 
 
-def validate_required_values(
+def validate_required_values(  # noqa: PLR0915
     data: SingleDataset, calculation_type: CalculationType | None = None, symmetric: bool = True
 ) -> list[MissingValueError]:
     """
@@ -631,7 +631,7 @@ def validate_branch3(data: SingleDataset, component: ComponentType) -> list[Vali
     return errors
 
 
-def validate_three_winding_transformer(data: SingleDataset) -> list[ValidationError]:
+def validate_three_winding_transformer(data: SingleDataset) -> list[ValidationError]:  # noqa: PLR0915
     errors = validate_branch3(data, ComponentType.three_winding_transformer)
     errors += _all_greater_than_zero(data, ComponentType.three_winding_transformer, "u1")
     errors += _all_greater_than_zero(data, ComponentType.three_winding_transformer, "u2")


pyproject.toml
```diff
@@ -113,8 +113,10 @@ select = [
     "FURB",
     "FLY",
     "SLOT",
+    "PL",
     "NPY",
 ]
+ignore = ["PLR2004"]
```
figueroa1395: Ohh! This one is about magic numbers, right? I think we should address this, not ignore it. What is the reasoning behind ignoring it here?

pyproject.toml
```diff
@@ -123,6 +125,9 @@ combine-as-imports = true
 [tool.ruff.lint.per-file-ignores]
 # Ignore `F811` (redefinition violations) in all examples notebooks since we use redefinition.
 "docs/examples/*.ipynb" = ["F811", "E402"]
+# Pylint was only run in src directory before moving to Ruff
+"tests/*" = ["PL"]
+"setup.py" = ["PL"]
```
figueroa1395: Do we want to consider extending this check at some point? Or this hasn't been discussed? I see the reasoning behind keeping as is, but I am just curious.

TonyXiang8787: `setup.py` will be deleted in https://github.com/PowerGridModel/power-grid-model/pull/1019.

Laurynas-Jagutis: This specifically hasn't been discussed. We initially agreed to make ruff and its checks as close as possible to the setup that was before, thus the choice here to only allow pylint in src.

Nonetheless, I have fixed some PL checks in the tests, mainly: PLR1736 and PLR5501.

There are still 40+ errors in tests. 30 of those are PLR2004(magic numbers). The rest are a mix of:
PLR0915 Too many statements (173 > 50)
PLR0912 Too many branches (18 > 12)
PLR0913 Too many arguments in function definition (7 > 5)
PLC0206 Extracting value from dictionary without calling `.items()`

mgovers: this one should be fixed

> PLC0206 Extracting value from dictionary without calling .items()

please ignore the remaining ones:

> PLR0915 Too many statements (173 > 50)
> PLR0912 Too many branches (18 > 12)
> PLR0913 Too many arguments in function definition (7 > 5)

src/power_grid_model/_core/power_grid_core.py
```diff
@@ -460,7 +460,7 @@ def destroy_dataset_mutable(self, dataset: MutableDatasetPtr) -> None:  # type:
         pass  # pragma: no cover
 
     @make_c_binding
-    def dataset_mutable_add_buffer(  # type: ignore[empty-body]
+    def dataset_mutable_add_buffer(  # type: ignore[empty-body]  # noqa: PLR0913
```
figueroa1395, line 463: We have `# nova: PLR0913` in many different places throughout the codebase, but that is by design. Should we consider ignoring it globally? Or do we still want to get a "warning" and try to keep function signatures as short as possible? @Laurynas-Jagutis @mgovers

src/power_grid_model/_core/buffer_handling.py
```diff
@@ -185,7 +186,9 @@ def _get_dense_buffer_properties(
     if actual_ndim not in (1, 2):
         raise ValueError(f"Array can only be 1D or 2D. {VALIDATOR_MSG}")
 
-    actual_is_batch = actual_ndim == 2
+    batch_ndim = 2
+
+    actual_is_batch = actual_ndim == batch_ndim
     actual_batch_size = shape[0] if actual_is_batch else 1
```
figueroa1395: Maybe a better name would be something explicitly suggesting 2d? and then in line 192, use the same name but with 1d explicit instead of `... else 1`? I don't know of an exact name, but reflecting the value in it would be nice.

src/power_grid_model/_core/utils.py
```diff
@@ -168,13 +171,13 @@ def get_batch_size(
                 break
             if array.ndim == 1:
                 raise TypeError("Incorrect dimension present in batch data.")
-            if array.ndim == 2:
+            if array.ndim == batch_ndim:
```
figueroa1395: Similar to above, maybe we can have some "globals" indicating the dimension of batch data, regular data, etc. I imagine these magic numbers (1, 2, 3) referring to the possible dimensions are spread in many places.

If in the end `batch_dim = 2` is just clear, let's define it that way then and put in somewhere. And then come with something for the other two.

Also, what does the return 1 below mean?

src/power_grid_model/validation/errors.py
```diff
@@ -241,6 +246,7 @@ class InvalidValueError(SingleFieldValidationError):
 
     _message = "Field {field} contains invalid values for {n} {objects}."
     values: list
+    __hash__ = None  # type: ignore[assignment]
```
figueroa1395, line 250: Why do we need this and the others below now? Is the intention to make the error type classes unhashable? What for?

src/power_grid_model/_core/power_grid_core.py
mgovers: i think having too many arguments is quite bad, so we should keep the warning.

src/power_grid_model/_core/utils.py
```diff
@@ -42,6 +42,9 @@
 from power_grid_model._core.power_grid_meta import initialize_array, power_grid_meta_data
 from power_grid_model._core.typing import ComponentAttributeMapping, _ComponentAttributeMappingDict
 
+batch_ndim = 2
+unsupported_dim = 3
```
mgovers:
```suggestion
unsupported_ndim = 3
```

tests/unit/test_buffer_handling.py
```diff
@@ -13,6 +13,10 @@
 from power_grid_model._core.dataset_definitions import ComponentType, DatasetType
 from power_grid_model._core.power_grid_meta import initialize_array, power_grid_meta_data
 
+batch_ndim = 2
+scenario_total_elements = 4
+batch_total_elements = 8
```
mgovers: here and elsewhere: module-level constants shall be capitalized according to https://peps.python.org/pep-0008/#constants

mgovers (commented):
as you may see in this PR, even giving magic values a localized name does not necessarily make the code nicer. However, it does highlight the underlying design issues, and provides a way out and a first step towards refactoring the code in the future.

src/power_grid_model/_core/buffer_handling.py
Laurynas-Jagutis: For now I still kept batch_ndim = 2, but I have added singular_ndim = 1. 
If you think that for example these would be more clear:
batch_2dim = 2 and singular_1dim = 1 I could edit them

src/power_grid_model/_core/buffer_handling.py
```diff
@@ -117,7 +117,8 @@ def _get_raw_attribute_data_view(data: np.ndarray, schema: ComponentMetaData, at
     Returns:
         a raw view on the data set.
     """
-    if schema.dtype[attribute].shape == (3,) and data.shape[-1] != 3:
+    supported_dim = 3
+    if schema.dtype[attribute].shape == (supported_dim,) and data.shape[-1] != supported_dim:
```
mgovers: This is both not descriptive enough and also not correct. This is exactly why you should be very careful with magic values. I myself did not understand what it was about until I dug more deeply into this.

1. `3` is NOT a supported amount of dimensions, UNLESS a very specific case. Therefore, `supported_dim` is not a good name.
2. Only 1-dimensional and 2-dimensional datastructures are supported. There is only 1 exception and that is when we're dealing with asymmetric values in dense batch attribute data, in which case you have 3 dimensions and the last dimension contains exactly 3 values
3. This also means that `if` statement is incorrect (i.e. we found a bug)

Here's my proposal:


```suggestion
    dense_batch_ndim = 2
    dense_batch_asym_attr_ndim = dense_batch_ndim + 1
    n_asym_values = 3
    shape = schema.dtype[attribute].shape
    if len(shape) > batch_ndim and not (len(shape) == dense_batch_asym_attr_ndim and shape[-1] != n_asym_values):
```
you can also invert the if statement:

```py
    dense_batch_ndim = 2
    dense_batch_asym_attr_ndim = dense_batch_ndim + 1
    n_asym_values = 3
    shape = schema.dtype[attribute].shape
    if len(shape) <= batch_ndim or (len(shape) == dense_batch_asym_attr_ndim and shape[-1] != n_asym_values):
        return _get_raw_data_view(data, dtype=schema.dtype[attribute].base)
    raise ValueError("Given data has a different schema than supported.")
```

Please do this fix in a separate PR off `main` together with a unit test that tries to trigger the issue from the lowest-level public function that calls this function (i.e., `get_buffer_view`).

I recommend first creating the unit tests (don't forget to commit when if fails correctly) and then applying the fix

src/power_grid_model/_core/utils.py
```diff
@@ -42,6 +42,10 @@
 from power_grid_model._core.power_grid_meta import initialize_array, power_grid_meta_data
 from power_grid_model._core.typing import ComponentAttributeMapping, _ComponentAttributeMappingDict
 
+SINGULAR_NDIM = 1
```
mgovers: not a fan of the name `SINGULAR`. Maybe something like `SINGLE_DATASET_NDIM` (and `BATCH_DATASET_NDIM`, resp.)?

src/power_grid_model/_core/utils.py
```diff
@@ -42,6 +42,10 @@
...
+BATCH_NDIM = 2
+UNSUPPORTED_NDIM = 3
```
mgovers: creating this on a global level doesn't make sense. In fact, all amount of dimensions above `BATCH_NDIM` (or below 0, for all that matters) are unsupported and it is very prone to bugs (as is shown in my other comment).

src/power_grid_model/_core/utils.py
```diff
@@ -166,17 +170,17 @@ def get_batch_size(
...
                 raise TypeError("Incorrect dimension present in batch data.")
-            if array.ndim == 2:
+            if array.ndim == BATCH_NDIM:
                 return 1
             return array.shape[0]
         sym_array = next(iter(batch_data.values()))
 
     sym_array = cast(DenseBatchArray | BatchColumn, sym_array)
-    if sym_array.ndim == 3:
+    if sym_array.ndim == UNSUPPORTED_NDIM:
```
mgovers: this seems like a bug to me

```suggestion
    if sym_array.ndim > BATCH_NDIM:
```
or (but please double-check whether `ndim == 0` is supported, or we may just introduce a new bug)
```suggestion
    if sym_array.ndim not in (SINGULAR_NDIM, BATCH_NDIM):
```

src/power_grid_model/validation/errors.py
Laurynas-Jagutis: This was the error raised: https://docs.astral.sh/ruff/rules/eq-without-hash/

A class that implements __eq__ but not __hash__ will have its hash method implicitly set to None - so this was the case in our code for a long time, the only difference is that ruff picked it up and raised a warning to make it explicit.

Then based on this: https://docs.python.org/3/reference/datamodel.html#object.__hash__  I checked the use cases for the errors, the only use cases I found were using them in lists, and not in dicts, sets or frozensets 😄 so I thought that unhashable is a good option.

mgovers: digging into this myself, rather than use cases, i think the main reason why `__hash__` should be `None` is that `ValidationError`s contain `ids` which may be updateable at runtime and affect the effective value that the validation error represents. Hence, it cannot be hashable.

I accept the new change.

figueroa1395: Clear, resolving.

src/power_grid_model/_core/utils.py
Laurynas-Jagutis: I chose the `if sym_array.ndim not in (SINGULAR_NDIM, BATCH_NDIM):` option

I have double checked whether ndim == 0 is supported using the test: test_get_batch_size

I supplied it a setup of `pytest.param(np.empty(shape=()), None, 0, id="re")`.
For that setup I ran into an error message `Expecting a 1D/2D Numpy structured array or a dictionary of such.`

src/power_grid_model/_core/utils.py
Laurynas-Jagutis: ruff checks magic values in comparisons only: https://docs.astral.sh/ruff/rules/magic-value-comparison/
Because of that there are no such errors/warning in return statements.

1 is a return value which shows the number of batches in the data, I don't think that we need to give it a name

src/power_grid_model/validation/errors.py
```diff
@@ -14,6 +14,9 @@
 from power_grid_model import ComponentType
 from power_grid_model._core.dataset_definitions import DatasetType
 
+MIN_FIELDS = 2
+MIN_COMPONENTS = 2
```
mgovers: please make private
```suggestion
_MIN_FIELDS = 2
_MIN_COMPONENTS = 2
```

mgovers (commented):
one small nitpick. Everything else looks OK

src/power_grid_model/_core/utils.py
mgovers: OK

mgovers (dismissed):

mgovers (dismissed):

mgovers (approved):

Issue #765, opened 2024-10-07
[FEATURE] Support local current measurement for Newton-Raphson SE


petersalemink95:
See #547 for the full feature request. When implementing this the local current measurement will already be supported for iterative linear SE, see #764.

This feature request implements only the local calculation for the Newton-Raphson SE calculation.

PR #1028, opened 2025-06-27
Current sensor: NRSE local angle implementation

figueroa1395:
Closes https://github.com/PowerGridModel/power-grid-model/issues/765 and related to https://github.com/PowerGridModel/power-grid-model/issues/547.

Implementation and tests for the local angle current sensor for Newton Raphson State Estimation (NRSE).

Checklist:

- [x] Implementation.
- [x] Unit tests.
- [x] Validation tests.
~~- [ ] Potentially remove experimental feature flag.~~

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
@@ -336,7 +336,7 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
                                                                          measurement);
                                 break;
                             case AngleMeasurementType::global_angle:
-                                throw SparseMatrixError{};
+                                throw NotImplementedError{};
                                 break;
                             default:
                                 assert(measurement.angle_measurement_type == AngleMeasurementType::local_angle ||
@@ -354,7 +354,7 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
                                                                          measurement);
                                 break;
                             case AngleMeasurementType::global_angle:
-                                throw SparseMatrixError{};
+                                throw NotImplementedError{};
                                 break;
                             default:
                                 assert(measurement.angle_measurement_type == AngleMeasurementType::local_angle ||
@@ -536,7 +536,7 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
     }
 
     void multiply_add_branch_blocks(NRSEGainBlock<sym>& block, NRSEGainBlock<sym>& diag_block, NRSERhs<sym>& rhs_block,
-                                    auto& left_block, auto const& right_block, auto const& measured_flow,
+                                    auto& left_block, auto const& right_block, DecomposedComplexRandVar<sym> const& measured_flow,
                                     auto const& f_x_complex) {
         auto const& block_F_T_k_w = transpose_multiply_weight(left_block, measured_flow);
 
@@ -659,7 +659,7 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
     /// @param jac_block F_k(u1, u2, y12)
     /// @param measured_flow object with members p_variance and q_variance
     /// @return  F_k(u1, u2, y12)^T . W
-    NRSEJacobian transpose_multiply_weight(NRSEJacobian const& jac_block, auto const& measured_flow) {
+    NRSEJacobian transpose_multiply_weight(NRSEJacobian const& jac_block, DecomposedComplexRandVar<sym> const& measured_flow) {
         auto const w_p = diagonal_inverse(measured_flow.real_component.variance);
         auto const w_q = diagonal_inverse(measured_flow.imag_component.variance);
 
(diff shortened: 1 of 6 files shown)


tests/cpp_unit_tests/test_math_solver_common.hpp
```diff
@@ -61,11 +61,12 @@ inline void assert_output(SolverOutput<sym> const& output, SolverOutput<sym> con
...
 template <symmetry_tag sym_type> struct SteadyStateSolverTestGrid {
     /*
-    network, v means voltage measured, p means power measured, pp means double measured
+    network, v means voltage measured, p means power measured, pp means double measured,
+    cl means local current measured
     variance always 1.0
                                                           shunt0 (ys) (p)
      (pp)                     (y0, ys0)           (y1)         |
-    source --yref-- bus0(vp) -p-branch0-pp- bus1 --branch1-p-  bus2(vv)
+    source --yref-- bus0(vp) -p-branch0-pp- bus1 -cl-branch1-p-  bus2(vv)
```
nitbharambe: update this

nitbharambe:
> Is it true that there is no way but do `auto` in functions like `process_branch_power_measurement`, `process_branch_local_current_measurement` etc? Either parameters list or inside.

To be addressed in upcoming PR

Jerry-Jinfeng-Guo (approved):
LGTM

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -311,15 +314,52 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
...
+                            auto const& measurement = measured_values.branch_from_current(obj);
+                            switch (measurement.angle_measurement_type) {
+                            case AngleMeasurementType::local_angle:
+                                process_branch_local_current_measurement(block, diag_block, rhs_block, y_branch.yff(),
+                                                                         y_branch.yft(), u_state, ij_voltage_order,
+                                                                         measurement);
+                                break;
+                            case AngleMeasurementType::global_angle:
+                                throw SparseMatrixError{};
+                                break;
```
mgovers: no need to break in unreachable code
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -311,15 +314,52 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
...
+                            auto const& measurement = measured_values.branch_to_current(obj);
+                            switch (measurement.angle_measurement_type) {
+                            case AngleMeasurementType::local_angle:
+                                process_branch_local_current_measurement(block, diag_block, rhs_block, y_branch.ytt(),
+                                                                         y_branch.ytf(), u_state, ij_voltage_order,
+                                                                         measurement);
+                                break;
+                            case AngleMeasurementType::global_angle:
+                                throw SparseMatrixError{};
+                                break;
```
mgovers: idem

```suggestion
```

tests/cpp_unit_tests/test_math_solver_se.hpp
```diff
@@ -374,22 +521,15 @@ TEST_CASE_TEMPLATE_DEFINE("Test math solver - SE, measurements", SolverType, tes
                     {.angle_measurement_type = AngleMeasurementType::local_angle,
                      .measurement = {.real_component = {.value = 1.97, .variance = 0.05},
                                      .imag_component = {.value = 0.0, .variance = 0.05}}}};
+                // We can't compare ILSE with NRSE one to one
+                if (!SolverType::is_NRSE_solver) {
```
mgovers, line 525: not sure whether this is the right way to resolve this. would it be possible to provide 1 optimum with 2 different error tolerances instead (like we do in the validation cases)? or does it really end up in a different local optimum?

in particular, i did not come up with these numbers randomly. Instead, i took the power sensor version and converted it to these values manually. So if you get different optima, it may indicate a bug somewhere in the code.

figueroa1395: Closing with summary of offline discussion:

1-1 testing with power and current sensors works straightforward for ILSE if the variances are the same and the voltage is set to 1, as both sensors become the same thing in practice. However, for NRSE, the Jacobian for power and current sensors is different by definition, so one can't related them even if the voltage is the to 1 and the variances are equal. For this reason, testing in this way was scratched for the case of current sensors with NRSE.

Furthermore, to test that this was done correctly, a manual test (https://github.com/PowerGridModel/power-grid-model/blob/b25a40d2c9ba9de3390b71443fc25da8129254a7/tests/cpp_unit_tests/test_math_solver_se.hpp#L179, https://github.com/PowerGridModel/power-grid-model/blob/b25a40d2c9ba9de3390b71443fc25da8129254a7/tests/cpp_unit_tests/test_math_solver_se.hpp#L269 and https://github.com/PowerGridModel/power-grid-model/blob/b25a40d2c9ba9de3390b71443fc25da8129254a7/tests/cpp_unit_tests/test_math_solver_se.hpp#L357) for current sensors was implemented. Where a grid is manually calculated and the power results from the grid are used as SE inputs, where then the NRSE results are compared with the by hand ones. This proved that the implementation was correct.

figueroa1395:
Experimental flag to be removed once global current sensor is implemented.

PR #1030, opened 2025-07-06
NRSE current sensor: Global angle current sensor

nitbharambe:
**Implements issue**:  #547 

- Current sensor implementation. Workout yet to be reviewed.


mgovers:
do-not-merge label added because the mathematical workout needs to be reviewed as well

tests/cpp_unit_tests/test_math_solver_se.hpp
```diff
@@ -366,6 +420,16 @@ TEST_CASE_TEMPLATE_DEFINE("Test math solver - SE", SolverType, test_math_solver_
...
         }
+
+        SUBCASE("Test se with global angle current sensors") {
+            SolverType solver{y_bus, topo_ptr};
+            CalculationInfo info;
+            SolverOutput<sym> output;
+
+            auto const se_input = grid.se_input_angle_global_current_sensors();
+            output = run_state_estimation(solver, y_bus, se_input, error_tolerance, num_iter, info);
+            assert_output(output, grid.output_ref());
```
mgovers:
```suggestion
            auto const se_input = grid.se_input_angle_global_current_sensors();
            SolverOutput<sym> const output = run_state_estimation(solver, y_bus, se_input, error_tolerance, num_iter, info);
            assert_output(output, grid.output_ref());
```

tests/cpp_unit_tests/test_math_solver_se.hpp
```diff
@@ -571,8 +635,7 @@ TEST_CASE_TEMPLATE_DEFINE("Test math solver - SE, measurements", SolverType, tes
                 CHECK(real(output.branch[0].i_f) == doctest::Approx(real(1.95 * global_shift)));
                 CHECK(imag(output.branch[0].i_f) == doctest::Approx(imag(1.95 * global_shift)));
             } else {
-                CHECK_THROWS_AS(run_state_estimation(solver, y_bus_sym, se_input, error_tolerance, num_iter, info),
-                                SparseMatrixError);
+                CHECK_NOTHROW(run_state_estimation(solver, y_bus_sym, se_input, error_tolerance, num_iter, info));
```
mgovers: we probably should do some more in-depth standalone value comparisons. i know we had this discussions before for the local current sensor NRSE, but only checking `NOTHROW` doesn't really make sense.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -533,6 +541,35 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
         }
     }
 
+    void process_branch_global_current_measurement(NRSEGainBlock<sym>& block, NRSEGainBlock<sym>& diag_block,
```
figueroa1395, line 545: Nitpick: Can you add a similar comment above the function definition? Similar to what is done for `process_branch_local_current_measurement`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -533,6 +541,35 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
...
+        block_rr_or_cc.dP_dt += -imag(current_chi_chi);
+        block_rr_or_cc.dQ_dt += real(current_chi_chi);
+        block_rr_or_cc.dP_dv += real(current_chi_chi);
+        block_rr_or_cc.dQ_dv += imag(current_chi_chi);
+
+        NRSEJacobian block_rc_or_cr{};
+        block_rc_or_cr.dP_dt += -imag(current_chi_psi);
+        block_rc_or_cr.dQ_dt += real(current_chi_psi);
+        block_rc_or_cr.dP_dv += real(current_chi_psi);
+        block_rc_or_cr.dQ_dv += imag(current_chi_psi);
```
figueroa1395: Can you make an overload of the function `calculate_jacobian` for this? Only relevant for global angle current sensors. This is to keep this function consistent with the rest, and would also avoid repeated code.

tests/cpp_unit_tests/test_math_solver_se.hpp
```diff
@@ -240,6 +240,60 @@ template <symmetry_tag sym_type> struct SESolverTestGrid : public SteadyStateSol
         return result;
     }
 
+    auto se_input_angle_global_current_sensors() const {
```
figueroa1395: Nitpick: Can we potentially combine this with above? Or maybe make a "utils" function that creates symmetric/asymmetric sensors (organized in this way), with an argument that defines if they are global or local, and another argument with the contents of the values (branch_from_local_current for local, and output_reference.branch for global)? If this is too much work or you think it would be convoluted, please ignore it. It is very nitpicky and tests should be as readable as possible.

tests/cpp_unit_tests/test_math_solver_se.hpp
figueroa1395: To add: can you also apply the suggested change to the above tests?

tests/cpp_unit_tests/test_math_solver_se.hpp
figueroa1395: Since above we have proper tests for the exact values, I would say we just remove this for NRSE then, and instead make a subtest intended only to test that the call is being done. If we come up with the exact values, we would have to "duplicate" the test above.

figueroa1395 (commented):
I just had a few minor comments, but the math part of the implementation follows the math workout to the dot. I see no problems regarding that.

In addition, I would say that we remove the experimental feature flag in a different PR after Tony has taken a look.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
nitbharambe: Ill add a new function. No need to overload as they are not related as much.

tests/cpp_unit_tests/test_math_solver_se.hpp
nitbharambe: Removed the nothrow check

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -533,6 +541,28 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
         }
     }
 
+    /// @brief Adds contribution of global angle branch current measurements. Logic is similar to power measurements,
+    /// but with an additional voltage division as per mathematical workout.
```
figueroa1395: In this case we don't follow the same logic as in the power sensor and divide by the voltage magnitude, right? We instead just follow the "direct" definition of current via the y-bus and the voltage. Maybe the message can be briefly adjusted.

tests/cpp_unit_tests/test_math_solver_se.hpp
```diff
@@ -176,59 +176,68 @@ template <symmetry_tag sym_type> struct SESolverTestGrid : public SteadyStateSol
...
-                                                 return cabs(b.i_t) * exp(1.0i * arg(b.s_t));
-                                             });
+            output_reference.branch | std::ranges::views::transform([angle_measurement_type](auto const& b) {
+                if (angle_measurement_type == AngleMeasurementType::local_angle) {
+                    return static_cast<ComplexValue<sym>>(cabs(b.i_f) * exp(1.0i * arg(b.s_f)));
+                }
+                return b.i_f;
+            });
+
+        auto const branch_to_local_current =
```
figueroa1395:
```suggestion
        auto const branch_to_current =
```

Same for `auto const branch_from_local_current =`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
```diff
@@ -549,10 +549,12 @@ template <symmetry_tag sym_type> class NewtonRaphsonSESolver {
...
         ComplexTensor<sym> const current_chi_chi = dot(y_xi_xi, ComplexDiagonalTensor<sym>{u_state.u_chi(order)});
         ComplexTensor<sym> const current_chi_psi = dot(y_xi_mu, ComplexDiagonalTensor<sym>{u_state.u_psi(order)});
+        ComplexTensor<sym> const current_chi_chi_v_inv = dot(current_chi_chi, u_state.abs_u_chi_inv(order));
+        ComplexTensor<sym> const current_chi_psi_v_inv = dot(current_chi_psi, u_state.abs_u_psi_inv(order));
         ComplexValue<sym> const f_x_complex = sum_row(current_chi_chi + current_chi_psi);
 
-        auto const block_rr_or_cc = calculate_jacobian_global_current(current_chi_chi);
-        auto const block_rc_or_cr = calculate_jacobian_global_current(current_chi_psi);
+        auto const block_rr_or_cc = calculate_jacobian(-current_chi_chi, current_chi_chi_v_inv);
+        auto const block_rc_or_cr = calculate_jacobian(-current_chi_psi, current_chi_psi_v_inv);
```
figueroa1395, line 556: Everything is now written in the same "language". Nice!

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_se_solver.hpp
nitbharambe: I removed the message. Cant think of something different to add. It follows the methodology as the workout. And no additional comments regarding implementation.

figueroa1395 (approved):

mgovers:
cfr. the discussion from yesterday, we can remove the `do-not-merge` label and merge as-is, right?

figueroa1395:
> cfr. the discussion from yesterday, we can remove the `do-not-merge` label and merge as-is, right?

Indeed, we just keep under experimental feature flag for now.

PR #1033, opened 2025-07-07
Clang tidy: fix configuration + resolve newly introduced issues

mgovers:
Apparently, we haven't been running clang-tidy properly for the past 4 months, because of a small typo in the `.clang-tidy` file. This file fixes that and its underlying warning + resolves the newly introduced warnings.

NOTE: the move from raw iterators to range-based for-loops was what enabled the original exclusion to be possible. See also https://github.com/llvm/llvm-project/issues/68213 for an interesting read on the problems with the original exclusion.


.clang-tidy
```diff
@@ -67,8 +67,4 @@ WarningsAsErrors: '*'
 HeaderFilterRegex: '.*'
 
 FormatStyle: file
-
-CheckOptions:
-  - key: performance-unnecessary-value-param.AllowedTypes
-  - value: 'ForwardIterator'
```
mgovers, line 73: this was a typo that caused us to not be able to use clang-tidy for 4 months without us knowing

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
```diff
@@ -20,19 +20,22 @@ constexpr std::array<Branch3Side, 3> const branch3_sides = {Branch3Side::side_1,
...
-    reserve_component<Component>(state, std::distance(begin, end));
+    reserve_component<Component>(state, std::ranges::size(component_inputs));
     // do sanity check on the transformer tap regulator
     std::vector<Idx2D> regulated_objects;
     // loop to add component
-    for (auto it = begin; it != end; ++it) {
-        ComponentView const input = *it;
+
+    for (auto const& input_proxy : component_inputs) {
+        ComponentView const input = [&input_proxy]() -> ComponentView { return input_proxy; }();
```
mgovers, line 37: 🪄 keep `const&` when it's already of type `ComponentInput const&`, or otherwise conversion to `const` value-type when it's a proxy

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
```diff
@@ -20,19 +20,22 @@ constexpr std::array<Branch3Side, 3> const branch3_sides = {Branch3Side::side_1,
 // template to construct components
 // using forward interators
 // different selection based on component type
-template <std::derived_from<Base> Component, class ComponentContainer, std::forward_iterator ForwardIterator>
+template <std::derived_from<Base> Component, class ComponentContainer, std::ranges::forward_range ComponentInputs>
     requires model_component_state_c<MainModelState, ComponentContainer, Component>
-inline void add_component(MainModelState<ComponentContainer>& state, ForwardIterator begin, ForwardIterator end,
+inline void add_component(MainModelState<ComponentContainer>& state, ComponentInputs component_inputs,
```
mgovers: since we now explicitly capture by range view, we do not have the warning about copying iterators anymore

Jerry-Jinfeng-Guo (changes requested):
PR title

tests/cpp_unit_tests/test_math_solver_se.hpp
```diff
@@ -354,7 +354,7 @@ TEST_CASE_TEMPLATE_DEFINE("Test math solver - SE, measurements", SolverType, tes
 
         */
         // to make the distinction between global and local angle current measurement
-        auto const global_shift = phase_shift(1.0 + 1.0i);
+        auto const global_shift = phase_shift(1.0 + 1.0i); // NOLINT(readability-uppercase-literal-suffix)
```
nitbharambe: Maybe we should global ignore this since we would have only 1.0i as a suffix in our code.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -188,25 +188,12 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
     // template to construct components
     // using forward interators
     // different selection based on component type
-    template <std::derived_from<Base> CompType, std::forward_iterator ForwardIterator>
-    void add_component(ForwardIterator begin, ForwardIterator end) {
+    template <std::derived_from<Base> CompType> void add_component(std::ranges::viewable_range auto&& components) {
         assert(!construction_complete_);
-        main_core::add_component<CompType>(state_, begin, end, system_frequency_);
+        main_core::add_component<CompType>(state_, components, system_frequency_);
     }
```
mgovers: less code that is more efficient 🎉

tests/cpp_unit_tests/test_math_solver_se.hpp
mgovers: thinking of that too. it's a false positive...

figueroa1395: I agree that we can ignore this as it is VERY specific. However, from a quick search, it looks like the best practice here would be to do
```
#include <complex_literals>

using namespace std::complex_literals;
```

And apparently that should "fix" the complain, as it makes it explicit that the `i` is intended for imaginary unit. But that is probably just overkill. I am in favor of ignoring globally unless such a specific change is easy (possibly globally as well) and doesn't bring any nasty side effects.

mgovers: unfortunately, the `std::complex_literals` is not working: 
image

ergo: ignoring globally, it is.

.clang-tidy
figueroa1395: So, if I understand correctly, not checking for `const`, `const&` or `&` only for forward iterators caused all this?

mgovers: or at least an attempt to do so, indeed. But it turns out that there's an issue on LLVM that complains about non-`const&` stuff for all template function arguments, even if they should represent iterators, and even if passing by value is actually more efficient than passing by reference. tony had a strong opinion about passing iterators by value, so that's why this change was added. however, there was a typo in that it should have been
```yml
CheckOptions:
  - key: performance-unnecessary-value-param.AllowedTypes
    value: 'ForwardIterator'
```
but it had one too many `-` before `value`, which caused an incomplete key/value pair, which caused the `.clang-tidy` file to not be parsed correctly and therefore, `clang-tidy` was completely skipped

.clang-tidy
```diff
@@ -28,7 +28,10 @@ cppcoreguidelines-*,
 -cppcoreguidelines-pro-bounds-pointer-arithmetic,
 -cppcoreguidelines-pro-type-reinterpret-cast,
 darwin-*,
-hiccp-*,
+hicpp-*,
+-hicpp-explicit-conversions,
+-hicpp-member-init,
```
figueroa1395, line 33: This one we had agreed before on disabling, am I correct? Any idea why are we disabling just now?

.clang-tidy
figueroa1395: But you removed it, shouldn't we keep it but with the typo fixed?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -73,6 +74,7 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
         Proxy(Idx idx, std::span<AttributeBuffer<Data> const> attribute_buffers)
             : idx_{idx}, attribute_buffers_{std::move(attribute_buffers)} {}
 
+        // NOLINTNEXTLINE(cppcoreguidelines-c-copy-assignment-signature,misc-unconventional-assign-operator)
```
figueroa1395: Remind me why is it okay to mark this overload as `const`?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -161,7 +163,6 @@ template <typename T, dataset_type_tag dataset_type> class ColumnarAttributeRang
     constexpr bool empty() const { return size_ == 0; }
     iterator begin() const { return get(0); }
     iterator end() const { return get(size_); }
-    auto iter() const { return std::ranges::subrange{begin(), end()}; }
     auto operator[](Idx idx) const { return *get(idx); }
```
figueroa1395, line 159: Since we now inherit from `std::ranges::view_interface`, I suppose we can now remove the `[]`, `empty()` and potentially other overloads? Or would this be problematic due to the custom `Idx`? Because if we are just inheriting to remove `iter`, maybe we don't have to.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -112,15 +112,15 @@ struct MetaAttribute {
 // meta component
 struct MetaComponent {
     // meta data
-    char const* name;
-    size_t size;
-    size_t alignment;
+    char const* name{nullptr};
```
figueroa1395, line 115: I am all for these, but didn't you just disable this with the addition of `-hicpp-member-init,`? Or is it just to keep clean starting from this, but due to the team decision we shouldn't necessary enforce it?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/input.hpp
figueroa1395: Neat!

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -16,16 +16,16 @@
 namespace power_grid_model::main_core::update {
 
 namespace detail {
-template <component_c Component, std::forward_iterator ForwardIterator, typename Func>
-    requires std::invocable<std::remove_cvref_t<Func>, typename Component::UpdateType, Idx2D const&>
-inline void iterate_component_sequence(Func func, ForwardIterator begin, ForwardIterator end,
-                                       std::span<Idx2D const> sequence_idx) {
-    assert(std::distance(begin, end) >= static_cast<ptrdiff_t>(sequence_idx.size()));
+template <component_c Component, std::ranges::view Elements, typename Func>
```
figueroa1395, line 19: Please share in the tech review all the "new" fancy introduced use of ranges to "replace" the iterators, really neat. However, before we were very specific about forward, backward, etc. iterators, but now we are just taking viewable ranges. It seems fine to me at first glance, and I assume the specificity of the iterators previously used was because we only implemented what was necessary, but could this potentially introduce issues as we are making things more generic in this sense?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
```diff
@@ -419,7 +419,7 @@ template <symmetry_tag sym_type> class IterativeLinearSESolver {
                                           ComplexValue<sym> const& voltage) const {
         using statistics::scale;
 
-        auto const measurement = static_cast<IndependentComplexRandVar<sym>>(current_measurement.measurement);
+        auto measurement = static_cast<IndependentComplexRandVar<sym>>(current_measurement.measurement);
```
figueroa1395, line 422: Why? This seems `const` in this scope.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -328,7 +328,8 @@ inline auto get_edge_weights(TransformerGraph const& graph) -> TrafoGraphEdgePro
             if ((edge_src_rank == infty) || (edge_tgt_rank == infty)) {
                 throw AutomaticTapInputError("The transformer is being controlled from non source side towards source "
                                              "side.\n");
-            } else if (edge_src_rank != edge_tgt_rank - 1) {
+            }
+            if (edge_src_rank != edge_tgt_rank - 1) {
```
figueroa1395, line 332: Did you split because we `throw` if the first branch is reached? Is this good practice?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -188,25 +188,13 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-        add_component<CompType>(components.begin(), components.end());
-    }
-
     // template to construct components
     // using forward interators
     // different selection based on component type
-    template <std::derived_from<Base> CompType, std::forward_iterator ForwardIterator>
-    void add_component(ForwardIterator begin, ForwardIterator end) {
+    template <std::derived_from<Base> CompType, std::ranges::viewable_range Inputs>
+    void add_component(Inputs&& components) {
```
figueroa1395, line 195: Very nice cleanup!

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: yes, indeed. exactly that. it means that there will be less stack unwinding to be done, and also RVO is enabled (because throwing is a separate control logic flow)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
mgovers: not really. the only "regression" is the fact that a viewable range is (input|output) iterator. since the lowest-level functions all `requires std::ranges::forward_range<Type>`, the compiler will stop you. from the usage-level, no such input range restriction was imposed, so it's OK to be generic there. the compiler will just complain when trying to compile the lowest-level stuff with the wrong input type, and because of the use of concepts, the explanation of the error contains a clear reason (because it does not satisfy the input iterator concept)

mgovers: and yes, i will share this in the next tech review

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
mgovers: i went over all cases that were actually bad/spurious and kept the "false positive" ones. NOTE: `hicpp-member-init` is an alias of `google-default-arguments` and it already was explicitly excluded from the configuration. that's also why we didn't hit on this before.

.clang-tidy
mgovers: `hicpp-member-init` and `-hicpp-explict-conversions` are aliases of other checks we have already explicitly disabled before.

the only reason that we only need to disable this now was because of a typo in the group enabler (`hiccp-*` vs `hicpp-*`)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers: i am not 100% satisfied with the current version yet... will get back to you.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers: `empty` overload was required by MacOS, unless we migrate from `std::empty` to the `std::ranges::empty` alternative, which I think is alright but there's not really any specific reason to. NOTE: `view_interface` automatically provides `empty` under certain conditions, but apparently, the MacOS implementation does not satisfy those conditions.

the `operator[]` we definitely need to keep, because it needs to be a `std::ranges::random_access_range`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/iterative_linear_se_solver.hpp
mgovers:
```
error: constness of 'measurement' prevents automatic move [performance-no-automatic-move,-warnings-as-errors]
[build]   426 |             return measurement; // no offset
[build]       |                    ^
```
if `measurement` is `const`, then a copy is enforced on `return`

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers:
```
dataset.hpp:91:20: error: binding reference of type 'Proxy' to value of type 'const power_grid_model::meta_data::ColumnarAttributeRange<power_grid_model::meta_data::(anonymous namespace)::AInput, power_grid_model::mutable_dataset_t>::Proxy' drops 'const' qualifier [clang-diagnostic-error]
[build]    91 |             return *this;
[build]       |
```

the whole thing is that we explicitly want to allow assigning to a `const Proxy`. Usually, the assignment operator can never be marked `const`, because, by definition, it mutates the object.

however, in this specific case, the assignment operator doesn't mutate the proxy itself, but only the underlying data. therefore, this assignment operator overload can exist in this specific context by exception

.clang-tidy
mgovers: 1. fixing the typo actually showed that it doesn't work as intended, because the configuration does not support template parameters.
2. allowing specific types is not necessary anymore because this PR removes the need for them in the first place

so i think removing is the way to go

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers: after investigating together, it seems like we can indeed remove this spurious thing. The reason is that `is_data_mutable` points to the property of the `ColumnarAttributeRange`, not to the `Proxy`. A `const Proxy` therefore should really be `const` and not have an assignment operator overload.

tests/cpp_unit_tests/test_asym_line.cpp
```diff
@@ -10,6 +10,9 @@ namespace power_grid_model {
 
 using namespace std::complex_literals;
 
+namespace {
+// this function exceeds the reasonable threshold of 800 statements
+// NOLINTNEXTLINE(readability-function-size,google-readability-function-size,hicpp-function-size)
```
nitbharambe, line 15: I see google-readability-function-size is an alias. Why is a separate mention of it required?
I also see same for `performance-move-const-arg,hicpp-move-const-arg`

mgovers: Unfortunately. Clang-tidy configuration is not smart enough to understand aliases. Instead, they showup as repeated rules.

PR #1034, opened 2025-07-08
Clang tidy: verify configuration file

mgovers:


mgovers:
ok, so `clang-tidy --verify-config` does not work for us, because:
1. only starting `clang-tidy-19` does it actually complain on bad key/value pairs in the configuration, but we only use `clang-tidy-18` in CI
2. it also complains on missing checks, but we may explicitly disable or enable certain checks for more recent clang-tidy versions that do not exist yet on older versions.

Therefore, this PR does not provide a good prevention for the issue found in #1033

PR #1035, opened 2025-07-09
Current sensor: NRSE local angle docs.

figueroa1395:
This PR adds documentation for NRSE with local angle current sensors, after it was implemented in #1028. In addition, it adds an example of global and local angle current sensors for ILSE, which was missing. It also relates to #547.

- [ ] Fix/add documentation.
- [ ] Add example of global and local angle current sensor for ILSE (this was missing).

### DIFF
--- a/docs/user_manual/components.md
+++ b/docs/user_manual/components.md
@@ -100,22 +100,22 @@ In this case, the attribute `from_status` and `to_status` is always 1.
 | --------- | ----------------- | -------------------------- | -------------------------------------------------------- |
 | `p_from`  | `RealValueOutput` | watt (W)                   | active power flowing into the branch at from-side        |
 | `q_from`  | `RealValueOutput` | volt-ampere-reactive (var) | reactive power flowing into the branch at from-side      |
-| `i_from`  | `RealValueOutput` | ampere (A)                 | current at from-side                                     |
+| `i_from`  | `RealValueOutput` | ampere (A)                 | magnitude of current at from-side                        |
 | `s_from`  | `RealValueOutput` | volt-ampere (VA)           | apparent power flowing at from-side                      |
 | `p_to`    | `RealValueOutput` | watt (W)                   | active power flowing into the branch at to-side          |
 | `q_to`    | `RealValueOutput` | volt-ampere-reactive (var) | reactive power flowing into the branch at to-side        |
-| `i_to`    | `RealValueOutput` | ampere (A)                 | current at to-side                                       |
+| `i_to`    | `RealValueOutput` | ampere (A)                 | magnitude of current at to-side                          |
 | `s_to`    | `RealValueOutput` | volt-ampere (VA)           | apparent power flowing at to-side                        |
 | `loading` | `double`          | -                          | relative loading of the line, `1.0` meaning 100% loaded. |
 
 #### Short circuit output
 
-| name           | data type         | unit       | description                |
-| -------------- | ----------------- | ---------- | -------------------------- |
-| `i_from`       | `RealValueOutput` | ampere (A) | current at from-side       |
-| `i_from_angle` | `RealValueOutput` | rad        | current angle at from-side |
-| `i_to`         | `RealValueOutput` | ampere (A) | current at to-side         |
-| `i_to_angle`   | `RealValueOutput` | rad        | current angle at to-side   |
+| name           | data type         | unit       | description                       |
+| -------------- | ----------------- | ---------- | --------------------------------- |
+| `i_from`       | `RealValueOutput` | ampere (A) | magnitude of current at from-side |
+| `i_from_angle` | `RealValueOutput` | rad        | current angle at from-side        |
+| `i_to`         | `RealValueOutput` | ampere (A) | magnitude of current at to-side   |
+| `i_to_angle`   | `RealValueOutput` | rad        | current angle at to-side          |
 
 ### Line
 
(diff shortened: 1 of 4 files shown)


figueroa1395:
I have made it a draft pull request initially, for a couple of reasons:

- I replaced some arbitrary power sensors with current sensors (both with global and local angles), but the current and angle measured values were entirely made up. It might be desirable (or not) to actually "convert" the values to keep more similar results.
- In the example, both ILSE and NRSE are run, but since current sensors for NRSE is still experimental, I had to replace back the current sensors for power sensors for the NRSE run in order to be able to run it altogether. Is this okay for the time being? Or do we prefer a different way to do this?
- I haven't added (yet) current sensors for the updates regarding the batch calculation and its validation. This just because I am not sure if we want to add more to that part, and because of the two above mentioned points, things might change quite a bit anyway.


Once I have some input from the team, I will proceed with the potential changes and then move away from draft.

docs/user_manual/calculations.md
```diff
@@ -548,9 +548,9 @@ appliances.
 It is not possible to mix [power sensors](./components.md#generic-current-sensor) with
 [current sensors](#./components.mdgeneric-current-sensor) on the same terminal of the same component.
 It is also not possible to mix
-[current sensors with global angle measurement type](#./components.mdgeneric-current-sensor) with
-[current sensors with local angle measurement type](#./components.mdgeneric-current-sensor) on the same terminal of the
-same component.
+[current sensors with global angle measurement type](./components.md#global-angle-current-sensors) with
+[current sensors with local angle measurement type](./components.md#local-angle-current-sensors) on the same terminal 
+of the same component.
```
mgovers, line 553: 😮

docs/user_manual/components.md
```diff
@@ -926,7 +926,8 @@ $$
 ### Generic Current Sensor
 
 ```{warning}
-At the time of writing, state estimation with current sensors is not supported by the Newton-Raphson calculation method.
+At the time of writing, state estimation with current sensors with `global_angle` `angle_measurement_type` is not
+supported by the Newton-Raphson calculation method.
```
mgovers, line 930: here and elsewhere: as long as local current sensor NRSE is experimental, that should also be reflected in the docs

docs/examples/State Estimation Example.ipynb
```diff
@@ -69,12 +70,12 @@
...
     "i.e. the unit of voltage is volt (V), not kV.\n",
     "\n",
-    "Please refer [Components](../user_manual/components.md) for detailed explanation of all component types and their input/output attributes. "
+    "Please refer to [Components](../user_manual/components.md) for a detailed explanation of all component types and their input/output attributes. "
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 4,
```
mgovers: please reset execution counts

docs/examples/State Estimation Example.ipynb
```diff
@@ -533,12 +819,20 @@
...
-    "**Following command does state estimation with Newton Raphson method. Note the difference of result for node compared to the results from iterative linear method.**"
+    "**Following, state estimation is performed with the Newton Raphson method. Note the difference in the results for the nodes compared to the results from the iterative linear method.**"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3b75b0d1",
+   "metadata": {},
+   "source": [
+    "***Note: At the time of writing, current sensors are not yet supported for state estimation with the Newton Raphson calculation method, so they are replaced in the input data with their corresponding power sensor equivalents.***"
```
mgovers: i am thinking that maybe we shouldn't release the new example notebook like this. it's not really nice.

an alternative could be to create a completely separate grid from the original input grid in a new section to show "look, current sensors work as well" but keep the existing documentation that shows the different calculation methods as-is.

Alternatively, we can wait with updating the example notebook until global current angle NRSE is finished

docs/examples/State Estimation Example.ipynb
```diff
@@ -1000,7 +1553,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 11,
+   "execution_count": null,
```
mgovers: autorunning after a failing cell is by default disabled. Instead, you have to manually complete the run

docs/examples/State Estimation Example.ipynb
```diff
@@ -119,24 +120,43 @@
...
     "    MeasuredTerminalType.node,\n",
     "]\n",
-    "sym_power_sensor[\"power_sigma\"] = [1.0e3, 1.0e3, 1.0e3, 1.0e3, 1.0e3, 1.0e3, 1.0e3, 1.0e3]\n",
-    "sym_power_sensor[\"p_measured\"] = [1.73e07, -1.66e07, -3.36e06, 3.39e06, 1.38e07, -1.33e07, 20e6, -10e6]\n",
-    "sym_power_sensor[\"q_measured\"] = [4.07e06, -3.82e06, -1.17e06, 8.86e05, 2.91e06, -2.88e06, 5e6, -2e6]\n",
+    "sym_power_sensor[\"power_sigma\"] = [1.0e3, 1.0e3, 1.0e3, 1.0e3, 1.0e3]\n",
+    "sym_power_sensor[\"p_measured\"] = [-1.66e07, -3.36e06, -1.33e07, 20e6, -10e6]\n",
+    "sym_power_sensor[\"q_measured\"] = [-3.82e06, -1.17e06, -2.88e06, 5e6, -2e6]\n",
+    "# sym_power_sensor[\"p_measured\"] = [1.73e07, -1.66e07, -3.36e06, 3.39e06, 1.38e07, -1.33e07, 20e6, -10e6]\n",
+    "# sym_power_sensor[\"q_measured\"] = [4.07e06, -3.82e06, -1.17e06, 8.86e05, 2.91e06, -2.88e06, 5e6, -2e6]\n",
```
mgovers:
```suggestion
```

docs/examples/State Estimation Example.ipynb
```diff
@@ -119,24 +120,43 @@
...
+    "]\n",
+    "sym_current_sensor[\"angle_measurement_type\"] = [\n",
+    "    AngleMeasurementType.local_angle,\n",
+    "    AngleMeasurementType.global_angle,\n",
+    "    AngleMeasurementType.local_angle,\n",
+    "]\n",
+    "sym_current_sensor[\"i_sigma\"] = [1.0e4, 1.0e4, 1.0e4]\n",
+    "sym_current_sensor[\"i_angle_sigma\"] = [0.1, 0.1, 0.1]\n",
+    "sym_current_sensor[\"i_measured\"] = [1.0e7, -1.0e7, 1.0e7]\n",
+    "sym_current_sensor[\"i_angle_measured\"] = [0.0, 0.0, 0.0]\n",
```
mgovers: that is some EXTREMELY high current :joy: these values (and the sigmas) should be obtained via `conj((p_measured + 1.0i * q_measured) / (voltage_magnitude * np.exp(1.0i * voltage_angle))`

docs/user_manual/components.md
figueroa1395: I'll add it, good catch!

docs/examples/State Estimation Example.ipynb
figueroa1395: Yup. I'll reset everything once I get the final version of the notebook right. Won't resolve until then.

docs/examples/State Estimation Example.ipynb
figueroa1395: This is one of the reasons I wanted some input first 😆, it does indeed look bad.  But since theoretically global is out of scope, but then it is "easy" to implement, let's wait for Peter's discussion with Nitish and possibly their input. A potential solution, after the discussion, can be just to remove the flag already and only show the example with local current sensors.

docs/examples/State Estimation Example.ipynb
figueroa1395: Good catch. Thanks!

docs/examples/State Estimation Example.ipynb
figueroa1395: I know, I just wanted to get a dummy notebook out fast to get feedback on the structure first. If that is set, I'll actually do the conversion to these. But it probably should be done indeed. I just put whatever initially to have a "template".

docs/examples/State Estimation Example.ipynb
figueroa1395: One alternative can be just to demo without explicitly calling global current sensors. The flow is exactly the same and that would keep the notebook clean. That way we can just go with it now in spite of how long would potentially global current sensors take to go out in prod. Thoughts @nitbharambe @petersalemink95?

Edit: And also remove the experimental feature flag on local current sensors.

docs/examples/State Estimation Example.ipynb
figueroa1395: I did the conversion already. It should look better now. The sigmas (that make sense) I just left as dummies as they aren't as important. I based the conversions of on https://github.com/PowerGridModel/power-grid-model/tree/main/tests/data/state_estimation/current-sensor/power-from-global-current-sensor and https://github.com/PowerGridModel/power-grid-model/tree/main/tests/data/state_estimation/current-sensor/power-from-local-current-sensor. Resolving.

docs/examples/State Estimation Example.ipynb
figueroa1395: I will fix it together with the execution count in the end 😃. Keeping this comment unresolved so I don't forget.

docs/examples/State Estimation Example.ipynb
figueroa1395: After offline discussion: Remove experimental feature flag from local current sensors, only show local current sensors in the notebook and add notes/disclaimers for global current sensors for NRSE. I will resolve the comment after that's done.

docs/examples/State Estimation Example.ipynb
figueroa1395: This should be fixed now.

docs/examples/State Estimation Example.ipynb
figueroa1395: Fixed.

docs/examples/State Estimation Example.ipynb
figueroa1395: Done in [460a9fc](https://github.com/PowerGridModel/power-grid-model/pull/1035/commits/460a9fc992672d5f0f6edee3bc530881b305e74f)

docs/user_manual/components.md
```diff
@@ -1019,6 +1021,11 @@ $$
 \underline{I} = \text{i_measured} \cdot e^{j \text{i_angle_measured}} \text{ .}
 $$
 
+```{warning}
+At the time of writing, state estimation with global angle current sensors is not supported by the Newton-Raphson 
+calculation method.
```
mgovers: not strictly necessary but for consistency (and ease of removal when we're finished with global angle as well)

```suggestion
At the time of writing, state estimation with current sensors with `global_angle` `angle_measurement_type` is not
supported by the Newton-Raphson calculation method.
```

docs/examples/State Estimation Example.ipynb
```diff
@@ -533,7 +817,15 @@
    "id": "00be7ae5",
    "metadata": {},
    "source": [
-    "**Following command does state estimation with Newton Raphson method. Note the difference of result for node compared to the results from iterative linear method.**"
+    "**Following, state estimation is performed with the Newton Raphson method. Note the difference in the results for the nodes compared to the results from the iterative linear method.**"
```
mgovers:
```suggestion
    "**In the following, state estimation is performed with the Newton-Raphson method. Note the difference in the results for the nodes compared to the results from the iterative linear method.**"
```

docs/examples/State Estimation Example.ipynb
```diff
@@ -46,6 +46,7 @@
     "import pandas as pd\n",
     "\n",
     "from power_grid_model import (\n",
+    "    AngleMeasurementType,\n",
```
mgovers, line 49: i can't comment on the specific line but in https://github.com/PowerGridModel/power-grid-model/pull/1035/files#diff-f7bceb110f42f15b499f521049d66f1b29c6905270620f16dae7d647010b2fa2L640 , there is one more typo:

`mesurements` > `measurements`

```md
It is also possible to specify the standard deviations of active and reactive power mesurements seperately. This is especially powerful when using the Newton Raphson method for state estimation. Refer to the [Documentation](https://power-grid-model--1035.org.readthedocs.build/en/1035/user_manual/calculations.html) for a detailed explanation. Below is an example of state estimation with seperate standard deviations for p and q measurements.
```

docs/examples/State Estimation Example.ipynb
```diff
@@ -731,32 +1185,26 @@
...
     "sym_power_sensor2[\"p_sigma\"] = [\n",
     "    1.0e3,\n",
-    "    1.0e3,\n",
-    "    1.0e4,\n",
-    "    1.0e3,\n",
     "    1.0e4,\n",
     "    1.0e3,\n",
     "    1.0e4,\n",
     "    1.0e3,\n",
     "]  # define different standard deviations for p and q measurements\n",
```
mgovers: please reformat for readability if possible

```suggestion
    "sym_power_sensor2[\"p_sigma\"] = [1.0e3, 1.0e4, 1.0e3, 1.0e4, 1.0e3]  # define different standard deviations for p and q measurements\n",
```

or

```suggestion
    "# define different standard deviations for p and q measurements\n",
    "sym_power_sensor2[\"p_sigma\"] = [1.0e3, 1.0e4, 1.0e3, 1.0e4, 1.0e3]\n",
```

mgovers (approved):

PR #1036, opened 2025-07-14
Cleanup main model: batch dispatch interface and adaptor

figueroa1395:


power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,120 @@
...
+    JobDispatchAdapter(JobDispatchAdapter const& other)
+        : model_copy_{std::make_unique<MainModel>(other.model_.get())}, model_{std::ref(*model_copy_)} {}
+
+    JobDispatchAdapter& operator=(JobDispatchAdapter const& other) {
+        if (this != &other) {
+            model_copy_ = std::make_unique<MainModel>(other.model_.get());
+            model_ = std::ref(*model_copy_);
+        }
+        return *this;
+    }
```
mgovers: rule of 5


```suggestion
    JobDispatchAdapter(JobDispatchAdapter const& other)
        : model_copy_{std::make_unique<MainModel>(other.model_.get())}, model_{std::ref(*model_copy_)} {}

    JobDispatchAdapter& operator=(JobDispatchAdapter const& other) {
        if (this != &other) {
            model_copy_ = std::make_unique<MainModel>(other.model_.get());
            model_ = std::ref(*model_copy_);
        }
        return *this;
    }

    JobDispatchAdapter(JobDispatchAdapter&& other) = default;
    JobDispatcherAdapter& operator=(JobDispatchAdapter const& other) = default;
    ~JobDispatcherAdapter() = default;
```

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+  private:
+    static constexpr Idx ignore_output{-1};
+
+    friend class JobDispatchInterface<JobDispatchAdapter>;
```
mgovers: first things first + clear comment

```suggestion
    friend class JobDispatchInterface<JobDispatchAdapter>; // CRTP

    static constexpr Idx ignore_output{-1};

```

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,120 @@
...
+    std::unique_ptr<MainModel> model_copy_;
+    std::reference_wrapper<MainModel> model_;
+
+    main_core::utils::SequenceIdx<ComponentType...> current_scenario_sequence_cache_{};
+    std::shared_ptr<const power_grid_model::main_core::utils::ComponentFlags<ComponentType...>> components_to_update_;
+    std::shared_ptr<const power_grid_model::main_core::update::independence::UpdateIndependence<ComponentType...>>
+        update_independence_;
+    std::shared_ptr<const main_core::utils::SequenceIdx<ComponentType...>> all_scenarios_sequence_;
```
mgovers: as mentioned before: not sure whether shared ptr is the way to go here.

Alternatively, make it one `unique_ptr` from the `main` thread, and several raw `&` from other threads

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,120 @@
...
+    void cache_calculate_impl(Calculate&& calculation_fn) const {
+        return std::forward<Calculate>(calculation_fn)(model_.get(),
+                                                       {
+                                                           false,
+                                                           1,
+                                                           "sym_output",
+                                                           model_.get().meta_data(),
+                                                       },
+                                                       ignore_output);
+    }
```
mgovers: 👍

mgovers: as discussed last thursday, we may want to defer this caching to the first scenario of each thread

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+        std::ranges::transform(*update_independence_, independence_flags_.begin(),
+                               [](auto const& comp) { return comp.is_independent(); });
+    }
+
+    void setup_impl(ConstDataset const& update_data, Idx scenario_idx) {
+        assert(components_to_update_ != nullptr);
+        assert(update_independence_ != nullptr);
+        current_scenario_sequence_cache_ =
+            main_core::update::get_all_sequence_idx_map<ComponentType...>( // something wrong here
```
mgovers: TODO

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -44,14 +41,7 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
-                               false,
-                               1,
-                               "sym_output",
-                               model.meta_data(),
-                           },
-                           ignore_output);
+            adapter.cache_calculate(std::forward<Calculate>(calculation_fn));
         } catch (SparseMatrixError const&) { // NOLINT(bugprone-empty-catch) // NOSONAR
             // missing entries are provided in the update data
         } catch (NotObservableError const&) { // NOLINT(bugprone-empty-catch) // NOSONAR
```
mgovers: the try/catch statements should probably also be part of the `cache_calculate`

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    void prepare_sub_batch_calculation_impl(ConstDataset const& update_data) {
+        // cache component update order where possible.
+        // the order for a cacheable (independent) component by definition is the same across all scenarios
+        components_to_update_ =
+            std::make_shared<const power_grid_model::main_core::utils::ComponentFlags<ComponentType...>>(
+                model_.get().get_components_to_update(update_data));
+        update_independence_ = std::make_shared<
+            const power_grid_model::main_core::update::independence::UpdateIndependence<ComponentType...>>(
+            main_core::update::independence::check_update_independence<ComponentType...>(model_.get().state(),
```
mgovers: these are thread-local in the original implementation, but now they are `shared_ptr`. What's the reason?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -62,59 +52,53 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
             };
-            auto model = copy_model_functor(start);
+            auto adapter = copy_adapter_functor(start);
 
-            auto current_scenario_sequence_cache = main_core::utils::SequenceIdx<ComponentType...>{};
-            auto [setup, winddown] =
-                scenario_update_restore(model, update_data, components_to_update, update_independence,
-                                        all_scenarios_sequence_, current_scenario_sequence_cache, infos);
+            auto setup = [&adapter, &update_data, &infos](Idx scenario_idx) {
+                Timer const t_update_model(infos[scenario_idx], 1200, "Update model");
```
mgovers: maybe it's a good idea to move the timing to the adapter as well

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -62,59 +52,53 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
 
         return BatchParameter{};
     }
 
-    template <typename Calculate>
-        requires std::invocable<std::remove_cvref_t<Calculate>, MainModel&, MutableDataset const&, Idx>
-    static auto sub_batch_calculation_(MainModel const& base_model, Calculate&& calculation_fn,
+  private:
+    template <typename Adapter, typename Calculate>
+    static auto sub_batch_calculation_(Adapter& base_adapter, Calculate&& calculation_fn,
```
mgovers: let's consider renaming this to `single_threaded_job`?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+    template <typename UpdateDataset> void setup(UpdateDataset const& update_data, Idx scenario_idx) {
+        return static_cast<Adapter*>(this)->setup_impl(update_data, scenario_idx);
+    }
+
+    void winddown() { return static_cast<Adapter*>(this)->winddown_impl(); }
+
+    CalculationInfo get_calculation_info() const {
+        return static_cast<const Adapter*>(this)->get_calculation_info_impl();
+    }
```
mgovers: here but also in the original implementation: currently returns by value (copy of the entire map). Is that correct?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Was this message because of my comment? I'll remove it, this is where the variadic template issue was being triggered for me, so it's no longer an issue.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: I actually thought about it as well, but decided to go this way instead and keep it at the dispatch level. I know that for testing, for instance, this is irrelevant... but, we are still timing steps for the job dispatcher. It also feels more natural when reading through the code, at least for me. Unless you or anyone has a strong opinion about it, I think it is a good idea to keep here.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: Nice one! Done.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: I like it. Done.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: I was delaying this because I thought perhaps `default` wasn't going to be enough. Let's give it a go. I still have to fix the tests, so... we might as well.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Let's discuss offline tomorrow.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Indeed. I won't resolve so we don't forget, but let's do that as step 2.b. First I want to make sure the tests pass. That is a follow-up improvement.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Incorrect reasoning. Both are no longer `shared_ptr`s and are now thread local.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: I actually removed all `shared_ptr` instances (I think), but `all_scenarios_sequence_` can potentially be "shared" accross threads, contrary to the original implementation. However, first I'll focus on the broken tests...

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
figueroa1395: I believe with the current implementation it is, because we end up using `std::map::merge` so we need a copy, and can't use a view. However, if we refactor a bit, perhaps we can get a view and use something like `std::map::insert` instead, although that would end up creating a copy anyways. This is a possible improvement point, but I'll leave it for latter (2.c). Won't resolve so I remember.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -42,79 +39,60 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
-            auto model = copy_model_functor(start);
+            auto adapter = copy_adapter_functor(start);
 
-            auto current_scenario_sequence_cache = main_core::utils::SequenceIdx<ComponentType...>{};
-            auto [setup, winddown] =
-                scenario_update_restore(model, update_data, components_to_update, update_independence,
-                                        all_scenarios_sequence_, current_scenario_sequence_cache, infos);
+            adapter.prepare_job(update_data);
+            auto setup = [&adapter, &update_data, &infos](Idx scenario_idx) {
+                Timer const t_update_model(infos[scenario_idx], 1200, "Update model");
```
figueroa1395: Even though the tests pass (locally), there is a possible (memory) issue around here. Most likely related to the copy constructor and the valid members in the original adaptor/underlying model. We need to investigate and fix this before considering merging.

figueroa1395:
Even though I am aware there are some known improvements (TODOs in comments - probably best to do at least some as follow ups in a separate PR) and a possible known issue (https://github.com/PowerGridModel/power-grid-model/pull/1036#discussion_r2228552348), I have removed from draft as the CI passes and I feel it is in a good state to gather some input.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Implementation changed and this is no longer relevant.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+    static constexpr Idx ignore_output{-1};
+
+    std::unique_ptr<MainModel> model_copy_;
+    std::reference_wrapper<MainModel> model_;
+
+    power_grid_model::main_core::utils::ComponentFlags<ComponentType...> components_to_update_{};
+    power_grid_model::main_core::update::independence::UpdateIndependence<ComponentType...> update_independence_{};
+    main_core::utils::ComponentFlags<ComponentType...> independence_flags_{};
+    std::shared_ptr<main_core::utils::SequenceIdx<ComponentType...>> all_scenarios_sequence_;
```
figueroa1395: This doesn't need to be a `shared_ptr`, but then we would have to calculate per thread instead of globally. Either is fine, but I wanted to try this. Any alternative to shared ptr is also suggested.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,148 @@
...
+class JobDispatchAdapter : public JobDispatchInterface<JobDispatchAdapter<MainModel, ComponentType...>> {
+  public:
+    JobDispatchAdapter(std::reference_wrapper<MainModel> model) : model_{std::move(model)} {}
+    JobDispatchAdapter(JobDispatchAdapter const& other)
+        : model_copy_{std::make_unique<MainModel>(other.model_.get())},
+          model_{std::ref(*model_copy_)},
+          components_to_update_{other.components_to_update_},
+          update_independence_{other.update_independence_},
+          independence_flags_{other.independence_flags_},
+          all_scenarios_sequence_{other.all_scenarios_sequence_} {}
```
figueroa1395: Applies for this and the rest of the constructors/assignment operators: It is either this or calculating once per thread. We can decide which is best/most efficient, but these 4 should be trivially copiable, so I think is fine. Thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: The problem was that I didn't completely specify the copy constructors... I had forgotten to add the other members. Now this is fixed. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: alright :+1:

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
mgovers: should we create a new GitHub issue and/or Jira ticket/subtask to keep track of this?

power_grid_model_c/power_grid_model/include/power_grid_model/common/two_d_array.hpp
Jerry-Jinfeng-Guo, line 1: See https://github.com/PowerGridModel/power-grid-model/pull/1057

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
mgovers: should we create a new GitHub issue and/or Jira ticket/subtask to keep track of this?

power_grid_model_c/power_grid_model/include/power_grid_model/common/two_d_array.hpp
Jerry-Jinfeng-Guo: Additionally, I do not think the name `two_d_array` is a representative name for the file

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
mgovers: i think using `shared_ptr` for now is fine. maybe, removal of the pre-calculate step will make this redundant anyways.

power_grid_model_c/power_grid_model/include/power_grid_model/common/two_d_array.hpp
mgovers: i may or may not have accidentally checked this in while resolving a merge conflict on an apparently dirty workspace... will revert

figueroa1395: Good catch, I completely missed this. It was reverted by @mgovers in [1a7a572](https://github.com/PowerGridModel/power-grid-model/pull/1036/commits/1a7a5724bb6638d15ef109142d7751ab070c1bfa). Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Potentially. We can evaluate what is best: pre-calculate, and copy the trivial types and keep this `shared_ptr`, or just calculate all per thread.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
figueroa1395: I will create a GitHub issue. We can keep track of what is already done here and what is missing. It can also be used to brainstorm on what is potentially left to do here.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
mgovers: for now, it's fine as-is

mgovers:
profile complete on Windows with the same minimum, slope and peak memory usage. The gradual increase is writing to previously uninitialized memory in the output data cfr. #1055 

## main
image

## this branch
image

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,148 @@
...
+            model_copy_ = std::move(other.model_copy_);
+            model_ = model_copy_ ? std::ref(*model_copy_) : std::move(other.model_);
+            components_to_update_ = std::move(other.components_to_update_);
+            update_independence_ = std::move(other.update_independence_);
+            independence_flags_ = std::move(other.independence_flags_);
+            all_scenarios_sequence_ = std::move(other.all_scenarios_sequence_);
+        }
+        return *this;
+    }
+    ~JobDispatchAdapter() = default;
```
mgovers: sonar-cloud seems to complain about the rule-of-5: https://sonarcloud.io/project/issues?pullRequest=1036&open=AZg8xD18fbtYvPdxb1yr&id=PowerGridModel_power-grid-model&tab=code.

please investigate

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -43,115 +40,85 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
+                adapter.winddown();
+            };
 
             auto calculate_scenario = JobDispatch::call_with<Idx>(
-                [&model, &calculation_fn_, &result_data, &thread_info](Idx scenario_idx) {
-                    calculation_fn_(model, result_data, scenario_idx);
-                    main_core::merge_into(thread_info, model.calculation_info());
+                [&adapter, &calculation_fn_, &result_data, &thread_info](Idx scenario_idx) {
+                    adapter.calculate(calculation_fn_, result_data, scenario_idx);
+                    thread_info.merge(adapter.get_calculation_info());
```
mgovers:
```suggestion
                    main_core::merge_into(thread_info, adapter.get_calculation_info());
```

mgovers:
~Benchmark seems to show a small performance hit, but i guess still within acceptable range (especially considering the `median_delta` remains the same~

EDIT: rerunning solved it. it's within system limitations

### First run
```
Statistics: {'mean_delta': np.float64(0.039678998231168036), 'std_delta': np.float64(1.9202797067740869), 'median_delta': np.float64(5.729999999999998e-05), '95th_percentile_delta': np.float64(0.23094399999999973)}
```
### Second run
```
Statistics: {'mean_delta': np.float64(0.011287913418725221), 'std_delta': np.float64(1.9076096652394496), 'median_delta': np.float64(8.9e-06), '95th_percentile_delta': np.float64(0.05559420000000001)}
```

## Visualization
### First run
benchmark
### Second run
benchmark_rerun

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Addressed. See [1769f1c](https://github.com/PowerGridModel/power-grid-model/pull/1036/commits/1769f1ca1254d2969048b3e5dc78cacc0e3e1de2)

figueroa1395:
For completeness, this is the speed benchmark in an M3 MacBook Pro. Old results are what's in main and the new results are what's in this PR. Looks good to me.

mac_benchmark

In addition, here is the memory profile in main, again in my Mac:
image

Compared to the memory profile in this PR:
image

Which looks fine as well. Hence, the PR is good to go from a performance perspective.

Note: The difference in the memory profiles compared to the ones in Windows come down to how MacOS handles memory differently. This behavior was double checked with `mprof` and were consistent.

mgovers:
ready for review by someone else than santiago and me

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,146 @@
...
+  private:
+    friend class JobDispatchInterface<JobDispatchAdapter>; // CRTP
+
+    static constexpr Idx ignore_output{-1};
+
+    std::unique_ptr<MainModel> model_copy_;
+    std::reference_wrapper<MainModel> model_;
+
+    power_grid_model::main_core::utils::ComponentFlags<ComponentType...> components_to_update_{};
+    power_grid_model::main_core::update::independence::UpdateIndependence<ComponentType...> update_independence_{};
```
nitbharambe:
```suggestion
    main_core::utils::ComponentFlags<ComponentType...> components_to_update_{};
    main_core::update::independence::UpdateIndependence<ComponentType...> update_independence_{};
```

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -43,115 +40,85 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
+    static auto single_thread_job(Adapter& base_adapter, Calculate&& calculation_fn, MutableDataset const& result_data,
+                                  ConstDataset const& update_data, std::vector<std::string>& exceptions) {
+        return [&base_adapter, &exceptions, calculation_fn_ = std::forward<Calculate>(calculation_fn), &result_data,
+                &update_data](Idx start, Idx stride, Idx n_scenarios) {
             assert(n_scenarios <= narrow_cast<Idx>(exceptions.size()));
 
             CalculationInfo thread_info;
 
-            Timer t_total(thread_info, 0000, "Total in thread");
+            Timer t_total(thread_info, 1000, "Total batch calculation in thread");
```
mgovers: this overlaps with `benchmark.cpp:115`:
```cpp
Timer const t_build(info, 1000, "Build model");
```

figueroa1395: `0000` is also taken in the benchmark. Any suggestion on which one to use?

mgovers: maybe `0100`? let's revisit all codes properly when tackling the calculation info

mgovers: nope, `0100` is also taken... `0200` then?

figueroa1395: Sure, I'll just add that one for now. Later we can properly think about this.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -43,115 +40,85 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
-                                        all_scenarios_sequence_, current_scenario_sequence_cache, thread_info);
+            auto adapter = copy_adapter_functor(start);
+
+            auto setup = [&adapter, &update_data, &thread_info](Idx scenario_idx) {
+                Timer const t_update_model(thread_info, 1200, "Update model");
+                adapter.setup(update_data, scenario_idx);
+            };
+
+            auto winddown = [&adapter, &thread_info](Idx /*scenario_idx*/) {
+                Timer const t_update_model(thread_info, 1201, "Restore model");
```
nitbharambe:
```suggestion
                Timer const t_restore_model(thread_info, 1201, "Restore model");
```

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
nitbharambe: Unrelated:

> nope, 0100 is also taken... 0200 then?

Feels like a cue to make the timer numbers as constants later.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,146 @@
...
+            update_independence_ = std::move(other.update_independence_);
+            independence_flags_ = std::move(other.independence_flags_);
+            all_scenarios_sequence_ = std::move(other.all_scenarios_sequence_);
+        }
+        return *this;
+    }
+    ~JobDispatchAdapter() { model_copy_.reset(); }
+
+  private:
+    friend class JobDispatchInterface<JobDispatchAdapter>; // CRTP
```
Jerry-Jinfeng-Guo: I am guessing this comment should appear near the `class JobDispatchAdapter` definition?

Jerry-Jinfeng-Guo:
Is there already test cases that demonstrate how easily this new adapter-interface scheme can be tested with mocks?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: :+1:

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
mgovers, line 1: i'm thinking maybe we should rename `job_dispatch_interface` to `job_interface`, and `job_dispatch_adapter` to `job_adapter`. That way, the `job_dispatch` uses a `job_interface`

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+    void thread_safe_add_calculation_info(CalculationInfo const& info) {
+        static_cast<Adapter*>(this)->thread_safe_add_calculation_info_impl(info);
+    }
+};
```
mgovers: make sure it's not accidentally constructible or destructible in a way that leaks memory

```suggestion

  private:
    JobDispatchInterface() = default;
    JobDispatchInterface(const JobDispatchInterface& /*other*/) = default;
    JobDispatchInterface(JobDispatchInterface&& /*other*/) = default;
    JobDispatchInterface& operator=(const JobDispatchInterface& /*other*/) = default;
    JobDispatchInterface& operator=(JobDispatchInterface&& /*other*/) = default;
    ~JobDispatchInterface() = default;

    friend Adapter;
};
```

or

```suggestion

  protected:
    JobDispatchInterface() = default;
    JobDispatchInterface(const JobDispatchInterface& /*other*/) = default;
    JobDispatchInterface(JobDispatchInterface&& /*other*/) = default;
    JobDispatchInterface& operator=(const JobDispatchInterface& /*other*/) = default;
    JobDispatchInterface& operator=(JobDispatchInterface&& /*other*/) = default;
    ~JobDispatchInterface() = default;
};
```

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
figueroa1395: Sure, I've got no problem with that. Although "job" becomes a bit generic. What about `dispatch_interface`, `dispatch_adapter` and `dispatch`?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: > Unrelated:
> 
> > nope, 0100 is also taken... 0200 then?
> 
> Feels like a cue to make the timer numbers as constants later.

I think so. Maybe we can create an issue for when we target `CalculationInfo` to keep track of this. Just out of the top of my head, making enums out of the codes (while organizing them and "setting a standrd") and removing the string from the lookup part come to mind.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
figueroa1395: We were just talking about this 😆 . Will do!

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
mgovers: I don't think that solves the naming issue. The problem that i have with the current naming is not the length, but the fact that `*dispatch_interface` makes it sound like the `dispatch`(er) **implements** the `dispatch_interface`, instead of **uses** it.

In a way, the `job_adapter` is a `worker` which is dispatched by the `dispatcher` in the main thread. Maybe the name `worker_interface` would suit its purpose better? I do like the name `job`, though.

figueroa1395:
> Is there already test cases that demonstrate how easily this new adapter-interface scheme can be tested with mocks?

Not yet. That would be part 3 of the saga. But in summary, we just need to test the `job_dispatch` calls (which should be easy to mock - just make a fake adapter that "implements" the interface and the result/update datasets can be "empty" `struct`s or something simple), and then potentially evaluate if we need to test the `job_dispatch_adapter` or to what extend to do so, as it is mostly a collection of calls to main model (which should be unit tested); probably constructors and such might need testing and if we do some actual relevant logic as well.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
figueroa1395: I see your point now, and that makes sense. We can go with `job*`. Maybe @nitbharambe or @Jerry-Jinfeng-Guo have any thoughts on this?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Currently this follows the same CRTP pattern as in the `serializer`/`deserializer`, so this comment is just to indicate that the `friend` clause is only because of CRTP and nothing else, rather than to point to CRTP being used. In the `class` definition, what gives it away is the pattern `class Adapter : Interface<Adapter> {...}`, in which the `class` inherits from another `class` that takes the first as a `template` parameter.

However, if you think we can/should add it as well at the class definition level, there is no reason to not do it.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -14,23 +14,20 @@
...
     static constexpr Idx sequential{-1};
 
-    template <typename Calculate>
-        requires std::invocable<std::remove_cvref_t<Calculate>, MainModel&, MutableDataset const&, Idx>
-    static BatchParameter batch_calculation_(MainModel& model, CalculationInfo& calculation_info,
-                                             Calculate&& calculation_fn, MutableDataset const& result_data,
-                                             ConstDataset const& update_data, Idx threading = sequential) {
-        // if the update dataset is empty without any component
-        // execute one power flow in the current instance, no batch calculation is needed
+    // TODO(figueroa1395): remove calculation_fn dependency
```
petersalemink95, line 21: Are these TODOs for a later stage or should they be addressed here?

mgovers: they are done in the immediate follow-up in #1071

mgovers: but the PR would become too big to do them now

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
petersalemink95: I like the idea of `job_interface`, `job_dispatch` and `job_adapter`

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: These are being resolved in follow up PRs (namely in #1066 and #1071) because this one is complex enough as is.

petersalemink95:
Just went through the PR - was quite hard to review. Not complaining - it is inherit to the nature of the PR -> shifting around code 😄 

That said: I don't see anything weird in the structure, so if both @figueroa1395 and @mgovers are happy with the content it would be good to go for me as well

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
mgovers: please fix the new sonar cloud issues: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=1036&id=PowerGridModel_power-grid-model&open=AZiIyo1L1VB51fmTBV-X

figueroa1395: @nitbharambe Will address them in his changes :)

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
nitbharambe: >Maybe @nitbharambe or @Jerry-Jinfeng-Guo have any thoughts on this?

Either would do for me.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
```diff
@@ -0,0 +1,151 @@
...
+          model_{model_copy_ ? std::ref(*model_copy_) : std::move(other.model_)} {
+        move_construction_remaining_members(std::move(other));
+        other.model_copy_.reset();
+    }
+    JobDispatchAdapter& operator=(JobDispatchAdapter&& other) noexcept {
+        if (this != &other) {
+            model_copy_ = std::move(other.model_copy_);
+            model_ = model_copy_ ? std::ref(*model_copy_) : std::move(other.model_);
+            move_construction_remaining_members(std::move(other));
+            other.model_copy_.reset();
```
nitbharambe: Added this line for resetting. Is this required or not?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
Jerry-Jinfeng-Guo: The comment of `CRTP` only is less clear. Perhaps it can be more descriptive here.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
Jerry-Jinfeng-Guo: `job_interface`, `job_dispatcher` and `job_adapter` for me. Both `interface` and `adapter` are obviously noun here but `dispatch` can be ambiguous depending on the context and persp.

Jerry-Jinfeng-Guo: OR more specifically here: if it is actually intended that `dispatch` here refers to a process (verb), I would say `job_dispatching` or even `dispatch_job`.

Jerry-Jinfeng-Guo (commented):
I know that most of the `requires` come from existing code, but when would be a good moment to replace them with named concepts?

figueroa1395:
> I know that most of the `requires` come from existing code, but when would be a good moment to replace them with named concepts?

It is good that you raised this question, because in a follow up PR (https://github.com/PowerGridModel/power-grid-model/pull/1071) I decided to silence sonar about that, so maybe that isn't the right approach.

My thought process was that these `requieres` clauses are very specific to this interface, and we probably won't ever use them elsewhere, so I don't think it is worth making them a named `concept`. On the other hand, if it would add more clarity or you have a different take, we can do it. However, if we change them, I would do it at https://github.com/PowerGridModel/power-grid-model/pull/1071, as the concepts in the interface are cleaned up a bit, so it's probably a good place to deal with that.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
figueroa1395: > `job_interface`, `job_dispatcher` and `job_adapter` for me. Both `interface` and `adapter` are obviously noun here but `dispatch` can be ambiguous depending on the context and persp.

Since most agree on this, we can go for it. Maybe it can be the last commit of this PR to close it up :)

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: It isn't, because the first thing that happens is `model_copy_ = std::move(other.model_copy_);`, which leaves `other.model_copy_` as a `nullptr`. I don't think it hurts because `std::unique_ptr::reset()` should handle `nullptr`s just fine, but it is redundant.

figueroa1395: https://github.com/PowerGridModel/power-grid-model/actions/runs/16830236808/job/47675925910?pr=1036 clang tidy already complained 😆

nitbharambe (dismissed):
LGTM post resolving of all comments.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -142,9 +146,9 @@ class JobDispatch {
               typename RecoverFromBadFn>
         requires std::invocable<std::remove_cvref_t<RunFn>, Args const&...> &&
                  std::invocable<std::remove_cvref_t<SetupFn>, Args const&...> &&
-                 std::invocable<std::remove_cvref_t<WinddownFn>, Args const&...> &&
+                 std::invocable<std::remove_cvref_t<WinddownFn>> &&
                  std::invocable<std::remove_cvref_t<HandleExceptionFn>, Args const&...> &&
-                 std::invocable<std::remove_cvref_t<RecoverFromBadFn>, Args const&...>
+                 std::invocable<std::remove_cvref_t<RecoverFromBadFn>>
```
mgovers, line 149: the original reason i kept the scenario index (and other things variadically`...`) in here (and elsewhere) was because it generalizes to any arbitrary job. For instance, a verbose logger wrapper implementation may implement logging both `Setting up scenario 123` and `Winding down scenario 123`. Asymmetry between setup and winddown removes that support.

It's fine to keep it like this if you prefer, but good to consider the implications.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: Created an [issue](https://github.com/PowerGridModel/power-grid-model/issues/1075) to keep track of this. Resolving the comment.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -43,115 +40,89 @@ template <class MainModel, class... ComponentType> class JobDispatch {
...
 
             CalculationInfo thread_info;
 
-            Timer t_total(thread_info, 0000, "Total in thread");
+            Timer t_total(thread_info, 0200, "Total batch calculation in thread");
 
-            auto const copy_model_functor = [&base_model, &thread_info] {
-                Timer const t_copy_model_functor(thread_info, 1100, "Copy model");
-                return MainModel{base_model};
+            // Can we remove this Idx?
```
nitbharambe:
```suggestion
```
My bad

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
nitbharambe: I see. It was becoming counter intuitive because copy_functor_adaptor needed an argument and sometimes we passed start and sometimes scenario_idx. Also removing it makes it clear the function is totally independent. It added to the confusion. I understand the context now but lets do away with the additional Idx.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "common/common.hpp"
+
+#include <concepts>
+#include <type_traits>
+#include <utility>
+
+namespace power_grid_model {
+template <typename Adapter> class JobDispatchInterface {
+  public:
+    template <typename Calculate, typename ResultDataset>
```
nitbharambe: To confirm, this is templated on ResultDataset (And UpdateDataset below) for testing purpose right? If not then this can be done away with.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_adapter.hpp
figueroa1395: Added clearer comments both on the interface and adapter side.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch_interface.hpp
figueroa1395: Indeed, the interface shouldn't know. what `ConstDataset`, `MutableDataset`, etc. are. This is generic as it is at the `job_dispatch` level as well.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
nitbharambe: We also had another idea of having it called using `non_existant_scenario=-1`. That feels okay. (Edit: ignore the comment.  That defeats the original intention too. Lets keep it removed).

Issue #1009, opened 2025-06-03
[IMPROVEMENT] Missing `DatasetType`s and `ComponentType`s


figueroa1395:
In #648, `DatasetType`s and `ComponentType`s were introduced to replace `string` based component and dataset types while keeping backwards compatibility - this closed issue #613. However, there are still several places across the codebase where `string`s are still used, i.e. test for the validation functions (https://github.com/PowerGridModel/power-grid-model/tree/main/tests/unit/validation), among others.

The goal of this issue is to fix this by homogenizing the codebase and replacing all the instances of `string`s with `DatasetType` and `ComponentType`, where this is applicable. Nevertheless, this is unrelated with #949, as `AttributeType`s have not been introduced yet.

PR #1037, opened 2025-07-15
Use of enumerations: Clean remaining ComponentType and DatasetType in data validator

nitbharambe:
**Fixes issue**: ` #1009 

This PR now concerns changes only regarding data validator (And a small utils function since it was already being discussed. Rest would be in a different PR)

All component and dataset type things should be replaced.. The data validator should shift entirely from strings to enums. 

* All internal functions should only work with enums. 
* All internal tests should **only** test for enums.
* Testing should check for a call of convert string to component type / dataset type at all public functions. (I imagine this is already present, but good to check again)

### Changes visible to user

* The data validator errors and error messages would now be corresponding to the enums instead of strings. This change was already introduced partially at some error types before. Now this PR should cover all places.

* All possible entry points to PGM are covered for both strings and enums usage.

### DIFF
--- a/src/power_grid_model/utils.py
+++ b/src/power_grid_model/utils.py
@@ -370,7 +370,9 @@ def self_test():
             "data": {
                 ComponentType.node: [{"id": 1, "u_rated": 10000}],
                 ComponentType.source: [{"id": 2, "node": 1, "u_ref": 1, "sk": 1e20}],
-                "sym_load": [{"id": 3, "node": 1, "status": 1, "type": 0, "p_specified": 0, "q_specified": 0}],
+                ComponentType.sym_load: [
+                    {"id": 3, "node": 1, "status": 1, "type": 0, "p_specified": 0, "q_specified": 0}
+                ],
             },
         }
 
--- a/src/power_grid_model/utils.py
+++ b/src/power_grid_model/utils.py
@@ -16,7 +16,7 @@ from typing import cast as cast_type
 import numpy as np
 
 from power_grid_model import CalculationMethod, PowerGridModel
-from power_grid_model._core.dataset_definitions import DatasetType, _map_to_component_types
+from power_grid_model._core.dataset_definitions import ComponentType, DatasetType, _map_to_component_types
 from power_grid_model._core.serialization import (
     json_deserialize,
     json_serialize,
@@ -368,8 +368,8 @@ def self_test():
             "is_batch": False,
             "attributes": {},
             "data": {
-                "node": [{"id": 1, "u_rated": 10000}],
-                "source": [{"id": 2, "node": 1, "u_ref": 1, "sk": 1e20}],
+                ComponentType.node: [{"id": 1, "u_rated": 10000}],
+                ComponentType.source: [{"id": 2, "node": 1, "u_ref": 1, "sk": 1e20}],
                 "sym_load": [{"id": 3, "node": 1, "status": 1, "type": 0, "p_specified": 0, "q_specified": 0}],
             },
         }
@@ -398,7 +398,9 @@ def self_test():
 
             assert output_data is not None
             assert math.isclose(
-                output_data["data"]["node"][0]["u"], input_data["data"]["node"][0]["u_rated"], abs_tol=1e-9
+                output_data["data"][ComponentType.node][0]["u"],
+                input_data["data"][ComponentType.node][0]["u_rated"],
+                abs_tol=1e-9,
             )
 
             print("Self test finished.")
--- a/src/power_grid_model/validation/_rules.py
+++ b/src/power_grid_model/validation/_rules.py
@@ -40,12 +40,10 @@ from typing import Any, Callable, Type, TypeVar
 
 import numpy as np
 
-from power_grid_model import ComponentType
-from power_grid_model._core.dataset_definitions import DatasetType
-from power_grid_model._core.enum import AngleMeasurementType
+from power_grid_model._core.dataset_definitions import ComponentType, DatasetType
+from power_grid_model._core.enum import AngleMeasurementType, FaultPhase, FaultType, WindingType
 from power_grid_model._core.utils import get_comp_size, is_nan_or_default
 from power_grid_model.data_types import SingleDataset
-from power_grid_model.enum import FaultPhase, FaultType, WindingType
 from power_grid_model.validation.errors import (
     ComparisonError,
     FaultPhaseError,
--- a/src/power_grid_model/validation/_rules.py
+++ b/src/power_grid_model/validation/_rules.py
@@ -41,6 +41,7 @@ from typing import Any, Callable, Type, TypeVar
 import numpy as np
 
 from power_grid_model import ComponentType
+from power_grid_model._core.dataset_definitions import DatasetType
 from power_grid_model._core.enum import AngleMeasurementType
 from power_grid_model._core.utils import get_comp_size, is_nan_or_default
 from power_grid_model.data_types import SingleDataset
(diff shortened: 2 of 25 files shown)


src/power_grid_model/utils.py
```diff
@@ -368,8 +368,8 @@ def self_test():
             "is_batch": False,
             "attributes": {},
             "data": {
-                "node": [{"id": 1, "u_rated": 10000}],
-                "source": [{"id": 2, "node": 1, "u_ref": 1, "sk": 1e20}],
+                ComponentType.node: [{"id": 1, "u_rated": 10000}],
+                ComponentType.source: [{"id": 2, "node": 1, "u_ref": 1, "sk": 1e20}],
                 "sym_load": [{"id": 3, "node": 1, "status": 1, "type": 0, "p_specified": 0, "q_specified": 0}],
```
nitbharambe: Left for entry point check

mgovers: since this is not in `tests`, I do believe that this should just be replaced as well. Only tests should do checks

```suggestion
                ComponentType.sym_load: [{"id": 3, "node": 1, "status": 1, "type": 0, "p_specified": 0, "q_specified": 0}],
```

src/power_grid_model/validation/_validation.py
```diff
@@ -191,15 +191,15 @@ def assert_valid_data_structure(data: Dataset, data_type: DatasetType) -> None:
...
+            raise KeyError(f"Unknown component '{component}' in {data_type} data.")
 
         # Check if component definition is as expected
         dtype = component_dtype[component]
         if isinstance(array, np.ndarray):
             if array.dtype != dtype:
                 if not hasattr(array.dtype, "names") or not array.dtype.names:
                     raise TypeError(
-                        f"Unexpected Numpy array ({array.dtype}) for '{component}' {data_type}_data "
+                        f"Unexpected Numpy array ({array.dtype}) for '{component}' {data_type} data "
```
mgovers, line 209: please double-check but i believe that this formats weirdly

```suggestion
                        f"Unexpected Numpy array ({array.dtype}) for '{component}' {data_type.value if isinstance(data_type, ComponentType) else data_type}_data "
```

or even (if it's internal only)


```suggestion
                        f"Unexpected Numpy array ({array.dtype}) for '{component}' {data_type.value}_data "
```

src/power_grid_model/validation/_validation.py
```diff
@@ -208,7 +208,7 @@ def assert_valid_data_structure(data: Dataset, data_type: DatasetType) -> None:
                 )
         else:
             raise TypeError(
-                f"Unexpected data type {type(array).__name__} for '{component}' {data_type}_data "
+                f"Unexpected data type {type(array).__name__} for '{component}' {data_type} data "
```
mgovers, line 218: idem

src/power_grid_model/validation/errors.py
```diff
@@ -317,9 +318,9 @@ class IdNotInDatasetError(SingleFieldValidationError):
     """
 
     _message = "ID does not exist in {ref_dataset} for {n} {objects}."
-    ref_dataset: str
+    ref_dataset: DatasetType
```
nitbharambe, line 321: A public change

src/power_grid_model/validation/_validation.py
nitbharambe: From our discussions, we should not go by first choice. We dont support  data_type to be str.
Hence we would have to change this message anyway.
Now, would we rather have the `{data_type.value} data` as output, or only the `'{data_type}' data`, is the question (where data_type is DatasetType enum). Appending `_data` directly might not make it clear where the issue is.

mgovers: > Now, would we rather have the {data_type.value} data as output, or only the '{data_type.value}' data, is the question

what is the difference?

>  Appending _data directly might not make it clear where the issue is.

good point.

nitbharambe: > what is the difference?

edited comment.

mgovers: Hmmm ok i see your point. i also see that the formatted string does an OK job

```py
>>> from power_grid_model import DatasetType
>>> i = DatasetType.input
>>> print(f"'{i}' data")
'DatasetType.input' data
>>> print(f"'{i.value}' data")
'input' data
>>> i
<DatasetType.input: 'input'>
```

mgovers: i guess we can keep as-is

src/power_grid_model/validation/utils.py
```diff
@@ -107,7 +101,7 @@ def _update_input_data(input_data: SingleDataset, update_data: SingleDataset):
 
 
 def _update_component_data(
-    component: _ComponentTypeLike, input_data: SingleComponentData, update_data: SingleComponentData
+    component: ComponentType, input_data: SingleComponentData, update_data: SingleComponentData
```
mgovers, line 104: here and elsewhere: please double-check that this is not backwards incompatible for users

mgovers (commented):
the remaining comment is also for me when i do some explorative reviewing

src/power_grid_model/validation/_rules.py
```diff
@@ -41,6 +41,7 @@
 import numpy as np
 
 from power_grid_model import ComponentType
+from power_grid_model._core.dataset_definitions import DatasetType
 from power_grid_model._core.enum import AngleMeasurementType
```
figueroa1395: Here and in other places: Shouldn't we import `ComponentType` and `DatasetType` like:
```suggestion
from power_grid_model import ComponentType, DatasetType
from power_grid_model._core.enum import AngleMeasurementType
```

src/power_grid_model/validation/errors.py
figueroa1395: I should be backwards compatible, hence alright. Correct?

src/power_grid_model/validation/utils.py
nitbharambe: Yes! All public functions should have a conversion to enum . This particular one is covered at validate_input_data and validate_update_data

src/power_grid_model/validation/_rules.py
nitbharambe: Why import from power_grid_model directly here? I think we can change the existing import to _core one right? To keep the imports detangled.
(I understand that validator vs non validator code would always be separate. But we end up using some internal _core functionality anyway. Hence lets keep all imports from internal instead of public+internal) What do you think?

src/power_grid_model/validation/errors.py
nitbharambe: Its not backwards compatible actually.

figueroa1395: Is it giving you squiggly lines? Would using `DatasetTypeLike` instead, help here?

src/power_grid_model/validation/_rules.py
figueroa1395: Alright, let's then import directly from core when inside the "core" ourselves.

src/power_grid_model/validation/errors.py
nitbharambe: Nope. Not giving squiggly lines. 
Correction: Yes, the change is backward compatible.

figueroa1395: Nice! From my side, feel free to resolve this.

src/power_grid_model/validation/_rules.py
nitbharambe: Ill make the change to make it import this way at all places.

nitbharambe: @figueroa1395 Last commit,  [f3bb805](https://github.com/PowerGridModel/power-grid-model/pull/1037/commits/f3bb8050755a478efb9ef51b0bd83a13a7ecbfe4), changes the import from public to internal for consistency throughout.

figueroa1395: Great. Resolving.

figueroa1395 (approved):

figueroa1395:
Approved but added `do-not-merge` label so we don't do releases on Friday. Ready to merge on Monday.

PR #1039, opened 2025-07-15
Buffer Handling: update obtaining raw data view, add tests

Laurynas-Jagutis:
based on: https://github.com/PowerGridModel/power-grid-model/pull/1025#discussion_r2204357330

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
@@ -13,19 +13,15 @@
 #include <string>
 
 namespace power_grid_model {
-namespace detail {
-inline auto to_string(std::floating_point auto x) {
-    std::ostringstream sstr{}; // NOLINT(misc-const-correctness) // https://github.com/llvm/llvm-project/issues/57297
-    sstr << x;
-    return sstr.str();
-}
-inline auto to_string(std::integral auto x) { return std::to_string(x); }
-} // namespace detail
-
 class PowerGridError : public std::exception {
   public:
-    void append_msg(std::string_view msg) { msg_ = std::format("{}{}", msg_, msg); }
     char const* what() const noexcept final { return msg_.c_str(); }
+
+  protected:
+    PowerGridError() = default;
+    PowerGridError(std::string msg) : msg_{std::move(msg)} {}
+
+    void append_msg(std::string_view msg) { msg_ = std::format("{} {}", msg_, msg); }
 
   private:
     std::string msg_;
@@ -38,20 +34,18 @@ class InvalidArguments : public PowerGridError {
         std::string value;
     };
 
-    template <std::same_as<TypeValuePair>... Options> InvalidArguments(std::string_view error_msg) {
-        append_msg(error_msg);
-    }
+    template <std::same_as<TypeValuePair>... Options>
+    InvalidArguments(std::string error_msg) : PowerGridError{std::move(error_msg)} {}
 
     template <std::same_as<TypeValuePair>... Options>
-    InvalidArguments(std::string_view method, std::string_view arguments) {
-        append_msg(std::format("{} is not implemented for {}!\n", method, arguments));
-    }
+    InvalidArguments(std::string_view method, std::string_view arguments)
+        : PowerGridError{std::format("{} is not implemented for {}!\n", method, arguments)} {}
 
     template <class... Options>
-        requires(std::same_as<std::remove_cvref_t<Options>, TypeValuePair> && ...)
+        requires((std::same_as<std::remove_cvref_t<Options>, TypeValuePair> && ...) && sizeof...(Options) > 0)
     InvalidArguments(std::string_view method, Options const&... options)
         : InvalidArguments{method, "the following combination of options"} {
-        (append_msg(std::format(" {}: {}\n", options.name, options.value)), ...);
+        (append_msg(std::format("{}: {}\n", options.name, options.value)), ...);
     }
 };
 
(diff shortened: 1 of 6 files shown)


src/power_grid_model/_core/buffer_handling.py
```diff
@@ -117,9 +117,16 @@ def _get_raw_attribute_data_view(data: np.ndarray, schema: ComponentMetaData, at
...
     """
-    if schema.dtype[attribute].shape == (3,) and data.shape[-1] != 3:
-        raise ValueError("Given data has a different schema than supported.")
-    return _get_raw_data_view(data, dtype=schema.dtype[attribute].base)
+    dense_batch_ndim = 2
+    dense_batch_asym_attr_ndim = dense_batch_ndim + 1
+    n_asym_values = 3
+    shape = schema.dtype[attribute].shape
+
+    true_shape = shape[0] if shape else 1
```
mgovers: ~what do you mean? what is a `true` shape? when does `shape` evaluate to `False`?~ EDIT: see my follow-up comment

src/power_grid_model/_core/buffer_handling.py
```diff
@@ -117,9 +117,16 @@ def _get_raw_attribute_data_view(data: np.ndarray, schema: ComponentMetaData, at
...
+
+    if true_shape <= dense_batch_ndim or (true_shape == dense_batch_asym_attr_ndim and data.shape[-1] == n_asym_values):
+        return _get_raw_data_view(data, dtype=schema.dtype[attribute].base)
```
mgovers, line 127: this still doesn't capture all edge cases. What if there's an asymmetric attribute that's provided as a single scenario (not batch) with length 10 instead of length 3?

src/power_grid_model/_core/buffer_handling.py
```diff
@@ -117,9 +117,16 @@ def _get_raw_attribute_data_view(data: np.ndarray, schema: ComponentMetaData, at
...
-        raise ValueError("Given data has a different schema than supported.")
-    return _get_raw_data_view(data, dtype=schema.dtype[attribute].base)
+    dense_batch_ndim = 2
+    dense_batch_asym_attr_ndim = dense_batch_ndim + 1
+    n_asym_values = 3
+    shape = schema.dtype[attribute].shape
+
+    true_shape = shape[0] if shape else 1
+
+    if true_shape <= dense_batch_ndim or (true_shape == dense_batch_asym_attr_ndim and data.shape[-1] == n_asym_values):
```
mgovers: probably better to do something along the lines of
```suggestion
    attr_schema = schema.dtype[attribute]
    attr_shape_start = data.ndim - attr_schema.ndim

    dataset_shape = data.shape[:attr_shape_start]
    attr_shape = data.shape[attr_shape_start:]

    if len(dataset_shape.ndim) <= dense_batch_ndim and attr_shape == attr_schema.shape:
```

src/power_grid_model/_core/buffer_handling.py
mgovers: See my other comment https://github.com/PowerGridModel/power-grid-model/pull/1039#discussion_r2207400691

tests/unit/test_buffer_handling.py
```diff
@@ -83,3 +87,60 @@ def test__get_sparse_buffer_properties(component_type, is_columnar):
         assert properties.columns == list(data["data"].keys())
     else:
         assert properties.columns is None
+
+
+@pytest.mark.parametrize(
+    "component, is_batch, is_columnar, is_sparse",
+    [
+        pytest.param(ComponentType.sym_load, True, True, False, id="sym_load"),
```
figueroa1395: Nitpick: 
```suggestion
        pytest.param(ComponentType.sym_load, True, True, False, id="sym_load_columnar"),
```

tests/unit/test_buffer_handling.py
```diff
@@ -83,3 +87,60 @@ def test__get_sparse_buffer_properties(component_type, is_columnar):
...
+    )
+
+
+@pytest.mark.parametrize(
+    "component, attribute",
+    [
+        pytest.param(ComponentType.asym_load, "p_specified", id="asym_load-shape_missmatch"),
+    ],
+)
+def test__get_raw_attribute_data_view_fail(component, attribute):
```
figueroa1395, line 129: I suppose this and the above test can be put into one?

Also, it isn't clear from the tests only (as this I think was done TDD), what was the issue.

Was the above test failing before the fix and now passes after the fix? And was the issue triggered by a behavior similar to what you replicate in this test? It seems a bit confusing.

tests/unit/test_buffer_handling.py
```diff
@@ -83,3 +88,126 @@ def test__get_sparse_buffer_properties(component_type, is_columnar):
...
+        is_columnar=is_columnar,
+        is_sparse=is_sparse,
+    )
+    if component == ComponentType.asym_load:
+        assert data["p_specified"].shape[-1] == 3
+
+    get_buffer_view(
+        data,
+        schema=schema,
+        is_batch=is_batch,
```
mgovers: please do a check here that asserts that the value is non-`null` and things like the buffer size and shape are correct

tests/unit/test_buffer_handling.py
mgovers: > I suppose this and the above test can be put into one?

it's usually considered best practice to split good-weather tests (that do not raise exceptions) from bad-weather tests (that do raise exceptions and/or contain fall-back logic)

Laurynas-Jagutis: The explanation can be found here: https://github.com/PowerGridModel/power-grid-model/pull/1025#discussion_r2204357330

tests/unit/test_buffer_handling.py
```diff
@@ -83,3 +88,153 @@ def test__get_sparse_buffer_properties(component_type, is_columnar):
...
+        pytest.param(ComponentType.asym_load, 0, "q_specified", id="asym_load-columnar"),
+        pytest.param(ComponentType.asym_load, (2, 0, 0), "q_specified", id="asym_load-columnar"),
+        pytest.param(ComponentType.asym_load, (2, 4, 0), "q_specified", id="asym_load-columnar"),
+        # sym component
+        pytest.param(ComponentType.sym_load, (2, 4, 4), "p_specified", id="sym_load-columnar"),
+        pytest.param(ComponentType.sym_load, (2, 4, 3), "p_specified", id="sym_load-columnar"),
+        pytest.param(ComponentType.sym_load, (0, 0, 1), "p_specified", id="sym_load-columnar"),
+        pytest.param(ComponentType.sym_load, (2, 4, 4), "q_specified", id="sym_load-columnar"),
+        pytest.param(ComponentType.sym_load, (2, 4, 3), "q_specified", id="sym_load-columnar"),
+        pytest.param(ComponentType.sym_load, (0, 0, 1), "q_specified", id="sym_load-columnar"),
```
mgovers, line 232: please double-check that it's OK to specify the same `id` multiple times. maybe it's better to just not provide an `id` at all

Laurynas-Jagutis: Whenever I run the tests in vscode, these ids are taken care of.
If we take a look at the cases of sym component, internally the ids become:

1. sym_load-columnar1
2. sym_load-columnar2
3. sym_load-columnar3
4. etc

so it is possible to distinguish the tests

mgovers: OK

PR #1040, opened 2025-07-15
Current sensor: Remove experimental feature flag - local angle NRSE

figueroa1395:
This PR removes the experimental feature flag for the local angle current sensors for NRSE. The documentation and example follows in https://github.com/PowerGridModel/power-grid-model/pull/1035

### DIFF
--- a/tests/native_api_tests/test_api_model.cpp
+++ b/tests/native_api_tests/test_api_model.cpp
@@ -1063,7 +1063,7 @@ TEST_CASE("API Model") {
         }
     }
 
-    SUBCASE("Current sensor for NRSE is experimental: not implemented") {
+    SUBCASE("Global angle current sensor for NRSE is experimental: not implemented") {
         auto const* const input_data_se_json = R"json({
   "version": "1.0",
   "type": "input",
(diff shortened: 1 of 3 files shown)


tests/native_api_tests/test_api_model.cpp
```diff
@@ -1099,9 +1100,10 @@ TEST_CASE("API Model") {
         for (auto const method : {PGM_default_method, PGM_iterative_linear, PGM_newton_raphson}) {
             CAPTURE(method);
             if (method == PGM_newton_raphson) {
-                CHECK_THROWS_WITH_AS(run_se_with_current_sensor(method, PGM_experimental_features_disabled),
-                                     "Newton-Raphson state estimation is not implemented for current sensors",
-                                     PowerGridRegularError);
+                CHECK_THROWS_WITH_AS(
+                    run_se_with_current_sensor(method, PGM_experimental_features_disabled),
+                    "Newton-Raphson state estimation is not implemented for global angle current sensors",
+                    PowerGridRegularError);
```
mgovers, line 1106: optional: maybe good to also loop to check that it is actually not experimental if you do have a local current sensor with NRSE but only if you really have a global one

mgovers (approved):
optional improvement. feel free to resolve

tests/native_api_tests/test_api_model.cpp
figueroa1395: Thought so as well. But I guess the way to do it would be to create a separate test with another json thingy with just that change, and personally I don't think it is worth it... Specially if we know this works because experimental features are now disabled for local current sensor in the validation tests. So I will resolve.

PR #1042, opened 2025-07-16
Use of enumerations: Clean remaining enums in other places

nitbharambe:
Continuation to #1037 

Contains all changes of renaming other than data validator.

### DIFF
--- a/tests/unit/test_0Z_model_validation.py
+++ b/tests/unit/test_0Z_model_validation.py
@@ -9,6 +9,7 @@ from typing import Callable
 import numpy as np
 import pytest
 
+from power_grid_model._core.dataset_definitions import ComponentType
 from power_grid_model._core.utils import convert_batch_dataset_to_batch_list
 from power_grid_model.enum import TapChangingStrategy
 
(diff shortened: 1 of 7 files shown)


figueroa1395:
~~Why can I just merge this PR even if there aren't any reviews? 👀 @nitbharambe @mgovers~~

Edit: I just realized it does not merge into main, hence I can just merge into the other branch 😆. All good!

tests/unit/test_0Z_model_validation.py
```diff
@@ -9,6 +9,7 @@
 import numpy as np
 import pytest
 
+from power_grid_model._core.dataset_definitions import ComponentType
```
figueroa1395, line 12: Same comment as before:
```suggestion
from power_grid_model import ComponentType
```

figueroa1395 (commented):
Nice job! It seems, once again, very thorough. A couple of additional comments:
- In `data_types.py` there are a bunch of docstrings with examples using string versions of `ComponentType`s or `DatasetType`s, do we wish to also edit those? I would say so, but I guess that is optional.
- In the serialization tests, a bunch of components are still strings as well. Was that intended? I suppose so, right?

tests/unit/test_0Z_model_validation.py
figueroa1395: Resolving after https://github.com/PowerGridModel/power-grid-model/pull/1037#discussion_r2213456769

nitbharambe:
> * In `data_types.py` there are a bunch of docstrings with examples using string versions of `ComponentType`s or `DatasetType`s, do we wish to also edit those? I would say so, but I guess that is optional.

Ok ill edit those.

> * In the serialization tests, a bunch of components are still strings as well. Was that intended? I suppose so, right?

That is intended. They get converted to json and then used.

figueroa1395 (approved):

figueroa1395 (dismissed):

figueroa1395 (approved):

PR #1043, opened 2025-07-18
Internal improvements: more constexpr + unit tests

mgovers:
In an attempt to better embrace the shift-left strategy, we add some additional `constexpr` statements in the core.

In some cases, runtime unit tests are replaced by static asserts. This is considered best-practice and may actually provide some performance benefits.

some metrics gathered on Windows using `clang-cl-debug` (clean rebuild all projects) and `ctest --test-dir=cpp_build/clang-cl-debug`

| branch/tag  | compile time | runtime all tests |
| ----------- | ------------ | ----------------- |
| `v1.11.26`  | 00:00:54.137 | 32.64 sec         |
| this branch | 00:00:50.384 | 32.00 sec         |

* [x] Benchmark to ensure no performance penalty


power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -13,20 +13,16 @@
...
 class PowerGridError : public std::exception {
   public:
-    void append_msg(std::string_view msg) { msg_ = std::format("{}{}", msg_, msg); }
+    PowerGridError() = default;
+    PowerGridError(std::string msg) : msg_{std::move(msg)} {}
+
     char const* what() const noexcept final { return msg_.c_str(); }
 
+  protected:
+    void append_msg(std::string_view msg) { msg_ = std::format("{}{}", msg_, msg); }
```
figueroa1395: I think you missed a whitespace:
```suggestion
    void append_msg(std::string_view msg) { msg_ = std::format("{} {}", msg_, msg); }
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -59,182 +53,152 @@ class MissingCaseForEnumError : public InvalidArguments {
...
-        }
-    }
+    NotObservableError() : PowerGridError{"Not enough measurements available for state estimation.\n"} {}
+    NotObservableError(std::string_view msg)
+        : PowerGridError{std::format("Not enough measurements available for state estimation.\n{}\n", msg)} {}
 };
 
 class IterationDiverge : public PowerGridError {
   public:
-    IterationDiverge() = default;
```
figueroa1395, line 134: I imagine we had the `default` constructor for when we wanted to `throw` "blank" `IterationDiverge` errors, isn't that the case? Just double checking.

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -59,182 +53,152 @@ class MissingCaseForEnumError : public InvalidArguments {
...
+              std::format("Iteration failed to converge after {} iterations! Max deviation: {}, error tolerance: {}.\n",
+                          num_iter, max_dev, err_tol)} {}
 };
 
 class MaxIterationReached : public IterationDiverge {
   public:
-    MaxIterationReached(std::string const& msg = "") {
-        append_msg(std::format("Maximum number of iterations reached {}\n", msg));
-    }
+    MaxIterationReached(std::string const& msg = "")
```
figueroa1395:
```suggestion
    MaxIterationReached(std::string_view msg = "")
```
right?

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -13,20 +13,16 @@
...
-    return sstr.str();
-}
-inline auto to_string(std::integral auto x) { return std::to_string(x); }
-} // namespace detail
-
 class PowerGridError : public std::exception {
   public:
-    void append_msg(std::string_view msg) { msg_ = std::format("{}{}", msg_, msg); }
+    PowerGridError() = default;
+    PowerGridError(std::string msg) : msg_{std::move(msg)} {}
```
figueroa1395: Do you know if this constructor is called directly anywhere else outside this file? If so, let's make sure that me move in the message from the caller side to take full advantage of this.

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
```diff
@@ -34,20 +34,23 @@ The input, i.e., [0, 1, 3] should be strictly increasing
...
     }
     return result;
 }
 
-inline auto sparse_decode(IdxVector const& indptr) {
-    auto result = IdxVector(indptr.back());
-    for (Idx const group : IdxRange{static_cast<Idx>(indptr.size()) - 1}) {
-        std::fill(std::begin(result) + indptr[group], std::begin(result) + indptr[group + 1], group);
+template <std::ranges::viewable_range IndPtr> constexpr auto sparse_decode(IndPtr&& indptr) {
+    auto indptr_view = std::views::all(std::forward<IndPtr>(indptr));
```
figueroa1395, line 50: With the addition of `template <std::ranges::viewable_range IndPtr>`, isn't this line redundant?

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
```diff
@@ -104,7 +107,7 @@ class SparseGroupedIdxVector {
         using difference_type = typename Base::difference_type;
         using reference = typename Base::reference;
 
-        constexpr GroupIterator() = default;
+        GroupIterator() = default;
```
figueroa1395, line 110: Why?

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
```diff
@@ -178,7 +181,7 @@ class DenseGroupedIdxVector {
...
```
figueroa1395, line 184: Why?

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -215,8 +215,8 @@ inline auto phase_mod_2pi(RealValue<asymmetric_t> const& phase) {
 }
 
 // calculate kron product of two vector
-inline double vector_outer_product(double x, double y) { return x * y; }
-inline DoubleComplex vector_outer_product(DoubleComplex x, DoubleComplex y) { return x * y; }
+constexpr double vector_outer_product(double x, double y) { return x * y; }
```
figueroa1395: Just double checking, but `constexpr` implies `inline`, right?

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -296,23 +296,23 @@ inline auto diag_mult(Eigen::ArrayBase<DerivedA> const& x, Eigen::ArrayBase<Deri
...
     return (val(0) + a * val(1) + a2 * val(2)) / 3.0;
 }
 
-inline auto pos_seq(DoubleComplex const& val) { return val; }
+constexpr auto pos_seq(DoubleComplex const& val) { return val; }
 
 // inverse of tensor
-inline auto inv(double val) { return 1.0 / val; }
-inline auto inv(DoubleComplex const& val) { return 1.0 / val; }
+template <std::floating_point Float> constexpr auto inv(Float val) { return Float{1.0} / val; }
```
figueroa1395, line 317: Was `template <std::floating_point Float> constexpr` needed? If so, why not change it everywhere else and just here?

tests/cpp_unit_tests/test_exceptions.cpp
```diff
@@ -8,10 +8,339 @@
...
+
+static_assert(static_cast<Idx>(TestCppStyleEnumClass::foo) == 0);
+static_assert(static_cast<Idx>(TestCppStyleEnumClass::bar) == 1);
+static_assert(static_cast<Idx>(TestCppStyleEnumClass::baz) == -1);
+static_assert(static_cast<Idx>(TestCppStyleEnumClass::nan) == na_IntS);
+
 TEST_CASE("Exceptions") {
-    SUBCASE("Iteration Diverge") {
-        REQUIRE(std::string{IterationDiverge{20, 1.0e20, 1.0e-8}.what()} ==
-                "Iteration failed to converge after 20 iterations! Max deviation: 1e+20, error tolerance: 1e-08.\n");
```
figueroa1395, line 14: I love how this was our exceptions tests 😆

tests/cpp_unit_tests/test_exceptions.cpp
```diff
@@ -8,10 +8,339 @@
...
+        CHECK(error.failed_scenarios() == failed_scenarios);
+        CHECK(error.err_msgs() == err_msgs);
+    }
+    SUBCASE("InvalidCalculationMethod") {
+        CHECK(std::string{InvalidCalculationMethod{}.what()} ==
+              "The calculation method is invalid for this calculation!");
+    }
+    SUBCASE("InvalidShortCircuitType") {
+        CHECK(std::string{InvalidShortCircuitType{FaultType::three_phase}.what()} ==
+              "The short circuit type (0) is invalid!\n");
```
figueroa1395, line 248: Question out of sheer curiosity: I see that, mostly in the short circuit related errors, the order of arguments in the constructor is used in the opposite order in the error message. Is there a reason it was done this way? Or just happened like this and we didn't notice? Honestly curious, I don't think the effort of changing the order is worth it just for the looks.

tests/cpp_unit_tests/test_exceptions.cpp
```diff
@@ -8,10 +8,339 @@
...
+        CHECK(std::string{InvalidShortCircuitPhases{FaultType::single_phase_to_ground, FaultPhase::abc}.what()} ==
+              "The short circuit phases (0) do not match the short circuit type (1)\n");
+        CHECK(std::string{InvalidShortCircuitPhases{FaultType::two_phase, FaultPhase::abc}.what()} ==
+              "The short circuit phases (0) do not match the short circuit type (2)\n");
+        CHECK(std::string{InvalidShortCircuitPhases{FaultType::two_phase_to_ground, FaultPhase::abc}.what()} ==
+              "The short circuit phases (0) do not match the short circuit type (3)\n");
+    }
+    SUBCASE("InvalidShortCircuitPhaseOrType") {
+        CHECK(std::string{InvalidShortCircuitPhaseOrType{}.what()} ==
+              "During one calculation the short circuit types phases should be similar for all faults\n");
```
figueroa1395, line 288: Maybe out of scope, but this message doesn't seem very clear to me. Specifically the phrasing _should be similar_... What does that mean?

tests/cpp_unit_tests/test_three_phase_tensor.cpp
```diff
@@ -44,8 +44,8 @@ TEST_CASE("Three phase tensor") {
         CHECK(cabs(vec7(0) - 1.0) < 1e-8);
         CHECK(cabs(vec7(1) - a2) < 1e-8);
         CHECK(cabs(vec7(2) - a) < 1e-8);
-        CHECK(RealValue<symmetric_t>{1.0} == 1.0);
-        CHECK(ComplexValue<symmetric_t>{1.0} == 1.0);
+        static_assert(RealValue<symmetric_t>{1.0} == 1.0);
+        static_assert(ComplexValue<symmetric_t>{1.0} == 1.0);
```
figueroa1395, line 48: Given that this is within a test, is there really a gain here? Idem for the others.

tests/cpp_unit_tests/test_exceptions.cpp
figueroa1395, line 1: Awesome, very clear and throughly tested.

tests/cpp_unit_tests/test_three_phase_tensor.cpp
```diff
@@ -140,32 +144,34 @@ TEST_CASE("Three phase tensor") {
         RealTensor<asymmetric_t> mat2;
         mat2 << -3, 2, 3, 4, -5, 6, 7, 8, -9;
         CHECK((mat1 == mat2).all());
-        double x = 5;
-        double const y = 10;
-        add_diag(x, -y);
-        CHECK(x == -5);
+        static_assert([] {
+            double x = 5;
+            double const y = 10;
```
figueroa1395, line 149: Can these two be `constexpr`?

tests/cpp_unit_tests/test_three_phase_tensor.cpp
```diff
@@ -191,10 +199,14 @@ TEST_CASE("Three phase tensor") {
     }
 
     SUBCASE("Test RealValue update - asym") {
-        RealValue<asymmetric_t> vec{1.0, nan, nan};
         RealValue<asymmetric_t> const vec_update_1{nan, nan, nan};
         RealValue<asymmetric_t> const vec_update_2{nan, nan, 2.0};
-        double const scalar = 3.0;
+        constexpr double scalar = 3.0;
+
+        RealValue<asymmetric_t> vec{1.0, nan, nan};
```
figueroa1395, line 206: Can't this one be `constexpr`? or `Eigen` under the hood doesn't like it?

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: can you elaborate, please?

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: yes and no... the original implementation deferred the whitespaces to the usages. I've now moved the usage to here.

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: maybe that was the original intention, but the only way it was used anymore was for the constructor of derived exception types. Also, because we have both `IterationDiverge` and `MaxIterationReached`, it's better to be explicit

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
mgovers: the fact that we're using perfect forwarding here is biting us. The latest usage (i.e., the one where `indptr` should be forwarded) is happening within the for-loop.

Taking the view (which should be cheap anyways) here, means that we don't have to do another magic command like `sink_into_the_void(std::forward<IndPtr>(indptr));` or a special case for the last element of the for-loop.

while **the** most pretty implementation, it at least is clean

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
mgovers: `= default` is a builtin phrase that cannot be `constexpr` (because the defaulted constructor might not be, depending on the member variable constructors)

That said, if the defaulted constructor is `constexpr`, then the declaration will also be `constexpr`

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
mgovers: idem

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
mgovers: yup

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
mgovers: no real reason, but it's simple enough here. I can extend to others if you like.

tests/cpp_unit_tests/test_exceptions.cpp
mgovers: yup 😆

tests/cpp_unit_tests/test_exceptions.cpp
mgovers: > Or just happened like this and we didn't notice?

this.

The same is actually true for the `NotImplemented` stuff, which is even not consistent between python and C++. I didn't want to "fix" it in this PR, though. This PR is mainly for the sake of feeling the pain and seeing what's wrong.

tests/cpp_unit_tests/test_exceptions.cpp
mgovers: yup, quite unclear and we should improve it. It's explained clearly in the docs and the code is very self-contained

> Multiple faults may exist within one calculation. Currently, all faults in one scenario are required to have the same fault_type and fault_phase. Across scenarios in a batch, the fault_type and fault_phase may differ.

see also https://github.com/PowerGridModel/power-grid-model/blob/feature/more-constexpr/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/short_circuit_solver.hpp#L413-L426

tests/cpp_unit_tests/test_three_phase_tensor.cpp
mgovers: 1. it makes sure that `constexpr` really is `constexpr` (that's how i found out that `constexpr GroupIterator() = default;` is not a valid `constexpr`.
2. shift-left: test compile-time rather than run-time.
3. also speeds up tests.

tests/cpp_unit_tests/test_three_phase_tensor.cpp
mgovers: the latter, yes. the former, no (because it is mutated later in the function).

however, there's no point in defining something `constexpr` that's happening in a `consteval` (like `static_assert`)

tests/cpp_unit_tests/test_three_phase_tensor.cpp
mgovers: `Eigen` indeed... `Eigen` is no-where `constexpr` and that severely limits the `constexpr`ability of this repo (and even this PR)

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
figueroa1395: Saw the change. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
figueroa1395: What I meant is that `PowerGridModel` is mostly used in other classes that derive from it via `PowerGridModel{"stuff"}` or `PowerGridModel{std::move(msg)}`. Or at least that is the case in this file. What I meant to ask was, if `PowerGridModel{}` is used anywhere else "directly" (not via the internal implementation of other exceptions, then make sure we are doing either `PowerGridModel{"stuff"}` or `PowerGridModel{std::move(msg)}`, as we do in this file. I don't expect that to be the case, but I wanted to double check.

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
figueroa1395: Right, I didn't think about the perfect forwarding. This is definitely cleaner that the `into_the_void` shenanigans.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
figueroa1395: I'd prefer to keep it uniform, so if extending it to others isn't a problem, I'd like that.

tests/cpp_unit_tests/test_exceptions.cpp
figueroa1395: Understood. Maybe We can make a quick issue out of it, so we don't forget about it. But definitely agree that it is better not to do it in this PR.

tests/cpp_unit_tests/test_exceptions.cpp
figueroa1395: Can you quickly fix it here then? If it is so clear from the docs and the code itself. Or would that potentially be breaking for users?

tests/cpp_unit_tests/test_three_phase_tensor.cpp
figueroa1395: At some point we have to address `Eigen`... But that's a different battle 😆.

tests/cpp_unit_tests/test_exceptions.cpp
mgovers: depending on how they extract their exceptions, it is potentially breaking if they are still using the old workflows or use a tool like Splunk for log analysis (in which case it introduces a new entry, which is alright, i guess, because this error should only pop up during development).

that said, i think changing messages is out of scope of this PR. we can try to come up with a better name offline today in a follow-up PR.

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: we can actually make it `protected` if you like. It's not used anywhere else. The only other location that explicitly uses `PowerGridError` is the C++ API wrapper, which has its own exception class different from the one declared here.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
mgovers: done

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
figueroa1395: Let's make it protected then.

tests/cpp_unit_tests/test_exceptions.cpp
mgovers: #1050

tests/cpp_unit_tests/test_exceptions.cpp
mgovers: #1050

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: done

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -60,14 +60,14 @@ template <scalar_value T> class Tensor : public Eigen3Tensor<T> {
   public:
     Tensor() { (*this) = Eigen3Tensor<T>::Zero(); }
     // additional constructors
-    explicit Tensor(T const& x) { (*this) << x, 0.0, 0.0, 0.0, x, 0.0, 0.0, 0.0, x; }
+    explicit Tensor(T const& x) { (*this) << x, T{0}, T{0}, T{0}, x, T{0}, T{0}, T{0}, x; }
```
figueroa1395, line 63: Commenting here just because: but could you also include `std::floating_point<T>` in the definition of `scalar_value`? As well as `std::complex<Float>`? Just to stay consistent. Or do we want to keep that one as is for some other reason?

In addition, wouldn't it make more sense to do `T{}` instead of `T{0}`?

mgovers: > Commenting here just because: but could you also include std::floating_point<T> in the definition of scalar_value? As well as std::complex<Float>? Just to stay consistent. Or do we want to keep that one as is for some other reason?

:eyes:

> In addition, wouldn't it make more sense to do T{} instead of T{0}?

although it doesn't really matter in this case, in general, `T{}` and `T{0}` have different use cases: `T{}` will give the default value, but if someone comes up with some kind of weird floating-point-like object (via customization points), then that will not result in the correct output. `T{0}` guarantees that the value is set to a value representing the zero-element. If the default value is sufficient (like in a default constructor), then `T{}` will do the trick, otherwise (like here), it's better to explicitly set `T{0}`

figueroa1395: That makes sense.

mgovers:
benchmark OK. Removing `do-not-merge` flag

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -287,32 +295,34 @@ template <symmetry_tag sym, class T> inline auto process_mean_val(T const& m) {
...
 }
 
-inline auto pos_seq(DoubleComplex const& val) { return val; }
+constexpr auto pos_seq(DoubleComplex const& val) { return val; }
 
 // inverse of tensor
-inline auto inv(double val) { return 1.0 / val; }
-inline auto inv(DoubleComplex const& val) { return 1.0 / val; }
+template <std::floating_point Float> constexpr auto inv(Float val) { return Float{1.0} / val; }
+template <std::floating_point Float> constexpr auto inv(std::complex<Float> const& val) { return 1.0 / val; }
```
Jerry-Jinfeng-Guo:
```suggestion
template <std::floating_point Float> constexpr auto inv(std::complex<Float> const& val) { return Float{1.0} / val; }
```

Jerry-Jinfeng-Guo (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
mgovers: good find

PR #1044, opened 2025-07-18
Full observability check: meshed validation cases

Jerry-Jinfeng-Guo:
This PR adds four validation test cases for the full observability check in meshed networks: observable with and without branch measurement; unobservable with and without branch measurement.

In the current code base, it is impossible to introduce the expected `NotImplementedError` as many SE related test cases assume the meshed observability check to pass.


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -221,7 +221,9 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     if (topo.is_radial) {
         is_sufficient_condition_met = detail::sufficient_observability_condition(y_bus_structure, observability_sensors,
                                                                                  n_voltage_phasor_sensors);
-    }
+    } /*else {
+        throw NotImplementedError{};
+    }*/
```
Jerry-Jinfeng-Guo: This is the place where an error should have been thrown.

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/params.json
```diff
@@ -0,0 +1,12 @@
...
+  "calculation_method": ["newton_raphson", "iterative_linear"],
+  "rtol": 1e-8,
+  "atol": {
+    "default": 1e-8,
+    ".+_residual": 5e-4
+  },
+  "fail": {
+    "raises": "NotImplementedError",
+    "reason": "Observability check for meshed network is not implemented yet."
+  }
```
Jerry-Jinfeng-Guo: This, and the following three, will not pass the Python test in its current form.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: I don't think so, because the observability check is always run when doing state estimation. If we would raise an error here, no state estimation calculation for meshed grids would be possible. The current intention is to skip it altogether if the grid is meshed and raise a `SparseMatrixError` always (if any).

On the other hand, I see that from the perspective of "testing" the provided test cases, this is going to be an issue, as there is no entry point. However, I would suggest that we then check for `SparseMatrixError` instead (for now), and mark it such that we would change to the correct error in the future.

tests/data/state_estimation/meshed-network-observability/README.md
```diff
@@ -0,0 +1,10 @@
+/**
+ * This module provides test cases for an N-node system with N-1 measurements.
+ * The grids in tests are illustrated in the `test_network_diagrams.svg`
+ * Test cases in this directory demonstrate the only three possible scenarios:
+ * 1. Observable system with branch measurement.
+ * 2. Observable system without branch measurement (using only nodal measurements).
+ * 3. Unobservable system with branch measurement (x2).
+ * 
+ * Note: In theory, it is impossible to construct an unobservable case with only nodal measurements (i.e., without branch measurement).
```
figueroa1395: Nitpick: I would emphasize in this line that this is true for if N-1 nodal measurements are present in an N-node system.

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/params.json
figueroa1395: As mentioned above, I think that if do something like:
```suggestion
  "fail": {
    "raises": "SparseMatrixError",
    "reason": "If a meshed grid is unobservable, SparseMatrixError is thrown. To be adjusted once the meshed observability check is introduced."
  }
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
mgovers: > In the current code base, it is impossible to introduce the expected NotImplementedError as many SE related test cases assume the meshed observability check to pass.

does that concern the validation cases or the unit tests? If the former, the validation cases should always be observable, so that would be problematic. If the latter, the unit tests can probably be made to pass relatively easily by code-injecting the observability check. Thoughts?

tests/data/state_estimation/meshed-network-observability/README.md
```diff
@@ -0,0 +1,10 @@
...
+ */
```
mgovers: since this is markdown, you don't need to guard it. it also requires a `license` header


```suggestion
<!--
SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>

SPDX-License-Identifier: MPL-2.0
-->

This module provides test cases for an N-node system with N-1 measurements.
The grids in tests are illustrated in the `test_network_diagrams.svg`
Test cases in this directory demonstrate the only three possible scenarios:

2. Observable system with branch measurement.
3. Observable system without branch measurement (using only nodal measurements).
4. Unobservable system with branch measurement (x2). * 

Note: In theory, it is impossible to construct an unobservable case with only nodal measurements (i.e., without branch measurement).
```

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/input.json
figueroa1395, line 1: I don't want to be annoying, but do we want to have these tests as validation tests instead of unit tests? I ask because we aren't really comparing with any output. If the answer is yes, that is fine by me.

tests/data/state_estimation/meshed-network-observability/README.md
mgovers: maybe you can ASCII-art the grids, e.g.

```
                line_8   v16
              l18/----- 3 --l19
        line_7 |/       |line_9 (c14)
s0-- 1 ------- 2 ------ 4
     p13       | line_10
        line_11|
               5 ------- 6 --l20
            v17  line_12 (c15)
```

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/input.json
```diff
@@ -0,0 +1,43 @@
...
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "power_sigma": 1, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_current_sensor": [
+      {"id": 14, "measured_object": 9, "measured_terminal_type": 0, "i_sigma": 1, "i_measured": 0},
+      {"id": 15, "measured_object": 12, "measured_terminal_type": 0, "i_sigma": 1, "i_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 16, "measured_object": 3, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null},
+      {"id": 17, "measured_object": 5, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null}
```
figueroa1395: Is `"u_angle_measured": null` equivalent to not specifying it at all? So are these equivalent to magnitude only voltage sensors instead of phasor ones? If so, the measured angle needs to be specified (even if `0`), as only phasor voltage measurements count towards the observability.

In addition, if those two are phasor voltage measurements, then they will be used to "make" the grid observable, but if phasor voltage measurements are present, we always need one "left untouched" such that it can be used as a reference. So you have to add one additional one anywhere for that purpose.

tests/data/state_estimation/meshed-network-observability/observable-without-branch-measurement-2/input.json
```diff
@@ -0,0 +1,39 @@
...
+      {"id":10, "from_node": 2, "to_node": 5, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":11, "from_node": 3, "to_node": 4, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":12, "from_node": 5, "to_node": 6, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500}
+    ],
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "power_sigma": 1, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 14, "measured_object": 2, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null},
+      {"id": 15, "measured_object": 3, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null}
```
figueroa1395: Same comment as above.

tests/data/state_estimation/meshed-network-observability/observable-without-branch-measurement/input.json
```diff
@@ -0,0 +1,38 @@
...
+      {"id":11, "from_node": 3, "to_node": 4, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":12, "from_node": 5, "to_node": 6, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500}
+    ],
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "power_sigma": 1, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 14, "measured_object": 2, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null},
+      {"id": 15, "measured_object": 4, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null},
+      {"id": 16, "measured_object": 5, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null}
```
figueroa1395: Same comment as above.

tests/data/state_estimation/meshed-network-observability/unobservable-with-branch-measurement/input.json
```diff
@@ -0,0 +1,45 @@
...
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "power_sigma": 1, "p_measured": 0, "q_measured": 0},
+      {"id": 14, "measured_object": 9, "measured_terminal_type": 0, "power_sigma": 1, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_current_sensor": [
+      {"id": 15, "measured_object": 12, "measured_terminal_type": 2, "i_sigma": 1, "i_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 16, "measured_object": 3, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null},
+      {"id": 17, "measured_object": 6, "u_sigma": 1, "u_measured": 10, "u_angle_measured": null}
```
figueroa1395: Same comment as above but with a different intention: we want this to fail because it is not topologically observable, not because the number of phasor voltage measurements is not enough.

tests/data/state_estimation/meshed-network-observability/test_network_diagrams.svg
figueroa1395, line 1: Can you add a label 1-3 to the "passing" grid configurations here? So in the future we know easily which test belongs to which grid.

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/params.json
Jerry-Jinfeng-Guo: This should be thrown regardless of the meshed grid being observable or not, which is the reason `NotImplementedError` is used in the first place.

tests/data/state_estimation/meshed-network-observability/README.md
Jerry-Jinfeng-Guo: The svg is added as per request by @nitbharambe

tests/data/state_estimation/meshed-network-observability/README.md
Jerry-Jinfeng-Guo: This is already apparent in the context.

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/input.json
Jerry-Jinfeng-Guo: Same argument as `tests/data/state_estimation/not-independent-sensor`, if sensible.

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/params.json
mgovers: we don't currently throw `NotImplementedError`, right? on the contrary: if it's a meshed grid, we just try and see what happens. Best case, it's observable and we produce output; worst case, we throw a `SparseMatrixError`.

Adding `NotImplementedError` would be a breaking change for people that provide observable meshed grids

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: > I don't think so, because the observability check is always run when doing state estimation. If we would raise an error here, no state estimation calculation for meshed grids would be possible.

That's the whole point of having this commented out. In fact, what should have happened here is that we explicit pass the check for meshed networks and return true as a temporary path.

tests/data/state_estimation/meshed-network-observability/test_network_diagrams.svg
Jerry-Jinfeng-Guo: You mean `2`? They all have labels with distinct names.

Jerry-Jinfeng-Guo: Added.

Jerry-Jinfeng-Guo:
The main concern here @mgovers @figueroa1395 is the error type here. If it is indeed the preferred way, I will adjust the tests. That is, to throw `SparseMatrixError`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: > That's the whole point of having this commented out. In fact, what should have happened here is that we explicit pass the check for meshed networks and return true as a temporary path.

I see your point now. And I think it is a good addition. Maybe can be done in this PR since it is small and relevant?

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/params.json
figueroa1395: > we don't currently throw `NotImplementedError`, right? on the contrary: if it's a meshed grid, we just try and see what happens. Best case, it's observable and we produce output; worst case, we throw a `SparseMatrixError`.
> 
> Adding `NotImplementedError` would be a breaking change for people that provide observable meshed grids

This is what I meant, yes.

tests/data/state_estimation/meshed-network-observability/test_network_diagrams.svg
figueroa1395: I see you added the 2 label to the image in the lower left corner, is that correct? can you add also 1 and 2 to the other two?

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/input.json
figueroa1395: You are right, I see that there are only unit tests for the necessary condition, actually. Out of scope of this PR. But do you think we should do a follow up to also unit test this? To be honest, those unit tests (from a quick glance) seem to be, effectively, "validation" tests. Just wanted to ask about your opinion before resolving.

figueroa1395:
> The main concern here @mgovers @figueroa1395 is the error type here. If it is indeed the preferred way, I will adjust the tests. That is, to throw `SparseMatrixError`.

I'd say that is the way to go until we implement the complete observability test for meshed grids. Nevertheless, I think doing what you suggested in https://github.com/PowerGridModel/power-grid-model/pull/1044#discussion_r2222697421 would go really well with this, as those failing tests should fail once again after the observability check is completed.

mgovers:
> The main concern here @mgovers @figueroa1395 is the error type here. If it is indeed the preferred way, I will adjust the tests. That is, to throw `SparseMatrixError`.

cfr. #1049 , you can mark it as such:

```json
{
  "raises": {
    "raises": "NotObservableError",
    "reason": "Some clear reason as to why this grid is not observable."
  },
  "xfail": {
    "raises": "SparseMatrixError",
    "reason": "A sufficient observability check for meshed grids is not yet implemented. See also https://github.com/PowerGridModel/power-grid-model/issues/864."
  }
}
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: I'll add them then

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/input.json
Jerry-Jinfeng-Guo: I personally am perfectly fine with unit test the observability check function, but it seems that the pipeline is now not easily unit-testable. Shall we first start with the follow-up of addressing that? Once that is in place, we can simply just unit test the observability check with connected graph, not a full pgm network.

tests/data/state_estimation/meshed-network-observability/test_network_diagrams.svg
Jerry-Jinfeng-Guo: The labels and the folder names are 1-1 corresponding. Shall I in that case still introduce a redundant symbol?

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/input.json
figueroa1395: Indeed. Let's keep this out of scope for now then.

tests/data/state_estimation/meshed-network-observability/test_network_diagrams.svg
figueroa1395: I'd prefer so, very easy to map mentally with a quick glance.

Jerry-Jinfeng-Guo: Labeled 01-04. Although I do believe this introduced some non-uniformity.

figueroa1395: I personally don't mind it, because the images are a lot more clear to me. Maybe @mgovers or @nitbharambe have any thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Saw the change. Resolving.

tests/data/state_estimation/meshed-network-observability/test_network_diagrams.svg
mgovers: ok for me

tests/data/state_estimation/meshed-network-observability/01-observable-with-branch-measurement/params.json
```diff
@@ -0,0 +1,12 @@
...
+  "calculation_method": ["newton_raphson", "iterative_linear"],
+  "rtol": 1e-8,
+  "atol": {
+    "default": 1e-8,
+    ".+_residual": 5e-4
+  },
+  "fail": {
+    "raises": "NotImplementedError",
+    "reason": "Observability check for meshed network is not implemented yet."
+  }
```
mgovers: do we want to merge this before merging #1049 ? if not, please update to the current format.

Jerry-Jinfeng-Guo:
@figueroa1395 @mgovers branch rebased, error types updated and fail-pass updated

docs/user_manual/calculations.md
```diff
@@ -147,8 +147,9 @@ state estimation calculation in `power-grid-model`.
 - If no voltage phasor sensors are available, then both the following conditions shall be satisfied:
   - There are no global angle current sensors.
   - `n_unique_power_or_current_sensor >= n_bus - 1`.
-- Otherwise (if there are voltage phasor sensors available), the following condition shall be satisfied:
-  - `n_unique_power_or_current_sensor + n_voltage_sensor_with_phasor >= n_bus`
+- Otherwise (if there are voltage phasor sensors available, deducting one that can not be used),
+the following condition shall be satisfied:
+  - `n_unique_power_or_current_sensor + n_voltage_sensor_with_phasor - 1 >= n_bus - 1`
```
figueroa1395: Maybe the wording "... deducting one that can not be used" is not so straightforward if one is not very involved with observability tests in the context of power systems. Perhaps it can be re-worded to something along the lines of "... deducting/minus/subtracting one that is taken as the reference"

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -114,7 +114,7 @@ class SparseMatrixError : public PowerGridError {
         append_msg("Sparse matrix error, possibly singular matrix!\n"
                    "If you get this error from state estimation, "
                    "it might mean the system is not fully observable, i.e. not enough measurements.\n"
-                   "It might also mean that you are running into a corner case where PGM cannot resolve yet."
+                   "It might also mean that you are running into a corner case where PGM cannot resolve yet.\n"
```
figueroa1395: 👍

tests/data/state_estimation/meshed-network-observability/01-observable-with-branch-measurement/input.json
```diff
@@ -0,0 +1,43 @@
...
+      {"id": 15, "measured_object": 9, "measured_terminal_type": 0, "angle_measurement_type": 0, "i_sigma": 1, "i_measured": 0, "i_angle_measured": 0},
+      {"id": 16, "measured_object": 12, "measured_terminal_type": 0, "angle_measurement_type": 0, "i_sigma": 1, "i_measured": 0, "i_angle_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 17, "measured_object": 3, "u_sigma": 1, "u_measured": 10, "u_angle_measured": 0}
+    ],
+    "sym_load": [
+      {"id": 18, "node": 2, "status": 1, "type": 0, "p_specified": 9, "q_specified": 0},
+      {"id": 19, "node": 3, "status": 1, "type": 0, "p_specified": 9, "q_specified": 0},
+      {"id": 20, "node": 6, "status": 1, "type": 0, "p_specified": 9, "q_specified": 0}
```
figueroa1395: Taking the test as the ground truth, I believe the images are outdated. For example, the `sym_load` at node 4 present in the image, is not present in the test; similarly the `sym_power_sensor` 14 measuring `line` 10 is not in the image.

It is very likely that this is also the case for other images as well. Can you double check?

tests/data/state_estimation/meshed-network-observability/01-observable-with-branch-measurement/input.json
```diff
@@ -0,0 +1,43 @@
...
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "power_sigma": 1, "p_measured": 0, "q_measured": 0},
+      {"id": 14, "measured_object": 10, "measured_terminal_type": 1, "power_sigma": 1, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_current_sensor": [
+      {"id": 15, "measured_object": 9, "measured_terminal_type": 0, "angle_measurement_type": 0, "i_sigma": 1, "i_measured": 0, "i_angle_measured": 0},
+      {"id": 16, "measured_object": 12, "measured_terminal_type": 0, "angle_measurement_type": 0, "i_sigma": 1, "i_measured": 0, "i_angle_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 17, "measured_object": 3, "u_sigma": 1, "u_measured": 10, "u_angle_measured": 0}
```
figueroa1395: This comment is not meant to change anything if there is no need to. But keep in mind that the `sym_voltage_sensor` isn't (or rather can't) being used for the observability test, as it is the reference voltage. This grid is only observable because `node` 4 has zero injection, so there is effectively an additional (virtual) sensor besides the ones here.

tests/data/state_estimation/meshed-network-observability/02-unobservable-with-branch-measurement/params.json
```diff
@@ -0,0 +1,15 @@
+{
+  "calculation_method": ["newton_raphson", "iterative_linear"],
+  "rtol": 1e-8,
+  "atol": {
+    "default": 1e-8,
+    ".+_residual": 5e-4
+  },
+  "raises": {
+    "raises": "NotObservableError",
+    "reason": "Some clear reason as to why this grid is not observable."
```
figueroa1395: Isn't the reason that test-2 is intended to be unobservable (According to the images)?

tests/data/state_estimation/meshed-network-observability/03-observable-without-branch-measurement/input.json
```diff
@@ -0,0 +1,37 @@
...
+    ],
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "power_sigma": 1, "p_measured": 0, "q_measured": 0},
+      {"id": 14, "measured_object": 10, "measured_terminal_type": 1, "power_sigma": 1, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 15, "measured_object": 3, "u_sigma": 1, "u_measured": 10, "u_angle_measured": 0}
+    ],
+    "sym_load": [
+      {"id": 16, "node": 3, "status": 1, "type": 0, "p_specified": 9, "q_specified": 0}
```
figueroa1395, line 34: Nitpick: Add another `sym_load` at `node` 6 so that leaf isn't so "empty". The grid should remain observable. 

Edit: I see now that perhaps this is intentional to "cover more cases". If so, ignore this comment and resolve.

tests/data/state_estimation/meshed-network-observability/03-observable-without-branch-measurement/input.json
```diff
@@ -0,0 +1,37 @@
...
+      {"id":10, "from_node": 2, "to_node": 5, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":11, "from_node": 3, "to_node": 4, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":12, "from_node": 5, "to_node": 6, "from_status": 1, "to_status": 1, "r1": 1.0, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500}
+    ],
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "power_sigma": 1, "p_measured": 0, "q_measured": 0},
+      {"id": 14, "measured_object": 10, "measured_terminal_type": 1, "power_sigma": 1, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 15, "measured_object": 3, "u_sigma": 1, "u_measured": 10, "u_angle_measured": 0}
```
figueroa1395: Same comment as for test grid 1. Just an FYI about the voltage phasor sensor here. In this case it is observable because `node` 2 has zero injection.

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/params.json
figueroa1395: This has been addressed with https://github.com/PowerGridModel/power-grid-model/pull/1049 and https://github.com/PowerGridModel/power-grid-model/blob/0626f9c4c425b0bae89106fbf62423b0850c7799/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp#L227, resolving.

tests/data/state_estimation/meshed-network-observability/observable-with-branch-measurement/input.json
figueroa1395: This was discussed offline, and the current tests don't have this problem. Resolving.

tests/data/state_estimation/meshed-network-observability/observable-without-branch-measurement-2/input.json
figueroa1395: Idem.

tests/data/state_estimation/meshed-network-observability/observable-without-branch-measurement/input.json
figueroa1395: Idem.

tests/data/state_estimation/meshed-network-observability/unobservable-with-branch-measurement/input.json
figueroa1395: Idem.

nitbharambe:
@Jerry-Jinfeng-Guo For now, lets change all branch current sensors to branch power sensors and think about the handling of current sensors later on.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -150,6 +150,8 @@ inline bool necessary_observability_condition(ObservabilitySensorsResult const&
         throw NotObservableError{
             "The total number of independent power sensors is not enough to make the grid observable."};
     }
+    // If there are any voltage phasor sensor, one will not be used:
```
Jerry-Jinfeng-Guo:
```suggestion
    // If there are any voltage phasor sensor, one will be reserved as reference and not be used:
```

docs/user_manual/calculations.md
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/pull/1044/commits/9fda52e86c15d917aa667a3b484498f7301cdac8

tests/data/state_estimation/meshed-network-observability/01-observable-with-branch-measurement/input.json
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/pull/1044/commits/7a35298eb7fca6c0f5c1109504aad54431c3582c

tests/data/state_estimation/meshed-network-observability/02-unobservable-with-branch-measurement/params.json
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/pull/1044/commits/4cf53856bbd2d4364d1caffcb12863dc8f1827f2

tests/data/state_estimation/meshed-network-observability/01-observable-with-branch-measurement/params.json
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/pull/1044/commits/6f06c34e20d4afecaf2da5af00b659d99eb0ba73

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -221,6 +223,9 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     if (topo.is_radial) {
         is_sufficient_condition_met = detail::sufficient_observability_condition(y_bus_structure, observability_sensors,
                                                                                  n_voltage_phasor_sensors);
+    } else {
+        // This is a temporary path for meshed grids
+        return ObservabilityResult{.is_observable = true, .is_possibly_ill_conditioned = false};
```
nitbharambe: Can the return below be moved inside the if statement?

nitbharambe: The temporary path would modify behaviour of currently meshed but NotObservableError cases because of low count of sensors to SparseMatrixError right? (Edit: I see it is an unused attribute. so nevermind this)

nitbharambe:
The cases look good. Ready to go post comments resolving.

figueroa1395:
> @Jerry-Jinfeng-Guo For now, lets change all branch current sensors to branch power sensors and think about the handling of current sensors later on.

I'm curious about this. I know the type of sensors is irrelevant for the purpose of the observability test, but why do we want to "not handle" current sensors for now?

Edit: I see some tests use current sensors anyway.

Edit 2: Discussed offline: Current sensors with `i_measured` and `i_angle_measured` both at zero are "invalidated", so this was triggering different behavior with the power sensors' counterpart. This is to be better documented, but after updating the values, this isn't a problem. Please add anything if I miss-understood the matter.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -221,9 +224,12 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     if (topo.is_radial) {
         is_sufficient_condition_met = detail::sufficient_observability_condition(y_bus_structure, observability_sensors,
                                                                                  n_voltage_phasor_sensors);
+        return ObservabilityResult{.is_observable = is_necessary_condition_met && is_sufficient_condition_met,
+                                   .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
     }
-    return ObservabilityResult{.is_observable = is_necessary_condition_met && is_sufficient_condition_met,
-                               .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
+    // This is a temporary path for meshed grids
+    // pos_ill_condition should be passed via observability_sensors but python side on ubuntu has trouble with defaults
```
figueroa1395: What do you mean? Why `.is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned` is not possible?

tests/data/state_estimation/meshed-network-observability/02-unobservable-with-branch-measurement/input.json
```diff
@@ -0,0 +1,45 @@
...
+      {"id": 7, "from_node": 1, "to_node": 2, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id": 8, "from_node": 2, "to_node": 3, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id": 9, "from_node": 2, "to_node": 4, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":10, "from_node": 2, "to_node": 5, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":11, "from_node": 3, "to_node": 4, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":12, "from_node": 5, "to_node": 6, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500}
+    ],
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "p_sigma": 1e6, "q_sigma": 1e6, "p_measured": 0, "q_measured": 0},
+      {"id": 14, "measured_object": 8, "measured_terminal_type": 1, "p_sigma": 1e6, "q_sigma": 1e6, "p_measured": 0, "q_measured": 0}
```
figueroa1395: In the image, I believe this sensor is wrongly positioned at node 3 instead of line 8.

tests/data/state_estimation/meshed-network-observability/03-observable-without-branch-measurement/input.json
```diff
@@ -0,0 +1,37 @@
...
+      {"id":10, "from_node": 2, "to_node": 5, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":11, "from_node": 3, "to_node": 4, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":12, "from_node": 5, "to_node": 6, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500}
+    ],
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "p_sigma": 1e6, "q_sigma": 1e6, "p_measured": 0, "q_measured": 0},
+      {"id": 14, "measured_object": 10, "measured_terminal_type": 1, "p_sigma": 1e6, "q_sigma": 1e6, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 15, "measured_object": 3, "u_sigma": 1, "u_measured": 10000, "u_angle_measured": 0}
```
figueroa1395: The sensors here don't correspond to the ones drawn in image 3.

tests/data/state_estimation/meshed-network-observability/04-observable-without-branch-measurement/input.json
```diff
@@ -0,0 +1,40 @@
...
+      {"id":11, "from_node": 3, "to_node": 4, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500},
+      {"id":12, "from_node": 5, "to_node": 6, "from_status": 1, "to_status": 1, "r1": 0.1, "x1": 0.0, "c1": 0.0, "tan1": 0.0, "i_n": 500}
+    ],
+    "sym_power_sensor": [
+      {"id": 13, "measured_object": 0, "measured_terminal_type": 2, "p_sigma": 1e6, "q_sigma": 1e6, "p_measured": 0, "q_measured": 0},
+      {"id": 14, "measured_object": 2, "measured_terminal_type": 9, "p_sigma": 1e6, "q_sigma": 1e6, "p_measured": 0, "q_measured": 0},
+      {"id": 15, "measured_object": 3, "measured_terminal_type": 9, "p_sigma": 1e6, "q_sigma": 1e6, "p_measured": 0, "q_measured": 0}
+    ],
+    "sym_voltage_sensor": [
+      {"id": 16, "measured_object": 5, "u_sigma": 1, "u_measured": 1000, "u_angle_measured": 0}
```
figueroa1395, line 32: This sensor is missing in the image. Or is it intentional because it isn't being used as an observability sensor, but as the voltage reference?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: Python test side on ubunty and macos would complain in ci

tests/data/state_estimation/meshed-network-observability/02-unobservable-with-branch-measurement/input.json
Jerry-Jinfeng-Guo: Moved from `branch_to` to `node` 'to'  https://github.com/PowerGridModel/power-grid-model/pull/1044/commits/0987f98e345b0efb984ffa7a315df00a96d6625d

tests/data/state_estimation/meshed-network-observability/04-observable-without-branch-measurement/input.json
Jerry-Jinfeng-Guo: All cases have but one voltage sensor, indeed, for the purpose of satisfying a voltage reference

Jerry-Jinfeng-Guo: Explicitly marked all reference voltage sensor https://github.com/PowerGridModel/power-grid-model/pull/1044/commits/6f49892d4ab20b46429b117067fd339ba014f4a2

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
nitbharambe: @Jerry-Jinfeng-Guo Unresolved. it is not unused. It determines if we use perturbation or not. 
This would mean we end up using perturbation where we previously did not, right?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Can you share how they complain? That seems weirdly odd. Specially at the Python side? Is it because the test cases are ill conditioned and trigger the pivot thingy?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: We will have to take a look at the Python on Ubuntu/Macos side when I did pass the attribute through the `observability_sensors`

tests/data/state_estimation/meshed-network-observability/02-unobservable-with-branch-measurement/params.json
```diff
@@ -0,0 +1,15 @@
+{
+  "calculation_method": ["newton_raphson", "iterative_linear"],
+  "rtol": 1e-8,
+  "atol": {
+    "default": 1e-8,
+    ".+_residual": 5e-4
+  },
+  "raises": {
+    "raises": "NotObservableError",
```
figueroa1395: This one should raise `SparseMatrixError`, shouldn't it? That is why CI is currently complaining. This is because we know it is not observable, but the check isn't implemented yet for meshed, so it will be unobservable but `throw` `SparseMatrixError` in this case.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: We decided here that the current behavior in the newest commit reflects what should happen: observability check return not observable for meshed network, but still pass to the calculation phase where real errors (`SparseMatrixError`) might be thrown, which matches the existing behavior.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: We decided here that the current behavior in the newest commit reflects what should happen: observability check return not observable for meshed network, but still pass to the calculation phase where real errors (SparseMatrixError) might be thrown, which matches the existing behavior.

tests/data/state_estimation/meshed-network-observability/02-unobservable-with-branch-measurement/params.json
Jerry-Jinfeng-Guo: Applied in https://github.com/PowerGridModel/power-grid-model/pull/1044/commits/eced1d51ff1dfddd52c02cb1cb3cbe8c5f1e6d3b

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -221,7 +224,12 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     if (topo.is_radial) {
         is_sufficient_condition_met = detail::sufficient_observability_condition(y_bus_structure, observability_sensors,
                                                                                  n_voltage_phasor_sensors);
+        return ObservabilityResult{.is_observable = is_necessary_condition_met && is_sufficient_condition_met,
+                                   .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
     }
+    // This is a temporary path for meshed grids
+    // pos_ill_condition should be passed via observability_sensors but python side on ubuntu has trouble with defaults
+    //  ToDo(JGuo): the `is_possibly_ill_conditioned` is set to false for now to avoid perturbation
```
figueroa1395: This can now be removed.

Edit: Just the TODO. Github visualization is not nice.

figueroa1395 (approved):
I believe this is good to go.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -222,6 +225,9 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
         is_sufficient_condition_met = detail::sufficient_observability_condition(y_bus_structure, observability_sensors,
                                                                                  n_voltage_phasor_sensors);
     }
+    // This is a temporary path for meshed grids
+    // pos_ill_condition should be passed via observability_sensors but python side on ubuntu has trouble with defaults
```
figueroa1395: Now I see that this should also been removed as it has been solved 😄

nitbharambe (approved):

PR #1046, opened 2025-07-22
Internal benchmark: add benchmarks for state estimation, short circuit and automatic tap changer

mgovers:
In prep for the main model refactoring, this PR extends the internal benchmark. This way, we can really be sure that we do not break anything performance-wise during the refactoring.

NOTE: only 1 3-phase fault and 1 automatic tap regulator are added. Scaling is not included in this PR.

In addition, tracking process memory usage is also not added in this PR.

### DIFF
--- a/tests/benchmark_cpp/fictional_grid_generator.hpp
+++ b/tests/benchmark_cpp/fictional_grid_generator.hpp
@@ -608,10 +608,8 @@ class FictionalGridGenerator {
     }
 
     void generate_fault() {
-        input_.fault.emplace_back(FaultInput{.id = id_gen_++,
-                                             .status = 1,
-                                             .fault_type = FaultType::three_phase,
-                                             .fault_object = input_.node.back().id});
+        input_.fault.emplace_back(FaultInput{
+            .id = id_gen_++, .status = 1, .fault_type = FaultType::three_phase, .fault_object = input_.node.back().id});
     }
 
     void generate_tap_changer() {


tests/benchmark_cpp/benchmark.cpp
```diff
@@ -140,25 +179,181 @@ int main(int /* argc */, char** /* argv */) {
...
+    //                                    .calculation_method = newton_raphson});
+    // benchmarker.run_benchmark(option, {.calculation_type = state_estimation,
+    //                                    .calculation_symmetry = symmetric,
+    //                                    .calculation_method = iterative_linear});
+    // benchmarker.run_benchmark(option, {.calculation_type = state_estimation,
+    //                                    .calculation_symmetry = asymmetric,
+    //                                    .calculation_method = newton_raphson});
+    // benchmarker.run_benchmark(option, {.calculation_type = state_estimation,
+    //                                    .calculation_symmetry = asymmetric,
+    //                                    .calculation_method = iterative_linear});
```
Jerry-Jinfeng-Guo: What is the purpose of this section?

mgovers: oh sorry, was meant to be uncommented. my bad.

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -20,57 +20,93 @@ MathSolverDispatcher const& get_math_solver_dispatcher() {
...
-        } else {
-            title += "Iterative current method";
-        }
-        std::cout << "=============" << title << "=============\n";
+        std::cout << get_benchmark_run_title(option, model_options, batch_size) << '\n';
+
+        auto const run = [this, &model_options, &info, batch_size](Idx batch_size_) {
+            switch (model_options.calculation_type) {
+            case short_circuit:
+                return run_calculation<ShortCircuitOutputData>(model_options, batch_size_, info);
```
figueroa1395, line 92: Fun fact: I didn't know the output data for short circuit was this special :o

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -140,25 +179,181 @@ int main(int /* argc */, char** /* argv */) {
...
     option.has_mv_ring = false;
     option.has_lv_ring = false;
-    benchmarker.run_benchmark<symmetric_t>(option, newton_raphson, batch_size);
-    benchmarker.run_benchmark<symmetric_t>(option, newton_raphson, batch_size, 6);
-    benchmarker.run_benchmark<symmetric_t>(option, linear);
-    benchmarker.run_benchmark<symmetric_t>(option, iterative_current);
-    benchmarker.run_benchmark<asymmetric_t>(option, newton_raphson);
-    benchmarker.run_benchmark<asymmetric_t>(option, linear);
-    // benchmarker.run_benchmark<asymmetric_t>(option, iterative_current);
+    benchmarker.run_benchmark(
```
figueroa1395, line 190: I see originally it was done sequential and with `6`, as you did here. However, why was `6` specifically the choice? Wouldn't it make more sense to do `-1`, and actually do it for both cases for each run? In any case, the benchmark is machine dependent and we are to run it at every refactoring step.

Also, why is only `NewtonRaphson` run in parallel? Why not the others?

Finally, why is `linear_current` not benchmarked?

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -140,25 +179,181 @@ int main(int /* argc */, char** /* argv */) {
...
+    benchmarker.run_benchmark(option, {.calculation_type = power_flow,
+                                       .calculation_symmetry = symmetric,
+                                       .calculation_method = iterative_current,
+                                       .max_iter = 100});
+    benchmarker.run_benchmark(
+        option,
+        {.calculation_type = power_flow, .calculation_symmetry = asymmetric, .calculation_method = newton_raphson});
+    benchmarker.run_benchmark(
+        option, {.calculation_type = power_flow, .calculation_symmetry = asymmetric, .calculation_method = linear});
+    // benchmarker.run_benchmark<asymmetric_t>(option, {.calculation_type = power_flow,
```
figueroa1395: Why was and is this one commented out?

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -140,25 +179,181 @@ int main(int /* argc */, char** /* argv */) {
...
+                               .calculation_method = linear,
+                               .optimizer_type = automatic_tap_adjustment},
+                              batch_size);
+    benchmarker.run_benchmark(option,
+                              {.calculation_type = power_flow,
+                               .calculation_symmetry = asymmetric,
+                               .calculation_method = linear,
+                               .optimizer_type = automatic_tap_adjustment},
+                              batch_size);
+
```
figueroa1395: If we are testing the tap adjustment here, shouldn't we just pick the fastest method, use it for both `symmetric` and `asymmetric`, and run over all the tap changing strategies? Why doing `linear` and `newton_raphson` for the `symmetric` case?

tests/benchmark_cpp/fictional_grid_generator.hpp
```diff
@@ -33,6 +37,12 @@ struct InputData {
     std::vector<SymLoadGenInput> sym_load;
     std::vector<AsymLoadGenInput> asym_load;
     std::vector<ShuntInput> shunt;
+    std::vector<SymVoltageSensorInput> sym_voltage_sensor;
+    std::vector<AsymVoltageSensorInput> asym_voltage_sensor;
+    std::vector<SymPowerSensorInput> sym_power_sensor;
+    std::vector<AsymPowerSensorInput> asym_power_sensor;
+    std::vector<FaultInput> fault;
+    std::vector<TransformerTapRegulatorInput> transformer_tap_regulator;
```
figueroa1395, line 45: Should we also include current sensors here? Probably not needed (as in no value added), but double checking.

tests/benchmark_cpp/fictional_grid_generator.hpp
```diff
@@ -459,6 +522,117 @@ class FictionalGridGenerator {
...
+                    .power_sigma = nan,
+                    .p_measured = load.p_specified,
+                    .q_measured = load.q_specified,
+                    .p_sigma = {power_tol * cabs(load.p_specified(0)), power_tol * cabs(load.p_specified(1)),
+                                power_tol * cabs(load.p_specified(2))},
+                    .q_sigma = {power_tol * cabs(load.q_specified(0)), power_tol * cabs(load.q_specified(1)),
+                                power_tol * cabs(load.q_specified(2))}};
+            });
+
+        std::ranges::transform(input_.line, std::back_inserter(input_.sym_power_sensor), [this](LineInput const& line) {
```
figueroa1395, line 600: Add a small `// branch power sensors` comment, as this isn't an appliance anymore. Additionally, do we need asymmetric line power sensors? Or the asymmetry here just comes from the loads, source, etc?

tests/benchmark_cpp/benchmark.cpp
mgovers: whoops hahahaha, that's actually why the output.hpp is so much larger than it probably should be

tests/benchmark_cpp/benchmark.cpp
mgovers: 1. i don't think there's a specific reason, but it's good to keep it 1 specific threading level across all devices.
2. threading should be independent of the actual model that is run. Since NR is usually the slowest one, it provides a decent benchmark. In fact, i have even thought about not testing state estimation and short circuit in parallel, but i went for it because they are different code paths
3. not sure but it's only for asymmetric. there is probably a good reason for it that tony found. i can try to see what happens when i uncomment it.

tests/benchmark_cpp/benchmark.cpp
mgovers: no idea... i didn't want to introduce more difficult stuff. i will try to see what happens when i uncomment it, but there must've been a good reason for it

tests/benchmark_cpp/benchmark.cpp
mgovers: i thought about doing that indeed, but the benchmark becomes even larger. i decided not to go with it because it's not our current focus point and the benchmark would become even larger.

because linear provides a fallback for NR, i came up with benchmarking these two... but you're right, it's probably better to benchmark the different optimizer strategies

tests/benchmark_cpp/fictional_grid_generator.hpp
mgovers: i thought so as well, but it's going to be very hard to obtain some kind of reasonable bounds on the current and indeed the added value is minimal

tests/benchmark_cpp/benchmark.cpp
figueroa1395: Alright. Give it a go when uncommenting it and let me know what's up. Besides that, feel free to resolve unless you want to post here some results of uncommenting.

tests/benchmark_cpp/fictional_grid_generator.hpp
figueroa1395: Yup, let's just omit them for now unless someone has a strong opinion about them. No need to overcomplicate this.

tests/benchmark_cpp/benchmark.cpp
nitbharambe: 3. I imagine linear and linear current result in same benchmark. So no reason to duplicate it.

tests/benchmark_cpp/fictional_grid_generator.hpp
nitbharambe: We dont need asymmetric power sensors on asymmetric line specially. asymmetric power sensors on normal line should be equivalent for benchmark.

tests/benchmark_cpp/benchmark.cpp
mgovers: ohhhhh i understand: `linear_current` was just forgotten, but `iterative_current` was explicitly disabled for asymmetric because can diverge and/or take forever (100 iterations is A LOT)

tests/benchmark_cpp/benchmark.cpp
mgovers: iteration diverge

tests/benchmark_cpp/fictional_grid_generator.hpp
mgovers: resolved the `//`

as @nitbharambe  mentioned, adding asymmetric stuff is not necessary

tests/benchmark_cpp/benchmark.cpp
mgovers: iteration diverge
```
============= Benchmark case: radial grid, asymmetric, Iterative current method =============

*****Run with initialization*****
Number of nodes: 2605

An exception was raised during execution: Iteration failed to converge after 100 iterations! Max deviation: 59.1282, error tolerance: 1e-08.

0000.Total: 0.0345428
1000.	Build model: 0.001319

*****Run without initialization*****
Number of nodes: 2605

An exception was raised during execution: Iteration failed to converge after 100 iterations! Max deviation: 59.1282, error tolerance: 1e-08.

0000.Total: 0.0280372
0000.Total: 0.0280372
```

tests/benchmark_cpp/benchmark.cpp
mgovers:
```
============= Benchmark case: meshed grid, asymmetric, Iterative current method =============

*****Run with initialization*****
Number of nodes: 1804

An exception was raised during execution: Iteration failed to converge after 100 iterations! Max deviation: 439.663, error tolerance: 1e-08.

0000.Total: 0.026041
1000.	Build model: 0.0006985

*****Run without initialization*****
Number of nodes: 1804

An exception was raised during execution: Iteration failed to converge after 100 iterations! Max deviation: 439.663, error tolerance: 1e-08.

0000.Total: 0.021756
0000.Total: 0.021756
```

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -20,57 +20,93 @@ MathSolverDispatcher const& get_math_solver_dispatcher() {
...
-            title += "Newton-Raphson method";
-        } else if (calculation_method == CalculationMethod::linear) {
-            title += "Linear method";
-        } else {
-            title += "Iterative current method";
-        }
-        std::cout << "=============" << title << "=============\n";
+        std::cout << get_benchmark_run_title(option, model_options, batch_size) << '\n';
+
+        auto const run = [this, &model_options, &info, batch_size](Idx batch_size_) {
```
nitbharambe: 2 different batch_sizes: batch_size and batch_size_. seems werid

mgovers: typo

tests/benchmark_cpp/benchmark.cpp
mgovers: done

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -359,9 +359,9 @@ int main(int /* argc */, char** /* argv */) {
     benchmarker.run_benchmark(option, {.calculation_type = state_estimation,
                                        .calculation_symmetry = symmetric,
                                        .calculation_method = iterative_linear});
-    benchmarker.run_benchmark(option, {.calculation_type = state_estimation,
-                                       .calculation_symmetry = asymmetric,
-                                       .calculation_method = newton_raphson});
+    // benchmarker.run_benchmark(option, {.calculation_type = state_estimation,
+    //                                    .calculation_symmetry = asymmetric,
+    //                                    .calculation_method = newton_raphson});
```
figueroa1395, line 364: Doing a final check. Is this and below intended to be commented out? I suppose yes because of the commit message, but why do we expect NRSE for asymmetric calculations to diverge here? Should that happen?

tests/benchmark_cpp/fictional_grid_generator.hpp
```diff
@@ -611,7 +611,7 @@ class FictionalGridGenerator {
         input_.fault.emplace_back(FaultInput{.id = id_gen_++,
                                              .status = 1,
                                              .fault_type = FaultType::three_phase,
-                                             .fault_object = input_.node.front().id});
+                                             .fault_object = input_.node.back().id});
```
figueroa1395: Was this a bug?

tests/benchmark_cpp/fictional_grid_generator.hpp
```diff
@@ -523,8 +523,8 @@ class FictionalGridGenerator {
     }
 
     void generate_sensors() {
-        constexpr double voltage_tol = 0.01;
-        constexpr double power_tol = 0.05;
+        constexpr double voltage_tol = 0.1;
+        constexpr double power_tol = 0.1;
```
figueroa1395, line 527: Why the specific change?

tests/benchmark_cpp/benchmark.cpp
mgovers: indeed iteration diverge for NRSE. i don't think there's anything we can do here except for changing the grid or coming up with a better initial condition.

tests/benchmark_cpp/fictional_grid_generator.hpp
mgovers: not a bug but if the fault is on the source node, there's basically nothing being investigated (short circuit between the source and itself). By putting it on another node, the grid is actually calculated

tests/benchmark_cpp/fictional_grid_generator.hpp
mgovers: attempt to improve the NRSE convergence. didn't work, though. i can revert if you prefer

figueroa1395: That's alright. Resolving.

tests/benchmark_cpp/benchmark.cpp
figueroa1395: Given the structure of the benchmark, it is fine by me. So I will resolve now.

nitbharambe (approved):

PR #1049, opened 2025-07-22
Validation cases: split validation case `fail` into `xfail` and `raises`

### LINKED ISSUE #1047: Split failing validation cases into separate "should fail" and "should not but does fail"
Also consider reporting cases that pass but should not (`xpass`?)

mgovers:
Closes #1047


tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -388,7 +388,7 @@ std::optional<CaseParam> construct_case(std::filesystem::path const& case_dir, j
         }
     }
 
-    param.fail = calculation_method_params.contains("fail");
+    param.fail = calculation_method_params.contains("xfail") || calculation_method_params.contains("raises");
```
mgovers, line 391: if you have a better naming, please LMK

tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-any-max-iter/params.json
```diff
@@ -3,7 +3,7 @@
   "tap_changing_strategy": "any_valid_tap",
   "rtol": 1e-05,
   "atol": 1e-05,
-  "fail": {
+  "raises": {
```
mgovers, line 6: this one may be up for discussion on whether it should be `raises` or `xfail`...

tests/unit/test_0Z_model_validation.py
```diff
@@ -100,42 +123,43 @@ def test_single_validation(
...
-        assert np.all(input_array["id"][indexer_array] == ids_array)
-
-    # test calculate with only node and source result
-    kwargs = dict(base_kwargs, **{"output_component_types": [ComponentType.node, ComponentType.source]})
-    result = calculation_function(model, **supported_kwargs(kwargs=kwargs, supported=calculation_args))
-    assert set(result.keys()) == {ComponentType.node, ComponentType.source}
-    kwargs = dict(base_kwargs, **{"output_component_types": {ComponentType.node, ComponentType.source}})
-    result = calculation_function(model, **supported_kwargs(kwargs=kwargs, supported=calculation_args))
-    assert set(result.keys()) == {ComponentType.node, ComponentType.source}
+    with maybe_raises(params):
```
mgovers, line 126: i did consider doing a `with` clause for every sub-step, but it is not known (in the current implementation) whether the error should happen during model creation or later during the calculation. I went with just wrapping everything in a single `with` clause for ease of implementation and to prevent overengineering

nitbharambe: Why would a with clause for every sub step be a solution? Did you mean to specify at which stage the raises is supposed to occur?

mgovers: it's the other way around: i'd like - for instance - to check that every scenario raises the same error. However, that would indeed require specifying the failing stage, and i did not like that solution. That's why i went with the one-clause-to-rule-them-all

tests/cpp_validation_tests/test_validation.cpp
figueroa1395: So `xfail` means that this is currently failing because it is not implemented yet, or there is a bug, etc.? And `raises` is used where we actually want it to fail and hence `raise` something on purpose, right?

Naming is fine by me.

tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-any-max-iter/params.json
figueroa1395: Does this test that with an "ill" constructed input (which we know in advance will reach the max iterations threshold), the `TapPositionOptimizer` will fail and raise `MaxIterationReached`? If so, I'd say we should keep `raises`.

tests/unit/test_0Z_model_validation.py
figueroa1395: That is definitely an improving step at some point. I would assume that in large batch calculations different errors are to be raised per scenario. Maybe not in the DSO context, but would definitely happen in a training AI context. But I'd say out of scope of this.

figueroa1395 (approved):
LGTM.

tests/cpp_validation_tests/test_validation.cpp
mgovers: > So xfail means that this is currently failing because it is not implemented yet, or there is a bug, etc.? And raises is used where we actually want it to fail and hence raise something on purpose, right?

idd

> Naming is fine by me.

:+1:

tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-any-max-iter/params.json
mgovers: alright

mgovers:
@Jerry-Jinfeng-Guo this will cause a simple merge conflict with #1044 . Renaming `fail` to either `raises` or `xfail` will do the trick. Probably best to merge this PR first.

mgovers:
> @Jerry-Jinfeng-Guo this will cause a simple merge conflict with #1044 . Renaming `fail` to either `raises` or `xfail` will do the trick. Probably best to merge this PR first.

EDIT: #1047 now also includes `xpass`, which is what you will need

tests/unit/utils.py
```diff
@@ -35,9 +35,21 @@
...
+    from _pytest.outcomes import Failed as _Failed  # pylint: disable=import-outside-toplevel
+except ImportError:
+    import warnings  # pylint: disable=import-outside-toplevel
+
+    warnings.warn(
+        """Failed exception is not available in the current pytest version.Some validation cases tests marked """
+        """ as {"xfail": {"raises": "Failed"}} may report as xfail even though they are actual problems."""
+    )
+    _Failed = None
+
```
figueroa1395: Let's discuss in tomorrow's knowledge sharing session.

mgovers:
adding do-not-merge label cfr. https://github.com/PowerGridModel/power-grid-model/pull/1049#discussion_r2224999073

docs/examples/Make Test Dataset.ipynb
```diff
@@ -422,6 +424,101 @@
...
+    "| raises `AError`   | raises `BError`   | `{\"raises\": {\"raises\": \"AError\"}, \"xfail\": {\"raises\": \"BError\"}}` |   <span style=\"color : red\">F (Failed)</span>   |  <span style=\"color : red\">F (XPASS)</span>   |                                                  <span style=\"color : yellow\">x (XFAIL)                                                   | <span style=\"color : red\">F (CError)</span> |\n",
+    "| raises `AError`   | pass              | `{\"raises\": {\"raises\": \"AError\"}, \"xfail\": {\"raises\": \"Failed\"}}` | <span style=\"color : yellow\">x (XFAIL) </span>  |  <span style=\"color : red\">F (XPASS)</span>   | <span style=\"color : red\">F (BError)</span> if `_pytest.outcomes.Failed` is available, else <span style=\"color : yellow\">x (XFAIL)</span> |                                             |\n",
```
mgovers: i've spent quite some time trying different orders but they are all meh. any thoughts on your end as to which order is preferred?

docs/examples/Make Test Dataset.ipynb
```diff
@@ -258,9 +260,9 @@
...
       "    ],\n",
       "    \"line\": [\n",
-      "      {\"id\": 3, \"from_node\": 1, \"to_node\": 2, \"from_status\": 1, \"to_status\": 1, \"r1\": 0.25, \"x1\": 0.2, \"c1\": 1e-05, \"tan1\": 0, \"i_n\": 1000},\n",
-      "      {\"id\": 5, \"from_node\": 2, \"to_node\": 6, \"from_status\": 1, \"to_status\": 1, \"r1\": 0.25, \"x1\": 0.2, \"c1\": 1e-05, \"tan1\": 0, \"i_n\": 1000},\n",
-      "      {\"id\": 8, \"from_node\": 1, \"to_node\": 6, \"from_status\": 1, \"to_status\": 1, \"r1\": 0.25, \"x1\": 0.2, \"c1\": 1e-05, \"tan1\": 0, \"i_n\": 1000}\n",
+      "      {\"id\": 3, \"from_node\": 1, \"to_node\": 2, \"from_status\": 1, \"to_status\": 1, \"r1\": 0.25, \"x1\": 0.20000000000000001, \"c1\": 1.0000000000000001e-05, \"tan1\": 0, \"i_n\": 1000},\n",
+      "      {\"id\": 5, \"from_node\": 2, \"to_node\": 6, \"from_status\": 1, \"to_status\": 1, \"r1\": 0.25, \"x1\": 0.20000000000000001, \"c1\": 1.0000000000000001e-05, \"tan1\": 0, \"i_n\": 1000},\n",
+      "      {\"id\": 8, \"from_node\": 1, \"to_node\": 6, \"from_status\": 1, \"to_status\": 1, \"r1\": 0.25, \"x1\": 0.20000000000000001, \"c1\": 1.0000000000000001e-05, \"tan1\": 0, \"i_n\": 1000}\n",
```
figueroa1395, line 265: lol at these.

docs/examples/Make Test Dataset.ipynb
```diff
@@ -422,6 +424,101 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "3c018413",
+   "metadata": {},
+   "source": [
+    "#### Detailed configuration with the params.json\n",
+    "\n",
+    "##### Validation cases with exceptions\n",
+    "\n",
+    "In certain cases, you may want to create test cases that either \"should raise an exception\" and/or \"currently has behavior that differs from the intended behavior\". Examples are:\n",
```
figueroa1395:
```suggestion
    "In certain cases, you may want to create test cases that either \"should raise an exception\" and/or \"currently have a behavior that differs from the intended one\". Examples are:\n",
```

docs/examples/Make Test Dataset.ipynb
```diff
@@ -422,6 +424,101 @@
...
+    "\n",
+    "In addition to the PGM exception types and some Python built-in exception types, there are two types worth explicitly mentioning:\n",
+    "\n",
+    "* `AssertionError` can be used to denote tests in which the actual values are different from the expected values.\n",
+    "* `Failed` can be used to denote tests that should raise (`raises`) but actually do not raise any exceptions at all (`xfail`).\n",
+    "  * **NOTE:** If `_pytest.outcomes.Failed` is not found, there is a fall-back to the default [`pytest.mark.xfail`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail) implementation. That means, that a test is marked `xfail` if it throws any exception that is not the intended exception, not `fail`. This is a known limitation of the current implementation.\n",
```
figueroa1395, line 454: 👍

docs/examples/Make Test Dataset.ipynb
figueroa1395: I would organize it how you showed me offline, i.e. first all the pass in intended behavior, and then all the raises... However, I don't think this is too important as the table is concise and small enough. So I'd say don't worry too much about it.

On the other hand, there are some blank spaces in the table, like the last two columns on the first row and the last column on rows 2,3 and 5.

mgovers: > I would organize it how you showed me offline, i.e. first all the pass in intended behavior, and then all the raises... However, I don't think this is too important as the table is concise and small enough. So I'd say don't worry too much about it.

:+1:

> On the other hand, there are some blank spaces in the table, like the last two columns on the first row and the last column on rows 2,3 and 5.

i mention that in the line before the table: `To reduce the verbosity, duplications have been removed.`. I can re-add them if you like

figueroa1395: I'd prefer them there, but if you have a strong opinion is okay to also leave them out.

mgovers:
image
oof, the rendering though...

docs/examples/Make Test Dataset.ipynb
mgovers: readded

tests/unit/utils.py
mgovers: decision was made to improve the warning message and otherwise keep the current behavior

mgovers:
> adding do-not-merge label cfr. [#1049 (comment)](https://github.com/PowerGridModel/power-grid-model/pull/1049#discussion_r2224999073)

resolved

figueroa1395 (approved):
LGTM

PR #1051, opened 2025-07-25
Validation cases: same behavior in C++ as Python for test cases with exceptions

### LINKED ISSUE #935: [FEATURE] Run C++ validation cases even if they are marked fail
Validation cases can be marked `fail`ing for a variety of reasons:

* The validation case is intended to throw an error, e.g.:
  * There is a known edge case that should not be marked
* The validation case should not fail but is currently expected to do so, e.g.:
  * The feature may not be implemented yet
  * The feature may be experimental
  * The feature may be expected to throw an error

There are 2 missing things in the current implementation:

* [The original C++ validation cases could not find the error type because it would require a lot of implementation effort to do so. #1048](https://github.com/PowerGridModel/power-grid-model/issues/1048)
  * However, after #688 and following, the validation cases can now at runtime match the error against the original message.
* [The current Python and C++ validation cases do not make any distinction between tests that are **intended** to fail and tests that currently are **expected** to fail. #1047](https://github.com/PowerGridModel/power-grid-model/issues/1047)
  * This raises confusion and also results in tests marked `xfail` even though they are actually correctly raising and should therefore pass.
  * This can be improved by making the distinction between `raises` for the former and `xfail` for the latter.
  * On both the Python side and the C++ side, the former should be marked `pass` and the latter as `xfail` on raising the exception.
  * Also note that the current Python exceptions cannot handle cases that pass but should actually thrown an error. This is also captured in #1047

This feature implements both of those improvements.

### LINKED ISSUE #1048: Handle C++ failing validation cases the same way as the Python way

mgovers:
Closes #1048
Closes #935

Checks to be done before merge:
* [x] Check all raises and xfail combinations in the table
* [x] Make decision on what to do with things like OSErrors or Python-only errors; some ideas:
  ~* [ ] add `"skip": "cpp"`~
  * [x] postpone for now
* [x] Make decision on design:
  * [x] current design is OK for now
  ~* [ ] go Lippincott pattern~


tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -145,20 +265,19 @@ template <typename T>
...
                                                                        possible_attribute_value, dynamic_atol, rtol)
             : compare_value<decltype(ref_attribute_value)>(ref_attribute_value, attribute_value, dynamic_atol, rtol);
     if (!match) {
-        std::stringstream case_sstr;
-        case_sstr << "dataset scenario: #" << scenario_idx << ", Component: " << component_name << " #" << obj
-                  << ", attribute: " << attribute_name
-                  << ": actual = " << get_as_string<decltype(attribute_value)>(attribute_value) + " vs. expected = "
-                  << get_as_string<decltype(ref_attribute_value)>(ref_attribute_value);
-        CHECK_MESSAGE(match, case_sstr.str());
+        auto const case_str = std::format(
```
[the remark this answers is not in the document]
mgovers: not necessary

Jerry-Jinfeng-Guo:
> Closes #1048 Closes #935
> 
> Checks to be done before merge:
> 
> * [x]  Check all raises and xfail combinations in the table
> * [ ]  Make decision on what to do with things like OSErrors or Python-only errors; some ideas:
>   
>   * [ ]  add `"skip": "cpp"`
>   * [ ]  postpone for now
> * [ ]  Make decision whether current design is OK for now or whether we want to go Lippincott pattern

From my side: 
* [x]  add `"skip": "cpp"`
* [x]  current design is OK for now

nitbharambe:
Whats a usecase of skip: cpp?

The duplication of regexes gets resolved in a later PR?

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -102,6 +99,128 @@ template <typename T> std::string get_as_string(T const& attribute_value) {
...
+    template <typename T>
+        requires std::invocable<std::remove_cvref_t<T>, Subcase&>
+    void execute_case(T&& statement) noexcept {
+        CHECK_NOTHROW(maybe_mark_xfail(maybe_with_raises(std::forward<T>(statement)))(*this));
+    }
+
+  private:
+    std::optional<std::string> raises_{};
+    std::optional<std::string> xfail_raises_{};
+    int fail_count{};
```
nitbharambe: Why do we count fails? why not just `is_failing`

mgovers: no deeper reason besides just making it easier to reason about when resetting the `fail_count`. if you want, i can change it

mgovers:
> Whats a usecase of skip: cpp?

as discussed offline: this is out of scope for now.

> The duplication of regexes gets resolved in a later PR?

yeah this would be an addition to the C API. I just briefly discussed with tony and we can also just extend the existing error codes (maybe using bitflags?) we would have to be very careful to be backwards compatible, though. Definitely out of scope for now.

tests/cpp_validation_tests/test_validation.cpp
mgovers: done

nitbharambe: Would be nice. It also becomes a bit confusing as one would search for scenario where the count reaches two or more.

PR #1054, opened 2025-07-28
CI: bump compilers

mgovers:
gcc-14 is available on Ubuntu 24.04: https://packages.ubuntu.com/noble/gcc-14
Even though we do not have to use new things, gcc-14 actually contained some backported performance improvements (that came from a bug filed to the C++26 standard but that was actually also retroactively applied to older C++ versions).

Given the following disclaimer in the build guide, we can bump the list it refers to without breaking any user workflows:

> Below is a list of tested compilers:

### DIFF
--- a/docs/advanced_documentation/build-guide.md
+++ b/docs/advanced_documentation/build-guide.md
@@ -59,7 +59,7 @@ You can define the environment variable `CXX` to for example `clang++` to specif
 
 * MSVC >= 19.0
   * Latest release tested in CI (e.g. Visual Studio 2022, IDE or build tools)
-* Clang CL >= 17.0
+* Clang CL >= 19.0
   * Latest release tested in CI (e.g. Visual Studio 2022, IDE or build tools)
 
 #### macOS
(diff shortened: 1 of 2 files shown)


docs/advanced_documentation/build-guide.md
```diff
@@ -45,26 +45,26 @@ Below is a list of tested compilers:
...
+* MSVC >= 19.0
   * Latest release tested in CI (e.g. Visual Studio 2022, IDE or build tools)
-* Clang CL >= 17.0
+* Clang CL >= 19.0
   * Latest release tested in CI (e.g. Visual Studio 2022, IDE or build tools)
 
 #### macOS
 
-* Clang >= 15.0
+* Clang >= 18.0
```
mgovers: we use the XCode (16.4 starting with this branch) not the Homebrew version

```suggestion
* Clang >= 17.0
```

docs/advanced_documentation/build-guide.md
```diff
@@ -45,26 +45,26 @@ Below is a list of tested compilers:
...
   * Latest release tested in CI (e.g. Visual Studio 2022, IDE or build tools)
-* Clang CL >= 17.0
+* Clang CL >= 19.0
   * Latest release tested in CI (e.g. Visual Studio 2022, IDE or build tools)
 
 #### macOS
 
-* Clang >= 15.0
+* Clang >= 17.0
   * Latest release tested in CI
```
mgovers:
```suggestion
  * Latest XCode release tested in CI
```

PR #1055, opened 2025-07-30
Performance: fix memory creep + false sharing in calculation info

mgovers:
This fixes a memory usage problem and a computational performance problem that has been in the code for a long time.

1. On current `main` (and long before that), the memory usage increases for every scenario. This is caused by the fact that the calculation info was tracked and stored for each scenario - something that is not necessary because we don't expose it to users anyways.
2. On current `main` (introduced over a year ago in 0115a90 ), usage of aligned memory was removed. This may have resulted in [false sharing](https://en.wikipedia.org/wiki/False_sharing) since, which comes with performance hits.

This PR moves from per-scenario tracking of the calculation info to per-thread tracking and re-introduces the alignment.

Unfortunately, we cannot remove the `#ifdef` feature test for the [hardware interference size](https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size.html) because Clang < 19 and AppleClang not supporting it yet at all, cfr. https://en.cppreference.com/w/cpp/compiler_support/17.

## Tests before merging:

* [x] Profile to make sure that the issue is resolved
  * [x] C++
  * [x] Python
* [x] Run internal benchmark and compare with current `main`


power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -11,6 +11,12 @@
...
 namespace power_grid_model {
 
-using CalculationInfo = std::map<std::string, double, std::less<>>;
+#if defined(__cpp_lib_hardware_interference_size)
+constexpr size_t cache_line_size = std::hardware_destructive_interference_size;
+#else
+constexpr size_t cache_line_size = 64;
+#endif
+
+class alignas(cache_line_size) CalculationInfo : public std::map<std::string, double, std::less<>> {};
```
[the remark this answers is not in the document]
mgovers: false positive

mgovers:
## C++ memory profiling

The following profiling was done using Visual Studio 17.14.9 on the first run of the internal benchmark.

### Main

image

### This branch

image

### Legend

| snapshot ID | breakpoint location                                    |
| :---------: | ------------------------------------------------------ |
| 1           | right before the first timer in PGM internal benchmark |
| 2           | at the end of the creation of the `sub_batch` job      |
| 3           | right before the `job_dispatch` call                   |
| 4           | right before the 1st call to `calculate_scenario`      |
| 5           | right after the 1st call to `calculate_scenario`       |
| 6           | right before the 2nd call to `calculate_scenario`      |
| 7           | right after the 2nd call to `calculate_scenario`       |
| 8           | right before the 3rd call to `calculate_scenario`      |
| 9           | right after the 3rd call to `calculate_scenario`       |
| 10          | right before the 4th call to `calculate_scenario`      |
| 11          | right after the 4th call to `calculate_scenario`       |

### Conclusion

The memory usage does not increase anymore after the first scenario is completely processed.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -11,6 +11,12 @@
 
 namespace power_grid_model {
 
-using CalculationInfo = std::map<std::string, double, std::less<>>;
+#if defined(__cpp_lib_hardware_interference_size)
```
figueroa1395: I believe you need to add `#include <new>` (https://en.cppreference.com/w/cpp/header/new.html) and that is why this is failing on linux.

mgovers: no it's caused by ABI compatibility (see my new commit)

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -93,52 +93,48 @@ template <class MainModel, class... ComponentType> class JobDispatch {
                 &result_data, &update_data, &all_scenarios_sequence_ = std::as_const(all_scenarios_sequence),
                 components_to_update, update_independence](Idx start, Idx stride, Idx n_scenarios) {
             assert(n_scenarios <= narrow_cast<Idx>(exceptions.size()));
-            assert(n_scenarios <= narrow_cast<Idx>(infos.size()));
+            assert(0 <= start && start < narrow_cast<Idx>(infos.size()));
 
-            Timer const t_total(infos[start], 0000, "Total in thread");
+            auto& thread_info = infos[start];
```
figueroa1395: Why this step? Or is it just for cleanup? Or does it affect thread "safeness"? Or is it something else?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -60,7 +60,7 @@ template <class MainModel, class... ComponentType> class JobDispatch {
 
         // error messages
         std::vector<std::string> exceptions(n_scenarios, "");
-        std::vector<CalculationInfo> infos(n_scenarios);
+        std::vector<CalculationInfo> infos(n_threads(n_scenarios, threading), CalculationInfo{});
```
figueroa1395: So this "solves" the per-scenario issue by just doing it per-thread (which is unavoidable with the current design), right?

figueroa1395 (commented):
Won't approve as some in-person discussion will happen tomorrow, but it looks good to me.

However, I do think we should address `CalculationInfo` not being thread safe right away as an immediate follow up.

Edit: I see clang tidy complaining about some padding issue with `MainModelImpl` now (https://github.com/PowerGridModel/power-grid-model/actions/runs/16617781791/job/47014484585?pr=1055) which was probably triggered by the new alignment of `ComponentInfo`.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -11,6 +11,12 @@
...
 namespace power_grid_model {
 
-using CalculationInfo = std::map<std::string, double, std::less<>>;
+// We would prefer to use std::hardware_destructive_interference_size, but as explained in
+// https://discourse.llvm.org/t/rfc-c-17-hardware-constructive-destructive-interference-size/48674/7
+// and https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Winterference-size, there is no
+// good way to determine the size of the cache line in a portable way at compile-time.
+constexpr size_t cache_line_size = 64;
+
+class alignas(cache_line_size) CalculationInfo : public std::map<std::string, double, std::less<>> {};
```
TonyXiang8787: If we use `mutex`, there is no need to align the `CalculationInfo` right? Each thread will just have a `CalculationInfo` locally.

TonyXiang8787: The forced alignment was needed if you put `CalculationInfo` into a `std::vector` and use each individual `info` by different thread. You want to make sure they are sufficiently away from each other.

In this new PR, we don't have that structure at all.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: it made it easier to reason about "this thread can only safely use this info" but since we now use the std::mutex approach, this is outdated. closing

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: cfr. offline discussion: yes but the new implementation uses `std::mutex` so this is no longer relevant

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
mgovers: yes this was a remnant that i just hadn't resolved yet (was going to, though). Fixed in eff593f

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/calculation_info.hpp
```diff
@@ -10,21 +10,16 @@
...
-            } else {
-                result[k] += v;
-            }
+inline CalculationInfo& merge_into(CalculationInfo& dest, CalculationInfo const& src) {
+    static auto const key = Timer::make_key(2226, "Max number of iterations");
+    for (auto const& [k, v] : src) {
+        if (k == key) {
+            dest[k] = std::max(dest[k], v);
+        } else {
+            dest[k] += v;
```
figueroa1395: Nitpick: `source` and `destination` instead of `src` and `dest`. Since this is stylistic, feel free to resolve if you prefer your way.

mgovers: done

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -106,7 +106,7 @@ template <class MainModel, class... ComponentType> class JobDispatch {
 
             CalculationInfo thread_info;
 
-            Timer const t_total(thread_info, 0000, "Total in thread");
```
figueroa1395: Why?

mgovers: I think this just moved locations slightly in the diff view. It's still there

mgovers: Ohhh sorry i see what you mean. I added a `t_total.stop()` below, which requires te tiner to be mutable. The stop was necessary, because the current thread is now responsible for reporting to the main thread. If i don't stop the timer before doing so, then the total time will not be reported to the main thread

figueroa1395: I meant, why is it no longer `const`?

mgovers:
Benchmark
```
Statistics: {'mean_delta': np.float64(0.019075246068759343), 'std_delta': np.float64(3.0313877773380136), 'median_delta': np.float64(-2.999999999999999e-07), '95th_percentile_delta': np.float64(0.14372659999999987)}
```
benchmark_comparison

PR #1061, opened 2025-08-04
Linting and Formatting: more ruff checks

Laurynas-Jagutis:
Adds W, TID, ARG, PTH, RET, SLF, PIE


pyproject.toml
```diff
@@ -126,6 +135,9 @@ combine-as-imports = true
 "docs/examples/*.ipynb" = ["F811", "E402"]
 # Pylint was only run in src directory before moving to Ruff
 "tests/*" = ["PLR0915", "PLR0912", "PLR0913"]
+# Ignore private member access: _model_ptr, _all_component_count...
+"src/power_grid_model/_core/*.py" = ["SLF001"]
+"tests/unit/test_error_handling.py" = ["SLF001"]
```
figueroa1395, line 140: `"tests/unit/test_error_handling.py" = ["SLF001"]` should be fine, as we are unit testing stuff, but should we do `"src/power_grid_model/_core/*.py" = ["SLF001"]`? Or are we better not having some members "private" from a "design" perspective? Given that it is already in `.../_core/...` I would assume the second, but maybe you can comment a bit about it?

Laurynas-Jagutis: I thought that we would want to keep them private, so that no user would make a mistake of using or editing those variables.

Laurynas-Jagutis: I see your point about it being in _core, which in general could mean "Do not modify", but I would keep it as is

figueroa1395: I am not sure, that is why I ask. Since they live under `..._core...` (note the `_`), no user should be messing with those anyways. Perhaps it is fine to keep them private, but maybe not. Just raising the question as I found it interesting.

figueroa1395: > I see your point about it being in _core, which in general could mean "Do not modify", but I would keep it as is

Sorry, I didn't see this reply before my previous comment. If you feel this is the best approach, that is alright with me. Just pointing it out :). Feel free to resolve.

PR #1065, opened 2025-08-05
Linting and Formatting: RUF, UP and RSE checks

Laurynas-Jagutis:
adds RUF, UP and RSE checks and fixes


src/power_grid_model/_core/data_handling.py
```diff
@@ -38,7 +38,7 @@ def get_output_type(*, calculation_type: CalculationType, symmetric: bool) -> Da
     if calculation_type == CalculationType.short_circuit:
         return DatasetType.sc_output
 
-    raise NotImplementedError()
+    raise NotImplementedError
```
mgovers, line 41: 😮

Laurynas-Jagutis: yep, this was interesting, the rule can be found here: https://docs.astral.sh/ruff/rules/unnecessary-paren-on-raise-exception/#unnecessary-paren-on-raise-exception-rse102

PR #1066, opened 2025-08-05
Clean-up main model: migrate job dispatch out of main model impl

mgovers:
Finally 🎉 

This was the intention behind the splitup in #1018 and #1036

Part 3 of the split-up of `MainModelImpl` into the following design: draw the arrow between `main_model` and `job_dispatch_adapter`
Untitled


power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -61,9 +63,25 @@ class MainModel {
...
+        = 0 parallel, use number of hardware threads
+        > 0 specify number of parallel threads
+    raise a BatchCalculationError if any of the calculations in the batch raised an exception
+    */
     BatchParameter calculate(Options const& options, MutableDataset const& result_data,
                              ConstDataset const& update_data) {
-        return impl().calculate(options, result_data, update_data);
+        JobDispatchAdapter<Impl, AllComponents> adapter{std::ref(impl())};
+        return JobDispatch::batch_calculation(adapter, Impl::calculator(options), result_data, update_data,
+                                              options.threading);
```
mgovers, line 84: alternative implementation: move the adapter creation back to `MainModelImpl` so that this becomes

```suggestion
        return JobDispatch::batch_calculation(impl().get_adapter(), Impl::calculator(options), result_data, update_data,
                                              options.threading);
```

input is more than welcome

figueroa1395: I think this is fine as is. If we are to get rid of `main_model_impl.hpp` at some point, we might as well start doing so.

figueroa1395:
Since it is in draft I won't "approve", but I like the current state. It does what it should, simplifies logic and removes one layer in `main_model_impl`. However, I would only merge until https://github.com/PowerGridModel/power-grid-model/pull/1036 is merged, as I don't want to make it more complex than it is already.

mgovers:
> Since it is in draft I won't "approve", but I like the current state. It does what it should, simplifies logic and removes one layer in `main_model_impl`. However, I would only merge until #1036 is merged, as I don't want to make it more complex than it is already.

yes that's indeed why it's still in draft 😛

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -511,36 +509,14 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-
-    Run the calculation function in batch on the provided update data.
-
-    The calculation function should be able to run standalone.
-    It should output to the provided result_data if the trailing argument is not ignore_output.
+    static auto calculator(Options options) {
+        return [options](MainModelImpl& model, MutableDataset const& target_data, Idx pos) {
+            auto sub_opt = options; // copy
+            sub_opt.err_tol = pos != ignore_output ? options.err_tol : std::numeric_limits<double>::max();
+            sub_opt.max_iter = pos != ignore_output ? options.max_iter : 1;
```
figueroa1395: Saw this while working on something else:
```suggestion
    static auto calculator(Options const& options) {
        return [&options](MainModelImpl& model, MutableDataset const& target_data, Idx pos) {
            Options sub_opt = options; // copy
            sub_opt.err_tol = pos != ignore_output ? options.err_tol : std::numeric_limits<double>::max();
            sub_opt.max_iter = pos != ignore_output ? options.max_iter : 1;
```

mgovers: i actually want to capture by value here to prevent dangling references. but if you want to, we can maybe do
```suggestion
    static auto calculator(Options const& options) {
        return [options](MainModelImpl& model, MutableDataset const& target_data, Idx pos) {
            auto sub_opt = options; // copy
            sub_opt.err_tol = pos != ignore_output ? options.err_tol : std::numeric_limits<double>::max();
            sub_opt.max_iter = pos != ignore_output ? options.max_iter : 1;
```

figueroa1395: I actually think we may have to leave as is, wouldn't it be potentially dangerous to pass by `const&` when multithreaded? As this is called in `single_thread_job`...

`Options` should also be trivially copy-able, right?

mgovers: it's indeed trivially copyable (that's why clang-tidy didn't like my `std::move(options)`). let's leave it as-is

mgovers:
do-not-merge label added because it depends on #1036

mgovers:
> do-not-merge label added because it depends on #1036

#1036 was merged; do-not-merge label removed

figueroa1395 (approved):

Issue #1063, opened 2025-08-05
[IMPROVEMENT] *Clean-up main model: Job dispatch*


figueroa1395:
**Problem description**

The main model has grown organically large and complex, resulting in very complicated code that is tightly coupled and monolithic, hence very difficult to understand, test, change and extend.

This issue is concerned specifically with cleaning up, refactoring, improving, and decoupling the job dispatch logic from the main model, one of the "components" that make it up.

**How to approach the problem**

We have made a four step process as a guide/skeleton on how to tackle the clean-up of each main model "component"; here we lay it down for the job dispatch logic:

- [x] Separate the job dispatch related logic into a separate file/module:#1018
- [ ] Clean up, refactor, decouple and improve the job dispatch logic. This is to be done in as many separate PRs as needed, with a quick feedback loop in mind, and with the design choices up to the developers working on this:
    - [ ] Decouple main model from the job dispatch logic: https://github.com/PowerGridModel/power-grid-model/pull/1036.
    - [ ] Decouple the calculation functions from the job dispatch logic (https://github.com/PowerGridModel/power-grid-model/blob/1769f1ca1254d2969048b3e5dc78cacc0e3e1de2/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp#L22).
    - [ ] Add/leverage concepts and generic template parameters in the main job dispatch file, its interface, and its adaptor (https://github.com/PowerGridModel/power-grid-model/blob/1769f1ca1254d2969048b3e5dc78cacc0e3e1de2/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp#L23, https://github.com/PowerGridModel/power-grid-model/blob/1769f1ca1254d2969048b3e5dc78cacc0e3e1de2/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp#L24).
    - [ ] Evaluate if pre-calculating and sharing per thread (current), or working only per thread is best (or something in between) - (https://github.com/PowerGridModel/power-grid-model/pull/1036/files#r2250673073).
    - [ ] Can `get_calculation_info()` be done by view? Or is by value the best we can do? (https://github.com/PowerGridModel/power-grid-model/pull/1036/files#r2225279545)
[...]

PR #1071, opened 2025-08-06
Clean-up main model: make job dispatch calculation agnostic

figueroa1395:
Relates to https://github.com/PowerGridModel/power-grid-model/issues/1063.

**Short description**
This PR makes the job dispatch logic agnostic on how to perform calculations. This is now directly handled by the adapter. Some additional small things like the better use of relevant concepts is also addressed here.

**Why?**
The job dispatch logic should not know about calculation details, this induces tight coupling. In addition, this lead to unnecessary layers and back-and-forth calls. These made the job dispatch interface logic more complex and forced us to do a lot of perfect forwarding for the `in-between` layers (lambdas). 

**How?**
The `calculator` function now lives in `main_model` directly (one single entry point), which is in brief, an orchestrator for the different calculation methods which only depends on the `Options`. Furthermore, reference wrapping (minimal additional overhead) the `Options` in the `JobDispatchAdapter`, allows us to directly call the `calculator` in the adapter, removing perfect forwarding and intermediate lambdas.

**Extra**
This comes as a follow up to https://github.com/PowerGridModel/power-grid-model/pull/1036. Specifically the following points in #1063:
- Decouple the calculation functions from the job dispatch logic (https://github.com/PowerGridModel/power-grid-model/blob/1769f1ca1254d2969048b3e5dc78cacc0e3e1de2/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp#L22).
- Add/leverage concepts and generic template parameters in the main job dispatch file, its interface, and its adaptor (https://github.com/PowerGridModel/power-grid-model/blob/1769f1ca1254d2969048b3e5dc78cacc0e3e1de2/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp#L23, https://github.com/PowerGridModel/power-grid-model/blob/1769f1ca1254d2969048b3e5dc78cacc0e3e1de2/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp#L24).
- Some general cleanup.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
@@ -21,10 +21,12 @@ template <class MainModel, class... ComponentType>
 class JobDispatchAdapter<MainModel, ComponentList<ComponentType...>>
     : public JobDispatchInterface<JobDispatchAdapter<MainModel, ComponentList<ComponentType...>>> {
   public:
-    JobDispatchAdapter(MainModel& model, std::reference_wrapper<MainModelOptions const> options)
-        : model_{std::make_unique<MainModel>(model)}, options_{options} {}
+    JobDispatchAdapter(std::reference_wrapper<MainModel> model_reference,
+                       std::reference_wrapper<MainModelOptions const> options)
+        : model_reference_{model_reference}, options_{options} {}
     JobDispatchAdapter(JobDispatchAdapter const& other)
-        : model_{std::make_unique<MainModel>(*other.model_)},
+        : model_copy_{std::make_unique<MainModel>(other.model_reference_.get())},
+          model_reference_{std::ref(*model_copy_)},
           options_{std::ref(other.options_)},
           components_to_update_{other.components_to_update_},
           update_independence_{other.update_independence_},
@@ -32,7 +34,8 @@ class JobDispatchAdapter<MainModel, ComponentList<ComponentType...>>
           all_scenarios_sequence_{other.all_scenarios_sequence_} {}
     JobDispatchAdapter& operator=(JobDispatchAdapter const& other) {
         if (this != &other) {
-            model_ = std::make_unique<MainModel>(*other.model_);
+            model_copy_ = std::make_unique<MainModel>(other.model_reference_.get());
+            model_reference_ = std::ref(*model_copy_);
             options_ = std::ref(other.options_);
             components_to_update_ = other.components_to_update_;
             update_independence_ = other.update_independence_;
@@ -42,7 +45,8 @@ class JobDispatchAdapter<MainModel, ComponentList<ComponentType...>>
         return *this;
     }
     JobDispatchAdapter(JobDispatchAdapter&& other) noexcept
-        : model_{std::move(other.model_)},
+        : model_copy_{std::move(other.model_copy_)},
+          model_reference_{model_copy_ ? std::ref(*model_copy_) : std::move(other.model_reference_)},
           options_{other.options_},
           components_to_update_{std::move(other.components_to_update_)},
           update_independence_{std::move(other.update_independence_)},
@@ -50,7 +54,8 @@ class JobDispatchAdapter<MainModel, ComponentList<ComponentType...>>
           all_scenarios_sequence_{std::move(other.all_scenarios_sequence_)} {}
     JobDispatchAdapter& operator=(JobDispatchAdapter&& other) noexcept {
         if (this != &other) {
-            model_ = std::move(other.model_);
+            model_copy_ = std::move(other.model_copy_);
+            model_reference_ = model_copy_ ? std::ref(*model_copy_) : std::move(other.model_reference_);
             options_ = other.options_;
             components_to_update_ = std::move(other.components_to_update_);
             update_independence_ = std::move(other.update_independence_);
@@ -59,7 +64,7 @@ class JobDispatchAdapter<MainModel, ComponentList<ComponentType...>>
         }
         return *this;
     }
-    ~JobDispatchAdapter() { model_.reset(); }
+    ~JobDispatchAdapter() { model_copy_.reset(); }
 
   private:
     // Grant the CRTP base (JobDispatchInterface<JobDispatchAdapter>) access to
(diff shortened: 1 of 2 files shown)


figueroa1395:
[c1b2d5a](https://github.com/PowerGridModel/power-grid-model/pull/1071/commits/c1b2d5aa3a635f9978b9c2a67ba0685531bd8e1c) was introduced after extensive discussion with @nitbharambe. Even though it introduces the overhead of one (in total) additional copy per calculation run, it cleans up the adapter nicely. The overhead should not be significant compared to the amount of scenarios involved in a batch calculation. Nevertheless, worth pointing out.

nitbharambe:
> [c1b2d5a](https://github.com/PowerGridModel/power-grid-model/pull/1071/commits/c1b2d5aa3a635f9978b9c2a67ba0685531bd8e1c) was introduced after extensive discussion with @nitbharambe. Even though it introduces the overhead of one (in total) additional copy per calculation run, it cleans up the adapter nicely. The overhead should not be significant compared to the amount of scenarios involved in a batch calculation. Nevertheless, worth pointing out.

Correctly said. We also benchmark and showcase a single calculation. So adding an extra note to look at the single calculation benchmark and a team decision ofcourse if it is a lot.

figueroa1395:
I did some benchmarking on MacOs. Here are the results:

**Memory**
pr1071_plot
This looks fine if compared to https://github.com/PowerGridModel/power-grid-model/pull/1036#issuecomment-3153799733.

**Speed**
1 12vs1071
We have a problem, somewhere along the way of removing the calculation functions we just don't time things anymore... Even though the final result seems okay, this needs fixing.

Jerry-Jinfeng-Guo:
> **How?** The `calculator` function now leaves in `main_model` directly (one single entry point)...
> 

Do you mean to say 'lives in `main_model`'?

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
```diff
@@ -110,33 +106,33 @@ class JobDispatchAdapter<MainModel, ComponentList<ComponentType...>>
     void prepare_job_dispatch_impl(ConstDataset const& update_data) {
         // cache component update order where possible.
         // the order for a cacheable (independent) component by definition is the same across all scenarios
-        components_to_update_ = model_.get().get_components_to_update(update_data);
-        update_independence_ = main_core::update::independence::check_update_independence<ComponentType...>(
-            model_.get().state(), update_data);
+        components_to_update_ = model_->get_components_to_update(update_data);
+        update_independence_ =
+            main_core::update::independence::check_update_independence<ComponentType...>(model_->state(), update_data);
```
Jerry-Jinfeng-Guo: Not directly related, but now that we look back, do we actually need an extra namespace `independence` in the `update.hpp`? https://github.com/PowerGridModel/power-grid-model/blob/ada4efc70ea0946dcdadb586013d53d50f51f638/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp#L46

figueroa1395:
> > **How?** The `calculator` function now leaves in `main_model` directly (one single entry point)...
> 
> Do you mean to say 'lives in `main_model`'?

Indeed! Good catch. I edited the description accordingly :)

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
figueroa1395: Let me check it out and I will get back to you :)

figueroa1395: I took a look but I don't see why not. Do you think it takes more than what it adds? What's the problem you now see here?

Jerry-Jinfeng-Guo: It reads (also types) quite long and it is actually the only real additional `namespace` within `update`. I personally think the content within this `namespace`: `UpdateCompProperties`, `process_buffer_span`, `check_component_independence`, `validate_update_data_independence` and `check_update_independence` are perfectly fine living directly inside `update`.

figueroa1395:
> [c1b2d5a](https://github.com/PowerGridModel/power-grid-model/pull/1071/commits/c1b2d5aa3a635f9978b9c2a67ba0685531bd8e1c) was introduced after extensive discussion with @nitbharambe. Even though it introduces the overhead of one (in total) additional copy per calculation run, it cleans up the adapter nicely. The overhead should not be significant compared to the amount of scenarios involved in a batch calculation. Nevertheless, worth pointing out.

I realize this is a problem now... if we do the additional copy now at the `MainModel` level (introduced with [c1b2d5a](https://github.com/PowerGridModel/power-grid-model/pull/1071/commits/c1b2d5aa3a635f9978b9c2a67ba0685531bd8e1c)), when we pass the copy to the `job_dispatch`, we are tracking the benchmark at a copy so we never "merge back" the metrics to the original model.

This has a couple of potential ways to go:
1. Simplest and easiest is to go back to the original implementation in the adapter ("revert" [c1b2d5a](https://github.com/PowerGridModel/power-grid-model/pull/1071/commits/c1b2d5aa3a635f9978b9c2a67ba0685531bd8e1c)) and just deal with a reference to the model and a copy. But perhaps better naming would be needed or at least some commenting. Then the suggested clean-up patter can be re done once we have refactored `CalculationInfo` or made a proper logger. - this is my preferred solution.
2. Figure out a way to merge back to the original `CalculationInfo`... The problem is that this will either couple again job dispatch (which defeats the whole purpose of this) or make it even easier (unless someone thinks of something very clever). - I don't like this.
3. We delay this PR until `CalculationInfo` is handled correctly. - I hate this.

I say we go with 1. @nitbharambe thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
figueroa1395: To be honest, the `*_independence` naming at the end of the calls comes a long way. If this is more polluting than anything we can just remove that `namespace`, I have no issue. Since it is a very tiny change, we can already do it here.

mgovers: while i do think that 3 layers of namespacing is not needed, so this may not be the right moment to give this comment, but, in general i'm actually a fan of adding more namespaces rather than less. 

Note: you can always add shorthands `using main_core::update::independence::check_update_indendence;` at the top of the function, or (but this only works when namespaces are used more consistently throughout the code) `namespace independence = main_core::update::independence;`

nitbharambe:
@figueroa1395 Agreed, lets go with 1.

figueroa1395:
After [8287cb8](https://github.com/PowerGridModel/power-grid-model/pull/1071/commits/8287cb887f419d0ca6483ff36f09ab1f0be0475b), the speed benchmark works fine. Here are the results on Mac:
Figure_2
The results look fine: we are ready to merge :)

PR #1073, opened 2025-08-07
Testing: split cpp unit tests

mgovers:
Splits the unit tests into individual targets. This has 2 benefits:

1. it is much faster to develop things deep in the core. We won't have to wait - for instance - to build the entire main model when you only want to make a simple unit test in `statistics.hpp` pass. Instead, set the build target to the respective CMake unit test target, build and iterate.
2. it is much clearer which files are tested and which ones are not. This will improve our confidence in the code-base.

This is an enabler for testability of the job dispatch logic

### DIFF
--- a/tests/cpp_unit_tests/CMakeLists.txt
+++ b/tests/cpp_unit_tests/CMakeLists.txt
@@ -37,7 +37,6 @@ target_link_libraries(power_grid_model_unit_tests_auxiliary
     PRIVATE
         power_grid_model
         doctest::doctest
-        nlohmann_json nlohmann_json::nlohmann_json
 )
 
 doctest_discover_tests(power_grid_model_unit_tests_auxiliary)


tests/cpp_unit_tests/CMakeLists.txt
mgovers, line 1: I can easily also change the file structure of the unit tests to follow the same structure (either in this PR or a follow-up). What do you think?

mgovers: quick discussion with tony: future improvement.

tests/cpp_unit_tests/CMakeLists.txt
```diff
@@ -2,18 +2,49 @@
...
+    "test_deserializer.cpp"
+    "test_serializer.cpp"
+)
+
+target_link_libraries(power_grid_model_unit_tests_auxiliary
+    PRIVATE
+        power_grid_model
+        doctest::doctest
+        nlohmann_json nlohmann_json::nlohmann_json
+)
```
figueroa1395: don't we also need to add `msgpack-cxx` here?

tests/cpp_unit_tests/CMakeLists.txt
```diff
@@ -23,44 +54,83 @@ set(PROJECT_SOURCES
...
+
+target_link_libraries(power_grid_model_unit_tests_component
+    PRIVATE
+        power_grid_model
+        doctest::doctest
+)
+
+doctest_discover_tests(power_grid_model_unit_tests_component)
+
+add_executable(power_grid_model_unit_tests_main_core
```
figueroa1395, line 72: Here is where all the new unit tests after splitting main model would go, right?

tests/cpp_unit_tests/CMakeLists.txt
```diff
@@ -23,44 +54,83 @@ set(PROJECT_SOURCES
...
+
+target_link_libraries(power_grid_model_unit_tests_optimizer
+    PRIVATE
+        power_grid_model
+        doctest::doctest
+)
+
+doctest_discover_tests(power_grid_model_unit_tests_optimizer)
+
+add_executable(power_grid_model_unit_tests
```
figueroa1395, line 121: Shall we just place this under main core tests? If not, let's come up with a better name.

figueroa1395 (dismissed):
LGTM, also tested locally and it works.

tests/cpp_unit_tests/CMakeLists.txt
mgovers: it's the other way around: `power_grid_model` interface library already contains `msgpack-cxx` and `nlohmann_json::nlohmann_json` as dependencies, so we can actually remove this

mgovers: `nlohamnn_json` IS required in the CPP validation tests, though... probably that's where this copy-paste mistake came from. Good find!

tests/cpp_unit_tests/CMakeLists.txt
mgovers: i don't want to put it under `main_core` cause it's supposed to contain only 'core' stuff, but i have been pondering about this as well... any input is welcome

tests/cpp_unit_tests/CMakeLists.txt
mgovers: i would guess so, yes

tests/cpp_unit_tests/CMakeLists.txt
figueroa1395: If I think of something I'll let you know. Can't think of anything clever atm.

mgovers: alternatively, we should move the current "base-level" stuff like job dispatch, main model, etc. into a separate directory as well

figueroa1395: Maybe it is good as is (for now) and we just have to rename this, and as we keep on dis-assembling main model, new targets can be created along the way. Things will make more sense then.

PR #1074, opened 2025-08-08
Testing: split cpp validation tests into targets per calculation type

mgovers:
Split the C++ validation cases into multiple test cases by calculation type (not only by whether it's a single calculation or a batch calculation).

This will allow faster development when only looking to focus on a specific test case.

As requested by @Jerry-Jinfeng-Guo


Jerry-Jinfeng-Guo:
Ready to merge next non-Friday work day.

mgovers:
> Ready to merge next non-Friday work day.

happy Monday

PR #1081, opened 2025-08-12
Data validator: relax requirement that p_sigma and q_sigma cannot mix normal and infinite values

mgovers:
The PGM documentation specifies the following regarding `p_sigma` and `q_sigma`:

> 1. If both p_sigma and q_sigma are provided, they represent the standard deviation of the active and reactive power, respectively, and the value of power_sigma is ignored. Any infinite component invalidates the entire measurement.
> 2. If neither p_sigma nor q_sigma are provided, power_sigma represents the standard deviation of the apparent power.
> 3. Providing only one of p_sigma and q_sigma results in undefined behaviour.

The underlying consequence is that:

> If any component of p_sigma and q_sigma is infinite, then they all shall be infinite.

There are a couple problems with this:

1. The phrasing is slightly misleading, because there is ambiguity between whether `invalidates the entire sensor` means `is not supported` or that it `disables the entire sensor`.
2. Above all, how the PGM actually handles infinite sigmas should be an implementation detail. The PGM core already handles this edge case correctly cfr. https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/measured_values.hpp#L380-L381 .
3. Whether to specify only 1 or all components as infinite is business logic from a user perspective - not a PGM requirement.

This requirement is therefore nonsensical and should not live in the generic PGM data validator. If anything, it should live in some kind of strict business-logic-aware use-case specific data validator.

This PR therefore relaxes the statement and updates the documentation to explain the intention it more clearly.

### DIFF
--- a/docs/user_manual/components.md
+++ b/docs/user_manual/components.md
@@ -885,11 +885,12 @@ Valid combinations of `power_sigma`, `p_sigma` and `q_sigma` are:
 |               |           |           | &#10060; |
 
 ```{note}
-1. If both `p_sigma` and `q_sigma` are provided, they represent the standard deviation of the active and reactive power,
-respectively, and the value of `power_sigma` is ignored. Any infinite component invalidates the entire measurement.
+1. If both `p_sigma` and `q_sigma` are provided (i.e., not NaN), they represent the standard deviation of the active and
+reactive power, respectively, and the value of `power_sigma` is ignored. Any infinite component disables the entire
+measurement.
 
 2. If neither `p_sigma` nor `q_sigma` are provided, `power_sigma` represents the standard deviation of the apparent
-power.
+power. If it is infinite, it disables the entire measurement.
 
 3. Providing only one of `p_sigma` and `q_sigma` results in undefined behaviour.
 ```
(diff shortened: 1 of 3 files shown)


docs/user_manual/components.md
```diff
@@ -885,11 +885,12 @@ Valid combinations of `power_sigma`, `p_sigma` and `q_sigma` are:
...
 ```{note}
-1. If both `p_sigma` and `q_sigma` are provided, they represent the standard deviation of the active and reactive power,
-respectively, and the value of `power_sigma` is ignored. Any infinite component invalidates the entire measurement.
+1. If both `p_sigma` and `q_sigma` are provided (i.e., not NaN), they represent the standard deviation of the active and
+reactive power, respectively, and the value of `power_sigma` is ignored. Any infinite component disables the entire
+measurement.
 
 2. If neither `p_sigma` nor `q_sigma` are provided, `power_sigma` represents the standard deviation of the apparent
-power.
+power. If it is infinite, it disables the entire measurement.
```
Jerry-Jinfeng-Guo:
```suggestion
power. In this case, infinite value of `power_sigma` disables the entire measurement.
```

tests/unit/validation/test_validation_functions.py
Jerry-Jinfeng-Guo, line 1: Should we mix `nan` and `inf` in data?

mgovers: we can but i don't think it will matter much. we kind of already validate that inf is handled as regular data while nan is not. Easy to add, though. Your call

Jerry-Jinfeng-Guo (approved):

PR #1011, opened 2025-06-05
Transformer: support periodic clock input

Jerry-Jinfeng-Guo:
Closes https://github.com/PowerGridModel/power-grid-model/issues/987

- [x] Core
- [x] Unit test
- [x] Python validation
- [x] Doc

Issue #987, opened 2025-05-07
[FEATURE] PGM Meet-up 2025-05-21 hackathon: attribute modification in C++ core


Jerry-Jinfeng-Guo:
# Attribute modification
This topic aims to familiarize contributors with development inside the C++ core. To make the initial step easier and the feedback loop shorter, we start with modification to existing attributes. We walk you through a first example of extending the attribute handling of attribute `clock_` in class `Transformer`.

## Background
the term "clock" refers to the phase displacement between the primary and secondary windings of a three-phase transformer. This phase displacement is crucial for understanding how the transformer will interact with other components in the power system.

### Clock Notation in Transformers
The clock notation is a method used to describe the phase displacement. Imagine a clock face where the hour hand represents the high voltage (HV) winding and the minute hand represents the low voltage (LV) winding. The position of the minute hand relative to the hour hand indicates the phase displacement.

#### Values Between 0 and 12
The values between 0 and 12 correspond to the positions on a clock face, where each hour represents a 30-degree phase shift, which is the amount of angle difference between the secondary side and the primary side of the transformer. For example:

- 0: No phase displacement (0 degrees)
- 1: 30 degrees phase displacement
- 2: 60 degrees phase displacement
- 3: 90 degrees phase displacement
- 6: 180 degrees phase displacement
- 12: 360 degrees phase displacement (which is effectively the same as 0 degrees)

This notation helps in standardizing the description of transformers and ensures compatibility and proper functioning within the electrical grid.

#### Practical Example
If a transformer has a clock notation of 1, it means that the LV winding leads the HV winding by 30 degrees. If the clock notation is 11, the LV winding lags the HV winding by 30 degrees.

This system is widely used because it provides a simple and intuitive way to understand and communicate the phase relationships in three-phase transformers, which is essential for the design and operation of electrical power systems.

## Goal
It is readily clear that the clock attribute is a cyclic variable. Meaning that values outside range $\[0, 12\]$ can be remapped back to this range to maintain a continuous domain. Mathematically, the cyclic clock function $\( \phi \)$ can be defined as follows:

$\[ \phi(h) = 30h \]$

where:
- $\( \phi(h) \)$ is the phase displacement in degrees.
- $\( h \)$ is the clock hour, which takes values from 0 to 12

The function is cyclic because after 12 o'clock, the phase displacement resets to 0 degrees. This cyclic behavior can be represented using modulo arithmetic:

[...]

PR #1083, opened 2025-08-13
Update (three_winding_transformer["clock_13"] = [5]) Transformer Examples.ipynb

WhiteSymmetry:
**Fixes issue**: `# three_winding_transformer["clock_13"] = [5]`

ValidationException: There is a validation error in input_data: Field 'clock_13' is not between (or at) 0 and 12 for 1 three_winding_transformer.

three_winding_transformer["clock_13"] = [-7] # supports periodic input
-->
three_winding_transformer["clock_13"] = [5] # supports periodic input

The "clock" parameter in a transformer's vector group represents the phase shift between the windings. This is specified using a number from 0 to 11, analogous to the hours on a clock face [1]. Each integer step corresponds to a 30-degree phase shift [2]. 

For example, a clock number of 1 signifies a -30 degree phase shift, and 11 signifies a +30 degree (or -330 degree) shift [3].

The value -7 is outside the valid 0-11 range required by the validation function. While comment suggests you expected periodic input to be handled (e.g., wrapping -7 to a valid number), the validation function requires the value to be explicitly within the standard range.

To represent a phase shift equivalent to "-7 hours" on a clock, you can use modular arithmetic. On a 12-hour clock, -7 is the same as +5 (12 - 7 = 5). Therefore, the correct value for clock_13 is 5.

1. https://electrical-engineering-portal.com/understanding-vector-group-transformer-1
2. https://www.quora.com/Why-can-the-phase-shifts-of-transformers-be-represented-by-the-clock-number-in-vector-groups
3. https://metapowersolutions.com/what-is-a-transformer-vector-group/

### DIFF
--- a/docs/examples/Transformer Examples.ipynb
+++ b/docs/examples/Transformer Examples.ipynb
@@ -386,7 +386,7 @@
     "three_winding_transformer[\"winding_2\"] = [1]\n",
     "three_winding_transformer[\"winding_3\"] = [1]\n",
     "three_winding_transformer[\"clock_12\"] = [5]\n",
-    "three_winding_transformer[\"clock_13\"] = [-7]  # supports periodic input\n",
+    "three_winding_transformer[\"clock_13\"] = [5]  # supports periodic input\n",
     "three_winding_transformer[\"tap_side\"] = [0]\n",
     "three_winding_transformer[\"tap_pos\"] = [0]\n",
     "three_winding_transformer[\"tap_min\"] = [-10]\n",


docs/examples/Transformer Examples.ipynb
```diff
@@ -386,7 +386,7 @@
     "three_winding_transformer[\"winding_2\"] = [1]\n",
     "three_winding_transformer[\"winding_3\"] = [1]\n",
     "three_winding_transformer[\"clock_12\"] = [5]\n",
-    "three_winding_transformer[\"clock_13\"] = [-7]  # supports periodic input\n",
+    "three_winding_transformer[\"clock_13\"] = [5]  # supports periodic input\n",
```
mgovers, line 389: This is a recently added feature (#987 and #1011), so the notebook itself is correct.

However, you are completely right in that it should not throw the validation error. You have found a bug in our data validator.

Would you like to fix that or would you prefer us to do so?

c.c. @Jerry-Jinfeng-Guo

mgovers (changes requested):
Hi @WhiteSymmetry ,

Thank you for your contribution. We are very happy with you taking the time to do so.

You have actually found a bug in the data validator (as mentioned below). Since we would like to give you all the credits you deserve for finding it, would you also like to do the fix? If not, feel free to open a bug report instead and we will fix it. Since we like to fix bugs as soon as possible, if we do not get an answer in the coming couple hours, we will take it upon ourselves to resolve it.

HINT:

* validation
  * This is correct (2-winding transformer): https://github.com/PowerGridModel/power-grid-model/blob/main/src/power_grid_model/validation/_validation.py#L610
  * [x] This is not (3-winding transformer): https://github.com/PowerGridModel/power-grid-model/blob/main/src/power_grid_model/validation/_validation.py#L675-L676
* tests
  * Here, negative clock numbers are added for 2-winding transformers: https://github.com/PowerGridModel/power-grid-model/blob/main/tests/unit/validation/test_input_validation.py#L137
  * Here, negative clock numbers are added for 3-winding transformers: https://github.com/PowerGridModel/power-grid-model/blob/main/tests/unit/validation/test_input_validation.py#L174-L175
  * [ ] Here, the checks for the 3w transformer clock attributes `clock_12` and `clock_13` are missing: https://github.com/PowerGridModel/power-grid-model/blob/main/tests/unit/validation/test_input_validation.py#L499
  * [x] It would also be good to revisit and add more comprehensive tests elsewhere (e.g., that -12 is supported, but -13 is not, conform the documentation).

docs/examples/Transformer Examples.ipynb
WhiteSymmetry: I’d prefer someone with expertise in this software and a good understanding of these tests to handle it for accuracy.

mgovers: OK, thank you for replying. @Jerry-Jinfeng-Guo will take it on. We will make sure to give you credits by leaving your commit in the commit history.

mgovers:
Closing in favor of #1084

PR #1087, opened 2025-08-15
Feature/timer using codes

mgovers:
* [ ] Directly pass only the code to `Timer` instead of code + name
* [ ] Homogenize some of the duplicated codes (to be followed up)
* [ ] Use the code to store the `CalculationInfo` instead of a string key constructed from code + name
* [ ] Format the `CalculationInfo` when actually producing the output

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
@@ -14,6 +14,8 @@ namespace common::logging {
 enum class LoggingTag : Idx {
     unknown = -1,
     log = 0,
+    total = 100000,       // TODO(mgovers): find other error code?
+    build_model = 101000, // TODO(mgovers): find other error code?
     timer = 1,
     total_single_calculation_in_thread = 0100,
     total_batch_calculation_in_thread = 0200,
@@ -23,11 +25,24 @@ enum class LoggingTag : Idx {
     scenario_exception = 1300,
     recover_from_bad = 1400,
     prepare = 2100,
+    create_math_solver = 2210,
     math_calculation = 2200,
-    iterative_pf_solver_max_num_iter = 2226,
-    ilse_max_num_iter = 2228,
-    nrse_max_num_iter = 2229, // used to be duplicate number
+    math_solver = 2220,
+    initialize_calculation = 2221,
+    preprocess_measured_value = 102221, // TODO(mgovers): find other error code + make plural?
+    prepare_matrix = 2222,
+    prepare_matrix_including_prefactorization = 102222, // TODO(mgovers): find other error code
+    prepare_matrices = 1002222,                         // TODO(mgovers): find other error code
+    initialize_voltages = 2223,
+    calculate_rhs = 2224,
+    prepare_lhs_rhs = 102224, // TODO(mgovers): find other error code
+    solve_sparse_linear_equation = 2225,
+    solve_sparse_linear_equation_prefactorized = 102225, // TODO(mgovers): find other error code
+    iterate_unknown = 2226,
+    calculate_math_result = 2227,
     produce_output = 3000,
+    iterative_pf_solver_max_num_iter = 1002226, // TODO(mgovers): find other error code
+    max_num_iter = 1002228,                     // TODO(mgovers): find other error code
 };
 
 constexpr auto to_string(LoggingTag tag) {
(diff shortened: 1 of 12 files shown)


mgovers:
benchmark results v.s. `v1.12.9`. Deviations are within system standard deviations

msvc_release_v1_12_9_vs_feature_timer-using-codes

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
mgovers, line 1: a lot of these TODO's require some additional input from the rest of the team: what to do with it?

also OK with making that a follow-up PR

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
```diff
@@ -0,0 +1,122 @@
...
+#pragma once
+
+#include "common.hpp"
+
+#include <string_view>
+
+namespace power_grid_model {
+namespace common::logging {
+
+enum class LoggingTag : Idx {
```
mgovers: if anyone has a better name, please LMK

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/calculation_info.hpp
```diff
@@ -11,7 +11,9 @@
 namespace power_grid_model::main_core {
 
 inline CalculationInfo& merge_into(CalculationInfo& destination, CalculationInfo const& source) {
-    static auto const key = Timer::make_key(2226, "Max number of iterations");
+    static constexpr auto key =
+        LoggingTag::iterative_pf_solver_max_num_iter; // TODO(mgovers) also add LoggingTag::max_num_iter; this is a bug
+                                                      // in main
```
mgovers: did not want to do this change in this PR.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
```diff
@@ -0,0 +1,123 @@
...
+    solve_sparse_linear_equation = 2225,
+    solve_sparse_linear_equation_prefactorized = 102225, // TODO(mgovers): find other error code
+    iterate_unknown = 2226,
+    calculate_math_result = 2227,
+    produce_output = 3000,
+    iterative_pf_solver_max_num_iter = 1002226, // TODO(mgovers): find other error code
+    max_num_iter = 1002228,                     // TODO(mgovers): find other error code
+};
+
+constexpr auto to_string(LoggingTag tag) {
```
mgovers: sonar cloud complains that this function is untested. I can move it to `benchmark.cpp` if you prefer. Thoughts?

TonyXiang8787: Do not use `auto` here. Make the return type explicit.

mgovers:
```suggestion
constexpr std::string to_string(LoggingTag tag) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
Jerry-Jinfeng-Guo: `LogEvent`, `LogEntry`

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -20,6 +20,20 @@ MathSolverDispatcher const& get_math_solver_dispatcher() {
     return math_solver_dispatcher;
 }
 
+std::string make_key(LoggingTag code) {
+    std::stringstream ss;
+    ss << std::setw(4) << std::setfill('0') << static_cast<std::underlying_type_t<LoggingTag>>(code) << ".";
```
Jerry-Jinfeng-Guo: Would this `4` be enough?

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -139,7 +153,7 @@ struct PowerGridBenchmark {
 
     static void print(CalculationInfo const& info) {
```
Jerry-Jinfeng-Guo: This is rather generic name. Shall we be specific here  and call it `print_info`?

tests/benchmark_cpp/benchmark.cpp
mgovers: no but that was the original code 😬 more than happy to change this, though

tests/benchmark_cpp/benchmark.cpp
mgovers: i think we can be generic here (overload resolution and it's also in an anonymous namespace in this translation unit) but no strong opinions. i'll change it.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
mgovers: i'll go with LogEvent

tests/benchmark_cpp/benchmark.cpp
mgovers: i've changed some of the codes to new codes that **do** fit in 4 digits. however, these should not be the final error codes, as there is no guideline as to how we calculate them

Jerry-Jinfeng-Guo (approved):

PR #1089, opened 2025-08-18
Linting and Formatting: S, FBT, PT, TRY checks

Laurynas-Jagutis:
Adds S, FBT, PT, TRY checks


docs/examples/Generic Branch Example.ipynb
```diff
@@ -410,7 +410,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
```
mgovers: please rerun

docs/examples/Generic Branch Example.ipynb
```diff
@@ -634,7 +634,8 @@
     "\n",
     "pd.set_option(\"display.max_columns\", None)\n",
     "pd.set_option(\"display.max_rows\", None)\n",
-    "pd.set_option(\"display.expand_frame_repr\", False)\n",
+    "expand_frame_repr = False\n",
+    "pd.set_option(\"display.expand_frame_repr\", expand_frame_repr)\n",
```
mgovers: i do not quite agree with this rule change... which one is it?

tests/unit/utils.py
```diff
@@ -260,7 +260,7 @@ def compare_result(actual: SingleDataset, expected: SingleDataset, rtol: float,
             elif expected_col.dtype == np.int32:
                 expect_all_nan = np.all(expected_col == np.iinfo("i4").min)
             else:
-                raise Exception(f"Unknown data type {expected_col.dtype}!")
+                raise Exception(f"Unknown data type {expected_col.dtype}!")  # noqa: TRY002
```
mgovers: or make a new exception type only in this file

pyproject.toml
```diff
@@ -110,7 +110,8 @@ show-fixes = true
 # Assume Python 3.11
 target-version = "py311"
 
-extend-exclude = ["setup.py"]
+[tool.ruff.lint.pydocstyle]
+convention = "google"
```
mgovers, line 114: are you sure we're using `google` and not `numpy` or `PEP257`? Feels to me like it should be `numpy`

pyproject.toml
```diff
@@ -145,10 +146,16 @@ select = [
...
+    "FBT003",
+    "TRY002",
+    "PT"
 ]
-# the rule is deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
-ignore = ["UP038"]
+# UP038 is deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
+# S701 can't be supported, autoescape=True generates the files incorrectly.
+# PT019 flags false positives
+ignore = ["UP038", "S701", "PT019"]
```
mgovers:
```suggestion
ignore = [
    "UP038", # deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
    "S701", # can't be supported: autoescape=True generates the files incorrectly.
    "PT019" # flags false positives
]
```
or
```suggestion
ignore = [
    # deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
    "UP038",
    # can't be supported: autoescape=True generates the files incorrectly.
    "S701",
    # flags false positives
    "PT019"
]
```

docs/examples/Generic Branch Example.ipynb
Laurynas-Jagutis: FBT003 Boolean positional value in function call

pyproject.toml
Laurynas-Jagutis: Google seemed the closest to what we use:

Google: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods
Numpy: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy-style-python-docstrings
PEP257: https://peps.python.org/pep-0257/#multi-line-docstrings

While running the "D" check with each of the options I got: 947 errors when google is chosen, 1073 errors when numpy is chosen and 730 errors when PEP257 is chosen.

mgovers: ok if i look at the style guide, indeed, it's google-style

#errors doesn't really mean anything here. see offline discussion

docs/examples/Generic Branch Example.ipynb
Laurynas-Jagutis: Reverted and ignored the check for example notebooks

code_generation/code_gen.py
```diff
@@ -12,7 +12,7 @@
 TEMPLATE_DIR = Path(__file__).parent / "templates"
 OUTPUT_PATH = Path(__file__).parents[1]
 
-JINJA_ENV = Environment(loader=FileSystemLoader(TEMPLATE_DIR))
+JINJA_ENV = Environment(loader=FileSystemLoader(TEMPLATE_DIR), autoescape=True)
```
mgovers, line 15: i think (but please verify) that this is no longer needed

```suggestion
JINJA_ENV = Environment(loader=FileSystemLoader(TEMPLATE_DIR))
```

Laurynas-Jagutis: by default it is set to False. Currently we set it to True and then add False autoescape overrides in specific files. 

If we removed this, the we could also remove the False autoescape overrides (they would be redundant) and then we would need to ignore this check in pyproject

Laurynas-Jagutis: In general only a single file does not have the False autoescape override, in other files the overrides are on a global scale so full files are overriden. Not sure if it is worth it then to keep that check in general

mgovers: ok then let's keep as-is :+1:

pyproject.toml
```diff
@@ -145,24 +146,30 @@ select = [
...
+    "TRY002",
+    "PT"
 ]
-# the rule is deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
-ignore = ["UP038"]
+
+ignore = [
+    "UP038", # deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
+    "PT019" # flags false positives
+    ]
```
mgovers: format
```suggestion
]
```

docs/examples/Generic Branch Example.ipynb
mgovers: not the right solution. please instead use the suggestion cfr. https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call/ to use `value=False` (kwarg) instead

src/power_grid_model/validation/_rules.py
```diff
@@ -791,11 +791,12 @@ def no_strict_subset_missing(data: SingleDataset, fields: list[str], component_t
...
         component_type: component type to check
     """
+    fill_value = False
     errors = []
     if component_type in data:
         component_data = data[component_type]
-        instances_with_nan_data = np.full_like([], False, shape=(len(component_data),), dtype=bool)
-        instances_with_non_nan_data = np.full_like([], False, shape=(len(component_data),), dtype=bool)
+        instances_with_nan_data = np.full_like([], fill_value, shape=(len(component_data),), dtype=bool)
+        instances_with_non_nan_data = np.full_like([], fill_value, shape=(len(component_data),), dtype=bool)
```
mgovers: same as before cfr. https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call/

please, instead, use `np.full_like([], fill_value=False, ...)`

src/power_grid_model/validation/_rules.py
```diff
@@ -847,10 +848,11 @@ def not_all_missing(data: SingleDataset, fields: list[str], component_type: Comp
             "The fields parameter must contain at least 2 fields. Otherwise use the none_missing function."
         )
 
+    fill_value = True
     errors = []
     if component_type in data:
         component_data = data[component_type]
-        instances_with_all_nan_data = np.full_like([], True, shape=(len(component_data),), dtype=bool)
+        instances_with_all_nan_data = np.full_like([], fill_value, shape=(len(component_data),), dtype=bool)
```
mgovers: same as before cfr. https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call/

please, instead, use `np.full_like([], fill_value=False, ...)`

docs/examples/Generic Branch Example.ipynb
Laurynas-Jagutis: running into ValueError: Must provide an even number of non-keyword arguments
https://pandas.pydata.org/docs/dev/reference/api/pandas.set_option.html

Laurynas-Jagutis: I will try one more thing

Laurynas-Jagutis: same thing

Laurynas-Jagutis: Updated it, please let me know if you agree with the solution

mgovers: way better

mgovers (approved):

PR #1092, opened 2025-08-19
Clean-up main model: move topology related construction

nitbharambe:
In the large effort of refactoring main model, one part is to move the topology and y bus related construction to a different place.

* Moved topology logic to main_core/topology.hpp
* Changed topology's dependency from state to component container. 
* Refactored ComponentConnections construction in similar way as ComponentTopology in https://github.com/PowerGridModel/power-grid-model/commit/1ab7ee82123506fcff37e8d725f987830a7a70c6
* Moved all small individual topology functions in `main_core/topology.hpp` to detail namespace.
* Edited `component_container_c` to include more container functionality and applicable for 1+ component type. Moved it to `power_grid_model::common` namespace


power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -191,4 +191,46 @@ constexpr void register_topology_components(MainModelState<ComponentContainer> c
     });
 }
 
+ComponentTopology construct_topology(main_model_state_c auto const& state) {
```
nitbharambe: This complete file should not depend on state.  Only ComponentContainer seems enough.

mgovers: awesome!

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -191,4 +191,46 @@ constexpr void register_topology_components(MainModelState<ComponentContainer> c
...
+}
+
+ComponentConnections construct_components_connections(main_model_state_c auto const& state) {
+    ComponentConnections comp_conn;
+    comp_conn.branch_connected.resize(state.comp_topo->branch_node_idx.size());
+    comp_conn.branch_phase_shift.resize(state.comp_topo->branch_node_idx.size());
+    comp_conn.branch3_connected.resize(state.comp_topo->branch3_node_idx.size());
+    comp_conn.branch3_phase_shift.resize(state.comp_topo->branch3_node_idx.size());
+    comp_conn.source_connected.resize(state.comp_topo->source_node_idx.size());
+    std::transform(state.components.template citer<Branch>().begin(), state.components.template citer<Branch>().end(),
```
mgovers: i think sonar cloud wants you to do

```suggestion
    std::ranges::transform(state.components.template citer<Branch>(),
```

nitbharambe: I also wish to make an attempt at removing the out parameter altogether.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state.hpp
```diff
@@ -37,6 +37,16 @@ concept component_container_c = requires(ContainerType const& c, ID id) {
     { c.template get_item<ComponentType>(id) } -> std::convertible_to<ComponentType const&>;
 };
 
+template <typename ContainerType, typename ComponentType>
+concept extended_component_container_c =
```
nitbharambe: Thoughts on moving this concept and component_container_c to container.hpp?
The concept is also used for any "`RetrievableType`"  like Branch etc. Should we rename the ComponentType?
RetrievableType seems Container specific.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
```diff
@@ -9,6 +9,23 @@
 #include "../all_components.hpp"
 
 namespace power_grid_model::main_core {
+
+// using only components and not state
+
+template <typename ComponentType, class ComponentContainer>
```
nitbharambe: Thoughts on deleting these and using underlying function directly?

I intend to move these to a new file container_queries.hpp otherwise.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state.hpp
mgovers: why not add these requirements directly to the `component_container_c` concept?

also, while i am 100% for moving it to its own file, I would not move it to `container.hpp` but to `container_fwd.hpp`. I want the concept to be independent of the object (actually, that's what makes it more testable: when testing the state, you don't want to create a full-fledged container but only a local translation unit-internal mock container that mimics the behavior).

In terms of dependency inversion, ComponentContainer implements the `component_container_c` as if it were an interface class, not the other way around

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
mgovers: if removing them cleans up the code, i'm all for removing them. However, the Container class as-is does not support things like `std::ranges` directly but only after using a `template <ComponentType>` (e.g. for `citer` and for `size`). That is why I added this abstraction

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state.hpp
nitbharambe: > why not add these requirements directly to the component_container_c concept?

Yes. Forgot to mention, I intend to merge them in the end. I wanted to keep them separate to start with.

>I want the concept to be independent of the object

> a local translation unit-internal mock container that mimics the behavior.
 
Nice idea. Ill move to `container_fwd.hpp`

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <concepts>
+
+namespace power_grid_model {
+
+template <typename ContainerType, typename ComponentType>
+concept component_container_c = requires(ContainerType const& c, ID id) {
+    { c.template citer<ComponentType>().begin() } -> std::forward_iterator;
+    { c.template citer<ComponentType>().end() } -> std::forward_iterator;
+    { *(c.template citer<ComponentType>().begin()) } -> std::same_as<ComponentType const&>;
+    { *(c.template citer<ComponentType>().end()) } -> std::same_as<ComponentType const&>;
```
nitbharambe:
```suggestion
```
This should not be present right since it makes no sense to dereference end(). Is this correct?

nitbharambe:
Need thoughts on https://github.com/PowerGridModel/power-grid-model/pull/1092/commits/1ab7ee82123506fcff37e8d725f987830a7a70c6. Should we keep or revert the commit?

We had earlier decided against it when it was dependent on state and lived in main model. Reason was that this function is small unlike construct_topology. In the context of current PR, we would be homogenizing both these functions.

mgovers:
> Need thoughts on [1ab7ee8](https://github.com/PowerGridModel/power-grid-model/commit/1ab7ee82123506fcff37e8d725f987830a7a70c6). Should we keep or revert the commit?
> 
> We had earlier decided against it when it was dependent on state and lived in main model. Reason was that this function is small unlike construct_topology. In the context of current PR, we would be homogenizing both these functions.

i kind of like the commit, so IMO you can keep it

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
nitbharambe: Removing them means we move from, for eg, this:

```
        return BranchIdx{get_component_sequence_idx<Node>(components, branch.from_node()),
                         get_component_sequence_idx<Node>(components, branch.to_node())};
```

to

```
        return BranchIdx{components.template get_seq<Node>(branch.from_node()),
                         components.template get_seq<Node>(branch.to_node())};
```
I suppose its less ugly to read in the original way. Ill keep these functions for now.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state.hpp
nitbharambe: Moved. Renamed to RetrievableType since this concept is part of container now.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
nitbharambe: >I also wish to make an attempt at removing the out parameter altogether.

Initial experiments failed. Ill try again in a different PR. The existing PR is getting complicated. Hence resolved.

nitbharambe:
Limiting this PR's scope. Continuing in new one.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/container_queries.hpp
```diff
@@ -0,0 +1,26 @@
+
```
mgovers: useless whiteline
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/container_queries.hpp
```diff
@@ -0,0 +1,26 @@
...
+    return components.template size<ComponentType>();
+}
+
+template <typename ComponentType, class ComponentContainer>
+    requires component_container_c<ComponentContainer, ComponentType>
+inline Idx get_component_sequence_idx(ComponentContainer const& components, auto const& id_or_index) {
+    return components.template get_seq<ComponentType>(id_or_index);
+}
+
+} // namespace power_grid_model::main_core
```
mgovers: missing whiteline at EOF

```suggestion
} // namespace power_grid_model::main_core

```

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
mgovers, line 1: can you please add `static_assert`s to the `test_container.cpp` to verify that the container indeed satisfies the `component_container_c` concept?

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
```diff
@@ -0,0 +1,27 @@
...
+    { *(c.template citer<RetrievableType>().end()) } -> std::same_as<RetrievableType const&>;
+    { c.template get_item<RetrievableType>(id) } -> std::convertible_to<RetrievableType const&>;
+    { c.template size<RetrievableType>() } -> std::same_as<Idx>;
+    { c.template get_seq<RetrievableType>(idx2d) } -> std::same_as<Idx>;
+};
+
+template <typename ContainerType, typename... RetrievableType>
+concept multi_component_container_c = (component_container_c<ContainerType, RetrievableType> && ...);
+
+} // namespace power_grid_model
```
mgovers:
```suggestion
} // namespace power_grid_model

```

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
```diff
@@ -0,0 +1,27 @@
...
+//
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+#include "common/common.hpp"
+
+#include <concepts>
+
+namespace power_grid_model {
```
mgovers: maybe

```suggestion
namespace power_grid_model::common {
```

what do you think?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -152,43 +144,96 @@ constexpr void register_topology_components(MainModelState<ComponentContainer> c
...
+    detail::apply_registration<Component>(components, comp_conn.branch3_phase_shift,
+                                          [](Branch3 const& branch3) { return branch3.phase_shift(); });
+}
+
+template <std::same_as<Source> Component, class ComponentContainer>
+    requires component_container_c<ComponentContainer, Component>
+constexpr void register_connections_components(ComponentContainer components, ComponentConnections& comp_conn) {
+    detail::apply_registration<Component>(components, comp_conn.source_connected,
+                                          [](Source const& source) { return source.status(); });
+}
```
mgovers, line 206: maybe everything above this in `namespace detail`?

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
nitbharambe: Its better for this to not be in power_grid_model namespace. Lets go with common. Later on as we add specific things in this namespace, we can rename into something more appropriate.

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
```diff
@@ -0,0 +1,27 @@
...
+    { c.template citer<RetrievableType>().end() } -> std::forward_iterator;
+    { *(c.template citer<RetrievableType>().begin()) } -> std::same_as<RetrievableType const&>;
+    { *(c.template citer<RetrievableType>().end()) } -> std::same_as<RetrievableType const&>;
+    { c.template get_item<RetrievableType>(id) } -> std::convertible_to<RetrievableType const&>;
+    { c.template size<RetrievableType>() } -> std::same_as<Idx>;
+    { c.template get_seq<RetrievableType>(idx2d) } -> std::same_as<Idx>;
+};
+
+template <typename ContainerType, typename... RetrievableType>
+concept multi_component_container_c = (component_container_c<ContainerType, RetrievableType> && ...);
```
nitbharambe: Okay, maybe I should merge both of these concepts together.

nitbharambe: Done

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
mgovers: for a concept, it's OK to have that. I do believe, however, that we should try to bump to use actual iterator checks. It's possible that this is still a remnant from back when we couldn't use STL concepts for ranges because of boost

mgovers: you are right in the sense that it is undefined behavior, though... Maybe you can replace it with something along the lines of


```suggestion
    { c.template citer<ComponentType>().end() } -> std::same_as<decltype(c.template citer<ComponentType>().begin())>;
```

power_grid_model_c/power_grid_model/include/power_grid_model/container_fwd.hpp
mgovers: don't know what happened but this is back

PR #1093, opened 2025-08-20
Clean-up main model: Add logger interface

mgovers:
* [x] Use a logger interface as an abstraction for calculation info
* [x] The entire math solver translation unit does not know calculation info anymore

Cfr. offline discussion, this experiment was cleaned up and taken to be the first step towards the calculation info cleanup


power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -4,15 +4,75 @@
...
+class CalculationInfo : public NoLogger {
+    using Data = std::map<LogEvent, double>;
 
-using CalculationInfo = std::map<LogEvent, double>;
+  public:
+    using Report = std::add_lvalue_reference_t<std::add_const_t<Data>>;
+    static_assert(std::same_as<Report, Data const&>);
 
+    void log(LogEvent tag, double value) override { log_impl(tag, value); }
+    void log(LogEvent tag, Idx value) override { log_impl(tag, static_cast<double>(value)); }
```
Jerry-Jinfeng-Guo, line 34: Should we add a super generic `log(LogEvent tag)` without value to just log the event taking place? 
```suggestion
    void log(LogEvent tag, double value) override { log_impl(tag, value); }
    void log(LogEvent tag, Idx value) override { log_impl(tag, static_cast<double>(value)); }
    void log(LogEvent tag) override { log_impl(tag); }
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
```diff
@@ -43,8 +43,26 @@ enum class LogEvent : int16_t {
     max_num_iter = 2248,                     // TODO(mgovers): find other error code
 };
 
+struct Logger {
```
nitbharambe: An interface is a good idea and definitely be incorporated in any design we agree on.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        if (loggers_.size() == 1 && loggers_.front() != nullptr) { // allows perfect forwarding to log messages
+            loggers_.front()->log(tag, std::forward<T>(values)...);
+        } else {
+            for (auto& logger : loggers_) {
+                if (logger != nullptr) {
+                    logger->log(tag, values...);
+                }
+            }
+            capturing::into_the_void(std::forward<T>(values)...);
```
nitbharambe: values can be any count of arguments and maybe not all get used in logging and hence unused ones get sent to void?

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
```diff
@@ -0,0 +1,68 @@
...
+namespace power_grid_model::common::logging {
+class NoLogger : public Logger {
+    void log(LogEvent /*tag*/) override {}
+    void log(LogEvent /*tag*/, std::string_view /*message*/) override {}
+    void log(LogEvent /*tag*/, double /*value*/) override {}
+    void log(LogEvent /*tag*/, Idx /*value*/) override {}
+    std::unique_ptr<Logger> clone() const override { return std::make_unique<NoLogger>(); }
+};
+
+class LogDispatcher final : public LogDispatch {
```
nitbharambe: Although Dispatcher is currently not agreed on, I think this is a good experiment that we should refer back to in future when we include new logs.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    LogDispatcher() = default;
+    LogDispatcher(const LogDispatcher&) = delete;
+    LogDispatcher& operator=(const LogDispatcher&) = delete;
+    LogDispatcher(LogDispatcher&&) = default;
+    LogDispatcher& operator=(LogDispatcher&&) = default;
+    ~LogDispatcher() override = default;
+
+  private:
+    std::vector<Logger*> loggers_;
```
nitbharambe: If multithreading related logic for log was to be moved to logger, it can be:
`std::vector<std::vector<Logger*>> loggers_;`
And there would be a separate aggregation function.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/calculation_info.hpp
```diff
@@ -10,16 +10,9 @@
...
-                                                    // in main
-    for (auto const& [k, v] : source) {
-        if (k == key) {
-            destination[k] = std::max(destination[k], v);
-        } else {
-            destination[k] += v;
-        }
+inline Logger& merge_into(Logger& destination, CalculationInfo const& source) {
+    for (const auto& [tag, value] : source.report()) {
+        destination.log(tag, value);
```
nitbharambe, line 15: Maybe this merge_into can call a different function merge_thread_into. And we can then have a maximum action for all rather than accumulate

nitbharambe: The base_log needs to be added and not maximized, but some similar logic can be a way to not have sum of threads as output of benchmark

mgovers: i believe the `merge_into` is a nice example of something where it depends on the event type whether to add or maximize. e.g., the total amount of time spent within a thread is indeed something that could use the $L_{\infty}$ norm (maximizing) but there is a different event which is e.g. total computation resources used, which should use the $L_1$ norm (summation) across threads.

However, that should be a multi-threaded logger decision (e.g., it can take 1 event of a specific event type and convert it into 2 events, one that uses $L_1$ norm and one that uses $L_{\infty}$ norm.)

mgovers: implementation-wise, that could look like something along the lines of

```cpp
class MultiThreadedCalculationInfo final : public Logger {
  private:
    using SubThreadLogger = CalculationInfo; // probably needs to be a wrapper class with mutex and/or thread id knowledge
    using MainThreadLogger = CalculationInfo;

  public:
    template <typename... Args>
    void log(Args&&... args) 
        requires requires(SubThreadLogger logger) {
            { logger.log(args) };
        }
    { 
        // assert it is the main thread and/or add a mutex
        main_log.log(std::forward<Args>(args));
    }

    void sync() {
        // assert it is the main thread and/or add a mutex
        for (auto& thread_log : thread_logs) {
            if (thread_log != nullptr) {
                sync_thread(*thread_log);
            }
        }
    }

    void report() const {
        return main_log.report();
    }

  private:
    MainThreadLogger main_log;
    std::vector<std::unique_ptr<SubThreadLogger>> thread_logs;

    void sync_thread(SubThreadLogger thread_log) {
        // guard with mutex
        for (auto& [tag, value] : thread_log.report) {
            switch(tag) {
                case total_time_in_thread:
                    log(maximum_time_in_thread, value);
                    log(tag, value);
                    break;
                default:
                    log(tag, value);
                    break;
            }
        }
    }
};
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
mgovers: apologies. i did add this but apparently didn't commit. reopening

mgovers: OHHH i see, i did add it to the logger interface but not here.

What should the report look like? integer counter (similar to how `maximum_num_iterations` is a double but represents a float)?

mgovers: thinking about it, it is tricky because it depends on the event whether it should time as `0`, or keep track of a counter... maybe it's good to add it when the first event comes into play that should just be an event counter

Jerry-Jinfeng-Guo: I imagine it makes sense that we would sometimes simply want to know that things have happened, without needing to know how long the duration of each thing. For timers it makes perfect sense to associate a value to tell the duration.

mgovers: i do think it would be useful. i just don't know how to report.

let's say: we have the following code
```cpp
CalculationInfo info;
try {
  Timer timer{info, LogEvent::some_event_timer};             // current way
  CodeHit{info, LogEvent::some_event_started};               // new event

  do_something();

  CodeHit{info, LogEvent::some_event_hit}; // other new event
  CodeHit{info, LogEvent::some_event_hit}; // other new event
} except (...) {
  CodeHit{info, LogEvent::some_event_threw_exception}; // another event
}

// what should the output look like here?
auto const report = info.report();
// example output:
// std::map<LogEvent, double>{
//   {some_event_timer: 0.2},
//   {some_event_started: 1.0}, // 1.0 because it is hit once
//   {some_event_hit: 2.0}      // 2.0 because it was hit twice
// }

// now let's consider accidentally using a timed log event
CodeHit{info, LogEvent::some_event_timer};

// what should the output look like here?
auto const report = info.report();
// example output
// std::map<LogEvent, double>{
//   {some_event_timer, 1.2},   // CodeHit accidentally adds 1 to timer
//   {some_event_started, 1.0},
//   {some_event_hit, 2.0}
// }
```

under the hood, what we could do is split up the report into separate maps, e.g.

```cpp
struct Timing {
    double timing{};
};
struct MaximizedInteger {
    Idx count{};
};
struct HitCount {
    Idx count{};
};

// member variables; maybe combine into one but maybe has performance hit
std::map<LogEvent, Timing> timings;
std::map<LogEvent, MaximizedInteger> num_iter;
std::map<LogEvent, HitCount> hit_count;        // or maybe just std::vector<LogEvent>

using Metric = std::variant<Timing, MaximizedInteger, HitCount>;

std::map<LogEvent, Metric> report(); // combines the 3 maps and throws error if there
                                     // are events logged in 2 different maps
```

feels like overengineering in the short term, though...

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
mgovers: no this is about how to do perfect forwarding (in particular about using a moved object). an alternative would be

```cpp
        if (auto const first_non_null = std::ranges::find_if(loggers_, [] (auto& logger) { return logger != nullptr; });
            first_non_null != loggers_.end()) {
            for (auto& logger : std::ranges::subrange(first_non_null + 1, loggers_.end());) {
                if (logger != nullptr) {
                    logger->log(tag, values...);
                }
            }
            (*first_non_null)->std::forward<T>(values)...);
        } else {
            capturing::into_the_void(std::forward<T>(values)...)
        }
```

mgovers: by C++ guidelines: every object that was moved into the current function shall be moved again (for performance reasons).

Forwarding means that the object could be passed by `const&` or be moved into this object (or captured by value for small objects). You don't know. Therefore, you need to forward objects provided to this function again. But you can only forward once (otherwise, you might be moving an already moved object) and exactly once (cfr. the guideline)

`capturing::into_the_void` is a sink that consumes objects that would otherwise not be moved into this function.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
mgovers: cfr. offline discussion: because this is quite a simple experiment, we can redo it when needed.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
mgovers: cfr. offline discussion: removed the `LogDispatcher` altogether

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
mgovers: removed

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
mgovers: indeed. for now, the log dispatcher was removed, but let's keep this in mind indeed

mgovers: each sub-logger instance would need its own sync logic

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
Jerry-Jinfeng-Guo: I think `std::vector<LogEvent>` suffices for recording what's happened. Since our logging is not recording on the go, and instead keep a list/map of things, I think this is indeed the best we could do for this purpose. Although ideally, it could look something like `Event #1234 hit at timestamp 12:34:56.789`.

mgovers: that would be a full-fledged logger, which has its own purpose (possible future addition). this calculation info is not a full-fledged logger but an aggregate, so indeed `vector` it is, then

mgovers: if `std::vector` will suffice, then we might as well log it as `tag, 0`. That way, the report is unambiguous (it will report that the event was hit, but it does not add any time to the function)

Jerry-Jinfeng-Guo: With `vector` one could still record multiple occurrences of single type event. If we stick to the map and give it value `0`, it will simple keep rewriting `0` with `0`.

mgovers: thinking about it, i really don't think we should add this right now unless there is a specific event we want to capture this way (we currently do not, but if you have a specific request, please LMK)

as you mention, indeed, `0` as a default value is a bad value, but it will add runtime overhead when having to check `is_nan` for every log, so even adding it to the map is not a good idea in the first place.

alternatively, a separate vector means that the report will have to be created (copied) at report time and passed by value, not by reference. While i'm ok with that, there's no need to build that now.

in addition, how should the reporting happen when there are multiple events in the vector, and also events listed in the other data.

all things considered, it feels like too much implementation details even though there's no concrete use case of it.

power_grid_model_c/power_grid_model/include/power_grid_model/component/asym_line.hpp
```diff
@@ -16,7 +14,9 @@
 #include "../common/common.hpp"
 #include "../common/matrix_utils.hpp"
 #include "../common/three_phase_tensor.hpp"
-#include "line_utils.hpp"
+
+#include <iostream>
```
TonyXiang8787, line 18: @mgovers can you explain why do we need `iostream` here?

mgovers: i missed this. this was auto-formatted (the `#include` statement was already there). will remove in a separate PR

PR #1094, opened 2025-08-20
Clean-up main model: migrate reset calculation info to job dispatch

mgovers:
The location where the calculation info was reset is quite opaque. To more explicitly show the calculation 

NOTE: this does not solve the actual issue that main model owns its own info, nor the fact that the location where the reset is done is somewhere halfway through the calculation. It merely makes it more visible what happens under the hood.

* [x] Run Benchmark and make sure that there are no regressions like the one introduced in #1087

Relates to #1082 (reinstigated the original logic to allow benchmarking against original main)
Relates to #1093 (this PR is an enabler for follow-up improvements similar to #1093, but it is not a blocker for it)

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
@@ -68,6 +68,13 @@ template <typename Adapter> class JobInterface {
     {
         return static_cast<const Adapter*>(this)->get_calculation_info_impl();
     }
+    CalculationInfo reset_calculation_info()
+        requires requires(Adapter& adapter) { // NOSONAR
+            { adapter.reset_calculation_info_impl() } -> std::same_as<CalculationInfo>;
+        }
+    {
+        return static_cast<Adapter*>(this)->reset_calculation_info_impl();
+    }
 
     void thread_safe_add_calculation_info(CalculationInfo const& info)
         requires requires(Adapter& adapter) { // NOSONAR
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
@@ -68,12 +68,12 @@ template <typename Adapter> class JobInterface {
     {
         return static_cast<const Adapter*>(this)->get_calculation_info_impl();
     }
-    CalculationInfo reset_calculation_info()
+    void reset_calculation_info()
         requires requires(Adapter& adapter) { // NOSONAR
-            { adapter.reset_calculation_info_impl() } -> std::same_as<CalculationInfo>;
+            { adapter.reset_calculation_info_impl() } -> std::same_as<void>;
         }
     {
-        return static_cast<Adapter*>(this)->reset_calculation_info_impl();
+        static_cast<Adapter*>(this)->reset_calculation_info_impl();
     }
 
     void thread_safe_add_calculation_info(CalculationInfo const& info)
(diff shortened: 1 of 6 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -68,6 +68,13 @@ template <typename Adapter> class JobInterface {
     {
         return static_cast<const Adapter*>(this)->get_calculation_info_impl();
     }
+    CalculationInfo reset_calculation_info()
+        requires requires(Adapter& adapter) { // NOSONAR
+            { adapter.reset_calculation_info_impl() } -> std::same_as<CalculationInfo>;
+        }
+    {
+        return static_cast<Adapter*>(this)->reset_calculation_info_impl();
+    }
```
mgovers: NOTE: this ugly dependency should be resolved in follow-ups after #1093

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -68,6 +68,13 @@ template <typename Adapter> class JobInterface {
     {
         return static_cast<const Adapter*>(this)->get_calculation_info_impl();
     }
+    CalculationInfo reset_calculation_info()
+        requires requires(Adapter& adapter) { // NOSONAR
+            { adapter.reset_calculation_info_impl() } -> std::same_as<CalculationInfo>;
```
nitbharambe: Does not return anything

mgovers: yup, you're right... CI is also complaining. I wonder why my own build didn't

mgovers:
still something weird going on with the benchmark. do-not-merge label added

mgovers:
> still something weird going on with the benchmark. do-not-merge label added

Found the bug on `main` that caused the new results to (incorrectly) appear much slower compared to the original one:
image

However, after digging into this, it turns out that the tap position optimizer contains a bug in the following line: https://github.com/PowerGridModel/power-grid-model/blob/feature/migrate-reset-calculation-info/power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp#L1033

On current `main`, the calculation info was cleaned up before every tap position optimizer subcalculation, and, as a result, the tap position optimizer was not fully benchmarked, because the intermediate timings were discarded - not great. Moving cleaning responsibility to the JobDispatch meant that we now DO benchmark the tap position optimizer, effectively fixing this bug.

NOTE: this also relates to #1082 .

TBD: we have 3 paths going forward:

* [x] merge https://github.com/PowerGridModel/power-grid-model/pull/1094 as-is, including the fix for the calculation info. that means that the benchmarks on versions after this PR will not be compatible with versions prior to this PR.
* [ ] add an info.clear() line in that location in  https://github.com/PowerGridModel/power-grid-model/pull/1094. that would actually just shift the problem from main model impl to the optimizer. i don't particularly like this solution.
* [ ] same as 2. but immediately after that PR, do a fix that removes the info.clear(). that way, there is a single PR that contains a single line bugfix, but we still have the downsides of 1.

mgovers:
Decision: merge this PR as-is and recalibrate our benchmark.

mgovers:
> still something weird going on with the benchmark. do-not-merge label added

do-not-merge label removed cfr. above

nitbharambe (approved):

PR #1095, opened 2025-08-21
Clang-format check: Add EOF whiteline check

nitbharambe:


nitbharambe:
It is possible to add this in VScode settings too.
Should I add it for our repository?
```
    "files.insertFinalNewline": true,
    "files.trimFinalNewlines": true
```

mgovers:
Cool! Be weary when running `pre-commit run --all-files clang-format` that some files generated by the code generator need formatting as well, and therefore also manually the templates. However, i don't know whether jinja allows that. We'll have to see, maybe we need an explicit ignore statement there

PR #1099, opened 2025-08-28
Error messages: clearer message when dataset is inconsistent

mgovers:


src/power_grid_model/_core/buffer_handling.py
```diff
@@ -200,9 +200,15 @@ def _get_dense_buffer_properties(
...
+        raise ValueError(
+            f"Incorrect/inconsistent data provided: {'batch' if actual_is_batch else 'single'} "
+            f"data provided but {'batch' if is_batch else 'single'} data expected. {VALIDATOR_MSG}"
+        )
     if batch_size is not None and batch_size != actual_batch_size:
-        raise ValueError(f"Provided 'batch size' is incorrect for the provided data. {VALIDATOR_MSG}")
+        raise ValueError(
+            f"Incorrect/inconsistent batch size provided: {actual_batch_size} scenarios provided "
+            f"but {batch_size} scenarios expected. {VALIDATOR_MSG}"
+        )
```
nitbharambe, line 211:
```suggestion
        raise ValueError(
            f"Incorrect/inconsistent batch size provided in the data: "
            f"one is {actual_batch_size} scenarios the other is {batch_size} scenarios. {VALIDATOR_MSG}"
        )
```
Since there is nothing "actual" or "expected".

mgovers: well, once we get to this function, there is: the input requests/expects `batch_size`, but then the actual data is different (`actual_batch_size`).

The only alternative would be to make a sanity check and/or a try/except block in the Dataset handlers. what are your thoughts?

nitbharambe: The sanity check would be better then. So this stays as an internal error while sanity check is something we provide for users. To our users, it is quite confusing what an actual/expected size is.

mgovers: went with a slightly different implementation than my original thought but it looks way better now

tests/unit/test_buffer_handling.py
```diff
@@ -95,6 +94,82 @@ def test__get_sparse_buffer_properties(component_type, is_columnar):
...
+    ],
+)
+def test__get_buffer_properties__batch_requested_for_single_data(component_type, is_columnar):
+    data = load_data(component_type, is_batch=False, is_columnar=is_columnar, is_sparse=False)
+    schema = power_grid_meta_data[DatasetType.update][component_type]
+
+    with pytest.raises(
+        ValueError, match="Incorrect/inconsistent data provided: single data provided but batch data expected."
+    ):
+        get_buffer_properties(data, schema=schema, is_batch=True, batch_size=BATCH_DATASET_NDIM)
```
nitbharambe, line 116: I see it doesn't originate from this PR, but can you make the load_data also use the constants like BATCH_DATASET_NDIM?

Everything else LGTM.

PR #1104, opened 2025-08-28
Compiler support: enable C++23 in PGM core compilation

mgovers:
This PR unlocks and enables features (that may be relevant to us) like:

* multi-dimensional subscript operator
* `std::to_underlying` for enums
* `std::views::zip` (shown as an example)
* monadic operations for `std::optional`
* `<expected>`
* `std::unreachable`
* `std::ranges::to` to construct new containers from another range
* formatting ranges
* `std::ranges::contains`

To show the usage, the dependency of the `common` directory on `boost` is entirely removed because `std::views::zip` was unlocked. It worked out of the box, and it even simplifies the code.

See also https://en.cppreference.com/w/cpp/compiler_support/23.html

### DIFF
--- a/tests/cpp_unit_tests/test_y_bus.cpp
+++ b/tests/cpp_unit_tests/test_y_bus.cpp
@@ -7,9 +7,6 @@
 
 #include <doctest/doctest.h>
 
-#include <boost/range/adaptors.hpp>
-#include <boost/range/irange.hpp>
-
 #include <ranges>
 
 namespace power_grid_model {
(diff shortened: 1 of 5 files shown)


tests/cpp_unit_tests/test_y_bus.cpp
```diff
@@ -7,9 +7,6 @@
 
 #include <doctest/doctest.h>
 
-#include <boost/range/adaptors.hpp>
-#include <boost/range/irange.hpp>
```
nitbharambe, line 11: The tests have boost/range includes as well.

nitbharambe (approved):
Is

PR #1107, opened 2025-09-01
Cleanup main model: dependency inversion main model impl info

mgovers:
* [ ] Make calculation info a proper logger
* [ ] Give main model impl and job adapter the logger
* [ ] Multithreaded logging is a specialized class and no longer is scattered throughout the job dispatch and adapter
* [ ] NOTE: the destructors are responsible for syncing and unregistering the child classes

relates to #1093

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -33,7 +33,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@v6
         with:
           python-version: "3.13"
 
@@ -308,7 +308,7 @@ jobs:
       - uses: actions/checkout@v5
 
       - name: Setup python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: '3.12'
           architecture: x64
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: '3.12'
 
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -43,7 +43,7 @@ jobs:
         run: |
           brew install boost eigen nlohmann-json msgpack-cxx doctest
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: "3.12"
       - name: Install build-wrapper
@@ -51,7 +51,7 @@ jobs:
         
       - name: Python test and coverage
         run: |
-          pip install -e .[dev]
+          CC=clang CXX=clang++ pip install -e .[dev]
           pytest
 
       - name: Run build-wrapper for C++


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -610,6 +604,11 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
     bool construction_complete_{false};
 #endif // !NDEBUG
 
+    Logger& logger() const {
+        static common::logging::NoLogger no_log{};
+        return log_ != nullptr ? *log_ : no_log;
+    }
```
mgovers, line 612: one future improvement could be to pass the logger not by reference but by pass-by-value view that does this same thing.

tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -76,21 +98,72 @@ class JobAdapterMock : public JobInterface<JobAdapterMock> {
     void prepare_job_dispatch_impl(MockUpdateDataset const& /*update_data*/) const { /* patch base class function */ }
     void setup_impl(MockUpdateDataset const& /*update_data*/, Idx /*scenario_idx*/) const { ++(counter_->setup_calls); }
     void winddown_impl() const { ++(counter_->winddown_calls); }
-    static CalculationInfo get_calculation_info_impl() {
-        CalculationInfo info;
-        info.log(LogEvent::total, 0.0);
-        return info;
+    void reset_logger_impl() {
+            nullptr) { // this if statement may be encountered when the destructor is called on a moved object
```
mgovers:
```suggestion
    void reset_logger_impl() {
        if (counter_ ==
            nullptr) { // this if statement may be encountered when the destructor is called on a moved object
```

mgovers:
do-not-merge label removed because i think the PR is in a good state for a first iteration

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <mutex>
+
+namespace power_grid_model::common::logging {
+
+template <std::derived_from<Logger> LoggerType>
+    requires requires(LoggerType& destination, LoggerType const& source) {
+        { merge_into(destination, source) };
+    }
+class MultiThreadedLoggerImpl : public MultiThreadedLogger {
+    using SubThreadLogger = LoggerType;
```
figueroa1395: Nitpick: I'm not a fan of `SubThreadLogger`, maybe `ParentThreadLogger` or something different sounds better?

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -0,0 +1,54 @@
...
+      public:
+        ThreadLogger(MultiThreadedLoggerImpl& parent) : parent_{&parent} {}
+        ~ThreadLogger() override { sync(); }
+        void sync() const { parent_->sync(*this); }
+
+      private:
+        MultiThreadedLoggerImpl* parent_;
+    };
+
+    virtual std::unique_ptr<Logger> create_child() override { return std::make_unique<ThreadLogger>(*this); }
```
figueroa1395: Nitpick: But I find `override` only cleaner and clearer.
```suggestion
    std::unique_ptr<Logger> create_child() override { return std::make_unique<ThreadLogger>(*this); }
```

On the other hand, who do we want to manage the lifetime of these? Is this to be managed by job dispatch? Maybe I'll answer this question along the way.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging_impl.hpp
```diff
@@ -26,4 +26,6 @@ class NoLogger : public Logger {
     }
 };
```
figueroa1395, line 27: Commenting here just because, but can we rename this file? `logging_impl.hpp` is a bit missleading to me. What about something like `dummy_logging.hpp`?

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "logging_impl.hpp"
+
+#include <mutex>
+
+namespace power_grid_model::common::logging {
+
+template <std::derived_from<Logger> LoggerType>
+    requires requires(LoggerType& destination, LoggerType const& source) {
+        { merge_into(destination, source) };
```
figueroa1395: Our loggers (`NoLogger` and `CalculationInfo`) "implement" `merge_into` via a free function, and maybe I just have to continue reading, but wouldn't this mean that our own loggers can't be used as `MultiThreadedLogger`s?

On the other hand, I suppose the idea of this is to make the users, implementing their own loggers, also add `merge_into` into them? It reads a bit as two different design choices.

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+  private:
+    friend class ThreadLogger;
+
+    LoggerType log_;
+    std::mutex mutex_;
+
+    void sync(ThreadLogger const& logger) {
+        std::lock_guard const lock{mutex_};
+        merge_into(log_, static_cast<SubThreadLogger const&>(logger));
```
figueroa1395: This `merge_into` in the free function one in the same logging `namespace`, hence why do we then need the requirement of a `LoggerType` to implement it?

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
```diff
@@ -152,5 +156,14 @@ class JobAdapter<MainModel, ComponentList<ComponentType...>>
...
     }
+
+    void reset_logger_impl() {
+        log_ = nullptr;
+        model_reference_.get().reset_logger();
+    }
+    void set_logger_impl(Logger& log) {
+        log_ = &log;
+        model_reference_.get().set_logger(*log_);
+    }
```
figueroa1395, line 167: This tells me now than `MainModelImpl` owns the logger? Is that what we want? Or maybe it just owns a pointer, but then should we still call this via main model instead of via whoeover owns the logger data?

Edit: I see that main model just holds a pointer to each logger, which the adapter does too, so we can just redirect on each thread... This seems fine, and can potentially help us clean the pattern used here in the adapter of having a reference wrapper and a copy of the model at the same time. However, even though this makes it clean having "copies" in two places, could that bite us in the butt later? Should we keep a pointer to the logger only here for instance, and call it via the adaptor in main model impl? Probably not, but maybe something worth thinking about for a couple of minutes.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -35,38 +38,76 @@ class JobDispatch {
         // if the batch_size is zero, it is a special case without doing any calculations at all
         // we consider in this case the batch set is independent but not topology cacheable
         if (n_scenarios == 0) {
+            adapter.reset_logger();
```
figueroa1395, line 41: This and above mean we don't log anything, right? Is that the intention? If there is "nothing" to calculate...

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -35,38 +38,76 @@ class JobDispatch {
...
+        common::logging::NoMultiThreadedLogger no_log{};
+        return batch_calculation(adapter, result_data, update_data, threading,
+                                 static_cast<common::logging::MultiThreadedLogger&>(no_log));
+    }
+    template <typename Adapter, typename ResultDataset, typename UpdateDataset>
+        requires std::is_base_of_v<JobInterface<Adapter>, Adapter>
+    static BatchParameter batch_calculation(Adapter& adapter, ResultDataset const& result_data,
+                                            UpdateDataset const& update_data) {
+        return batch_calculation(adapter, result_data, update_data, sequential);
+    }
```
figueroa1395: Nice.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -35,38 +38,76 @@ class JobDispatch {
...
+                messages[scenario_idx] = "unknown exception";
+            }
+        };
+    }
 
     template <typename Adapter, typename ResultDataset, typename UpdateDataset>
     static auto single_thread_job(Adapter& base_adapter, ResultDataset const& result_data,
-                                  UpdateDataset const& update_data, std::vector<std::string>& exceptions) {
-        return [&base_adapter, &exceptions, &result_data, &update_data](Idx start, Idx stride, Idx n_scenarios) {
+                                  UpdateDataset const& update_data, std::vector<std::string>& exceptions,
```
figueroa1395, line 82: Commenting here because for some reason doesn't allow me to do so in the right line:

Nitpick: This is about:
```
auto thread_info_ptr = base_log.create_child();
Logger& thread_info = *thread_info_ptr;
```

Should we keep the `base_log` - `log` language? Something like:
```
auto thread_log_ptr = base_log.create_child();
Logger& thread_log = *thread_info_ptr;
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -78,11 +78,12 @@ class MainModel {
...
     BatchParameter calculate(Options const& options, MutableDataset const& result_data,
                              ConstDataset const& update_data) {
+        info_.clear();
         JobAdapter<Impl, AllComponents> adapter{std::ref(impl()), std::ref(options)};
-        return JobDispatch::batch_calculation(adapter, result_data, update_data, options.threading);
+        return JobDispatch::batch_calculation(adapter, result_data, update_data, options.threading, info_);
     }
 
-    CalculationInfo calculation_info() const { return impl().calculation_info(); }
+    CalculationInfo calculation_info() const { return info_.get(); }
```
figueroa1395, line 88: There are some places where this is now called `info_` and others `log`. Should we homogenize to one?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -159,7 +158,7 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
           math_solver_dispatcher_{&math_solver_dispatcher} {}
 
     MainModelImpl(MainModelImpl const& other)
```
figueroa1395, line 161: One of the goals of doing the `CalculationInfo` refactor was to get rid of these verbose and error prone constructors and assignment operators for main model. Maybe we should keep an eye on that. With the `log_` pointer living in `MainModelImpl` that may not be the case.

Maybe we can come up with a way to access the info from `MainModelImpl` without having to "own" a copy of the pointer there, just for having the above mentioned cleanup in mind.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -99,6 +100,7 @@ class MainModel {
     }
 
     std::unique_ptr<Impl> impl_;
+    MultiThreadedCalculationInfo info_;
```
figueroa1395, line 105: So `MainModel` owns the multi-threaded logger which is just an "upgraded" version of a regular logger, right? And then the job dispatch itself will own the children, correct? Just making sure my train of thought is correct.

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
```diff
@@ -83,7 +95,7 @@ class JobAdapter<MainModel, ComponentList<ComponentType...>>
     // current_scenario_sequence_cache_ is calculated per scenario, so it is excluded from the constructors.
     main_core::utils::SequenceIdx<ComponentType...> current_scenario_sequence_cache_{};
 
-    std::mutex calculation_info_mutex_;
+    Logger* log_{nullptr};
```
nitbharambe, line 98: Why does job adapter also need to store it?

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -86,7 +86,24 @@ class CalculationInfo final : public Logger {
...
 };
+
+inline Logger& merge_into(Logger& destination, CalculationInfo const& source) {
+    for (const auto& [tag, value] : source.report()) {
+        destination.log(tag, value);
+    }
+    return destination;
+}
+
+class MultiThreadedCalculationInfo : public MultiThreadedLoggerImpl<CalculationInfo> {
```
nitbharambe, line 100: The SubThreadLogger being CalculationInfo was kind of confusing to me before review.
It is now a SubThreadLogger which is a "map".

Lets throw all notions of earlier CalculationInfo out? Or atleast rename te calculation_info in places like MathSolver.
We do that in a later PR or now?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: exactly

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: one of the thoughts i've had is whether to do (symbolically):

* [ ] the current way
      ```cpp
      reset_loger(log)   // register
      calculate(args...) // run
      reset_logger()     // unregister
      ```
* [ ] proposed
      ```cpp
      calculate(args..., log) // run with logger
      ```

the advantage of the former is having less `log` arguments throughout the code. the advantage of the latter is having less verbose constructors and less complex logic flow

i'd consider the current implementation to be a good intermediate state to split up the PRs if we want to go for the latter instead.

c.c. @nitbharambe

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: i think `log` is better for most locations, but `info` makes sense where explicitly `CalculationInfo` is used. The current line of code is merely an intermediate state before we push the logger's responsibility across the C API. Follow-ups are to make both the `model.cpp` and the `benchmark.cpp` own their own loggers and pass them to the `MainModel`

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: yes, good one

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: Definitely do this in a follow up PR, but I think we should keep in mind the goal of simplifying the overall logic. If the logger logic itself is a bit complex, that should be alright, as long as it is self contained and its complexity doesn't permeate the rest, but rather makes the actual PGM simpler.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: IMO we should move to one single line for these. i can clean up in this PR if you prefer, or wait a little bit.

I also think, e.g., that `MainModelImpl` should only have one constructor

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: exactly

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
figueroa1395: I just commented on this line because it is where I thought about this. But I have the feeling that in other places `info` is used where the "generic" `log` may be more adequate. Something to keep in mind that can also be polished later.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: the main thing is that there should be a `reset_logger()` call before every `return` statement with the current implementation (note: i don't particularly like that; see my earlier comment)

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
mgovers: if you like, we can either go with the `log` in every function call OR we can create a logger view, which takes a logger but returns `NoLogger` by default if it's not set

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: We can do these changes in a follow up PR. Let's keep this one self contained with the dependency injection via the new logger stuff.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: do you mean https://github.com/PowerGridModel/power-grid-model/pull/1107#discussion_r2316130298?

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
figueroa1395: I don't know if either convince me, but I don't have any alternative idea either. So we can try the view and see how it looks like (puns intended 😆)

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: because it is not known a priori how to take one logger and merge it into another logger.

E.g., if one has a `NoLogger`, `merge_into` will be the no-op.
but let's say we have a logger that keeps track of the time order of logs, then we may have 2 overloads, e.g.

```cpp
Logger& merge_into(Logger& dest, SubType const& source) {
    for (auto const& [k, v] : source) {
        dest.log(k, v);
    }
    return dest;
}
SubType& merge_into(SubType& dest, SubType const& source) {
    dest.insert(source);
    return dest;
}
```
or the source may e.g. keep track of the variants of logging. Only the type of the source knows how to handle itself for a specific destination type

mgovers: btw: it does not need to be a free function anymore, if that is what you mean

figueroa1395: What I mean is, in line 52, that call is done via the free function `merge_into` that lives in the same `namespace`, hence it is very confusing, because you are not using the `SubType` implementation, which makes it seem useless to have that `concept`. I foresee that to be very confusing later on. For these reasons, maybe it is better not to have `merge_into` as a free function (?) and just have each logger instance implement it (perhaps)? There may be some duplication and a little bit more writing, but it becomes clearer to me. Or maybe there is something I'm missing.

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: > Out loggers (NoLogger and CalculationInfo) "implement" merge_into via a free function, and maybe I just have to continue reading, but wouldn't this mean that our own loggers can't be used as MultiThreadedLoggers?

I ran into the limitations of inheritance and did not want to go down the path of diamond inheritance (e.g. `using MultiThreadedCalculationInfo = MultiThreadedLogger<CalculationInfo>` can only function as a calculation info in itself if it derives from both CalculationInfo **and** MultiThreadedLogger at the same time, which results in diamond inheritance)

I therefore unfortunately had to let go of the idea of an exposure-only pattern.

`merge_into` may indeed be too restrictive as is... ideas are welcome

figueroa1395: I'll think about it.

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: `virtual` was redundant in the first place 😬 

> Is this to be managed by job dispatch? Maybe I'll answer this question along the way.

as you may have noticed: indeed, the lifetime is managed by every thread individually

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: let me think about it a little bit

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
mgovers: math solver was already completely made agnostic of the calculation info

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
figueroa1395: > `virtual` was redundant in the first place 😬

Yup, I meant to ask to remove it :)

> as you may have noticed: indeed, the lifetime is managed by every thread individually,

Yup, this is nice.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
mgovers: the only locations that IMO should know of calculation info are the objects/functions that manage a calculation info in their own scope, that is: the benchmark and the C API (`main_model.hpp` for now)

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: or i can just remove it. it's a private member type anyways

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: i can probably make it a member

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
mgovers: let's do it in a separate PR

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: i meant https://github.com/PowerGridModel/power-grid-model/pull/1107#discussion_r2316113388

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
mgovers: image
most of them are dataset, the rest are already migrated

figueroa1395: Let's keep as is for now then and refine if needed along the follow-ups.

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
figueroa1395: I think that'd be good, in my opinion. Just for the sake of clarity.

mgovers: done

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
mgovers: i've thought about his for a bit and this relates to the registering/unregistering discussion from elsewhere in this PR (specifically the one in https://github.com/PowerGridModel/power-grid-model/pull/1107/#discussion_r2316113388)

i've described the fact that the registering/unregistering is quite tricky in itself. an alternative could be to use `model.calculate(..., log_)`. Now we can also do a hybrid situation here: the adapter could be the boundary between registering/unregistering (to the job adapter). The model itself could be `model.calculate(..., log_)`. Then, the adapter still contains a logger, but the model no longer does.

thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: i've made it such that the implementation class itself can decide whether it implements `merge_into`, rather than making it a virtual function. That way, for things like a `file_logger` that logs to one file per thread cannot be merged into another file. if we ever want to implement such a logger, we can decide how to do the multithreading at that point (maybe just empty sync statement)

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
mgovers: Experiment in https://github.com/PowerGridModel/power-grid-model/pull/1115

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: resolved

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: Experiment in https://github.com/PowerGridModel/power-grid-model/pull/1115

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: What do you mean with pass-by-value view in this context?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: I'll keep an eye on https://github.com/PowerGridModel/power-grid-model/pull/1115, I'll resolve this comment and take a look at the experiment PR when you say it's ready for a look.

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
figueroa1395: I guess this comment can be answered as well by https://github.com/PowerGridModel/power-grid-model/pull/1115, right?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -22,10 +22,13 @@ class JobDispatch {
     template <typename Adapter, typename ResultDataset, typename UpdateDataset>
         requires std::is_base_of_v<JobInterface<Adapter>, Adapter>
     static BatchParameter batch_calculation(Adapter& adapter, ResultDataset const& result_data,
-                                            UpdateDataset const& update_data, Idx threading = sequential) {
+                                            UpdateDataset const& update_data, Idx threading,
+                                            common::logging::MultiThreadedLogger& log) {
```
figueroa1395, line 26: Side note, but it may be a good idea to update our internal UML around  `job_dispatch` with this logger dependency (and include the logger stuff into it). I mean a simple version, just to show progress there.

tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -76,21 +93,79 @@ class JobAdapterMock : public JobInterface<JobAdapterMock> {
...
     }
-    void reset_calculation_info_impl() const { ++(counter_->reset_calculation_info_calls); }
 };
 
 class SomeTestException : public std::runtime_error {
   public:
     using std::runtime_error::runtime_error;
 };
+
+class TestLogger : public common::logging::Logger {
```
figueroa1395, line 113: Maybe this is a next step, but I don't think `TestLogger` belongs here. I'd say we put it at the same level as `NoLogger` (or something similar), and then use it here but latter on also to unit test the logger stuff... As in our own loggers should be directly tested, but the multithreaded ones and the interface (as we expect users to make their own loggers - theoretically) should be tested through something like this. Just food for thought.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: Leaving this one open as a reminder for a follow up. Feel free to resolve.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers:
```cpp
class LoggerView : public Logger {
  public:
    LoggerView() = default;
    LoggerView(Logger& logger): logger_{logger} {}

    // all the overrides
    void log(...) override { log_impl(...); }

  private:
    Logger* logger_{};

    template <typename... Args>
    void log_impl(Args&&... args)
        requires(Logger& logger) {
            { logger.log(std::forward<Args>(args)...); }
        }
    {
        if (log_ != nullptr) {
            log_->log(std::forward<Args>(args)...);
        }
    }
};
```

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: yes we need to refresh that after this PR and #1115 (and follow-ups) anyways

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: Given #1115, this should not be needed anymore.

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
figueroa1395: In #1115 the paradigm changes a bit, now only the adapter "owns" the logger. This makes sense as main model should not have to deal with logging, it should just know it can log something. Whereas the adapter is the layer in between the job dispatch (which manages the threads that own the logger) and the main model. Potentially, another experiment can be done after #1115 in which we try to "get rid" of the "ownership" of the logger from the adapter, and just keep the job dispatch in charge, but that is to be seen. I personally like the current state after #1115.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -6,6 +6,7 @@
 
 #include <cmath>
 #include <complex>
+#include <concepts>
```
nitbharambe, line 9:
```suggestion
```

mgovers: it's required. it just wasn't included in this list before because of include order, but if you include it as a standalone file, then it's required

PR #1109, opened 2025-09-02
Cleanup main model: Clean up templates and variadics

nitbharambe:
Goals:

* Clean MainModelImpl of duplicate/repeated templates.
* Make MainModelImpl Testable for any collection of components.


power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -29,4 +30,43 @@ template <class... Types, class Functor> constexpr auto run_functor_with_all_typ
     return std::array { functor.template operator()<Types>()... };
 }
 
+template <typename Tuple> struct tuple_type_identities_to_tuple_types;
+
+template <typename... Ts> struct tuple_type_identities_to_tuple_types<std::tuple<std::type_identity<Ts>...>> {
+    using type = std::tuple<Ts...>;
+};
```
mgovers: 😮

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -205,23 +205,24 @@ constexpr void register_connections_components(ComponentContainer components, Co
...
 }
 
+template <typename Tuple, std::size_t... Is>
+void register_all_components_impl(auto const& components, ComponentTopology& comp_topo, std::index_sequence<Is...>) {
+    (detail::register_topology_components<std::tuple_element_t<Is, Tuple>>(components, comp_topo), ...);
+}
+
+template <typename Tuple> void register_all_components(auto const& components, ComponentTopology& comp_topo) {
+    register_all_components_impl<Tuple>(components, comp_topo, std::make_index_sequence<std::tuple_size_v<Tuple>>{});
+}
```
mgovers: awesome

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -205,23 +205,24 @@ constexpr void register_connections_components(ComponentContainer components, Co
                                   [](Source const& source) { return source.status(); });
 }
 
+template <typename Tuple, std::size_t... Is>
+void register_all_components_impl(auto const& components, ComponentTopology& comp_topo, std::index_sequence<Is...>) {
+    (detail::register_topology_components<std::tuple_element_t<Is, Tuple>>(components, comp_topo), ...);
+}
+
+template <typename Tuple> void register_all_components(auto const& components, ComponentTopology& comp_topo) {
```
nitbharambe: We can have a common run_functor_with_tuple too.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -29,4 +30,43 @@ template <class... Types, class Functor> constexpr auto run_functor_with_all_typ
...
+
+    return std::tuple_cat(std::conditional_t<sub_type_in_type.template operator()<Types>(),
+                                             std::tuple<typename std::type_identity<Types>>, std::tuple<>>{}...);
+}
+
+// main model implementation template
+template <class T, class U> struct MainModelType;
+
+template <class... ExtraRetrievableType, class... ComponentType>
+struct MainModelType<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentList<ComponentType...>> {
```
nitbharambe: A struct which should contain all information about the MainModel being created.
The above utilities can be incorporated here as well.

One downside is the addition of `typename` at many places since this is a struct.

Also this is start of the experiment, it might be possible to clean up or come up with different ways altogether.

nitbharambe:
The build/tests seem to fail. I had cleaned up before pushing. Checking issues.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: For future reference: We also have similar utility of get_cls_pos in container (Which is not container specific). They can be combined. (Also I read addressing via tuple is better than recursive)

mgovers: > One downside is the addition of typename at many places since this is a struct.

for most standard library objects, `foo_t` is a shorthand for `typename foo::type`. maybe that can clean up that part a little bit? I personally am not really concerned about the amount of typenames

mgovers: > Also I read addressing via tuple is better than recursive

100% 💯

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -29,4 +30,43 @@ template <class... Types, class Functor> constexpr auto run_functor_with_all_typ
...
+    // using MainModelState = main_core::MainModelState<ComponentContainer>;
+    using ComponentTypesTuple = std::tuple<std::type_identity<ComponentType>...>;
+
+    using TopologyTypesTuple =
+        std::tuple<std::type_identity<Node>, std::type_identity<Branch>, std::type_identity<Branch3>,
+                   std::type_identity<Source>, std::type_identity<Shunt>, std::type_identity<GenericLoadGen>,
+                   std::type_identity<GenericVoltageSensor>, std::type_identity<GenericPowerSensor>,
+                   std::type_identity<GenericCurrentSensor>, std::type_identity<Regulator>>;
+    // using TopologyConnectionTypes = std::tuple<std::type_identity<Branch>, std::type_identity<Branch3>,
+    // std::type_identity<Source>>;
```
figueroa1395: Just a thought, but since this is very verbose/explicit, maybe it can be code generated so we don't forget to add stuff manually here. Or somehow link it to a list elsewhere, so we have only one explicit list, like in `all_components.hpp`. Just a thought for later cleanup.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
figueroa1395: This is very cool!

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
figueroa1395: I don't mind this at all, I actually find it very neat.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -205,23 +205,24 @@ constexpr void register_connections_components(ComponentContainer components, Co
                                   [](Source const& source) { return source.status(); });
 }
 
+template <typename Tuple, std::size_t... Is>
```
figueroa1395: Nitpick: Can you rename `Is` to something more expresive?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -29,4 +30,43 @@ template <class... Types, class Functor> constexpr auto run_functor_with_all_typ
...
+}
+
+// main model implementation template
+template <class T, class U> struct MainModelType;
+
+template <class... ExtraRetrievableType, class... ComponentType>
+struct MainModelType<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentList<ComponentType...>> {
+
+    using ComponentContainer = Container<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentType...>;
+    // using MainModelState = main_core::MainModelState<ComponentContainer>;
```
figueroa1395: Maybe these two (included the commented out line) can potentially be used "more" in `main_model_impl.hpp` to clean it up a bit (there is some "duplication" in the other file now); just like it is done with `MainModelType`.

figueroa1395 (commented):
I really like where this PR is going :). Very good ideas here.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -36,35 +36,59 @@ template <typename... Ts> struct tuple_type_identities_to_tuple_types<std::tuple
...
-constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...> /*types_tuple*/,
-                                  std::tuple<std::type_identity<SubTypes>...> /*sub_types_tuple*/) {
-    constexpr auto sub_type_in_type = []<typename T>() { return (std::is_same_v<T, SubTypes> || ...); };
+template <typename Tuple>
+using tuple_type_identities_to_tuple_types_t = typename tuple_type_identities_to_tuple_types<Tuple>::type;
+
+template <typename... Types, typename... SelectTypes>
+constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...>,
+                                  std::tuple<std::type_identity<SelectTypes>...>) {
+    constexpr auto sub_type_in_type = []<typename T>() { return (std::is_same_v<T, SelectTypes> || ...); };
```
figueroa1395: Nitpick: Maybe `sub_type_in_type` can be renamed to reflect the new naming `SubTypes` -> `SelectTypes`?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -36,35 +36,59 @@ template <typename... Ts> struct tuple_type_identities_to_tuple_types<std::tuple
...
 struct MainModelType<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentList<ComponentType...>> {
 
     using ComponentContainer = Container<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentType...>;
-    // using MainModelState = main_core::MainModelState<ComponentContainer>;
-    using ComponentTypesTuple = std::tuple<std::type_identity<ComponentType>...>;
-
-    using TopologyTypesTuple =
+    using MainModelState = main_core::MainModelState<ComponentContainer>;
+    using ExtraRetrievableTypesTuple = std::tuple<std::type_identity<ExtraRetrievableType>...>;
+    using ComponentTypesTuple = std::tuple<ComponentType...>;
```
figueroa1395: Should we make a distinction in the naming when something is just a `tuple` or a `tuple` of `std::type_identity`s?  Or did you mean using `ComponentTypesTuple = std::tuple<std::type_identity<ComponentType>...>;`?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -36,35 +36,59 @@ template <typename... Ts> struct tuple_type_identities_to_tuple_types<std::tuple
...
+
+    // TODO: Clean up using-s
+    using _AllTypesTuple =
+        std::tuple<std::type_identity<ComponentType>..., std::type_identity<ExtraRetrievableType>...>;
+    // TODO: Would making a unique be necessary? We have Node mentioned as a ExtraRetrievableType and ComponentType so
+    // summing up is possible but maybe not appropriate. Would it be better to handle them both separately? using
+    // _AllTypesTupleUnique = typename tuple_type_identities_to_tuple_types<decltype(filter_tuple_types(
+    // _TopologyTypesTuple{}, _AllTypesTuple{}))>::type;
+
+    using _ComponentTypesTuple = std::tuple<std::type_identity<ComponentType>...>;
```
figueroa1395: This and around, tied to https://github.com/PowerGridModel/power-grid-model/pull/1109/commits/f717e8cea02ed8696d48e68abbd4eb27cd5edfad#r2329356329. Maybe `_*` is not descriptive enough when using `std::type_identity` in the `tuple`s.

figueroa1395:
It looks like Sonar isn't identifying c++23 https://github.com/PowerGridModel/power-grid-model/actions/runs/17494449107/job/49691592601?pr=1109 (see complain about `std::ranges::to`)? Maybe it can be configured correctly? Perhaps https://github.com/PowerGridModel/power-grid-model/pull/1116 will fix it.

c.c @mgovers

mgovers:
> It looks like Sonar isn't identifying c++23 https://github.com/PowerGridModel/power-grid-model/actions/runs/17494449107/job/49691592601?pr=1109 (see complain about `std::ranges::to`)? Maybe it can be configured correctly? Perhaps #1116 will fix it.
> 
> c.c @mgovers

that's exactly the intention of #1116 indeed 😬

tests/cpp_unit_tests/test_core_utils.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+        static_assert(
+            std::is_same_v<typename ModelType::ComponentContainer,
+                           Container<ExtraRetrievableTypes<Base, Node, Branch, Appliance>, Node, Line, Source>>);
+        static_assert(std::is_same_v<typename ModelType::ComponentTypesTuple, std::tuple<Node, Line, Source>>);
+        static_assert(std::is_same_v<typename ModelType::TopologyTypesTuple, std::tuple<Node, Branch, Source>>);
+        static_assert(std::is_same_v<typename ModelType::TopologyConnectionTypesTuple, std::tuple<Branch, Source>>);
+        static_assert(ModelType::n_component_types == 3);
+    }
+    SUBCASE("Different component order: Line Source Node") {
```
figueroa1395: So this one proves with the current implementation, the order doesn't matter, right? awesome!

tests/cpp_unit_tests/test_core_utils.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+        static_assert(
+            std::is_same_v<typename ModelType::ComponentContainer,
+                           Container<ExtraRetrievableTypes<Base, Node, Branch, Appliance>, Line, Source, Node>>);
+        static_assert(std::is_same_v<typename ModelType::ComponentTypesTuple, std::tuple<Line, Source, Node>>);
+        static_assert(std::is_same_v<typename ModelType::TopologyTypesTuple, std::tuple<Node, Branch, Source>>);
+        static_assert(std::is_same_v<typename ModelType::TopologyConnectionTypesTuple, std::tuple<Branch, Source>>);
+        static_assert(ModelType::n_component_types == 3);
+    }
+    SUBCASE("Bad case: Line Source") {
```
figueroa1395: Shouldn't this fail already as `Node` isn't present? shouldn't the model be not "constructible" according to our discussions?

tests/cpp_unit_tests/test_core_utils.cpp
mgovers: 1. `static_assert`s don't need to be in a subcase/test case (because they are static)
2. only that the main model template itself can be instantiated with arbitrary order, not that the model itself can be correctly created

tests/cpp_unit_tests/test_core_utils.cpp
figueroa1395: I suppose the answer to this is related to https://github.com/PowerGridModel/power-grid-model/pull/1109#discussion_r2329434879 item 2? It is still fine to instantiate this, but now we have to show that it can't be constructed? Would it be possible to make it fail at this stage, would that be desirable? I would guess so.

mgovers: this is about whether the type is valid, not about whether the code actually works. if you want to make this `static_assert`s, you'll also have to make the constructor `constexpr` and actually instantiate

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -6,14 +6,43 @@
...
+    constexpr bool is_empty_component() const { return !has_any_elements; }
+    constexpr bool is_independent() const { return qualify_for_optional_id() || provided_ids_valid(); }
+    constexpr Idx get_n_elements() const {
+        assert(uniform || elements_ps_in_update == utils::invalid_index);
+
+        return qualify_for_optional_id() ? elements_ps_in_update : na_Idx;
+    }
+};
+
+/////////////////// To remove ///////////////////
```
nitbharambe, line 25: The usage of these utils with `ComponentTypes...` is for now (and mostly in future) limited only to ComponentTypes. So nice to just move them inside.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: > for most standard library objects, foo_t is a shorthand for typename foo::type. maybe that can clean up that part a little bit? I personally am not really concerned about the amount of typenames

I meant in the usage that we see in main_model_impl.hpp. We have `typename MainModelType::SequenceIdx`. 
But now that the changes are done, I dont see it cluttering a lot.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: The alias here are the ones which shall be used outside. Ive changed the rest to *_ values signifying private members (and added private specifier too).

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: Marked them with private. They are named only to make the logic clear.

tests/cpp_unit_tests/test_core_utils.cpp
nitbharambe: > static_asserts don't need to be in a subcase/test case (because they are static)

True. I understand that the subcase/test case is redundant here. Intention was to classify them instead of plain statements because I thought there would be way too many. What would be the way to go about classification?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: Thats the goal. It should be used at all places now.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: This doesnt seem to be viable to be code generated. 
Also the rules of the order are quite specific and dependent on the corresponding usage in main_model_impl.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: These would be removed in final stage after all usages are cleared. which would be after https://github.com/PowerGridModel/power-grid-model/pull/1119/

tests/cpp_unit_tests/test_core_utils.cpp
nitbharambe: This case is now tested in static asserts in https://github.com/PowerGridModel/power-grid-model/pull/1120/files

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
figueroa1395: Do you mean that all functions between lines 16 - 32 will be removed, or that `ComponentTypes...` is to be removed? Because I see that in #1119 (at least some of) these functions are being used, that the parameter pack is not needed no more, which I believe is the intention?

nitbharambe: They are now included in MainModelType. They are not used independently anywhere else. (Since ComponenntTypes... is not directly available anywhere)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -45,52 +46,81 @@ constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...>,
...
-                   std::type_identity<GenericCurrentSensor>, std::type_identity<Regulator>>;
-    using _TopologyConnectionTypes =
-        std::tuple<std::type_identity<Branch>, std::type_identity<Branch3>, std::type_identity<Source>>;
+                   std::type_identity<GenericCurrentSensor>, std::type_identity<Regulator>>{};
 
-    using TopologyTypesTuple =
-        tuple_type_identities_to_tuple_types_t<decltype(filter_tuple_types(_TopologyTypesTuple{}, _AllTypesTuple{}))>;
-    using TopologyConnectionTypesTuple = tuple_type_identities_to_tuple_types_t<decltype(filter_tuple_types(
-        _TopologyConnectionTypes{}, _AllTypesTuple{}))>;
+    static constexpr auto topology_connection_types_tuple_v_ =
```
figueroa1395: Just as a general question, why are these in `topology_connection_types...`?

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
figueroa1395, line 1: This is fantastic!

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -13,6 +13,34 @@
 namespace power_grid_model::main_core::utils {
 
 constexpr Idx invalid_index{-1};
+
+struct UpdateCompProperties {
```
figueroa1395: Why was this removed from `update.hpp`?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -122,17 +122,19 @@ template <class T, class U> class MainModelImpl;
...
 template <class... ExtraRetrievableType, class... ComponentType>
 class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentList<ComponentType...>> {
-  private:
-    // internal type traits
-    // container class
-    using ComponentContainer = Container<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentType...>;
-    using MainModelState = main_core::MainModelState<ComponentContainer>;
 
+  public:
     using MainModelType = main_core::utils::MainModelType<ExtraRetrievableTypes<ExtraRetrievableType...>,
```
figueroa1395: Maybe done in a following commit, but can this one be cleaned up a bit as well?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -213,12 +215,10 @@ class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentLis
...
-    std::array<bool, main_core::utils::n_types<ComponentType...>>
-    get_components_to_update(ConstDataset const& update_data) const {
-        return main_core::utils::run_functor_with_all_types_return_array<ComponentType...>(
-            [&update_data]<typename CompType>() {
-                return (update_data.find_component(CompType::name, false) != main_core::utils::invalid_index);
-            });
+    std::array<bool, MainModelType::n_types> get_components_to_update(ConstDataset const& update_data) const {
+        return MainModelType::run_functor_with_all_component_types_return_array([&update_data]<typename CompType>() {
+            return (update_data.find_component(CompType::name, false) != main_core::utils::invalid_index);
+        });
```
figueroa1395: Once again, so neat! Nice job!

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -12,6 +12,38 @@
 #include <vector>
 namespace power_grid_model::main_core::utils {
 
+namespace detail {
```
figueroa1395, line 14: Maybe all the `run_functor_*` functions could now live here. And just keep them in one place.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -46,7 +46,32 @@ template <typename T> bool check_id_na(T const& obj) {
 
 namespace independence {
 
-using UpdateCompProperties = utils::UpdateCompProperties;
+struct UpdateCompProperties {
```
figueroa1395, line 49: I see this is back :) ignore my previous comment about it, I am reviewing by commits so I can follow your train of thoughts :)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
figueroa1395: Saw this was addressed later... resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
nitbharambe: Forward declaring the struct was not enough for it to be used in the MainModelType struct. (And I cant include update because of circular dependency) So I had moved it to core_utils instead.
And later moving to a separate file made it possible to revert the change.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+    // Update related types
+    using OwnedUpdateDataset = std::tuple<std::vector<typename ComponentType::UpdateType>...>;
+    using SequenceIdxView = std::array<std::span<Idx2D const>, n_types>;
+    using UpdateIndependence = std::array<main_core::update::independence::UpdateCompProperties, n_types>;
+    using SequenceIdx = std::array<std::vector<Idx2D>, n_types>;
+    using SequenceIdxRefWrappers = std::array<std::reference_wrapper<std::vector<Idx2D> const>, n_types>;
+    using ComponentFlags = std::array<bool, n_types>;
+
+    // Clean these 2. They are unused
```
figueroa1395, line 84: This is to be done in the follow up after #1119?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
figueroa1395: After taking a better look, I guess the only remaining is https://github.com/PowerGridModel/power-grid-model/pull/1109/files#r2343888619, right?

nitbharambe: Since those run_functor_* were only accessed for entire `ComponentTypes...` with no use case for running on arbitrary `Types...`, they are moved to MainModelType. 
If it is to be ran for arbitrary `Types...` we can create the util and use that util in MainModelType.

The run_functor_tuple however works on arbitrary `Types...` without using any template from MainModelType. Hence I put it as tuple.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: Ignore this, this is actually the "entry point" to the rest of the cleanup.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: https://github.com/PowerGridModel/power-grid-model/pull/1109#discussion_r2343882777

figueroa1395:
I have no further comments. Great job. To my understanding the few remaining todos are only to be addressed in followups, correct? I believe we should try and merge soon, so this doesn't grow anymore, it is already a massive improvement.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
nitbharambe: topology_connection_types = Order for creating state.comp_conn
topology_types_= Order for creating state.comp_topo

figueroa1395: So `topology_connection_types` refers to the types which connection can change and hence affect the topology?

nitbharambe:
> I have no further comments. Great job. To my understanding the few remaining todos are only to be addressed in followups, correct? I believe we should try and merge soon, so this doesn't grow anymore, it is already a massive improvement.

Correct. Remaining PRs should address those TODOs

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -28,5 +39,11 @@ template <class... Types, class Functor> constexpr void run_functor_with_all_typ
 template <class... Types, class Functor> constexpr auto run_functor_with_all_types_return_array(Functor functor) {
     return std::array { functor.template operator()<Types>()... };
 }
+/////////////////// To remove ///////////////////
+
+template <typename Tuple, class Functor> constexpr void run_functor_with_tuple_return_void(Functor&& functor) {
+    detail::run_functor_with_tuple_index_return_void<Tuple>(functor,
```
mgovers: perfect forwarding

```suggestion
    detail::run_functor_with_tuple_index_return_void<Tuple>(std::forward<Functor>(functor),
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -6,14 +6,25 @@
...
 #include <array>
 #include <vector>
-
 namespace power_grid_model::main_core::utils {
 
+namespace detail {
+
+template <typename Tuple, class Functor, std::size_t... Indices>
+constexpr void run_functor_with_tuple_index_return_void(Functor functor, std::index_sequence<Indices...> /*unused*/) {
+    (functor.template operator()<std::tuple_element_t<Indices, Tuple>>(), ...);
```
mgovers:
```suggestion
constexpr void run_functor_with_tuple_index_return_void(Functor&& functor, std::index_sequence<Indices...> /*unused*/) {
    (std::forward<Functor>(functor).template operator()<std::tuple_element_t<Indices, Tuple>>(), ...);
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -0,0 +1,97 @@
...
+                                             std::tuple<std::type_identity<Types>>, std::tuple<>>{}...);
+}
+
+} // namespace detail
+
+template <class T, class U> struct MainModelType;
+
+// TODO: discussion on checking dependent types can also be done here.
+template <class... ExtraRetrievableType, class... ComponentType>
+struct MainModelType<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentList<ComponentType...>> {
```
mgovers: since it contains `private` members, let's make it a class

```suggestion
class MainModelType<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentList<ComponentType...>> {
  public:
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -0,0 +1,97 @@
...
+    static constexpr auto all_types_tuple_v_ =
+        std::tuple<std::type_identity<ComponentType>..., std::type_identity<ExtraRetrievableType>...>{};
+    // TODO: Would making a unique be necessary? We have Node mentioned as a ExtraRetrievableType and ComponentType so
+    // summing up is possible but maybe not appropriate. Would it be better to handle them both separately? using
+
+    static constexpr auto topology_types_tuple_v_ =
+        std::tuple<std::type_identity<Node>, std::type_identity<Branch>, std::type_identity<Branch3>,
+                   std::type_identity<Source>, std::type_identity<Shunt>, std::type_identity<GenericLoadGen>,
+                   std::type_identity<GenericVoltageSensor>, std::type_identity<GenericPowerSensor>,
+                   std::type_identity<GenericCurrentSensor>, std::type_identity<Regulator>>{};
```
mgovers, line 64: would it be possible to forward-declare these component types? if so, we can construct a main model without including the respective component types... would be nice

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
mgovers, line 1: this is beautiful

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
```diff
@@ -15,12 +15,12 @@
 
 namespace power_grid_model {
 
-template <class MainModel, typename... ComponentType> class JobAdapter;
+template <class MainModel> class JobAdapter;
 
-template <class MainModel, class... ComponentType>
-class JobAdapter<MainModel, ComponentList<ComponentType...>>
-    : public JobInterface<JobAdapter<MainModel, ComponentList<ComponentType...>>> {
+template <class MainModel> class JobAdapter : public JobInterface<JobAdapter<MainModel>> {
```
mgovers, line 20: can we make it such that the `JobInterface` does not use a reference to the `MainModel` template parameter anymore? i think this CRTP pattern is not mixin but interface type, so in that case, C++23 `deducing this` would solve this!!! (knowledge sharing session incoming)

tests/cpp_unit_tests/test_main_model_type.cpp
mgovers, line 1: it's beautiful 🥹

tests/cpp_unit_tests/test_main_model_type.cpp
```diff
@@ -0,0 +1,155 @@
...
+        CHECK(calls == std::vector<std::string_view>{"node", "a_component", "source"});
+
+        calls.clear();
+        utils::run_functor_with_tuple_return_void<typename ModelType::TopologyTypesTuple>(
+            [&calls]<typename CompType>() { calls.push_back(std::string_view(CompType::name)); });
+        CHECK(calls == std::vector<std::string_view>{"node", "source"});
+    }
+
+    SUBCASE("Bad case: Line Source") {
+        // TODO rewrite for checking fail instead of pass
```
mgovers, line 140: remaining TODO: when do we want to resolve this?

tests/cpp_unit_tests/test_main_model_type.cpp
```diff
@@ -0,0 +1,155 @@
...
+        static_assert(std::is_same_v<typename ModelType::ComponentContainer,
+                                     Container<ExtraRetrievableTypes<Base, Branch, Appliance>, Line, Source>>);
+        static_assert(std::is_same_v<typename ModelType::ComponentTypesTuple, std::tuple<Line, Source>>);
+        static_assert(std::is_same_v<typename ModelType::TopologyTypesTuple, std::tuple<Branch, Source>>);
+        static_assert(std::is_same_v<typename ModelType::TopologyConnectionTypesTuple, std::tuple<Branch, Source>>);
+        static_assert(ModelType::n_types == 2);
+    }
+
+    // TODO add static_assert(std::constructible_from<ModelType, double, meta_data::MetaData const&,
+    // MathSolverDispatcher const&>);
```
mgovers, line 152: remaining TODO: when do we want to resolve this?

mgovers:
FYI: IMO this PR is good to go after clang-tidy and sonar cloud pass. everything else can wait.

tests/cpp_unit_tests/test_main_model_type.cpp
nitbharambe: In https://github.com/PowerGridModel/power-grid-model/pull/1120

tests/cpp_unit_tests/test_main_model_type.cpp
nitbharambe: In a final PR following #1119 and #1120

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
nitbharambe: Included this in #1119 for completion.

tests/cpp_unit_tests/test_core_utils.cpp
figueroa1395: Resolving as the concepts and its tests introduced in #1120 should cover this.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
figueroa1395: When going through `topology.hpp`, this is clear. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
figueroa1395: This would then be as a follow up and scoped out of this PR, correct?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
nitbharambe: Forward declaring these at the moment would not result in decoupling at the current state of PR. The definition slips in via for eg. update.hpp. It seems possible to address this and other issues but requires many changes. Hence in a separate PR.

An observation: 
Here and at other places we make use of nested type `ComponentType::UpdateType` or some other information available from definition. A way to go about it would be to maybe include only `Base`, and make it known that the types are derived from Base.

power_grid_model_c/power_grid_model/include/power_grid_model/job_adapter.hpp
nitbharambe: Yes, scoped out. We can change that for all CRTP usages in the repo too.

mgovers: alright, sounds good. resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
mgovers: alright, good to keep in mind. let's scope out for now

mgovers:
failed status checks seem to be caused by connection resets, not because of this PR

figueroa1395:
@nitbharambe @mgovers removed from merge queue to do benchmark beforehand

mgovers:
do-not-merge label added for benchmark run

nitbharambe:
image

```
                                           metric     old_sum     new_sum
0                                     Build model    0.081213    0.077572
1                           Calculate math result    0.079664    0.074877
2                                   Calculate rhs    0.025084    0.028427
3                                      Copy model    0.063022    0.059020
4                              Create math solver    0.009057    0.009050
5                          Initialize calculation    0.050788    0.048282
6                             Initialize voltages    0.001017    0.001066
7                                 Iterate unknown    0.039090    0.038466
8                                Math Calculation    0.770783    0.774556
9                                     Math solver    0.725347    0.731657
10                     Pre-process measured value    0.012138    0.012446
11                                        Prepare    0.369775    0.351605
12                                Prepare LHS rhs    0.043439    0.044815
13                                 Prepare matrix    0.012371    0.012112
14    Prepare matrix, including pre-factorization    0.037772    0.040377
15                           Prepare the matrices    0.060864    0.058170
16                                 Produce output    0.029647    0.029644
17                                  Restore model    0.435006    0.435215
18                   Solve sparse linear equation    0.316834    0.319764
19  Solve sparse linear equation (pre-factorized)    0.045426    0.052123
20                                          Total  123.227696  122.255050
21              Total batch calculation in thread   84.543940   83.022636
22             Total single calculation in thread   84.476222   82.960251
23                                   Update model    1.259807    1.225683
```

PR #1111, opened 2025-09-03
Clean-up main model: fix windows clang-tidy

mgovers:
Run full clang-tidy on Windows and resolve issues.

## Checks before merge

* [x] Run full clang-tidy on Windows (including `power_grid_model_benchmark_cpp` target)
* [x] Full run in https://github.com/PowerGridModel/power-grid-model/actions/runs/17428105940

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
@@ -84,8 +84,8 @@ template <typename Adapter> class JobInterface {
         static_cast<Adapter*>(this)->thread_safe_add_calculation_info_impl(info);
     }
 
-  protected:
-    // Protected & defaulted special members — CRTP: only the derived can create/copy/move this base
+  private:
+    friend Adapter;
     JobInterface() = default;
     JobInterface(const JobInterface& /*other*/) = default;
     JobInterface& operator=(const JobInterface& /*other*/) = default;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/y_bus.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/y_bus.hpp
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include "../common/common.hpp"
+
 namespace power_grid_model::main_core {
 
 template <symmetry_tag sym, typename ComponentContainer, typename... ComponentType>
@@ -47,7 +49,7 @@ void prepare_y_bus(MainModelState<ComponentContainer> const& state_, Idx n_math_
 
 template <symmetry_tag sym, typename ComponentContainer, typename... ComponentType>
 static std::vector<MathModelParamIncrement> get_math_param_increment(
-    MainModelState<ComponentContainer> received_state, Idx n_math_solvers_,
+    MainModelState<ComponentContainer>& received_state, Idx n_math_solvers_,
     std::array<std::vector<Idx2D>, main_core::utils::n_types<ComponentType...>> const& parameter_changed_components_) {
     using AddToIncrement =
         void (*)(std::vector<MathModelParamIncrement>&, MainModelState<ComponentContainer> const&, Idx2D const&);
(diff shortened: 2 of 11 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/main_core/y_bus.hpp
```diff
@@ -47,7 +49,7 @@ void prepare_y_bus(MainModelState<ComponentContainer> const& state_, Idx n_math_
 
 template <symmetry_tag sym, typename ComponentContainer, typename... ComponentType>
 static std::vector<MathModelParamIncrement> get_math_param_increment(
-    MainModelState<ComponentContainer> received_state, Idx n_math_solvers_,
+    MainModelState<ComponentContainer>& received_state, Idx n_math_solvers_,
```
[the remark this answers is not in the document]
mgovers: > The parameter name 'received_state' with a non-const reference is misleading. Since this appears to be input data, it should either be const or renamed to indicate it will be modified (e.g., 'mutable_state').

should not be `const`. Maybe the name should be changed. Deferred to #1109

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
```diff
@@ -264,7 +264,8 @@ struct YBusStructure {
         // construct transpose entry
         lu_transpose_entry.resize(nnz_counter_lu);
         // default transpose_entry[i] = i
-        std::iota(lu_transpose_entry.begin(), lu_transpose_entry.end(), 0);
+        std::iota( // NOLINT(modernize-use-ranges) // not yet supported by some compilers
```
mgovers, line 267: 2 reasons:

1. `std::ranges::iota` is not yet supported by the latest XCode, so `std::ranges::iota(lu_transpose_entry, 0)` is not an option
2. `std::ranges::to` is not supported by GCC-13, which is used in the `conda` and `check-code-quality` test pipelines, so `lu_transpose_entry = IdxRange{nnz_counter_lu} | std::views::to<IdxVector>();` is not an option

My preference goes to the latter because we don't need the `resize` anymore. In addition, it is something we probably should fix in the short term, as our documentation claims that GCC >= 14 is required.

* For the code quality pipeline, upgrading just requires adding a line to install `gcc-14 g++-14` (or later)
* For the conda pipeline, we need to decide on whether to use the conda version (`libgcc-15.1` is installed) and how to do so, or whether to also just install `gcc-14`.

that is out of scope of this PR, though

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -84,8 +84,8 @@ template <typename Adapter> class JobInterface {
         static_cast<Adapter*>(this)->thread_safe_add_calculation_info_impl(info);
     }
 
-  protected:
-    // Protected & defaulted special members — CRTP: only the derived can create/copy/move this base
+  private:
+    friend Adapter;
```
figueroa1395, line 88: Why was clang complaining about this?

figueroa1395 (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
mgovers: Proposal in https://github.com/PowerGridModel/power-grid-model/pull/1113

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
mgovers: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/crtp-constructor-accessibility.html

it allows the constructor be called from a derived class that is different from the one provided in the template parameters, e.g.

```cpp
template <typename Derived>
class Base {
  protected:
    Base() = default;
    ~Base() = default;
    // etc.
};

struct A: public Base<B>{};  // OK by CRTP design
struct B: public Base<A>{};  // should be illegal but is allowed when protected
```

PR #1113, opened 2025-09-03
CI: update ci build tools

mgovers:
* [x] we should not install PGM when we only want to check the code quality.
  * Therefore, we also do not need GCC/G++
* [x] install GCC/G++ 14 in conda pipeline
* [x] install GCC/G++ 14 in readthedocs build

This solves the issues in https://github.com/PowerGridModel/power-grid-model/pull/1111#discussion_r2318459595

Relates to #1111

PR #1114, opened 2025-09-03
Clean-up main model: clang tidy bis

mgovers:
Re-applies the fixes as proposed in #1111 that were unable to be resolved because of the old GCC/G++ versions in the conda build and code quality checks now that #1113 has upgraded the GCC/G++ version in the conda build and removed the GCC/G++ dependency from the code quality build.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
@@ -261,11 +261,8 @@ struct YBusStructure {
         // end of y_bus_entry_indptr is same as size of entry
         assert(y_bus_entry_indptr.back() == static_cast<Idx>(y_bus_element.size()));
 
-        // construct transpose entry
-        lu_transpose_entry.resize(nnz_counter_lu);
-        // default transpose_entry[i] = i
-        std::iota( // NOLINT(modernize-use-ranges) // not yet supported by some compilers
-            lu_transpose_entry.begin(), lu_transpose_entry.end(), 0);
+        lu_transpose_entry = IdxRange{nnz_counter_lu} | std::ranges::to<IdxVector>();
+
         // fill off-diagonal, loop all the branches
         for (auto const [entry_1, entry_2] : off_diag_map) {
             // for each branch entry tf and ft, they are transpose to each other
--- a/tests/cpp_unit_tests/test_dataset.cpp
+++ b/tests/cpp_unit_tests/test_dataset.cpp
@@ -1067,14 +1067,11 @@ TEST_CASE_TEMPLATE("Test dataset (common)", DatasetType, ConstDataset, MutableDa
                 check_get_individual_scenario();
             }
             SUBCASE("columnar") {
-                auto a_id_buffer = std::vector<ID>(a_elements_per_scenario * batch_size);
-                auto a_a1_buffer = std::vector<double>(a_elements_per_scenario * batch_size);
+                auto a_id_buffer =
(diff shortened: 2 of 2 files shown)


.readthedocs.yaml
```diff
@@ -11,13 +11,15 @@ sphinx:
...
+    python: "3.13"
   apt_packages:
     - graphviz
+    - gcc-14
+    - g++-14
   jobs:
     post_install:
-      # Build package with doc requirements from pyproject.optional-dependencies
-      - pip install --verbose --editable .[doc]
+      # select compiler
```
mgovers:
```suggestion
      # Build package with doc requirements from pyproject.optional-dependencies
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
```diff
@@ -261,11 +261,8 @@ struct YBusStructure {
         // end of y_bus_entry_indptr is same as size of entry
         assert(y_bus_entry_indptr.back() == static_cast<Idx>(y_bus_element.size()));
 
-        // construct transpose entry
-        lu_transpose_entry.resize(nnz_counter_lu);
-        // default transpose_entry[i] = i
-        std::iota( // NOLINT(modernize-use-ranges) // not yet supported by some compilers
-            lu_transpose_entry.begin(), lu_transpose_entry.end(), 0);
+        lu_transpose_entry = IdxRange{nnz_counter_lu} | std::ranges::to<IdxVector>();
```
figueroa1395, line 264: I really like this, although it is still not straightforward for my brain at a glance 😆 .

tests/cpp_unit_tests/test_dataset.cpp
```diff
@@ -1067,15 +1067,12 @@ TEST_CASE_TEMPLATE("Test dataset (common)", DatasetType, ConstDataset, MutableDa
                 check_get_individual_scenario();
             }
             SUBCASE("columnar") {
-                auto a_id_buffer = std::vector<ID>(a_elements_per_scenario * batch_size);
-                auto a_a1_buffer = std::vector<double>(a_elements_per_scenario * batch_size);
+                auto a_id_buffer =
+                    std::views::iota(0, a_elements_per_scenario * batch_size) | std::ranges::to<std::vector<ID>>();
+                auto a_a1_buffer =
+                    std::views::iota(0, a_elements_per_scenario * batch_size) | std::ranges::to<std::vector<double>>();
```
figueroa1395, line 1073: Might be intentional, but double checking just in case:
`auto a = std::vector<T>(N)` creates a default initialized vector of type `T` and size `N`, right? However, `std::views::iota(0, N)` generates a lazy view of a range `[0, N)` which is then piped into a vector, so there is a slight difference, right? Is that intentional?

nitbharambe (approved):

tests/cpp_unit_tests/test_dataset.cpp
mgovers: yes because the following line used to do `std::iota(a_id_buffer.begin(), a_id_buffer.end(), 0);`, which is exactly what `std::views::iota(0, a_elements_per_scenario * batch_size) | std::ranges::transform(a_id_buffer.begin())` does.

but there's no point in default initializing and then overwriting if you can also construct in-place using `std::ranges::to`

figueroa1395 (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
mgovers: yeah it's left-to-right reading, which is standard for functional programming but not for old-school C/C++ 😬

PR #1118, opened 2025-09-08
Sparse matrix error: validation cases

figueroa1395:
This PR includes validation test cases that trigger a `SparseMatrixError` for a simple meshed grid.


TonyXiang8787:
@figueroa1395  why is this a bug? We get sparse matrix error because we do not have sufficient condition check for observability in meshed grids. So the pivot perturbation is not enabled.

figueroa1395:
> @figueroa1395 why is this a bug? We get sparse matrix error because we do not have sufficient condition check for observability in meshed grids. So the pivot perturbation is not enabled.

Even if we hack-force the perturbation from the observability, the pivot is still not triggered and we continue to have `SparseMatrixError`. This happens because the matrix's (y-bus) norm is smaller than the largest pivot on the sub-block (which is an edge case as the norm excludes the diagonal elements), triggered by a link being coupled directly to the source with a simple mesh after.

In summary, the pivot is too large to be considered for perturbation, but it is too small so it triggers `SparseMatrixError`; even if the fully observability test for meshed grids was completed, this simple grid would still present `SparseMatrixError`.

Let's have an offline discussion about this.

TonyXiang8787:
Some idea to fix this issue:

# Record the actual possible ill-conditioned rows in the observability check

Currently we only record if the whole matrix could be ill-conditioned if one of the rows is ill-conditioned:

https://github.com/PowerGridModel/power-grid-model/blob/7615d7ea960ce0beeb5ced2812986640d1a0bc5a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp#L83-L85

We can choose to record a list of rows which can be ill-conditioned.

# In LU solver, change the matrix norm to full norm including diagonal

We now exclude diagonal for calculating matrix norm. To avoid some rows are wrongly perturbed. But now we can use two step verification.

https://github.com/PowerGridModel/power-grid-model/blob/6c9605a3a7534e4f077addc2599958fff9ddc96d/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp#L539-L543

# Two step verification of pivot perturbation

In Pivot perturbation judgment here:

https://github.com/PowerGridModel/power-grid-model/blob/6c9605a3a7534e4f077addc2599958fff9ddc96d/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp#L99-L100

Only trigger this if the row is actually one of the rows which are determined to be possibly ill-conditioned.

PR #1120, opened 2025-09-09
Type dependent checks: Checks at all components

nitbharambe:
From the discussions around https://github.com/PowerGridModel/power-grid-model/pull/1109 , we bumped into the question about dependent types. Hence a separate PR for discussing/solving this issue.

I had initially thought of having this constraint on MainModelType. But what if we included this check at all_components.hpp only? (Edit: Moving back to MainModelType)


power_grid_model_c/power_grid_model/include/power_grid_model/all_components.hpp
```diff
@@ -27,9 +27,56 @@
 #include "component/voltage_sensor.hpp"
 
 namespace power_grid_model {
+namespace detail {
+template <typename List, typename T, typename... NeedsTypes>
+concept dependent_type_check = !IsInList<T, List>::value || (IsInList<NeedsTypes, List>::value && ...);
```
nitbharambe: I was checking a different condition previously. It now seems to be possible to move these checks to MainModelType. Ill do so now.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -33,12 +33,48 @@ constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...> /*unu
...
+constexpr void run_functor_with_tuple_index_return_void(Functor functor, std::index_sequence<Indices...>) {
+    (functor.template operator()<std::tuple_element_t<Indices, Tuple>>(), ...);
+}
+
+template <typename List, typename T, typename... NeedsTypes>
+concept dependent_type_check = !IsInList<T, List>::value || (IsInList<NeedsTypes, List>::value && ...);
+
+template <typename CompList>
+concept validate_component_types_c =
+    dependent_type_check<CompList, Source, Node> &&                  //
```
nitbharambe, line 41: // in the end to avoid formatting

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -33,12 +33,48 @@ constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...> /*unu
                                              std::tuple<std::type_identity<Types>>, std::tuple<>>{}...);
 }
 
+template <typename Tuple, class Functor, std::size_t... Indices>
+constexpr void run_functor_with_tuple_index_return_void(Functor functor, std::index_sequence<Indices...>) {
+    (functor.template operator()<std::tuple_element_t<Indices, Tuple>>(), ...);
+}
```
nitbharambe:
```suggestion
```
Didnt merge properly

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -33,12 +33,44 @@ constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...> /*unu
...
+    dependent_type_check<CompList, ThreeWindingTransformer, Node> && //
+    dependent_type_check<CompList, Shunt, Node> &&                   //
+    dependent_type_check<CompList, SymGenerator, Node> &&            //
+    dependent_type_check<CompList, AsymGenerator, Node> &&           //
+    dependent_type_check<CompList, SymLoad, Node> &&                 //
+    dependent_type_check<CompList, AsymLoad, Node> &&                //
+    dependent_type_check<CompList, SymVoltageSensor, Node> &&        //
+    dependent_type_check<CompList, AsymVoltageSensor, Node> &&       //
+    dependent_type_check<CompList, SymPowerSensor, Node, Line, AsymLine, Link, GenericBranch, Transformer,
+                         ThreeWindingTransformer, SymGenerator, AsymGenerator, SymLoad, AsymLoad> && //
```
figueroa1395: 1. Shouldn't `Link` not be present here? Don't we forbid sensors on links?
2. Also, shouldn't here we use a slightly different concept? Something like
```
template <typename List, typename T, typename... NeedsTypes>
concept dependent_type_check_or = !IsInList<T, List>::value || (IsInList<NeedsTypes, List>::value || ...);
```
Notice the `||` before `...` in the end contrary to the original `&&`. I ask because a `SymPowerSensor` only needs to be attached to either of those, right? And if we have a previous check of the dependencies of the individual "attachment" locations, that should be enough, no? Contrary to what is here, we shouldn't need all those components to have a `SymPowerSensor` in a `Node` for instance. Or am I understanding something incorrectly?
3. Question 2. also applies to `AsymPowerSensor` and `*CurrentSensor`s.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -33,12 +33,44 @@ constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...> /*unu
                                              std::tuple<std::type_identity<Types>>, std::tuple<>>{}...);
 }
 
+
+template <typename List, typename T, typename... NeedsTypes>
+concept dependent_type_check = !IsInList<T, List>::value || (IsInList<NeedsTypes, List>::value && ...);
```
figueroa1395, line 37: What is the use of `!IsInList<T, List>::value`? Is it for things like `Node`, `Branch`, `Branch3`, etc?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
nitbharambe: 1. Yes. Copy paste. Link needs to be removed.
2. The && is the requirement currently. We have switch cases on measured_terminal_type in main_core/topology.hpp. We should add a `if constexpr ...` for such places and then this requirement can be changed to ||. But this is scoped out of this PR.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
nitbharambe: If the said component is not in the list, then skip checking for the types it needs

mgovers:
> I had initially thought of having this constraint on MainModelType. But what if we included this check at all_components.hpp only? (Edit: Moving back to MainModelType)

`all_components` may not be a good location for that... what if we let every component for itself decide + forward declare the other required components?

e.g.:

```cpp
struct Node; // forward declare dependent types

class Line {
  public:
    using uses_component_types = std::tuple<Node>;
};
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
```diff
@@ -33,12 +33,43 @@ constexpr auto filter_tuple_types(std::tuple<std::type_identity<Types>...> /*unu
                                              std::tuple<std::type_identity<Types>>, std::tuple<>>{}...);
 }
 
+template <typename List, typename T, typename... NeedsTypes>
+concept dependent_type_check = !IsInList<T, List>::value || (IsInList<NeedsTypes, List>::value && ...);
+
+template <typename CompList>
+concept validate_component_types_c =
+    dependent_type_check<CompList, Source, Node> &&                  //
```
mgovers, line 41: all of these should be forward declared if possible

tests/cpp_unit_tests/test_main_model_type.cpp
```diff
@@ -17,6 +18,24 @@ struct AComponent {
 };
 } // namespace
 
+template <typename T>
+concept Constructible = requires { T{}; };
```
mgovers: is `std::constructible_from<>` or `std::constructible` not possible?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
figueroa1395: What would be the advantage of forward declaring them? Specially if the declaration and implementation live on the same `*.hpp` file?

tests/cpp_unit_tests/test_main_model_type.cpp
figueroa1395: Unfortunately not. In Clang's implementation `std::constructible_from` needs instantiation and even "raw" `std::is_constructible_v<T, Args...>` does as well.

However, after discussion with Nitish yesterday, this is redundant, as the `static_assert`s that check `detail::validate_component_types_c<*>` already achieve the same. Since we don't want to instantiate explicitly but rather check that the type combination is valid, testing the concepts is enough.

On the other hand, `std::constructible_from<>` can and should be tested for `MainModelImpl`, not `MainModelType`. That is a TODO.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/main_model_type.hpp
mgovers: i would like to make it such that if we create a `test_minimal_main_model.cpp` that only creates a main model with `Node`, `GenericBranch`, `Source`, `Load` then I want to be able to compile that without including any other components. Idem for the topology itself.

nitbharambe: I had tried removing dependency from all_components. Referring discussion in  https://github.com/PowerGridModel/power-grid-model/pull/1109#discussion_r2362144120

Most of the decoupling was possible. It was difficult to have a PR for without being a git headache. So to be continued in a new PR. Resolving.

mgovers: awesome

mgovers:
@nitbharambe does this need a benchmark run?

nitbharambe:
```cpp
class Line {
  public:
    using uses_component_types = std::tuple<Node>;
};
```
This feels like a good idea. Things would fall in their right place. What advantage can it bring though compared to simply forward declaring all the components here? We can explore this question in that next PR.

nitbharambe:
> @nitbharambe does this need a benchmark run?

I suppose not. Its simply adding constraint. So all compile time

tests/cpp_unit_tests/test_main_model_type.cpp
```diff
@@ -44,8 +55,6 @@ TEST_CASE("MainModelType") {
         utils::run_functor_with_tuple_return_void<typename ModelType::TopologyTypesTuple>(
             [&calls]<typename CompType>() { calls.push_back(std::string_view(CompType::name)); });
         CHECK(calls == std::vector<std::string_view>{"node", "source"});
-
-        // static_assert(is_constructible_v<MainModelImpl<ModelType>>);
```
figueroa1395, line 48: Is this to be tested/done in a separate PR?

mgovers:
> ```c++
> class Line {
>   public:
>     using uses_component_types = std::tuple<Node>;
> };
> ```
> 
> This feels like a good idea. Things would fall in their right place. What advantage can it bring though compared to simply forward declaring all the components here? We can explore this question in that next PR.

My preference goes to forward declaring. If that is not possible, this could also be a good path going forward. the most dangerous downside about deciding it on the per-component level is that we forget to update e.g. one of the sensor types when we add a new branch/branch3 component type. The only benefit is that you do not include what you do not need, but that would also be achieved by forward declaring.

> I don't see the value right now, but if you think is a good idea, let's give it a go in a follow PR, maybe we'll keep "unlocking" nice stuff :).

out of scope of this PR, definitely.

tests/cpp_unit_tests/test_main_model_type.cpp
nitbharambe: Yes, In a separate PR: https://github.com/PowerGridModel/power-grid-model/pull/1135

PR #1115, opened 2025-09-04
Clean-up main model: pass logger by reference to main model

mgovers:
This PR is an experiment to see whether removing ownership of the `Logger* log_` member variable of `MainModelImpl` makes sense (cfr. https://github.com/PowerGridModel/power-grid-model/pull/1107#discussion_r2321103874).

Relates to #1107 .

PR #1121, opened 2025-09-12
Fix nightly: RUF043

figueroa1395:
Nightly is failing due to `ruff`s rule RUF043 introduced in its last release. See https://github.com/PowerGridModel/power-grid-model/pull/1115#issuecomment-3284977965 for more details.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -154,7 +154,6 @@ select = [
 ]
 
 ignore = [
-    "UP038", # deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
     "PT019" # flags false positives
 ]
 
--- a/tests/unit/validation/test_rules.py
+++ b/tests/unit/validation/test_rules.py
@@ -673,7 +673,7 @@ def test_not_all_missing():
         assert len(errors) == 1
         assert errors == [MultiFieldValidationError("foo_test", ["foo", "bar", "baz"], [1])]
 
-        with pytest.raises(ValueError, match="The fields parameter must contain at least 2 fields.") as excinfo:
+        with pytest.raises(ValueError, match="The fields parameter must contain at least 2 fields") as excinfo:
             not_all_missing(invalid, ["bar"], "foo_test")
 
         assert excinfo.type is ValueError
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -154,7 +154,6 @@ select = [
 ]
 
 ignore = [
-    "UP038", # deprecated, https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation
```
figueroa1395, line 157: See https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#deprecation

tests/unit/validation/test_rules.py
```diff
@@ -673,7 +673,7 @@ def _mock_nan_type(component: ComponentType, field: str):
         assert len(errors) == 1
         assert errors == [MultiFieldValidationError("foo_test", ["foo", "bar", "baz"], [1])]
 
-        with pytest.raises(ValueError, match="The fields parameter must contain at least 2 fields.") as excinfo:
+        with pytest.raises(ValueError, match="The fields parameter must contain at least 2 fields") as excinfo:
```
mgovers, line 676: Here and below: we stil need to check for the period mark (although the original clearly was wrong)

```suggestion
        with pytest.raises(ValueError, match="The fields parameter must contain at least 2 fields\.") as excinfo:
```

figueroa1395: Why? The match is only "contains" not "equal to". This is seen across the many other test cases as well. We may want to homogenize. In any case, I can just make all of the ones here literal strings and add `\.`, that is no problem either.

mgovers: we should definitely homogenize. i remember that we had the C++ exceptions inconsistently using with and without period at the ending as well. to see the consistency, we need to check for the period mark (not even exact match for the entire exception; just that it is there for the substring)

that said, i'm also ok with doing that in a separate PR

figueroa1395: Resolving in favor of https://github.com/PowerGridModel/power-grid-model/pull/1121#discussion_r2347947166

PR #1122, opened 2025-09-12
Clean-up main model: Pass logger by reference to adapter

figueroa1395:
Passes the logger by reference to the adapter. It's a continuation of #1115

### DIFF
--- a/docs/advanced_documentation/build-guide.md
+++ b/docs/advanced_documentation/build-guide.md
@@ -27,7 +27,7 @@ In this repository there are three builds:
 
 To build the library from source, you need to first prepare the compiler toolchains and the build dependencies.
 In this section a list of general requirements are given.
-After this section there are examples of setup in Linux (Ubuntu 22.04), Windows 10, and macOS (Big Sur).
+After this section there are examples of setup in Linux (Ubuntu 24.04), Windows 11, and macOS (Sequoia).
 
 ### Architecture Support
 


tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -56,53 +57,37 @@ class JobAdapterMock : public JobInterface<JobAdapterMock> {
     JobAdapterMock& operator=(JobAdapterMock const& other) {
         if (this != &other) {
             counter_ = other.counter_;
-            logger_ = nullptr; // reset logger
```
figueroa1395, line 59: Removing logger related stuff on this test, made me realize that we should probably test the logger by itself, as it feels we are doing it here... Also, we may possibly, if we give this the yes at some point, want to rethink the logger counters here.

tests/benchmark_cpp/benchmark.cpp
```diff
@@ -10,6 +10,7 @@
 #include <power_grid_model/main_model.hpp>
 #include <power_grid_model/math_solver/math_solver.hpp>
 
+#include <iomanip>
```
figueroa1395, line 13: This was required to compile the benchmark.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -18,20 +19,31 @@ template <typename Adapter> class JobInterface {
...
     {
-        return static_cast<Adapter*>(this)->calculate_impl(result_data, pos);
+        static_cast<Adapter*>(this)->calculate_impl(result_data, pos, logger);
+    }
+    template <typename ResultDataset> void calculate(ResultDataset const& result_data, Idx pos = 0) {
+        static constexpr NoLogger logger{};
+        calculate(result_data, pos, logger);
+    }
+    template <typename ResultDataset> void calculate(ResultDataset const& result_data, Logger& logger) {
+        calculate(result_data, Idx{}, logger);
```
figueroa1395, line 29: Looking at this, I am not sure anymore if I like this step... However, seems the logger is anyways necessary for the entry point in job dispatch, these can potentially be reduced and simplified (removing the `NoLogger` overload). Same for `cache_calculate`.

figueroa1395:
This is very much experimental, but let's gather some thoughts :)

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
mgovers: what about adding the logger view we've been talking about? (capture by &, default NoLogger)

mgovers (commented):
i personally think it's a nice cleanup, but let's see what the rest thinks

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
figueroa1395: Is a way to go. If you and/or others think this is not so nice, we can go that route. If this is deemed fine, then we can keep the overloads. They are overloads at last :laughing

mgovers: i have no strong opinion. i actually prefer the explicit path (less entrypoints to handle) but ok with anything

figueroa1395: For now I will leave in the two explicit paths
```
    template <typename ResultDataset> void calculate(ResultDataset const& result_data, Idx pos = 0) {
        static constexpr NoLogger logger{};
        calculate(result_data, pos, logger);
    }
```
and
```
    void cache_calculate() {
        static constexpr NoLogger logger{};
        cache_calculate(logger);
    }
```
because they may come useful when we expose the logger to the python, C, Cpp APIs, and the user doesn't care about logging at all. However, they certainly could be removed in the current state without any impact, as we always manually pass `NoLogger` when we don't intend to log.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -18,20 +19,31 @@ template <typename Adapter> class JobInterface {
...
         requires requires(Adapter& adapter) { // NOSONAR
-            { adapter.calculate_impl(result_data, pos) } -> std::same_as<void>;
+            { adapter.calculate_impl(result_data, pos, logger) } -> std::same_as<void>;
         }
     {
-        return static_cast<Adapter*>(this)->calculate_impl(result_data, pos);
+        static_cast<Adapter*>(this)->calculate_impl(result_data, pos, logger);
+    }
+    template <typename ResultDataset> void calculate(ResultDataset const& result_data, Idx pos = 0) {
+        static constexpr NoLogger logger{};
```
nitbharambe:
```suggestion
        static constexpr NoLogger no_logger{};
```
And, we can also have only one no_logger object and use it evereywhere, right?

figueroa1395: I would prefer to keep the `static constexpr NoLogger no_logger{}` local to avoid potential thread safety issues and so on. Right now it shouldn't be a problem, but it may be in the future.

To be honest, I actually think I should remove this overload alltogether and always request a logger to be explicitly passed in, let it be a `NoLogger` or an actual one. If the user is not interested in logging, we can just have a check for that and create a `NoLogger` explicitly.

The more I look at this I dislike it more. I'll remove these overloads and let's see how does it look like.

figueroa1395: See [c4e66a8](https://github.com/PowerGridModel/power-grid-model/pull/1122/commits/c4e66a8c7928da1560497605c411eda8f7941ca7), this should be better.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
figueroa1395: Ended up removing the overload with no logger, I just didn't like that design choice. See [c4e66a8](https://github.com/PowerGridModel/power-grid-model/pull/1122/commits/c4e66a8c7928da1560497605c411eda8f7941ca7). Resolving for now.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -83,6 +72,8 @@ template <typename Adapter> class JobInterface {
     JobInterface(JobInterface&& /*other*/) noexcept = default;
     JobInterface& operator=(JobInterface&& /*other*/) noexcept = default;
     ~JobInterface() = default;
+
+    using NoLogger = power_grid_model::common::logging::NoLogger;
```
nitbharambe: This got left out

figueroa1395: Oops. Thanks.

figueroa1395:
`do-not-merge` label added so we don't merge on Friday. To be merged next Monday.

figueroa1395:
Happy merging Monday!

PR #1124, opened 2025-09-15
Sonar Fail: Try to change args of compile commands

nitbharambe:

- Try changing as per readme https://github.com/SonarSource/sonarqube-scan-action?tab=readme-ov-file

### Checks

- [ ] If changes are related to CI/CD, are they verified via a manual run on this branch?

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -79,4 +79,4 @@ jobs:
         uses: SonarSource/sonarqube-scan-action@v5
         with:
           args: >
-            --define sonar.cfamily.compile-commands="${{ env.BUILD_WRAPPER_OUT_DIR }}/compile_commands.json"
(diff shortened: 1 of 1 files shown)


nitbharambe:
Experiment failed.  Closing PR :D

.github/workflows/sonar.yml
```diff
@@ -79,4 +79,4 @@ jobs:
         uses: SonarSource/sonarqube-scan-action@v5
         with:
           args: >
-            --define sonar.cfamily.compile-commands="${{ env.BUILD_WRAPPER_OUT_DIR }}/compile_commands.json"
+            --define sonar.cfamily.build-wrapper-output="${{ env.BUILD_WRAPPER_OUT_DIR }}"
```
nitbharambe, line 82: Are these extra " quotes the issue?

figueroa1395: I have the naive feeling they are, but the problem is that they shouldn't be 😆

Issue #1085, opened 2025-08-15
[FEATURE] Full observability check without voltage sensor with phasor


Jerry-Jinfeng-Guo:
# [FEATURE] Full observability check for **meshed** networks  (no voltage phasor sensors)

**Related:** #84, #838, #914, #508 

---

## Background & Motivation

Power Grid Model recently introduced/extended observability logic for **radial** grids and added recognition of **voltage phasor sensors** in the observability check pipeline. This makes non‑observable cases fail consistently with `NotObservableError`, instead of surfacing numerical factorization errors (e.g., `SparseMatrixError`). Extending a **full** observability check to **meshed networks** is the next step to give clear, deterministic feedback before state estimation runs, independent of the chosen algorithm (ILSE/NRSE).

This feature request proposes a **graph‑based, measurement‑aware** algorithm for **meshed** topologies driven primarily by **edge measurements** (branch current/power sensors) with optional node measurements, aligning with the sensor semantics already in PGM (e.g., current sensor as magnitude, edge vs. node power). Importantly, the algorithm excludes the presence of **voltage phasor sensors**, the addition of which will be one of the next steps.

> **Precondition (policy):** As with earlier design notes, we maintain the rule that **at least one voltage sensor without phasor** (i.e., magnitude‑only) must exist somewhere in the network. This provides a minimal reference independent of the presence/absence of voltage phasors. (This issue enforces/validates that precondition in the meshed‑grid check as well.)

---

## Scope

Implement a **full observability** checker for **meshed networks** that:

- Treats **edge measurements** (current/power sensors on branches) as first‑class constraints for spanning coverage over loops as well as trees.  
- Incorporates **node measurements** (net injections; optional) and **voltage magnitude** sensors for anchoring as needed.  
- **Does not rely on voltage phasor sensors** being present. If present in a **meshed** network, the observability check process should abort and pass to calculation.  
- Produces **deterministic, user‑facing results** (`Observable` vs. `NotObservableError`) prior to running the estimator, consistent with recent radial‑grid behavior.

---

## Proposed Algorithm (from workout)

The workout in **Full Observability Check in Meshed Networks with edge measurement.pdf** specifies a three‑part procedure: (1) build a **connectivity/measurement matrix** `m`, (2) run a **observability traversal** over the measurement graph, and (3) decide **full coverage** based on exhausted measurements and visited nodes.

### Key Steps:

1. **Build connectivity/measurement matrix**  
[...]

Issue #35, opened 2022-04-05
[FEATURE] merge nodes connected by link into one bus in mathematical model


TonyXiang8787:
Currently the link is modeled as a small impedance. This can cause numerical issues.

It is proposed to merge the nodes connected by link into one single bus. 

**Acceptance criteria**
- [ ] Nodes in topology that are connected by links are merged automatically.
- [ ] To be discussed: do we still calculate the power flow in links?
- [ ] Results extrapolated to original nodes
- [ ] Unit tests
- [ ] Validation tests [#1507]

Full release is blocked by #1418 [DONE]

scud-soptim:
Good morning, does that mean you calculate the current of the link via KCL?”

TonyXiang8787:
Hi @scud-soptim,

Yes, we are experimenting the KCL principles for the post-processing method to distribute the current/power flow for links.

figueroa1395:
The super-node post processing algorithm was implemented in https://github.com/PowerGridModel/power-grid-model/pull/1326, https://github.com/PowerGridModel/power-grid-model/pull/1335, https://github.com/PowerGridModel/power-grid-model/pull/1344, https://github.com/PowerGridModel/power-grid-model/pull/1359, and https://github.com/PowerGridModel/power-grid-model/pull/1370.

This is just one of the multiple steps involving this implementation.

PR #1125, opened 2025-09-16
Trigger pivot perturbation based on each possibly ill-conditioned one

figueroa1395:
This PR aims to improve the pivot perturbation scheme currently used, based on a finer grained criteria.

Before, we would judge if the whole matrix was ill-conditioned (prior to performing the LU decomposition), whereas with this PR we aim to judge the ill-condition on a per pivot basis. Furthermore, because of this we had to use the off diagonal infinite norm of the matrix to judge the pivot's size in order to avoid introducing fake observability, but now, we would be able to use the full infinite norm without having that issue.

In addition, with this PR we aim to resolve the edge cases reported in #1118 (to be solved only once the full observability check is in place, see https://github.com/PowerGridModel/power-grid-model/issues/1085), as well as make pivot perturbation more robust for radial grids. The scheme used is described in more detail in https://github.com/PowerGridModel/power-grid-model/pull/1118#issuecomment-3290768879.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
@@ -24,9 +24,9 @@ constexpr double epsilon_sqrt = 1.49011745e-8;      // sqrt(epsilon), sqrt does 
 // it will get modified if perturbation happens
 // also has_pivot_perturbation will be updated if perturbation happens
 template <scalar_value Scalar>
-inline void perturb_pivot_if_needed(double perturb_threshold, Scalar& value, double& abs_value,
+inline void perturb_pivot_if_needed(int8_t pertub_pivot, double perturb_threshold, Scalar& value, double& abs_value,
                                     bool& has_pivot_perturbation) {
-    if (abs_value < perturb_threshold) {
+    if (pertub_pivot == 1 && abs_value < perturb_threshold) {
         Scalar const scale = (abs_value == 0.0) ? Scalar{1.0} : (value / abs_value);
         value = scale * perturb_threshold;
         has_pivot_perturbation = true;
@@ -60,8 +60,9 @@ template <rk2_tensor Matrix> class DenseLUFactor {
     // put pivot perturbation in has_pivot_perturbation in place
     template <class Derived>
     static void factorize_block_in_place(Eigen::MatrixBase<Derived>&& matrix, BlockPerm& block_perm,
-                                         double perturb_threshold, bool use_pivot_perturbation,
-                                         bool& has_pivot_perturbation)
+                                         double perturb_threshold,
+                                         std::pair<bool, std::vector<int8_t>> possibly_ill_conditioned_pivots,
+                                         Idx block_node_idx, bool& has_pivot_perturbation)
         requires(std::same_as<typename Derived::Scalar, Scalar> && rk2_tensor<Derived> &&
                  (Derived::RowsAtCompileTime == size) && (Derived::ColsAtCompileTime == size))
     {
@@ -84,7 +85,7 @@ template <rk2_tensor Matrix> class DenseLUFactor {
             assert(col_biggest_eigen + pivot < size);
 
             // check absolute singular matrix
-            if (biggest_score == 0.0 && !use_pivot_perturbation) {
+            if (biggest_score == 0.0 && !possibly_ill_conditioned_pivots.first) {
                 // pivot perturbation not possible, cannot proceed
                 // set identity permutation and break the loop
                 for (int8_t remaining_rows_cols = pivot; remaining_rows_cols != size; ++remaining_rows_cols) {
@@ -96,8 +97,10 @@ template <rk2_tensor Matrix> class DenseLUFactor {
 
             // perturb pivot if needed
             double abs_pivot = sqrt(biggest_score);
-            perturb_pivot_if_needed(perturb_threshold, matrix(row_biggest, col_biggest), abs_pivot,
-                                    has_pivot_perturbation);
+            int8_t const possibly_ill_conditioned_pivot =
+                possibly_ill_conditioned_pivots.first ? possibly_ill_conditioned_pivots.second[block_node_idx] : 0;
+            perturb_pivot_if_needed(possibly_ill_conditioned_pivot, perturb_threshold, matrix(row_biggest, col_biggest),
+                                    abs_pivot, has_pivot_perturbation);
             max_pivot = std::max(max_pivot, abs_pivot);
 
             // swap rows and columns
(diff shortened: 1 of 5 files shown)


figueroa1395:
Converting back to draft as the validation cases created in #1118 are still unresolved if I force in the perturbation... Several things happen:
1. The ill-conditioned check is not strong enough, as the validation cases in #1118 have enough sensors. In order for this to work after meshed observability is implemented, we would first need to come up with a stronger ill-condition.
1. Even if I manually mark the pivots as ill-conditioned, they aren't considered small enough according to the current threshold (related to the matrix norm), so the pivot perturbation is not triggered still. This raises questions and is tied to the following point.
1. Even if the pivot is not considered small, we are still triggering `SparseMatrixError` after Gaussian elimination. This other threshold is based on the maximum pivot present in the matrix after the transformations are performed. The problem here is that we end up indirectly comparing two distinct things: the norm of the original matrix (used to determine if a pivot perturbation is to be triggered) with the max pivot of the matrix after several transformations have been done.
This is problematic because even though the norm of the matrix (including the diagonal) should always be bigger than the largest pivot for the original matrix, we are now comparing it with the max pivot of a transformed matrix for which the pivots can become a lot larger than the original norm. This is the edge case we are currently running into. So we end up doing a comparison that isn't fair.

To summarize, this PR still has value, as it improves how we do perturbation (more fine grained and the whole matrix norm is now taken into account without generating fake observability), but it doesn't resolve (at least yet) the edge cases we are running into. I don't have a concrete proposal on how to go forward, so thoughts are welcome, but I will try some experimentation with how we determine the max pivot (maybe calculating it from the original matrix makes more sense, but it still doesn't seem satisfactory) and comment back here if I find any relevant results.

figueroa1395:
~~[c1fa814](https://github.com/PowerGridModel/power-grid-model/pull/1125/commits/c1fa814ce36dd8868e8a85c5d9688433991a1b63) and [a9f0158](https://github.com/PowerGridModel/power-grid-model/pull/1125/commits/a9f0158e14dd6f6abbf1ae439865a4c098bb4e50) is a quick and dirty implementation of an idea I had to address this issue. This idea is actually independent of pivot perturbation - hence it is independent of the meshed-ness of the grid and the observability check.~~

~~In short, I replaced the `max_pivot` (previously calculated after gaussian elimination has taken place), with one that is calculated based on the original matrix. Now the validation tests cases seem to pass locally (but other test cases based on big `sk` are now broken). Leaving here to collect thoughts in case it is worth to keep investigating.~~

Edit: This doesn't work for ill-conditioned cases generated by the opposite compared to the ones in the validation tests, i.e. for cases in which the source short circuit voltage is extremely high (very large `sk`). This idea is scratched.

figueroa1395:
The final idea from this exploration is applied in [5d379a9](https://github.com/PowerGridModel/power-grid-model/pull/1125/commits/5d379a908e71eb30d08ecb0f0419d96a93abf23e). In short, the threshold to raise `SparseMatrixError` when the matrix is ill-conditioned has been split into two: relative and absolute thresholds.

- Relative threshold: It replaces the original, `epsilon * max_pivot` where `max_pivot` is given by the max pivot in the dense block after gaussian elimination. The new threshold is given by `epsilon * matrix_norm` where `matrix_norm` is calculated based in the whole original matrix (before any factorization or gaussian decomposition takes place). I will give a somewhat heuristic explanation for why I think this strategy makes sense.

The condition number $\kappa(\mathbf{A})$ of a matrix $\mathbf{A}$ is defined as 
```math
\kappa(\mathbf{A}) = \|\mathbf{A}\| \times \|\mathbf{A^{-1}}\| .
```
 Considering we are interested in the LU decomposition of the matrix, we can re-write the above as
```math
\kappa(\mathbf{A}) = \|\mathbf{A}\| \times \|\mathbf{U^{-1}} \times \mathbf{L^{-1}}\| \leq \|\mathbf{A}\| \times \|\mathbf{U^{-1}} \| \times \| \mathbf{L^{-1}}\| \approx \|\mathbf{A}\| \times \|\mathbf{U^{-1}} \| ,
```
as $\mathbf{L^{-1}}$  has 1's in the diagonal and is usually well behaved. Considering now 
```math
\|\mathbf{U^{-1}}\| \sim 1 / \text{min pivot} ,
```
as that is the element that affects the norm the most, we can approximate the condition number by
```math
\kappa(\mathbf{A}) \lesssim \|\mathbf{A}\| \times 1 / \text{min pivot} ,
```
and for the matrix to not be ill-conditioned, it needs to satisfy
```math
\kappa(\mathbf{A}) <  1 / \epsilon ,
```
which translates into the introduced relative threshold check
```math
\text{min pivot} < \|\mathbf{A}\| \times \epsilon .
```
This works with the norm of the original matrix two ways. The condition number is defined over the original matrix, as this is the original problem we are trying to solve, and we homogenize the perturbation threshold with this one. This also takes care of both extreme ill-conditioned cases, when `sk` is very large or when we have links.

- Absolute threshold: This one just checks against machine epsilon. This is by no means needed, but I still thought it is worth it. Furthermore, all tests pass with this one removed.

With the introduction of the above, the introduced ill condition validation cases pass, while the old ones do as well.

Note 1: This still makes use of the (not so strong) "topological" check of possibly ill conditioned pivots during the observability check, as the full norm of the matrix (including its diagonal) is to be used. This shouldn't be problem once meshed observability is introduced, as it can be made into a standalone routine.

Note 2: I will cleanup the PR only after input is received, as this is very much experimental from my side.

TonyXiang8787:
@figueroa1395 if you use relative threshold to raise `SparseMatrixError`, does it mean the matrix norm has to be always calculated? Even for power flow equations where you do not foresee ill-conditioned pivots at all?

figueroa1395:
> @figueroa1395 if you use relative threshold to raise `SparseMatrixError`, does it mean the matrix norm has to be always calculated? Even for power flow equations where you do not foresee ill-conditioned pivots at all?

@TonyXiang8787 The norm is only calculated with `initialize_pivot_perturbation()`, which is only called when there are possibly ill conditioned pivots as determined (on a per pivot basis) at the observability level, hence the norm is only relevant for state estimation.

In short, no. Power flow calculations should not be affected.

Note: I just manually checked, and the power flow validation cases never trigger it.

Jerry-Jinfeng-Guo:
`do-not-merge` added due to the experimental nature of the PR.

figueroa1395:
Outcome of offline meeting:
- The idea is sound, so we now should implement it properly and clean up the PR.
- Try to implement this in such a way that power flow does not get a performance hit with the calculation of the norm.
- Benchmark state estimation but specially power flow.
- Compare/simulate the values of the `max_pivot` vs `matrix_norm` to determine a possible combined or intermediate check.
- Test thoroughly internally.

After the above are finished, more input is to be requested as this is potentially a big change that may (not) have a big performance impact.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -539,19 +549,14 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
         // calculate the block-wise non-diagonal infinite norm of the matrix
```
mgovers:
```suggestion
        // calculate the block-wise infinite norm of the matrix
```

mgovers:
in accordance to the decisions based on the investigation in https://github.com/PowerGridModel/power-grid-model/pull/1125 and its follow-ups based off #1167 , we decided to abandon this path in favor of #35 

The cleanup in #1167 will also be closed.

PR #1127, opened 2025-09-19
Sonar cloud issues: resolve reliability issues + several other things

mgovers:
Now that sonar cloud finally works again, we see that we unfortunately introduced some new warnings. This PR fixes some of them and also gets rid of some other long outstanding issues

Removes:

* [x] reliability issue:
  * [x] destructor not marked `noexcept` 
  * [x] unused `assert` when `enumerated_zip_sequence` is used without additional arguments
* [x] remove unreadable `static_cast<IntS>` in favor of
  * [x] `std::underlying` for enums
  * [x] `int_status` for boolean statuses
* [x] `Timer` and `ThreadLogger` copy/move constructors participate in resource management
* [x] `if consteval` over `if (std::is_constant_evaluated())` because the latter is a runtime check rather than compiletime


power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
```diff
@@ -53,7 +54,11 @@ class MissingCaseForEnumError : public InvalidArguments {
   public:
     template <typename T>
     MissingCaseForEnumError(std::string_view method, T const& value)
-        : InvalidArguments{method, std::format("{} #{:d}", typeid(T).name(), static_cast<IntS>(value))} {}
+        : InvalidArguments{method, std::format("{} #{:d}", typeid(T).name(), value)} {}
+    template <typename T>
+        requires std::is_enum_v<T>
+    MissingCaseForEnumError(std::string_view method, T const& value)
+        : InvalidArguments{method, std::format("{} #{:d}", typeid(T).name(), std::to_underlying(value))} {}
```
figueroa1395: We don't need the non-concept "overload" above, right?

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
```diff
@@ -266,10 +266,10 @@ static_assert(grouped_idx_vector_type<DenseGroupedIdxVector>);
 
 inline auto enumerated_zip_sequence(grouped_idx_vector_type auto const& first,
                                     grouped_idx_vector_type auto const&... rest) {
-    assert(((first.size() == rest.size()) && ...));
-
-    auto const indices = IdxRange{first.size()};
-    return std::views::zip(indices, first, rest...);
+    if constexpr (sizeof...(rest) > 0) {
```
figueroa1395, line 269: Why was this `constexpr` introduced?

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -20,11 +20,26 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
...
+            return *this;
+        };
+        ThreadLogger(ThreadLogger&& other) noexcept
+            : LoggerType{std::forward<LoggerType>(other)}, parent_{other.parent_} {};
+        ThreadLogger& operator=(ThreadLogger&& other) noexcept {
+            if (this != &other) {
+                auto* const parent = other.parent_;
+                other.parent_ = nullptr;
+                LoggerType::operator=(std::move(other));
+                parent_ = parent;
```
figueroa1395: Why not directly `parent_ = other.parent_;`? and skip the intermediate `auto* const parent = other.parent_` step?

power_grid_model_c/power_grid_model/include/power_grid_model/component/base.hpp
```diff
@@ -28,7 +29,7 @@ class Base {
     virtual ~Base() = default;
     constexpr ID id() const noexcept { return id_; }
     constexpr BaseOutput base_output(bool is_energized) const {
-        return BaseOutput{.id = id_, .energized = static_cast<IntS>(is_energized)};
+        return BaseOutput{.id = id_, .energized = int_status(is_energized)};
```
figueroa1395: I'm not sure about this one... Since `IntS` is a simple `int` type, the static cast feels very expressive already. Also, it may even be more readable that `int_status` (perhaps some other name). The additional function call is probably inlined anyway, so that is no concern, but this choice is probably subjective. Just wanted to raise it anyway.

TonyXiang8787:
It seems to be a huge change with many stuff. Can we do it one category at a time?

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
mgovers: unneeded assert when rest is empty (expands to `assert()`)

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: initialization order v.s. use moved value

mgovers:
> It seems to be a huge change with many stuff. Can we do it one category at a time?

i will cherry-pick

power_grid_model_c/power_grid_model/include/power_grid_model/component/base.hpp
mgovers: cfr. offline discussion: changed to `status_to_int` + deferred to https://github.com/PowerGridModel/power-grid-model/pull/1128

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: deferred to #1129

power_grid_model_c/power_grid_model/include/power_grid_model/common/exception.hpp
mgovers: deferred + resolved in #1128

mgovers: we actually need the overload because the C API uses raw `Idx` rather than enums

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -24,7 +24,7 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
         ThreadLogger& operator=(ThreadLogger const&) = default;
         ThreadLogger(ThreadLogger&&) noexcept = default;
         ThreadLogger& operator=(ThreadLogger&&) noexcept = default;
-        ~ThreadLogger() override { sync(); }
+        ~ThreadLogger() noexcept override { sync(); }
```
mgovers, line 27: @figueroa1395 this apparently did nog solve the issue. We need an explicit try/catch block here. Can you add one?

PR #1130, opened 2025-09-19
Sonar cloud issues: destructors must be noexcept

mgovers:
Attempt 2 to resolve the reliability issue addressed in #1127

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
@@ -27,8 +27,8 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
         ~ThreadLogger() noexcept override {
             try {
                 sync();
-            } catch (...) {
-               // NOSONAR // we can't sync so we need to drop
+            } catch (...) { // NOLINT(bugprone-empty-catch) // NOSONAR
+                // we can't sync so we need to ignore the error
             }
         }
         void sync() const { parent_->sync(*this); }


power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -24,7 +24,13 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
         ThreadLogger& operator=(ThreadLogger const&) = default;
         ThreadLogger(ThreadLogger&&) noexcept = default;
         ThreadLogger& operator=(ThreadLogger&&) noexcept = default;
-        ~ThreadLogger() noexcept override { sync(); }
+        ~ThreadLogger() noexcept override {
+            try {
+                sync();
+            } catch (...) {
+                // we can't sync so we need to drop
```
mgovers:
```suggestion
               // NOSONAR // we can't sync so we need to drop
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -24,7 +24,13 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
         ThreadLogger& operator=(ThreadLogger const&) = default;
         ThreadLogger(ThreadLogger&&) noexcept = default;
         ThreadLogger& operator=(ThreadLogger&&) noexcept = default;
-        ~ThreadLogger() noexcept override { sync(); }
+        ~ThreadLogger() noexcept override {
+            try {
+                sync();
+            } catch (...) { // NOLINT(bugprone-empty-catch) // NOSONAR
+                // we can't sync so we need to ignore the error
+            }
```
TonyXiang8787, line 32: I do not understand the problem here? Why the original code get complained by Sonar?

mgovers: Sync contains a `lock{mutex}` which may (implementation and architecture defined) throw if OS raises an error. No idea why but the regular compilers do not catch this but sonar does (correctly)

figueroa1395 (approved):

PR #1133, opened 2025-09-22
Fix `brew install` on linux

figueroa1395:
Fixes the recent issues where `brew install` was having problems in `linux` build, `clang-tidy` build and `sonar-qube`.

Current `ubuntu-latest` from GitHub may ship with a `curl` version that is not up to date, which was generating issues. Nevertheless, `homebrew` ships with a more up to date version of `curl` which they test against. Hence, with the flag `HOMEBREW_FORCE_BREWED_CURL: 1` we can force `homebrew`'s own `curl` version which should solve most issues; this is the fix implemented here.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -45,6 +45,7 @@ jobs:
       CMAKE_PREFIX_PATH: /home/linuxbrew/.linuxbrew
       PRESET: ci-clang-tidy-${{ matrix.build-option }}
       TARGET: nonexistent  # will be overwritten later in this action
+      HOMEBREW_FORCE_BREWED_CURL: 1
 
     steps:
       - uses: actions/checkout@v5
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test-release.yml
```diff
@@ -74,6 +74,7 @@ jobs:
     env:
       CMAKE_PREFIX_PATH: /home/linuxbrew/.linuxbrew
       PRESET: ci-${{ matrix.compiler }}-${{ matrix.build-option }}
+      HOMEBREW_FORCE_BREWED_CURL: 1
```
figueroa1395, line 77: This is the only relevant line. The rest is uncommenting for testing purposes.

mgovers: please cherry-pick only this line (e.g. using `git rebase -i` and `git push --force` once you're sure it's OK) so that the temporary changes don't end up in the commit history

figueroa1395: done

PR #1119, opened 2025-09-09
Cleanup main model: Clean up templates and variadics from y bus

nitbharambe:
**Fixes issue**: Continuation of https://github.com/PowerGridModel/power-grid-model/pull/1109

### Changes proposed in this PR include:

Change templates to use MainModelType. Moved to a new PR as this includes logic change too.
Splitting the function does not seem less performant but increases readability.

PR #1135, opened 2025-09-23
Cleanup main model: Finish migrating to main model type

nitbharambe:
Continuation from https://github.com/PowerGridModel/power-grid-model/pull/1109, https://github.com/PowerGridModel/power-grid-model/pull/1119 and https://github.com/PowerGridModel/power-grid-model/pull/1120

* Final steps in moving to MainModelType. Now MainModelImpl is instantiated via MainModelType and constrained to only that.
* Change of JobInterface CRTP to deducing this.
* Moved JobDispatch::sequential to core_utils.hpp and ExtraRetrievableTypes from container.hpp to component_list.hpp
* Final cleanup and clearing of TODOs

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
@@ -26,6 +26,11 @@ class JobInterface {
         return std::forward<Self>(self).calculate_impl(result_data, pos, logger);
     }
 
+    template <typename Self, typename ResultDataset>
+    void calculate(this Self&& self, ResultDataset const& result_data, Logger& logger) {
+        std::forward<Self>(self).calculate(result_data, Idx{}, logger);
+    }
+
     template <typename Self>
     void cache_calculate(this Self&& self, Logger& logger)
         requires requires { // NOSONAR


power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -14,75 +14,71 @@
...
   public:
     // the multiple  NOSONARs are used to avoid the complaints about the unnamed concepts
-    template <typename ResultDataset>
-    void calculate(ResultDataset const& result_data, Idx pos = 0)
-        requires requires(Adapter& adapter) { // NOSONAR
-            { adapter.calculate_impl(result_data, pos) } -> std::same_as<void>;
+    template <typename Self, typename ResultDataset>
+    void calculate(this Self&& self, ResultDataset const& result_data, Idx pos = 0)
+        requires requires(this Self&& self) { // NOSONAR
+            { self.calculate_impl(result_data, pos) } -> std::same_as<void>;
```
mgovers: Here and below: Either use perfect forwarding here or just `Self` by `const&`

My guess is that `cons&` is sufficient.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -14,75 +14,71 @@
...
 namespace power_grid_model {
-template <typename Adapter> class JobInterface {
+class JobInterface {
   public:
     // the multiple  NOSONARs are used to avoid the complaints about the unnamed concepts
-    template <typename ResultDataset>
-    void calculate(ResultDataset const& result_data, Idx pos = 0)
-        requires requires(Adapter& adapter) { // NOSONAR
-            { adapter.calculate_impl(result_data, pos) } -> std::same_as<void>;
+    template <typename Self, typename ResultDataset>
```
mgovers, line 20: Should we add a requires clause or guard against missing implementations?

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: It was a bit confusing as to how this type gets deduced. 
I see this: https://en.cppreference.com/w/cpp/language/crtp.html. They used auto&& and then `self.` directly.

Either way: Changed to only & / const& which is sufficient.

nitbharambe:
> Beware: `deducing this` is a relatively recently implemented C++23 feature. Please double-check that the compiler versions referenced in our documentation support it.

You are right: Xcode's website says supported but cppreference does not. Our compiler versions should support it.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -14,75 +14,71 @@
...
-        return static_cast<Adapter*>(this)->calculate_impl(result_data, pos);
+        return self.calculate_impl(result_data, pos);
     }
 
-    void cache_calculate()
-        requires requires(Adapter& adapter) { // NOSONAR
-            { adapter.cache_calculate_impl() } -> std::same_as<void>;
+    template <typename Self>
+    void cache_calculate(this Self const& self)
+        requires requires(this Self const& self) { // NOSONAR
```
nitbharambe: We have build fails though because "explicit object parameter (aka this) in requires" is not allowed. Need to think a bit more on how to tackle this. Do you know @mgovers ?

mgovers: you can directly do
```cpp
        requires requires(Self const& self) {
```
the `this Self` is used for overload resolution as if it were a constraint in itself. you do not need it in additional constraints.

Alternatively, you can use `decltype(self)` or even directly `()` as `self` as a type is already fully determined by the time the requires clause is checked, because it comes after the function declaration `void cache_calculate(this Self const& self)`

e.g.

```cpp
template <typename Self>
    requires requires(Self const& s) { // here, Self is a type that is constrained,
                                       // but no constraint on `this` is added yet,
                                       // and `self` hasn't been declared yet
        { s.foo() } -> std::same_as<FooResult>; // you need to use `s` instead of `self`
        { s.bar() }; // no output constraint
    }
void cache_calculate(this Self const& self) // now, a `this` constraint is added;
                                            // `self` is fully declared
    requires requires { // constraints on `self`
        { self.baz() } -> std::same_as<BazResult>; // you can use `self` here because it is declared
        { self.bla() };
    }
{
    // body of function
}
```

in this case:

```cpp
        requires requires {
```

mgovers:
> > Beware: `deducing this` is a relatively recently implemented C++23 feature. Please double-check that the compiler versions referenced in our documentation support it.
> 
> You are right: Xcode's website says supported but cppreference does not. Our compiler versions should support it.

image

we will need to bump the build guide for macos

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: requires on each function are already present. So they should cover it right? Or you mean something more?

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
mgovers: if you need additional `requires` based on the **type** (i.e., `Self`), then using `template <typename Self> auto foo(this Self&& self);` is better,

if you do not need additional `requires` on the **type** (but only on the constraints), then you can just use `auto&&`, e.g., `auto foo(this auto&& self);`:

```cpp
auto foo_ref(this auto& self) { self.foo_impl(); }
auto foo_cref(this auto const& self) { self.foo_impl(); }
auto foo_fwd(this auto&& self) { std::forward<decltype(self)>(self).foo_impl(); }
```

or, with constraints
```cpp
auto foo_ref(this auto& self) 
    requires requires {
        { self.foo_impl() };
    }
{
    self.foo_impl();
}
auto foo_cref(this auto const& self) 
    requires requires {
        { self.foo_impl() };
    }
{
    self.foo_impl();
}
auto foo_fwd(this auto&& self)
    requires requires {
        { self.foo_impl() };
    }
{
    std::forward<decltype(self)>(self)();
}
```

the main reason to use perfect forwarding is if there may be a reason to invoke the function directly on an object after which the object can disappear. The way to think about it is: "would it make sense to call this on a moved object?" If so, then use forwarding. If not, then use `const&` or `&`

figueroa1395:
> we will need to bump the build guide for macos

Without having dug into the PR yet, why? Clang 17 already [supports ](https://developer.apple.com/documentation/xcode-release-notes/xcode-16_3-release-notes)`deducing this`

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: Oh I see it. That extra `this` in `requires` is redundant. as the function parameter does the deduction.

tests/cpp_unit_tests/test_main_model_type.cpp
```diff
@@ -5,19 +5,33 @@
...
+    // This dependency exists because of main_core/update.hpp
+    using UpdateType = AType;
     static constexpr char const* name = "a_component";
 };
+
 } // namespace
 
+static_assert(is_main_model_type_v<MainModelType<AllExtraRetrievableTypes, AllComponents>>);
+static_assert(!is_main_model_type_v<MainModelType<AllComponents, AllComponents>>);
+static_assert(!is_main_model_type_v<MainModelType<AType, AllComponents>>);
```
figueroa1395, line 30: The `concept`s in place for `MainModelType` should already take care of `!std:constructible_from<...>` for these non constructible types when trying them on `MainModelImpl`, right? Or is there any value in testing for some not constructible combinations for `MainModelImpl`?

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: This fixed the issue. Resolving

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
figueroa1395: The `this` parameter should already take care of it, together with the `requieres` for the implementation should be enough.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -14,75 +14,71 @@
...
-            { adapter.calculate_impl(result_data, pos) } -> std::same_as<void>;
+    template <typename Self, typename ResultDataset>
+    void calculate(this Self const& self, ResultDataset const& result_data, Idx pos = 0)
+        requires requires(Self const& self) { // NOSONAR
+            { self.calculate_impl(result_data, pos) } -> std::same_as<void>;
         }
     {
-        return static_cast<Adapter*>(this)->calculate_impl(result_data, pos);
+        return self.calculate_impl(result_data, pos);
     }
```
figueroa1395, line 27: I believe here and in the rest of the functions using this pattern, they can be simplified by removing the `Self const& self` inside the inner `requires`, as the parameter list already provides it and this is just shadowing it.
```suggestion
    template <typename Self, typename ResultDataset>
    void calculate(this Self const& self, ResultDataset const& result_data, Idx pos = 0)
        requires requires { // NOSONAR
            { self.calculate_impl(result_data, pos) } -> std::same_as<void>;
        }
    {
        return self.calculate_impl(result_data, pos);
    }
```

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
figueroa1395: Wouldn't doing something like the following cover all cases?
```
template<typename Self>
auto foo(this Self&& self, ...)  // I'm really abusing ... notation here
    requires requires {  
        { std::forward<Self>(self).foo_impl(...) };  
    }  
{  
    std::forward<Self>(self).foo_impl(...);  
} 
```

I also think using explicit template feels more readable.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -14,75 +14,71 @@
...
+            { self.set_logger_impl(log) } -> std::same_as<void>;
         }
     {
-        static_cast<Adapter*>(this)->set_logger_impl(log);
+        self.set_logger_impl(log);
     }
-
-  private:
-    friend Adapter;
-    JobInterface() = default;
```
figueroa1395, line 68: I would argue to keep at least the constructor explicitly `private`, as we should never instantiate the interface itself, which is theoretically possible now; albeit the `concept`s will probably forbid it, I'd prefer to make the intention explicit.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -116,17 +116,12 @@ decltype(auto) calculation_type_symmetry_func_selector(CalculationType calculati
...
 }
 
-// main model implementation template
-template <class T, class U> class MainModelImpl;
-
-template <class... ExtraRetrievableType, class... ComponentType>
-class MainModelImpl<ExtraRetrievableTypes<ExtraRetrievableType...>, ComponentList<ComponentType...>> {
+template <class ModelType>
+    requires(main_core::is_main_model_type_v<ModelType>)
+class MainModelImpl {
```
figueroa1395, line 121: <3

figueroa1395 (commented):
Awesome improvements! Definitely going in the right direction. 

In addition, I think that we should "replace" all of our CRTP with `deducing this` everywhere else (in a follow up PR), it's supper clean and the performance is the same.

Finally, could you do a knowledge sharing about `deducing this`? I think it would be really worth it.

tests/cpp_unit_tests/test_main_model_type.cpp
nitbharambe: These `static_assert`s are to test `is_main_model_type_v`. 

The static asserts of `is_constructible_from` is an additional final confirmation that everything is okay and we can instantiate the MainModelImpl without any issues.

figueroa1395: Indeed, my question is, should we add any `static_assert`s for `!is_constructible_from` for `MainModelImpl` for combination of types and such that aren't constructible? Would that add any value? (Note the `!`)

nitbharambe: The template for MainModelType would not be possible because of the validate_component_check, hence constructible_from does not even work. Hence they should not be present.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: The reason for making it private+friend in CRTP was that we disallow mistakes in CRTP which compile silently. With the 'this ' pattern now, the interface itself is free enough to be created on its own.

Friending here is unnecessary. Plus we would have to declare the Mock object as friend too if we were to do it.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: No opinion here. I imagined we would never have a use case for other value categories so why reduce readability with forwards. We can also now see that some functions are const while some are not. What is style recommendation here?

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
figueroa1395: What if we remove `friend`, because you are right, but still keep the constructor `protected`? That way this serves strictly as an interface; this is just for intentionality. If it still doesn't convince you, feel free to resolve as this would be a matter of style only.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
figueroa1395: I think the issue is that this can be a problem when instantiating the `Adapter` in a lambda and utilizing NRVO, for example like it happens in `job_dispatch.hpp`, because in such cases the instance can be treated as an rvalue. So this would cover such a case. But I would try to go for the shortest, easiest, most readable path.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: Agreed. Done.

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
nitbharambe: Changed to &&

figueroa1395 (dismissed):

mgovers:
@nitbharambe there's a merge conflict with `main`.

PR #1136, opened 2025-09-24
Observability: full observability check meshed network without voltage phasor

Jerry-Jinfeng-Guo:
# Implement Meshed Network Observability Check

## Summary

This PR implements https://github.com/PowerGridModel/power-grid-model/issues/1085 a comprehensive observability checking algorithm for meshed power grid networks without voltage phasor sensors. The implementation adds a spanning tree algorithm based on nodal measurements and connectivity analysis to determine network observability for non-radial (meshed) topologies.

### Algorithm Implementation

**New Spanning Tree Algorithm for Meshed Networks**
- **`sufficient_condition_meshed_without_voltage_phasor()`**: Main entry point for meshed network observability checking
- **`starting_from_node()`**: Core spanning tree construction algorithm that attempts to build a spanning tree using available measurements
- **`prepare_starting_nodes()`**: Smart starting point selection algorithm that prioritizes nodes without measurements and no edge measurements

### Data Structures

**New Connectivity Status Enum**
- `ConnectivityStatus` enum with 7 states to track measurement usage:
  - `is_not_connected` (-1): Not connected, redundant
  - `has_no_measurement` (0): Connected but no measurement available  
  - `node_measured` (0b010): Node has unused measurement
  - `node_downstream_measured` (0b001): Branch discovered with downstream measurement
  - `node_upstream_measured` (0b100): Branch discovered with upstream measurement
  - `branch_native_measured` (0b111): Branch has its own measurement, unused
  - `branch_measured_used` (0b101): Branch measurement already used

**New Neighbor Network Structure**  
- `ObservabilityNNResult`: Represents network topology with direct neighbor connections
- Bidirectional neighbor list expansion via `expand_neighbour_list()`
- Integration with Y-bus structure for efficient connectivity analysis

### Algorithm Features

**Multi-Priority Spanning Tree Construction**
1. **Native Edge Measurements** (Priority 1): Utilizes branch flow/power sensors directly
2. **Downwind Measurements** (Priority 2): Uses nodal measurements in downwind mode  
3. **General Connection Rules** (Priority 3): Handles complex measurement assignment scenarios
4. **Backtracking with Reassignment** (Priority 4): Implements measurement reassignment during backtracking

**Smart Measurement Reassignment**
- `reassign_nodal_measurement()`: Dynamically reassigns nodal measurements during backtracking
- Optimizes measurement usage when algorithm gets stuck
- Maintains edge direction consistency through connectivity status updates

### Refactored Codfe

**Enhanced Network Scanning**
- Renamed `count_observability_sensors()` → `scan_network_sensors()`  
- Added bus injection counting with early-out optimization
- Integrated neighbor list construction during network scanning
- Added comprehensive sensor classification (flow, voltage phasor, bus injection)

**Improved Function Organization**
- Renamed `necessary_observability_condition()` → `necessary_condition()`
- Renamed `sufficient_observability_condition()` → `sufficient_condition_radial_with_voltage_phasor()`
- Added namespace `observability` for better code organization
- Eliminated code duplication through parameterized helper functions

### Some Performance Optimizations

**Early Exit Conditions**
- **Sufficient nodal measurements**: If `bus_injections.back() > n_bus - 2`, network is immediately observable
- **Multiple starting points**: Algorithm tries different starting nodes to find valid spanning trees
- **Iteration limiting**: Prevents infinite loops with `max_iterations = n_bus²`

### Improvements

**Variable Naming Consistency**
- Fixed loop variable names: `row` → `bus` for clarity
- Consistent use of `n_bus` sizing  
- Improved variable scoping and const correctness

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -71,21 +71,21 @@ ObservabilitySensorsResult scan_network_sensors(MeasuredValues<sym> const& measu
         return topo.branch_bus_idx[branch][0] != -1 && topo.branch_bus_idx[branch][1] != -1;
     };
 
-    for (Idx row = 0; row != n_bus; ++row) {
+    for (Idx bus = 0; bus != n_bus; ++bus) {
         bool has_at_least_one_sensor{false};
-        Idx const current_bus = y_bus_structure.bus_entry[row];
-        neighbour_result[row].bus = current_bus;
+        Idx const current_bus_entry = y_bus_structure.bus_entry[bus];
+        neighbour_result[bus].bus = bus;
         // lower triangle is ignored ~~and kept as zero~~
         // diagonal for bus injection measurement
-        if (measured_values.has_bus_injection(row)) {
-            result.bus_injections[row] = 1;
+        if (measured_values.has_bus_injection(bus)) {
+            result.bus_injections[bus] = 1;
             result.bus_injections.back() += 1;
-            result.flow_sensors[current_bus] = 1;
+            result.flow_sensors[current_bus_entry] = 1;
             has_at_least_one_sensor = true;
-            neighbour_result[row].status = ConnectivityStatus::node_measured; // only treat power/0 injection
+            neighbour_result[bus].status = ConnectivityStatus::node_measured; // only treat power/0 injection
         }
         // upper triangle for branch flow measurement
-        for (Idx ybus_index = current_bus + 1; ybus_index != y_bus_structure.row_indptr[row + 1]; ++ybus_index) {
+        for (Idx ybus_index = current_bus_entry + 1; ybus_index != y_bus_structure.row_indptr[bus + 1]; ++ybus_index) {
             for (Idx element_index = y_bus_structure.y_bus_entry_indptr[ybus_index];
                  element_index != y_bus_structure.y_bus_entry_indptr[ybus_index + 1]; ++element_index) {
                 auto const& element = y_bus_structure.y_bus_element[element_index];
@@ -96,24 +96,24 @@ ObservabilitySensorsResult scan_network_sensors(MeasuredValues<sym> const& measu
                 // if the branch is fully connected and measured, we consider it as a valid flow sensor
                 // we only need one flow sensor, so the loop will break
                 Idx const branch = element.idx;
-                // Q j.guo: which one is the correct index for the branch, element_index or element.index
-                ObservabilityNNResult::neighbour neighbour_info{element_index, ConnectivityStatus::has_no_measurement};
-                neighbour_result[row].direct_neighbours.push_back(neighbour_info);
+                Idx const neighbour_bus = y_bus_structure.col_indices[ybus_index];
+                ObservabilityNNResult::neighbour neighbour_info{neighbour_bus, ConnectivityStatus::has_no_measurement};
+                neighbour_result[bus].direct_neighbours.push_back(neighbour_info);
                 if (has_flow_sensor(branch) && is_branch_connected(branch)) {
                     result.flow_sensors[ybus_index] = 1;
                     has_at_least_one_sensor = true;
-                    neighbour_result[row].direct_neighbours.back().status = ConnectivityStatus::branch_native_measured;
+                    neighbour_result[bus].direct_neighbours.back().status = ConnectivityStatus::branch_native_measured;
                     break;
                 }
             }
         }
         // diagonal for voltage phasor sensors
-        if (measured_values.has_voltage(row) && measured_values.has_angle_measurement(row)) {
+        if (measured_values.has_voltage(bus) && measured_values.has_angle_measurement(bus)) {
             has_at_least_one_sensor = true;
-            result.voltage_phasor_sensors[row] = 1;
+            result.voltage_phasor_sensors[bus] = 1;
         }
         // the system could be ill-conditioned if there is no flow sensor for one bus, except the last bus
-        if (!has_at_least_one_sensor && row != n_bus - 1) {
+        if (!has_at_least_one_sensor && bus != n_bus - 1) {
             result.is_possibly_ill_conditioned = true;
         }
     }
@@ -131,8 +131,8 @@ inline void assign_independent_sensors_radial(YBusStructure const& y_bus_structu
                                               std::vector<int8_t>& voltage_phasor_sensors) {
     Idx const n_bus{std::ssize(y_bus_structure.row_indptr) - 1};
     // loop the row without the last bus
-    for (Idx row = 0; row != n_bus - 1; ++row) {
-        Idx const current_bus = row;
+    for (Idx bus = 0; bus != n_bus - 1; ++bus) {
+        Idx const current_bus = bus;
         // upstream_bus_diagonal, concerns only the voltage phasor sensors as they are only on
         // the buses (diagonal entries)
         Idx const upstream_bus_diagonal = current_bus + 1;
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -23,8 +23,8 @@ struct ObservabilitySensorsResult {
     bool is_possibly_ill_conditioned{false};
 };
 enum class ConnectivityStatus {
-    not_connected = -1,               // not connected, redundant
-    connected_no_measurement = 0,     // connected (for branch) or node, no measurement available
+    is_not_connected = -1,            // not connected, redundant
+    has_no_measurement = 0,           // connected branch or node, but no measurement (may have been used)
     node_measured = 0b010,            // the node has measurement and is not yet used
     node_downstream_measured = 0b001, // branch is discovered with node measurement at from side; >>
     node_upstream_measured = 0b100,   // branch is discovered with node measurement at to side; <<
@@ -34,12 +34,12 @@ enum class ConnectivityStatus {
 // direct connected neighbour list
 struct ObservabilityNNResult {
     struct neighbour {
-        Idx bus;
-        ConnectivityStatus status;
+        Idx bus;                                                         // the bus index
+        ConnectivityStatus status{ConnectivityStatus::is_not_connected}; // the neighbour connectivity status
     };
-    Idx bus;
-    ConnectivityStatus status;
-    std::vector<neighbour> direct_neighbours;
+    Idx bus;                                                           // this bus index
+    ConnectivityStatus status{ConnectivityStatus::has_no_measurement}; // this bus connectivity status
+    std::vector<neighbour> direct_neighbours;                          // list of direct connected neighbours
 };
 
 // count flow and voltage phasor sensors for the observability check
@@ -52,9 +52,9 @@ struct ObservabilityNNResult {
 //      a vector of voltage phasor sensor count
 //      a boolean indicating if the system is possibly ill-conditioned
 template <symmetry_tag sym>
-ObservabilitySensorsResult count_sensors(MeasuredValues<sym> const& measured_values, MathModelTopology const& topo,
-                                         YBusStructure const& y_bus_structure,
-                                         ObservabilityNNResult& neighbour_result) {
+ObservabilitySensorsResult scan_network_sensors(MeasuredValues<sym> const& measured_values,
+                                                MathModelTopology const& topo, YBusStructure const& y_bus_structure,
+                                                std::vector<ObservabilityNNResult>& neighbour_result) {
     Idx const n_bus{topo.n_bus()};
 
     ObservabilitySensorsResult result{.flow_sensors = std::vector<int8_t>(y_bus_structure.row_indptr.back(), 0),
@@ -73,17 +73,19 @@ ObservabilitySensorsResult count_sensors(MeasuredValues<sym> const& measured_val
 
     for (Idx row = 0; row != n_bus; ++row) {
         bool has_at_least_one_sensor{false};
+        Idx const current_bus = y_bus_structure.bus_entry[row];
+        neighbour_result[row].bus = current_bus;
         // lower triangle is ignored ~~and kept as zero~~
         // diagonal for bus injection measurement
         if (measured_values.has_bus_injection(row)) {
             result.bus_injections[row] = 1;
             result.bus_injections.back() += 1;
-            result.flow_sensors[y_bus_structure.bus_entry[row]] = 1;
+            result.flow_sensors[current_bus] = 1;
             has_at_least_one_sensor = true;
+            neighbour_result[row].status = ConnectivityStatus::node_measured; // only treat power/0 injection
         }
         // upper triangle for branch flow measurement
-        for (Idx ybus_index = y_bus_structure.bus_entry[row] + 1; ybus_index != y_bus_structure.row_indptr[row + 1];
-             ++ybus_index) {
+        for (Idx ybus_index = current_bus + 1; ybus_index != y_bus_structure.row_indptr[row + 1]; ++ybus_index) {
             for (Idx element_index = y_bus_structure.y_bus_entry_indptr[ybus_index];
                  element_index != y_bus_structure.y_bus_entry_indptr[ybus_index + 1]; ++element_index) {
                 auto const& element = y_bus_structure.y_bus_element[element_index];
@@ -94,9 +96,13 @@ ObservabilitySensorsResult count_sensors(MeasuredValues<sym> const& measured_val
                 // if the branch is fully connected and measured, we consider it as a valid flow sensor
                 // we only need one flow sensor, so the loop will break
                 Idx const branch = element.idx;
+                // Q j.guo: which one is the correct index for the branch, element_index or element.index
+                ObservabilityNNResult::neighbour neighbour_info{element_index, ConnectivityStatus::has_no_measurement};
+                neighbour_result[row].direct_neighbours.push_back(neighbour_info);
                 if (has_flow_sensor(branch) && is_branch_connected(branch)) {
                     result.flow_sensors[ybus_index] = 1;
                     has_at_least_one_sensor = true;
+                    neighbour_result[row].direct_neighbours.back().status = ConnectivityStatus::branch_native_measured;
                     break;
                 }
             }
@@ -241,9 +247,9 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
         throw NotObservableError{"No voltage sensor found!\n"};
     }
 
-    detail::ObservabilityNNResult neighbour_result{};
+    std::vector<detail::ObservabilityNNResult> neighbour_results(static_cast<std::size_t>(topo.n_bus()));
     detail::ObservabilitySensorsResult observability_sensors =
-        detail::count_sensors(measured_values, topo, y_bus_structure, neighbour_result);
+        detail::scan_network_sensors(measured_values, topo, y_bus_structure, neighbour_results);
 
     //  sufficient & necessary early out, enough nodal measurement equals observable
     if (observability_sensors.bus_injections.back() > n_bus - 2) {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -13,8 +13,9 @@ namespace power_grid_model::math_solver {
 
 namespace detail {
 struct ObservabilitySensorsResult {
-    std::vector<int8_t> flow_sensors;
-    std::vector<int8_t> voltage_phasor_sensors;
+    std::vector<int8_t> flow_sensors;           // power sensor and current sensor
+    std::vector<int8_t> voltage_phasor_sensors; // voltage phasor sensors
+    std::vector<int8_t> bus_injections;         // bus injections, zero injection and power sensors at buses
     bool is_possibly_ill_conditioned{false};
 };
 
@@ -35,6 +36,7 @@ ObservabilitySensorsResult count_observability_sensors(MeasuredValues<sym> const
 
     ObservabilitySensorsResult result{.flow_sensors = std::vector<int8_t>(y_bus_structure.row_indptr.back(), 0),
                                       .voltage_phasor_sensors = std::vector<int8_t>(n_bus, 0),
+                                      .bus_injections = std::vector<int8_t>(n_bus + 1, 0),
                                       .is_possibly_ill_conditioned = false};
 
     auto has_flow_sensor = [&measured_values](Idx branch) {
@@ -51,6 +53,8 @@ ObservabilitySensorsResult count_observability_sensors(MeasuredValues<sym> const
         // lower triangle is ignored and kept as zero
         // diagonal for bus injection measurement
         if (measured_values.has_bus_injection(row)) {
+            result.bus_injections[row] = 1;
+            result.bus_injections.back() += 1;
             result.flow_sensors[y_bus_structure.bus_entry[row]] = 1;
             has_at_least_one_sensor = true;
         }
@@ -164,9 +168,9 @@ inline bool necessary_observability_condition(ObservabilitySensorsResult const& 
     return true;
 }
 
-inline bool sufficient_observability_condition(YBusStructure const& y_bus_structure,
-                                               ObservabilitySensorsResult& observability_sensors,
-                                               Idx const n_voltage_phasor_sensors) {
+inline bool sufficient_observability_condition_radial(YBusStructure const& y_bus_structure,
+                                                      ObservabilitySensorsResult& observability_sensors,
+                                                      Idx const n_voltage_phasor_sensors) {
     std::vector<int8_t>& flow_sensors = observability_sensors.flow_sensors;
     std::vector<int8_t>& voltage_phasor_sensors = observability_sensors.voltage_phasor_sensors;
     Idx const n_bus{std::ssize(y_bus_structure.row_indptr) - 1};
@@ -213,6 +217,13 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
 
     detail::ObservabilitySensorsResult observability_sensors =
         detail::count_observability_sensors(measured_values, topo, y_bus_structure);
+
+    // //  sufficient & necessary early out
+    // if (observability_sensors.bus_injections.back() > n_bus - 2) {
+    //     return ObservabilityResult{.is_observable = true,
+    //                                .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
+    // }
+
     Idx n_voltage_phasor_sensors{};
 
     // check necessary condition for observability
@@ -222,8 +233,14 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     // check the sufficient condition for observability
     // the check is currently only implemented for radial grids
     if (topo.is_radial) {
-        is_sufficient_condition_met = detail::sufficient_observability_condition(y_bus_structure, observability_sensors,
-                                                                                 n_voltage_phasor_sensors);
+        is_sufficient_condition_met = detail::sufficient_observability_condition_radial(
+            y_bus_structure, observability_sensors, n_voltage_phasor_sensors);
+    }
+    // DEBUG
+    if ((observability_sensors.bus_injections.back() > n_bus - 2) !=
+        (is_necessary_condition_met && is_sufficient_condition_met)) {
+        // Log or handle the inconsistency
+        std::cerr << "Inconsistency detected in observability conditions.\n";
     }
     //  ToDo(JGuo): meshed network will require a different treatment
     return ObservabilityResult{.is_observable = is_necessary_condition_met && is_sufficient_condition_met,
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -9,6 +9,10 @@
 
 #include "../common/exception.hpp"
 
+// DEBUG
+#include <iostream>
+// DEBUG
+
 namespace power_grid_model::math_solver {
 
 namespace detail {
@@ -218,11 +222,11 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     detail::ObservabilitySensorsResult observability_sensors =
         detail::count_observability_sensors(measured_values, topo, y_bus_structure);
 
-    // //  sufficient & necessary early out
-    // if (observability_sensors.bus_injections.back() > n_bus - 2) {
-    //     return ObservabilityResult{.is_observable = true,
-    //                                .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
-    // }
+    //  sufficient & necessary early out
+    if (observability_sensors.bus_injections.back() > n_bus - 2) {
+        return ObservabilityResult{.is_observable = true,
+                                   .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
+    }
 
     Idx n_voltage_phasor_sensors{};
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -239,7 +239,278 @@ inline void expand_neighbour_list(std::vector<ObservabilityNNResult>& neighbour_
     }
 }
 
-inline bool sufficient_condition_meshed_without_voltage_phasor() { return false; }
+inline void prepare_starting_nodes(std::vector<detail::ObservabilityNNResult> const& neighbour_list, Idx n_bus,
+                                   std::vector<Idx>& starting_candidates) {
+    // First find a list of starting points. These are nodes without measurements and all edges connecting to it has no
+    // edge measurements.
+    for (Idx bus = 0; bus < n_bus; ++bus) {
+        if (neighbour_list[bus].status == ConnectivityStatus::has_no_measurement) {
+            bool all_neighbours_no_edge_measurement = true;
+            for (const auto& neighbour : neighbour_list[bus].direct_neighbours) {
+                if (neighbour.status == ConnectivityStatus::branch_native_measured) {
+                    all_neighbours_no_edge_measurement = false;
+                    break;
+                }
+            }
+            if (all_neighbours_no_edge_measurement) {
+                starting_candidates.push_back(bus);
+            }
+        }
+    }
+
+    // If no such starting point, find nodes without measurements
+    if (starting_candidates.empty()) {
+        for (Idx bus = 0; bus < n_bus; ++bus) {
+            if (neighbour_list[bus].status == ConnectivityStatus::has_no_measurement) {
+                starting_candidates.push_back(bus);
+            }
+        }
+    }
+
+    // If no nodes without measurements, start from first node
+    // (but network should be observable, so this is just a fallback)
+    if (starting_candidates.empty()) {
+        starting_candidates.push_back(0);
+    }
+}
+
+inline bool starting_from_node(Idx start_bus, Idx n_bus,
+                               std::vector<detail::ObservabilityNNResult> const& _neighbour_list) {
+    // Make a fresh copy for this attempt
+    std::vector<detail::ObservabilityNNResult> local_neighbour_list = _neighbour_list;
+
+    // Initialize tracking structures
+    std::vector<bool> visited(n_bus, false);
+    std::vector<std::pair<Idx, Idx>> discovered_edges;
+    std::vector<std::pair<Idx, Idx>> edge_track; // for backtracking
+    bool downwind = false;                       // downwind flag
+
+    Idx current_bus = start_bus;
+    Idx const max_iterations = n_bus * n_bus; // prevent infinite loops
+    Idx iteration = 0;
+
+    // Define lambda functions for the different priorities and backtracking
+    auto try_native_edge_measurements = [&](Idx& current_bus, bool& found, bool& downwind) -> bool {
+        for (auto& neighbour : local_neighbour_list[current_bus].direct_neighbours) {
+            if (neighbour.status == ConnectivityStatus::branch_native_measured && !visited[neighbour.bus]) {
+                // Mark edge as discovered and neighbour as visited
+                discovered_edges.emplace_back(current_bus, neighbour.bus);
+                edge_track.emplace_back(current_bus, neighbour.bus);
+                visited[current_bus] = true;
+                visited[neighbour.bus] = true;
+
+                // Update status to branch_measured_used
+                neighbour.status = ConnectivityStatus::branch_measured_used;
+                // Update reverse connection
+                for (auto& reverse_neighbour : local_neighbour_list[neighbour.bus].direct_neighbours) {
+                    if (reverse_neighbour.bus == current_bus) {
+                        reverse_neighbour.status = ConnectivityStatus::branch_measured_used;
+                        break;
+                    }
+                }
+
+                downwind = true; // downwind = true
+                current_bus = neighbour.bus;
+                found = true;
+                return true;
+            }
+        }
+        return false;
+    };
+
+    auto try_downwind_measurement = [&](Idx& current_bus, bool& found, bool this_empty, bool downwind) -> bool {
+        if (!this_empty && downwind) {
+            for (auto& neighbour : local_neighbour_list[current_bus].direct_neighbours) {
+                if (neighbour.status == ConnectivityStatus::has_no_measurement && !visited[neighbour.bus]) {
+                    discovered_edges.emplace_back(current_bus, neighbour.bus);
+                    edge_track.emplace_back(current_bus, neighbour.bus);
+                    visited[current_bus] = true;
+                    visited[neighbour.bus] = true;
+
+                    // Update status to node_downstream_measured
+                    neighbour.status = ConnectivityStatus::node_downstream_measured;
+                    // Update reverse connection
+                    for (auto& reverse_neighbour : local_neighbour_list[neighbour.bus].direct_neighbours) {
+                        if (reverse_neighbour.bus == current_bus) {
+                            reverse_neighbour.status = ConnectivityStatus::node_upstream_measured;
+                            break;
+                        }
+                    }
+
+                    // Use current node's measurement
+                    local_neighbour_list[current_bus].status = ConnectivityStatus::has_no_measurement;
+                    current_bus = neighbour.bus;
+                    found = true;
+                    return true;
+                }
+            }
+        }
+        return false;
+    };
+
+    auto try_general_connection_rules = [&](Idx& current_bus, bool& found, bool& downwind, bool this_empty) -> bool {
+        // Helper lambda to handle common edge processing logic
+        auto process_edge = [&](auto& neighbour, ConnectivityStatus neighbour_status, ConnectivityStatus reverse_status,
+                                bool use_current_node, bool set_upwind = false) -> bool {
+            discovered_edges.emplace_back(current_bus, neighbour.bus);
+            edge_track.emplace_back(current_bus, neighbour.bus);
+            visited[current_bus] = true;
+            visited[neighbour.bus] = true;
+
+            // Update neighbour status
+            neighbour.status = neighbour_status;
+            // Update reverse connection
+            for (auto& reverse_neighbour : local_neighbour_list[neighbour.bus].direct_neighbours) {
+                if (reverse_neighbour.bus == current_bus) {
+                    reverse_neighbour.status = reverse_status;
+                    break;
+                }
+            }
+
+            // Use measurement from appropriate node
+            if (use_current_node) {
+                local_neighbour_list[current_bus].status = ConnectivityStatus::has_no_measurement;
+            } else {
+                local_neighbour_list[neighbour.bus].status = ConnectivityStatus::has_no_measurement;
+            }
+
+            // Set direction flag if needed
+            if (set_upwind) {
+                downwind = false; // upwind
+            }
+
+            current_bus = neighbour.bus;
+            found = true;
+            return true;
+        };
+
+        for (auto& neighbour : local_neighbour_list[current_bus].direct_neighbours) {
+            if (visited[neighbour.bus])
+                continue;
+
+            bool const next_empty =
+                local_neighbour_list[neighbour.bus].status == ConnectivityStatus::has_no_measurement;
+
+            if (!this_empty && (downwind && next_empty)) {
+                // Case: current has measurement, downwind, neighbour empty
+                return process_edge(neighbour, ConnectivityStatus::node_upstream_measured,
+                                    ConnectivityStatus::node_downstream_measured, true);
+            } else if (!this_empty && (downwind || next_empty)) {
+                // Case: current has measurement, (downwind OR neighbour empty)
+                return process_edge(neighbour, ConnectivityStatus::node_downstream_measured,
+                                    ConnectivityStatus::node_upstream_measured, true);
+            } else if (!next_empty) {
+                // Case: neighbour has measurement
+                return process_edge(neighbour, ConnectivityStatus::node_downstream_measured,
+                                    ConnectivityStatus::node_upstream_measured, false, true);
+            }
+        }
+        return false;
+    };
+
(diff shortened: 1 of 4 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -9,14 +9,38 @@
...
     bool is_possibly_ill_conditioned{false};
 };
+enum class ConnectivityStatus {
+    is_not_connected = -1,            // not connected, redundant
+    has_no_measurement = 0,           // connected branch or node, but no measurement (may have been used)
+    node_measured = 0b010,            // the node has measurement and is not yet used
+    node_downstream_measured = 0b001, // branch is discovered with node measurement at from side; >>
+    node_upstream_measured = 0b100,   // branch is discovered with node measurement at to side; <<
+    branch_native_measured = 0b111,   // branch has its own measurement, unused; & 0b101
+    branch_measured_used = 0b101      // branch discovered with its own measurement, used; | 0b010
```
figueroa1395: Or something similar? The naming wasn't clear to me until I went through the logic.
```suggestion
    branch_native_measured_unused = 0b111,   // branch has its own measurement, unused; & 0b101
    branch_native_measured_used = 0b101      // branch discovered with its own measurement, used; | 0b010
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -9,14 +9,38 @@
...
 namespace detail {
 struct ObservabilitySensorsResult {
-    std::vector<int8_t> flow_sensors;
-    std::vector<int8_t> voltage_phasor_sensors;
+    std::vector<int8_t> flow_sensors;           // power sensor and current sensor
+    std::vector<int8_t> voltage_phasor_sensors; // voltage phasor sensors
+    std::vector<int8_t> bus_injections;         // bus injections, zero injection and power sensors at buses
     bool is_possibly_ill_conditioned{false};
 };
+enum class ConnectivityStatus {
```
figueroa1395: I'm not very familiar with using bit integers, could you explain a bit what are the advantages of using them here? Or why did you introduce them instead of plain integers?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -28,13 +52,14 @@ struct ObservabilitySensorsResult {
...
-                                                       MathModelTopology const& topo,
-                                                       YBusStructure const& y_bus_structure) {
+ObservabilitySensorsResult scan_network_sensors(MeasuredValues<sym> const& measured_values,
+                                                MathModelTopology const& topo, YBusStructure const& y_bus_structure,
+                                                std::vector<ObservabilityNNResult>& neighbour_result) {
     Idx const n_bus{topo.n_bus()};
 
     ObservabilitySensorsResult result{.flow_sensors = std::vector<int8_t>(y_bus_structure.row_indptr.back(), 0),
                                       .voltage_phasor_sensors = std::vector<int8_t>(n_bus, 0),
+                                      .bus_injections = std::vector<int8_t>(n_bus + 1, 0),
```
figueroa1395, line 59: Nitpick: The `n+1` trick with the number of sensors at the `n+1` spot might be useful for `voltage_phasor_sensors` and `flow_sensors` as well, since we count them anyway later on.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -46,17 +71,21 @@ ObservabilitySensorsResult count_observability_sensors(MeasuredValues<sym> const
         return topo.branch_bus_idx[branch][0] != -1 && topo.branch_bus_idx[branch][1] != -1;
     };
 
-    for (Idx row = 0; row != n_bus; ++row) {
+    for (Idx bus = 0; bus != n_bus; ++bus) {
         bool has_at_least_one_sensor{false};
-        // lower triangle is ignored and kept as zero
+        Idx const current_bus_entry = y_bus_structure.bus_entry[bus];
+        neighbour_result[bus].bus = bus;
```
figueroa1395: Nitpick: Seeing the use here, maybe different naming is better? Something like `bus_neighborhood_info` or something along those lines?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -46,17 +71,21 @@ ObservabilitySensorsResult count_observability_sensors(MeasuredValues<sym> const
...
+        // lower triangle is ignored ~~and kept as zero~~
         // diagonal for bus injection measurement
-        if (measured_values.has_bus_injection(row)) {
-            result.flow_sensors[y_bus_structure.bus_entry[row]] = 1;
+        if (measured_values.has_bus_injection(bus)) {
+            result.bus_injections[bus] = 1;
+            result.bus_injections.back() += 1;
+            result.flow_sensors[current_bus_entry] = 1;
```
figueroa1395, line 80: Shouldn't this line be removed now that `bus_injections` holds the power sensors at nodes? Or does `flow_sensors` still includes node sensors + branch sensors? We shouldn't need to "double count", right? We may able to reduce the size of `flow_sensors` as well, but that might may logic too complex and not worth it, but still mentioning it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -67,20 +96,24 @@ ObservabilitySensorsResult count_observability_sensors(MeasuredValues<sym> const
                 // if the branch is fully connected and measured, we consider it as a valid flow sensor
                 // we only need one flow sensor, so the loop will break
                 Idx const branch = element.idx;
+                Idx const neighbour_bus = y_bus_structure.col_indices[ybus_index];
+                ObservabilityNNResult::neighbour neighbour_info{neighbour_bus, ConnectivityStatus::has_no_measurement};
+                neighbour_result[bus].direct_neighbours.push_back(neighbour_info);
                 if (has_flow_sensor(branch) && is_branch_connected(branch)) {
                     result.flow_sensors[ybus_index] = 1;
                     has_at_least_one_sensor = true;
+                    neighbour_result[bus].direct_neighbours.back().status = ConnectivityStatus::branch_native_measured;
```
figueroa1395: Considering the comment and the `break` within the `if` block, this would not work well for meshed grids, right? As this only considers the first neighbor that has a sensor. If there are multiple neighbors with sensors, only the first one is added and then continues onto the next bus. Probably a good idea to think about a test case to try to specifically break this.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
     return true;
 }
 
+inline void expand_neighbour_list(std::vector<ObservabilityNNResult>& neighbour_result) {
+    Idx const n_bus{static_cast<Idx>(neighbour_result.size())};
+    for (Idx bus = 0; bus != n_bus; ++bus) {
+        for (auto const& neighbour : neighbour_result[bus].direct_neighbours) {
+            auto& reverse_neighbour_list = neighbour_result[neighbour.bus].direct_neighbours;
+            auto it = std::find_if(reverse_neighbour_list.begin(), reverse_neighbour_list.end(),
+                                   [&bus](auto const& x) { return x.bus == bus; });
```
figueroa1395:
```suggestion
            auto it = std::ranges::find_if(reverse_neighbour_list,
                                   [&bus](auto const& reverse_neighbor) { return reverse_neighbor.bus == bus; });
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
     return true;
 }
 
+inline void expand_neighbour_list(std::vector<ObservabilityNNResult>& neighbour_result) {
```
figueroa1395: Nitpick: The `*neighbor*` naming convention seems a bit difficult to follow for me in this function specifically, perhaps it is worth doing some renaming in here? It's easy to lose track which neighbor chaing is being modified.

Note: We should unit test this function, specially because of https://github.com/PowerGridModel/power-grid-model/pull/1136#discussion_r2377974419. The other place may generate issues here. Also, if the mentioned comment is addressed, I wonder if this function is necessary.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+inline void prepare_starting_nodes(std::vector<detail::ObservabilityNNResult> const& neighbour_list, Idx n_bus,
+                                   std::vector<Idx>& starting_candidates) {
+    // First find a list of starting points. These are nodes without measurements and all edges connecting to it has no
+    // edge measurements.
+    for (Idx bus = 0; bus < n_bus; ++bus) {
+        if (neighbour_list[bus].status == ConnectivityStatus::has_no_measurement) {
+            bool all_neighbours_no_edge_measurement = true;
+            for (const auto& neighbour : neighbour_list[bus].direct_neighbours) {
+                if (neighbour.status == ConnectivityStatus::branch_native_measured) {
+                    all_neighbours_no_edge_measurement = false;
```
figueroa1395, line 250: https://github.com/PowerGridModel/power-grid-model/pull/1136#discussion_r2377974419 may potentially result in bugs here, I believe. This is because maybe not all neighbors are correctly found. On the other hand, maybe I'm overlooking something at the above quoted comment is not relevant for this?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+            }
+            if (all_neighbours_no_edge_measurement) {
+                starting_candidates.push_back(bus);
+            }
+        }
+    }
+
+    // If no such starting point, find nodes without measurements
+    if (starting_candidates.empty()) {
+        for (Idx bus = 0; bus < n_bus; ++bus) {
```
figueroa1395, line 262: We may already want to keep track of this list in the above loop. It's a tradeoff, but maybe it's worth it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+    }
+}
+
+inline bool starting_from_node(Idx start_bus, Idx n_bus,
+                               std::vector<detail::ObservabilityNNResult> const& _neighbour_list) {
+    // Make a fresh copy for this attempt
+    std::vector<detail::ObservabilityNNResult> local_neighbour_list = _neighbour_list;
+
+    // Initialize tracking structures
+    std::vector<bool> visited(n_bus, false);
```
figueroa1395: `std::vector<bool>` is problematic as it is not really a vector (https://stackoverflow.com/questions/17794569/why-isnt-vectorbool-a-stl-container), to the point that it is discouraged. Yyou may want to consider `std::vector<int8_t>` instead like in other places in this file.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+    std::vector<std::pair<Idx, Idx>> discovered_edges;
+    std::vector<std::pair<Idx, Idx>> edge_track; // for backtracking
+    bool downwind = false;                       // downwind flag
+
+    Idx current_bus = start_bus;
+    Idx const max_iterations = n_bus * n_bus; // prevent infinite loops
```
figueroa1395, line 289: Why is this the selected threshold? Curious of the logic behind it.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+    Idx iteration = 0;
+
+    // Define lambda functions for the different priorities and backtracking
+    auto try_native_edge_measurements = [&](Idx& current_bus, bool& found, bool& downwind) -> bool {
```
figueroa1395: Nitpick: Can you capture explicitly? Or rename the parameters to not shadow the variables above? Same for other lambdas.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+            }
+
+            current_bus = backtrack_to_bus;
+            found = true; // We made progress by backtracking
+            return true;
+        }
+        return false;
+    };
+
+    while (std::count(visited.begin(), visited.end(), true) < n_bus && iteration < max_iterations) {
```
figueroa1395: Nitpick: 
```suggestion
    while (std::ranges::count(visited, true) < n_bus && iteration < max_iterations) {
```

On the other hand, do we know if this algorithm will actually visit all the nodes given an arbitrary observable grid? `max_iteration` seems like the safeguard because we don't really know (maybe we actually do). If we don't know, should we then only consider unobservability as a hard wall? But if it is observable or we don't know, still performe calculations?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+        ++iteration;
+        bool found = false;
+        bool const this_empty = local_neighbour_list[current_bus].status == ConnectivityStatus::has_no_measurement;
```
figueroa1395: Nitpick: Can you rename `this_empty` to something more expresive? Maybe something like `is_current_bus_measured` perhaps?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+            if (!try_downwind_measurement(current_bus, found, this_empty, downwind)) {
+                // Third priority: General connection rules
+                if (!try_general_connection_rules(current_bus, found, downwind, this_empty)) {
+                    // If no progress, try backtracking
+                    try_backtrack(current_bus, found, downwind);
+                }
+            }
+        }
+
+        if (!found) {
```
figueroa1395: `found` then represents that the branch between the current node and the following one could be measured, right? Maybe other more suggestive name? As I didn't know what was this about until I reached this point.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+inline bool starting_from_node(Idx start_bus, Idx n_bus,
+                               std::vector<detail::ObservabilityNNResult> const& _neighbour_list) {
+    // Make a fresh copy for this attempt
+    std::vector<detail::ObservabilityNNResult> local_neighbour_list = _neighbour_list;
+
+    // Initialize tracking structures
+    std::vector<bool> visited(n_bus, false);
+    std::vector<std::pair<Idx, Idx>> discovered_edges;
+    std::vector<std::pair<Idx, Idx>> edge_track; // for backtracking
+    bool downwind = false;                       // downwind flag
```
figueroa1395: Let's add a comment on what is the purpose of this flag. Once you know, it's obvious, but it was hard to make sense of it initially.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+        for (auto& neighbour : local_neighbour_list[current_bus].direct_neighbours) {
+            if (visited[neighbour.bus])
+                continue;
+
+            bool const next_empty =
+                local_neighbour_list[neighbour.bus].status == ConnectivityStatus::has_no_measurement;
+
+            if (!this_empty && (downwind && next_empty)) {
+                // Case: current has measurement, downwind, neighbour empty
+                return process_edge(neighbour, ConnectivityStatus::node_upstream_measured,
```
figueroa1395: Isn't this covered (the same as) by `try_downwind_measurement`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +224,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+                    return true;
+                }
+            }
+        }
+        return false;
+    };
+
+    auto try_general_connection_rules = [&](Idx& current_bus, bool& found, bool& downwind, bool this_empty) -> bool {
+        // Helper lambda to handle common edge processing logic
+        auto process_edge = [&](auto& neighbour, ConnectivityStatus neighbour_status, ConnectivityStatus reverse_status,
```
figueroa1395: Naively, `process_edge` can probably be used not only in `try_general_connection_rules`, but in `try_downwind_measurement` and perhaps in `try_native_edge_measurements` as well. It may need a little refactoring and generalization, but it may be interesting to consider to reduce code duplication and lenght, and also keep this processing logic in one place, and keep the core logic of the `try*()` functions "cleaner".

Jerry-Jinfeng-Guo:
Validation test case with meshed topology in `ill-conditioned-system` is wrongly configured. The network is observable.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+            } else if (last_edge.second == current_bus) {
+                backtrack_to_bus = last_edge.first;
+            } else {
+                // Find connected node
+                backtrack_to_bus = last_edge.first;
+            }
+
+            // Consider reassignment if needed (downwind and current node still has measurement unused)
+            bool reassign = downwind && local_neighbour_list[current_bus].status == ConnectivityStatus::node_measured;
+            if (reassign) {
```
figueroa1395: Doesn't the `else` include the `else if`?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -13,10 +13,30 @@ namespace power_grid_model::math_solver {
...
+    is_not_connected = -1,            // not connected, redundant
+    has_no_measurement = 0,           // connected branch or node, but no measurement (may have been used)
+    node_measured = 0b010,            // the node has measurement and is not yet used
+    node_downstream_measured = 0b001, // branch is discovered with node measurement at from side; >>
+    node_upstream_measured = 0b100,   // branch is discovered with node measurement at to side; <<
+    branch_native_measured = 0b111,   // branch has its own measurement, unused; & 0b101
+    branch_measured_used = 0b101      // branch discovered with its own measurement, used; | 0b010
+};
+// direct connected neighbour list
+struct ObservabilityNNResult {
```
figueroa1395: Maybe a more descriptive name?
```suggestion
struct BusConnectivityInfo {
```
Or something like that?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -164,9 +193,9 @@ inline bool necessary_observability_condition(ObservabilitySensorsResult const&
     return true;
 }
 
-inline bool sufficient_observability_condition(YBusStructure const& y_bus_structure,
-                                               ObservabilitySensorsResult& observability_sensors,
-                                               Idx const n_voltage_phasor_sensors) {
+inline bool sufficient_condition_radial_with_voltage_phasor(YBusStructure const& y_bus_structure,
```
figueroa1395, line 199: Nitpick: Here, in `n_independent_flow_sensors` and `n_remaining_voltage_phasor_sensors` is where the `n+1` trick can simplify stuff, also removing logic from this check.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+    }
+
+    // If no nodes without measurements, start from first node
+    // (but network should be observable, so this is just a fallback)
+    if (starting_candidates.empty()) {
+        starting_candidates.push_back(0);
+    }
+}
+
+inline bool starting_from_node(Idx start_bus, Idx n_bus,
```
figueroa1395: At the end stage of this PR, or even better in a separate one, we should split up this function. It's very long and the nested lambdas make it more difficult to follow in my opinion.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+                               std::vector<detail::ObservabilityNNResult> const& _neighbour_list) {
+    // Make a fresh copy for this attempt
+    std::vector<detail::ObservabilityNNResult> local_neighbour_list = _neighbour_list;
+
+    // Initialize tracking structures
+    std::vector<bool> visited(n_bus, false);
+    std::vector<std::pair<Idx, Idx>> discovered_edges;
+    std::vector<std::pair<Idx, Idx>> edge_track; // for backtracking
```
figueroa1395: Aren't these two doing the same? But `discovered_edges` is never used after it's assigned, is it? Whereas `edge_track` indeed is.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+
+    // Define lambda functions for the different priorities and backtracking
+    auto try_native_edge_measurements = [&](Idx& current_bus, bool& found, bool& downwind) -> bool {
+        for (auto& neighbour : local_neighbour_list[current_bus].direct_neighbours) {
+            if (neighbour.status == ConnectivityStatus::branch_native_measured && !visited[neighbour.bus]) {
+                // Mark edge as discovered and neighbour as visited
+                discovered_edges.emplace_back(current_bus, neighbour.bus);
+                edge_track.emplace_back(current_bus, neighbour.bus);
+                visited[current_bus] = true;
+                visited[neighbour.bus] = true;
```
figueroa1395: Minor: Since `neighor_bus` will possibly be set to `true` again on the next run, perhaps `current_bus` assignment can be ommited and done prior to the call to this function.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+    Idx const n_bus{static_cast<Idx>(neighbour_result.size())};
+    for (Idx bus = 0; bus != n_bus; ++bus) {
+        for (auto const& neighbour : neighbour_result[bus].direct_neighbours) {
+            auto& reverse_neighbour_list = neighbour_result[neighbour.bus].direct_neighbours;
+            auto it = std::find_if(reverse_neighbour_list.begin(), reverse_neighbour_list.end(),
+                                   [&bus](auto const& x) { return x.bus == bus; });
+            if (it == reverse_neighbour_list.end()) {
+                ObservabilityNNResult::neighbour reverse_neighbour_info{bus, neighbour.status};
+                reverse_neighbour_list.push_back(reverse_neighbour_info);
+            }
```
figueroa1395, line 236: Can't this `if` block be ommited and the `reverse_neighbour_list` always be directly created? This is because in `scan_network_sensors` we only consider upstream nodes (the upper triangular y-bus matrix only). Or is this necessary because when traversing the tree we may not necessarily go in the "upstream direction" direction established by the y-bus matrix?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+
+                // Update status to branch_measured_used
+                neighbour.status = ConnectivityStatus::branch_measured_used;
+                // Update reverse connection
+                for (auto& reverse_neighbour : local_neighbour_list[neighbour.bus].direct_neighbours) {
+                    if (reverse_neighbour.bus == current_bus) {
+                        reverse_neighbour.status = ConnectivityStatus::branch_measured_used;
+                        break;
+                    }
+                }
```
figueroa1395, line 312: This works because of `expand_neighour_list` is run before, right? Hence completing the "half" picture from the y-bus structure in `scan_network_sensors`. Is that correct?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+                return true;
+            }
+        }
+        return false;
+    };
+
+    auto try_downwind_measurement = [&](Idx& current_bus, bool& found, bool this_empty, bool downwind) -> bool {
+        if (!this_empty && downwind) {
+            for (auto& neighbour : local_neighbour_list[current_bus].direct_neighbours) {
+                if (neighbour.status == ConnectivityStatus::has_no_measurement && !visited[neighbour.bus]) {
```
figueroa1395: It is not clear to me why we need the other neighbor to have not measurement. Is it such that we always prioritize using the farthest possible nodal measurement to avoid leaving some sensors ununsed?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+                    discovered_edges.emplace_back(current_bus, neighbour.bus);
+                    edge_track.emplace_back(current_bus, neighbour.bus);
+                    visited[current_bus] = true;
+                    visited[neighbour.bus] = true;
+
+                    // Update status to node_downstream_measured
+                    neighbour.status = ConnectivityStatus::node_downstream_measured;
```
figueroa1395: From this, I wonder if we should rename to `ConnectivityStatus::node_downwind_measured` because this direction might be different to the downstream direction as defined by the y-bus, however it will be in the same direction as the downwind.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+                continue;
+
+            bool const next_empty =
+                local_neighbour_list[neighbour.bus].status == ConnectivityStatus::has_no_measurement;
+
+            if (!this_empty && (downwind && next_empty)) {
+                // Case: current has measurement, downwind, neighbour empty
+                return process_edge(neighbour, ConnectivityStatus::node_upstream_measured,
+                                    ConnectivityStatus::node_downstream_measured, true);
+            } else if (!this_empty && (downwind || next_empty)) {
```
figueroa1395: From this it agains makes more sense to name the connectivity status in terms of downwind/upwind instead of downstream/upstream. But maybe it is subjective.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+        }
+
+        // Find and update the reverse connection from to_node to from_node
+        for (auto& neighbour : local_neighbour_list[to_node].direct_neighbours) {
+            if (neighbour.bus == from_node) {
+                // Change to downstream connection (from from_node to to_node perspective)
+                neighbour.status = ConnectivityStatus::node_downstream_measured;
+                break;
+            }
+        }
```
figueroa1395, line 444: This pattern happens a lot, maybe a good idea to make it a helper function.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +220,297 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+    // make a copy of the neighbour list
+    std::vector<detail::ObservabilityNNResult> neighbour_list = _neighbour_list;
+
+    Idx const n_bus = static_cast<Idx>(neighbour_list.size());
+    std::vector<Idx> starting_candidates;
+    prepare_starting_nodes(neighbour_list, n_bus, starting_candidates);
+
+    // Try each starting candidate
+    for (Idx start_bus : starting_candidates) {
+        if (starting_from_node(start_bus, n_bus, neighbour_list)) {
```
figueroa1395: Maybe rename this function from `starting_from_node` to `find_spanning_tree` or something along those lines? The current name doesn't really give hints on what it does.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: The intention is that the use of bit integers helps visualize what is the intention of each variable with the position of the ones and zeroes.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: This is probably a good follow up, as it depends on what will happen with the radial check. Resolving for now.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Changing this will mean significant refactoring that can be done in a follow up as it is not directly related to the meshed observability itself, but rather to the radial check. Resolving for now.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Offline discussion: this way is more readable. Let's only consider if it pops up again. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Offline discussion: This is a reasonable threshold as it covers a large enough number of permutations if the grid was to be fully connected and meshed. It resembles a first order approximation on the number of permutations. Going beyond this probably won't take us anywhere, as all the nodes would have been visited already. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: To be considered in a follow up. Resolving for now.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Answer is yes to all.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: It follows the logic flow.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -46,17 +67,21 @@ ObservabilitySensorsResult count_observability_sensors(MeasuredValues<sym> const
...
+            result.bus_injections[bus] = 1;
+            result.bus_injections.back() += 1;
+            result.flow_sensors[current_bus_entry] = 1;
             has_at_least_one_sensor = true;
+            neighbour_result[bus].status = ConnectivityStatus::node_measured; // only treat power/0 injection
         }
         // upper triangle for branch flow measurement
-        for (Idx ybus_index = y_bus_structure.bus_entry[row] + 1; ybus_index != y_bus_structure.row_indptr[row + 1];
-             ++ybus_index) {
+        for (Idx ybus_index = current_bus_entry + 1; ybus_index != y_bus_structure.row_indptr[bus + 1]; ++ybus_index) {
```
nitbharambe, line 85: We can fill the "expand_neighbour_list" info in this loop too using the col_indices

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
nitbharambe: Unresolving: I dont really like the n+1 trick. The extra .back() makes the intent really confusing. Why not have a counter within the struct itself?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -13,10 +13,30 @@ namespace power_grid_model::math_solver {
...
+};
+// direct connected neighbour list
+struct ObservabilityNNResult {
+    struct neighbour {
+        Idx bus;                                                         // the bus index
+        ConnectivityStatus status{ConnectivityStatus::is_not_connected}; // the neighbour connectivity status
+    };
+    Idx bus;                                                           // this bus index
+    ConnectivityStatus status{ConnectivityStatus::has_no_measurement}; // this bus connectivity status
+    std::vector<neighbour> direct_neighbours;                          // list of direct connected neighbours
```
nitbharambe, line 39: As discussed yesterday, we can store the neighbour list in YBusStructure order. Adding a note here, but lets think about it later.

nitbharambe: Out of scope here. And into future improvement ideas discussion.

Jerry-Jinfeng-Guo: Correction: the list is the product/output of a wrapper function/object in between the YBus and the Observability check module. Not **in** the YBus.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: The logic is sound. The `break` is meant for the element `for` loop. Once a branch is processed, it will always go to the next branch, which is the outer loop. This is tested in the case: https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/9e888df604eee7f7f527670a4cc5433d078d189c

figueroa1395: I have double checked and I can confirm this is indeed true. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: The "potential" bug was confirmed to not exist. Logic is sound. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Only the code change suggestion nitpick is still valid about this comment. The rest can be ignored as it was proved that the logic is sound from the unit tests.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: `downwind` naming makes sense, resolving. A comment is to be added explaining the meaning of `downwind` next to its definition.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: This can probably already be done inside the `scan_network_sensors()` function as pointed out by https://github.com/PowerGridModel/power-grid-model/pull/1136#discussion_r2398726311. This would safe the additional loop altogether.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Left in for documenting/debugging.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved in [56deb2c](https://github.com/PowerGridModel/power-grid-model/pull/1136/commits/56deb2c477038a4a8f309d459d0a9ea3570dcacc).

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: As mentioned during our discussions, it (one loop less) doesn't save operations. I view it as matter of preference.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/pull/1136#discussion_r2415868878

This also contradicts the 'single responsibility' paradigm you are promoting.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: I deem out of scope. Feel free to refactor in the upcoming refactoring.

tests/data/state_estimation/ill-conditioned-system/leaf-without-power-sensor-meshed/params.json
Jerry-Jinfeng-Guo, line 1: This file was already reviewed. See https://github.com/PowerGridModel/power-grid-model/pull/1139.

tests/data/state_estimation/ill-conditioned-system/leaf-without-power-sensor-meshed/sym_output.json
Jerry-Jinfeng-Guo, line 1: This file was already reviewed. See https://github.com/PowerGridModel/power-grid-model/pull/1139.

tests/data/state_estimation/meshed-network-observability/02-unobservable-with-branch-measurement/params.json
Jerry-Jinfeng-Guo, line 1: This file was already reviewed. See https://github.com/PowerGridModel/power-grid-model/pull/1139.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: The loop yes, it's probably even more readable here as we separate concerns, but I'm advocating for removing the `if` statement as I believe it would always go through anyway.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Resolved.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: Upon digging, this is confirmed to be not the case and the check for it not being contained in the list is necessary. Suggested changes would fail unit tests.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Offline discussion: similar patterns appear, but with slight differences. Not worth making generic if it becomes un-readable. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: Offline discussion: It's been made as general as possible for the current scope. If more is to be done, it shall be done in the follow up refactoring. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: We feel that even though there is nothing wrong with the use of the `n+1` slot to keep track of the number of bus injections, it hinders readability. Can you make a separate variable for this and just do
```suggestion
                                      .bus_injections = std::vector<int8_t>(n_bus, 0),
```
instead?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -191,8 +222,299 @@ inline bool sufficient_observability_condition(YBusStructure const& y_bus_struct
...
+    }
+
+    // Check if all nodes were visited (spanning tree found)
+    return std::ranges::count(visited, BusVisited::Visited) == n_bus;
+}
+
+inline bool
+sufficient_condition_meshed_without_voltage_phasor(std::vector<detail::BusNeighbourhoodInfo> const& _neighbour_list) {
+    // make a copy of the neighbour list
+    std::vector<detail::BusNeighbourhoodInfo> const neighbour_list = _neighbour_list;
```
figueroa1395: I don't think we need this extra copy, do we?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
figueroa1395: This logic has been adjusted accordingly.  Resolving.

PR #1137, opened 2025-09-25
State estimation: Infinite sigma appliance measurement with valid injection sensor

nitbharambe:
### Description

Currently we allow sigma=inf to disable any measurement.
The situation in the test case is:
` Source -- node_1 -- branch -- node_2 -- load`
where we have a node injection power measurement on a node_2 and a load power measurement on the load.
We expect the load measurement to be disabled here while the result of estimation follows node injection sensor.

The output of the sym_load currently is `p=0` and `q=0`. The desired behaviour should be `p=-p_node_injection_sensor` and `q=-q_node_injection_sensor` (And similar for sensor residuals). We should be receiving this desired result if we do not add a load measurement at all.

* [x] Fix for the issue
* [x] Validation Test case

To discuss something unrelated to this bug:
Currently, our validator flags sigma=inf measurements as invalid. Shall we allow it? Or atleast allow measured=0 and sigma=inf? 
Or do we recommend filtering these validator errors out manually?
Motivation is that disabling measurements should not lead to user skipping validation.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/measured_values.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/measured_values.hpp
@@ -525,7 +525,7 @@ template <symmetry_tag sym> class MeasuredValues {
             }
         }
 
-        result_data.push_back(combined_measurement);
+        result_data.push_back(std::move(combined_measurement));
         return static_cast<Idx>(result_data.size()) - 1;
     }
 
--- a/tests/cpp_validation_tests/test_validation.cpp
+++ b/tests/cpp_validation_tests/test_validation.cpp
@@ -512,15 +512,6 @@ std::optional<CaseParam> construct_case(std::filesystem::path const& case_dir, j
         }
     }
 
-    if (calculation_method_params.contains("atol")) {
-        json const& j_extra_atol = calculation_method_params.at("atol");
-        if (j_extra_atol.type() != json::value_t::object) {
-            param.atol = {{"default", j_extra_atol.get<double>()}};
-        } else {
-            j_extra_atol.get_to(param.atol);
-        }
-    }
-
     if (calculation_method_params.contains("raises")) {
         if (json const& raises = calculation_method_params.at("raises"); raises.contains("raises")) {
             param.raises = raises.at("raises").get<std::string>();


TonyXiang8787:
@nitbharambe 

Sigma = inf means disabling the measurement. This is a well defined behavior.

nitbharambe:
@TonyXiang8787 Indeed. Added a description in the PR. 

TLDR: Issue is that in the disabled appliance measurement case, node injection sensor's output is not reflected in load/gen's result

tests/data/state_estimation/high-variance-measurement/sym_output.json
```diff
@@ -35,7 +35,7 @@
       {
         "id": 31,
         "energized": 1,
-        "p": 0.0,
+        "p": 999092.970521542,
```
figueroa1395, line 38: This is part of the bug as well. We've had it for a long time now.

tests/data/state_estimation/inf-measurement-with-injection/params.json
```diff
@@ -0,0 +1,8 @@
+{
+  "calculation_method": [
+    "iterative_linear",
+    "newton_raphson"
+  ],
+  "rtol": 1e-08,
+  "atol": 1e-02
```
figueroa1395: This is only due to the differences between ILSE and NRSE for the asymmetric case. For the symmetric case, `"atol": 1e-05` works well.

nitbharambe:
<s>Another situation for unexpected behaviour was found:

```
 Source -- node_1 -- branch -- node_2(measured_inj) -- load (disabled)
                                                    -- gen (measured)
```

This situation does not lead to consistency in injection. ie. `node["p"] != asym_gen["p"] - asym_load["p"]`
do-not-merge added for this PR because what if the solution is related to both the issues. </s>

Edit: Current solution addresses this issue too

tests/data/state_estimation/inf-measurement-with-injection-measured-unmeasured-appliances/params.json
```diff
@@ -0,0 +1,8 @@
+{
+  "calculation_method": [
+    "iterative_linear",
+    "newton_raphson"
+  ],
+  "rtol": 1e-08,
+  "atol": 1e-02
```
figueroa1395: Since this doesn't look so nice, we may consider removing `iterative_linear` or `asym_output` as because of either of those (preferably the asymmetric test) we have to lower the `atol` and they aren't really needed for this test.

mgovers: you can provide separate `atol` for different calculation methods (if i remember correctly, you can use the `extra_param` key with a dict with calculation methods as values for that)

figueroa1395: I addressed this in [b55a78c](https://github.com/PowerGridModel/power-grid-model/pull/1137/commits/b55a78c2bb11e8972b97e866a0388a36d939d3aa), but I actually found that it wasn't supported by default already, so I made a quick PR to allow it: https://github.com/PowerGridModel/power-grid-model/pull/1145

tests/data/state_estimation/inf-measurement-with-injection/params.json
figueroa1395: See https://github.com/PowerGridModel/power-grid-model/pull/1137#discussion_r2393814413

tests/data/state_estimation/inf-measurement-with-injection/params.json
```diff
@@ -0,0 +1,18 @@
+{
+  "calculation_method": [
+    "iterative_linear",
+    "newton_raphson"
+  ],
+  "rtol": 1e-8,
+  "atol": {
+    "default": 1e-5
+  },
```
mgovers: shouldn't this just be like before? `extra_params` is a patch over the original params, so you don't need to explicitly specify a `default`, right?

```suggestion
  "atol": 1e-5
```

tests/data/state_estimation/inf-measurement-with-injection/params.json
```diff
@@ -0,0 +1,18 @@
...
+  "extra_params": {
+    "newton_raphson": {
+      "atol": 1e-5
+    },
+    "iterative_linear": {
+      "atol": 1e-2
```
mgovers: you can even go as far as to specify a larger `atol` only for a specific attribute. I think we do that for `*_residual` somewhere

figueroa1395: No needed here, it happens for many attributes for the asymmetric case and it would just be cumbersome when that is not the intention of this test

tests/data/state_estimation/inf-measurement-with-injection/params.json
figueroa1395: Done

tests/data/state_estimation/inf-measurement-with-injection-measured-unmeasured-appliances/params.json
figueroa1395: `atol` has been homogenized after replacing `link` with `line`. Either way, this is going beyond the scope of this bug fix. So I'm resolving and I will keep homogeneous in either case.

nitbharambe (commented):
LGTM.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/measured_values.hpp
```diff
@@ -510,7 +510,22 @@ template <symmetry_tag sym> class MeasuredValues {
...
+                is_inf(combined_measurement.imag_component.variance)) {
+                return unmeasured;
+            }
+        } else if constexpr (std::is_same_v<CalcParam, VoltageSensorCalcParam<sym>>) {
+            if (is_inf(combined_measurement.variance)) {
+                return unmeasured;
+            }
+        }
+
+        result_data.push_back(combined_measurement);
```
mgovers: optionally:

```suggestion
        result_data.push_back(std::move(combined_measurement));
```

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -512,6 +512,15 @@ std::optional<CaseParam> construct_case(std::filesystem::path const& case_dir, j
...
 
+    if (calculation_method_params.contains("atol")) {
+        json const& j_extra_atol = calculation_method_params.at("atol");
+        if (j_extra_atol.type() != json::value_t::object) {
+            param.atol = {{"default", j_extra_atol.get<double>()}};
+        } else {
+            j_extra_atol.get_to(param.atol);
+        }
+    }
+
```
mgovers: shouldn't this be removed from this PR as well?

figueroa1395: Indeed! Idk how this is still here 😆

PR #1142, opened 2025-09-29
Cleanup main model: Clean all includes and make some components to be forward declared

nitbharambe:
Idea is to make components forward declared to reduce coupling. 

This draft PR is created for gathering thoughts and requirements as of now.

* Forward declared all components
* Removed unecessary exception and logging includes.

We shall maintain include discipline from now. Include order would be:

For include / source files

* Important files (Main subjects). (Intention is to highlight for readability and not be strict)
* Other required files
* External dependencies
* Standard library

For tests:

* Helper includes of test
* Important files to be tested (Main subjects). (Intention is to highlight for readability and not be strict)
* Other required files
* External dependencies
* Testing dependencies
* Standard library

In general, use <> for external. "" for internal. Note: PGM includes in test are treated as external, hence they go as <>.


figueroa1395:
I like the cleanup. Now I wonder, should we move `all_components.hpp` to the `component` folder? Also seems intuitive to me to rename `component.hpp` to `all_components_fwd.hpp`, or somehow homogenize the two.

In addition, can we also include `component_list` in this "revolution"? Perhaps it now belongs in `component.hpp` and we try to keep all component related stuff in one location only.

.clang-tidy
```diff
@@ -68,6 +67,7 @@ readability-*,
 
 WarningsAsErrors: '*'
 
+IgnoreHeaders: 'Eigen/.*'
```
figueroa1395: I'm curious. Why was this necessary? If I understand correctly, this prevents clang-tidy from inserting/removing headers in our Eigen dependency?

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -9,6 +9,7 @@
 // eigen properties
 #include <Eigen/Dense>
 
+#include <cassert>
```
figueroa1395, line 13: Why is this needed here? It is now included in `common.hpp`.

nitbharambe: The branch is yet to be updated. Found this issue simultaneously along with https://github.com/PowerGridModel/power-grid-model/pull/1253

Didnt remove it becaause the purpose of this PR was to include everything from which symbols gets used. 
(Atleast as much as possible. This PR only enforces that for the tests.)

.clang-tidy
nitbharambe: We would not want eigen to check includes missing within the Eigen library. It is an option only for misc-include-cleaner.

https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html

mgovers: but what you're doing now is a global configuration, not a configuration for `misc-include-cleaner` only

nitbharambe: The option name indeed sounded very generic. 
But this option is placed under the include cleaner as per https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html. 
I think it applies to only the include-cleaner because:
* Check for eg. https://clang.llvm.org/extra/clang-tidy/checks/misc/header-include-cycle.html. There is an ignored files mentioned there.  Obviously these options might collide with each other.
* Check this search: https://clang.llvm.org/extra/search.html?q=IgnoreHeaders. The option gets mentioned only in the include-cleaner

mgovers: The correct syntax for only configuring for a single check is

```.clang-tidy
CheckOptions:
  - key: misc-include-cleaner.IgnoreHeaders
    value: 'Eigen/.*'
```

Configuring globally is bad practice and even dangerous in the future

tests/cpp_unit_tests/test_node.cpp
```diff
@@ -4,6 +4,11 @@
 
 #include <power_grid_model/component/node.hpp>
 
+#include "power_grid_model/auxiliary/update.hpp"
+#include <power_grid_model/common/common.hpp>
+#include <power_grid_model/common/enum.hpp>
+#include <power_grid_model/component/component.hpp>
```
mgovers: consistency between `""` and `<>`

tests/cpp_unit_tests/test_power_sensor.cpp
```diff
@@ -4,8 +4,20 @@
 
 #include <power_grid_model/component/power_sensor.hpp>
 
+#include "power_grid_model/auxiliary/input.hpp"
+#include "power_grid_model/auxiliary/output.hpp"
+#include "power_grid_model/auxiliary/update.hpp"
+#include <power_grid_model/calculation_parameters.hpp>
```
mgovers: consistency between `""` and `<>` (please check all files)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -10,7 +10,6 @@
 #include "meta_data.hpp"
 
 #include "../common/common.hpp"
-#include "../common/counting_iterator.hpp"
```
figueroa1395, line 13: As pointed out by the CI, add this via Jinja instead of directly editing here.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/output.hpp
```diff
@@ -7,10 +7,8 @@
 // clang-format off
 #pragma once
 
-#include "meta_data.hpp"
 
 #include "../common/common.hpp"
-#include "../common/enum.hpp"
```
figueroa1395: Idem.

tests/native_api_tests/test_api_serialization.cpp
```diff
@@ -2,12 +2,19 @@
...
 // SPDX-License-Identifier: MPL-2.0
 
-#include "power_grid_model_cpp.hpp"
-
+#include "power_grid_model_cpp/basics.hpp"
+#include "power_grid_model_cpp/buffer.hpp"
+#include "power_grid_model_cpp/dataset.hpp"
+#include "power_grid_model_cpp/handle.hpp"
+#include "power_grid_model_cpp/model.hpp"
+#include "power_grid_model_cpp/serialization.hpp"
```
figueroa1395: From `""` to `<>`.

tests/native_api_tests/test_api_serialization.cpp
```diff
@@ -2,12 +2,19 @@
...
+
+#include <power_grid_model_c/basics.h>
 #include <power_grid_model_c/dataset_definitions.h>
 
 #include <doctest/doctest.h>
 
+// NOLINTBEGIN(misc-include-cleaner)
```
figueroa1395, line 19: I see many of these, probably related to `nlohmann/json`. Shouldn't we instead exclude it via `misc-include-cleaner.IgnoreHeaders` as we did with `Eigen`?

tests/native_api_tests/test_api_model_update.cpp
```diff
@@ -310,7 +314,7 @@ auto const complete_state_json = R"json({
       {"id": 9, "node": 3, "status": 1, "g1": 0.015, "b1": 0, "g0": 0.015, "b0": 0}
     ]
   }
-})json"s;
+})json"s; // NOLINT(misc-include-cleaner) https://github.com/llvm/llvm-project/issues/98122
```
figueroa1395, line 317: These are annoying but I guess there is nothing else to do about them...

tests/native_api_tests/test_api_model.cpp
```diff
@@ -4,22 +4,31 @@
...
 
-#include "power_grid_model_cpp.hpp"
+#include "power_grid_model_cpp/basics.hpp"
+#include "power_grid_model_cpp/buffer.hpp"
+#include "power_grid_model_cpp/dataset.hpp"
+#include "power_grid_model_cpp/handle.hpp"
+#include "power_grid_model_cpp/model.hpp"
+#include "power_grid_model_cpp/options.hpp"
+
+#include "power_grid_model_c/basics.h"
```
figueroa1395: `""` vs `<>`.

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -315,7 +333,7 @@ void assert_result(OwningDataset const& owning_result, OwningDataset const& owni
                 auto attribute_type = MetaData::attribute_ctype(attribute_meta);
                 auto const& attribute_name = MetaData::attribute_name(attribute_meta);
                 // TODO need a way for common angle: u angle skipped for now
-                if (attribute_name == "u_angle"s) {
+                if (attribute_name == "u_angle"s) { // NOLINT(misc-include-cleaner)
```
figueroa1395, line 336: Why this one?

tests/cpp_unit_tests/test_meta_data_generation.cpp
```diff
@@ -17,50 +17,50 @@ TEST_CASE("Test column row conversion") {
         auto const& meta_map = meta_data::meta_data_gen::meta_data.get_dataset("input");
         auto const& node = meta_map.get_component("node");
         auto const& node_attr = node.attributes;
-        CHECK(node_attr[0].name == "id"s);
+        CHECK(node_attr[0].name == "id"s); // NOLINT(misc-include-cleaner)
```
figueroa1395, line 20: Why are these a problem here?

tests/cpp_unit_tests/test_math_solver_sc.cpp
```diff
@@ -4,7 +4,20 @@
...
 #include "test_math_solver_common.hpp"
 
-#include <power_grid_model/math_solver/short_circuit_solver.hpp>
+#include <power_grid_model/calculation_parameters.hpp>
+#include <power_grid_model/common/common.hpp>
+#include <power_grid_model/common/enum.hpp>
+#include <power_grid_model/common/exception.hpp>
+#include <power_grid_model/common/grouped_index_vector.hpp>
+#include <power_grid_model/common/three_phase_tensor.hpp>
+#include <power_grid_model/math_solver/short_circuit_solver.hpp> // NOLINT(misc-include-cleaner)
```
figueroa1395: Is it asking to directly import things called in the short circuit module? Or what's the issue?

tests/cpp_unit_tests/test_math_solver_pf_iterative_current.cpp
```diff
@@ -2,8 +2,9 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-#include "test_math_solver_pf.hpp"
+#include "test_math_solver_pf.hpp" // NOLINT(misc-include-cleaner)
```
figueroa1395, line 5: I'm on board with this one and the other similar ones that are used in the template testing.

power_grid_model_c_example/main.c
```diff
@@ -25,6 +25,7 @@ We do batch calculation with 3 scenarios, with the following mutation
 #2: source: u_ref = 1.10, sym_load_3: -200 kW
 */
 
+// NOLINTBEGIN(misc-include-cleaner)
```
figueroa1395, line 29: There is no `// NOLINTEND(misc-include-cleaner)`. Also, shouldn't we also be explicit like with the CPP API here? Why no lint here?

tests/cpp_validation_tests/test_validation.cpp
nitbharambe: Missed adding github issue name there. Same as others. Its with `operator s`

power_grid_model_c_example/main.c
nitbharambe: Ahh, I put the end at the bottom. This file is completely ignored as it as an example.

.clangd
mgovers, line 1: 😮

nitbharambe: I find clangd giving a lot more insights than microsoft c++. It spotted many of the cleaned up things of header files which is not possible via clang-tidy.

nitbharambe: I will bring up configuration in a knowledge sharing session.

mgovers: in that case, maybe add the extension to the workspace recommendations?

tests/cpp_unit_tests/test_meta_data_generation.cpp
nitbharambe: operator s

tests/cpp_unit_tests/test_math_solver_sc.cpp
nitbharambe: The symbol from that file is being used.
Suppressed warning. These set of test files dont sit well with the include-cleaner.

tests/native_api_tests/test_api_serialization.cpp
nitbharambe: Added

.clangd
nitbharambe: The microsoft C++ and clangd extentions cannot work together. We need to use either of them and disable the other. I disabled C_Cpp intellisense via vs code's settings. `"C_Cpp.intelliSenseEngine": "disabled"`

Since this is developer specific, we shall discuss whether to use it by default or not.

figueroa1395: Shall this knowledge be shared before merging?

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
```diff
@@ -4,16 +4,37 @@
...
 
 #include <boost/graph/adjacency_list.hpp>
 #include <boost/graph/compressed_sparse_row_graph.hpp>
 #include <boost/graph/depth_first_search.hpp>
+#include <boost/graph/graph_selectors.hpp>
+#include <boost/graph/graph_traits.hpp>
 #include <boost/graph/iteration_macros.hpp>
+#include <boost/graph/properties.hpp>
+#include <boost/iterator/counting_iterator.hpp>
+#include <boost/pending/property.hpp>
```
figueroa1395, line 24: Should we add boost as well to the ignored list? Here there seem to be some "hidden" things added which we don't really care about because they are boost internals. Or am I wrong?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data_gen.hpp
```diff
@@ -6,20 +6,16 @@
...
-#include "meta_gen/input.hpp"
-#include "meta_gen/output.hpp"
-#include "meta_gen/update.hpp"
-#include "output.hpp"
-#include "update.hpp"
+#include "meta_gen/input.hpp"  // IWYU pragma: keep
+#include "meta_gen/output.hpp" // IWYU pragma: keep
+#include "meta_gen/update.hpp" // IWYU pragma: keep
+#include "output.hpp"          // IWYU pragma: keep
+#include "update.hpp"          // IWYU pragma: keep
```
figueroa1395, line 16: So for meta data related stuff we accept we'll use the include all approach? This is fine as it is barely touched and doesn't provide much, but just confirming.

.clang-tidy
```diff
@@ -68,8 +70,10 @@ readability-*,
 
 WarningsAsErrors: '*'
 
-
 HeaderFilterRegex: '.*'
 
 FormatStyle: file
+
+CheckOptions:
+  misc-include-cleaner.IgnoreHeaders: 'Eigen/.*;nlohmann/.*;msgpack/.*'
```
figueroa1395: Follow up comment, but perhaps `boost` deserves a spot here too.

.clangd
nitbharambe: Should not be required as this configuration is non disruptive.

.clang-tidy
nitbharambe: Added

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
figueroa1395: After [b31ddc6](https://github.com/PowerGridModel/power-grid-model/pull/1142/commits/b31ddc63744f0a0e603273eea49ea6b3578c142c), can some of these be reverted/removed?

nitbharambe: We keep them because we do use symbols from those files. Adding in clangd / clang tidy only suppresses any warnings coming out from those libraries.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data_gen.hpp
nitbharambe: Yup. For now, its include all approach via this header.

nitbharambe: There would have been significant changes otherwise. But there is indeed potential to move these includes down at their usage site. At the usage as well we might have to suppress warnings as we use templating tricks to them. Hence this was the easier choice for now.

figueroa1395 (dismissed):

tests/native_api_tests/test_api_serialization.cpp
Jerry-Jinfeng-Guo, line 1: Sonar cloud complains here, perhaps we could add something like the following to the `// NOLINT`:
`
// NOSONAR - Required for nlohmann/json operations, clang-tidy suppression
`

nitbharambe: We have many commented code issues in our sonar. I think it's best not to add an ignore for them specially just for this case since it's a false positive. We need to deal with it in other way collectively.

.clangd
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+
+#
+
```
mgovers:
```suggestion
#
```
?

figueroa1395 (approved):

PR #1145, opened 2025-10-01
Validation tests: allow different atols per calculation method

figueroa1395:
This is a quick PR that allows specifying different `atol`s per calculation method on the validation tests via the `extra_params`.

Note: The use case of this is [b55a78c](https://github.com/PowerGridModel/power-grid-model/pull/1137/commits/b55a78c2bb11e8972b97e866a0388a36d939d3aa) on https://github.com/PowerGridModel/power-grid-model/pull/1137.


figueroa1395:
Closing because this patch is no longer needed. If we are ever to need this, we should do it properly.

PR #1146, opened 2025-10-01
Observability: add finer grained unit test for both existing radial check and new meshed check

Jerry-Jinfeng-Guo:
Unit test for all functions within the observability module, related to https://github.com/PowerGridModel/power-grid-model/pull/1136

- [x] radial check functions
- [x] meshed check functions

The unit tests are added with still dependency on mathematical topology and `YBusStructure`. This choice was due to the fact that when trying to wrap certain functionalities from the observability module, any mocking attempt will have to be redone. Therefore, the most straight-forward path is chosen to extend this unit test for the objective of function-level unit testing this module.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -16,6 +16,7 @@ struct ObservabilitySensorsResult {
     std::vector<int8_t> flow_sensors;           // power sensor and current sensor
     std::vector<int8_t> voltage_phasor_sensors; // voltage phasor sensors
     std::vector<int8_t> bus_injections;         // bus injections, zero injection and power sensors at buses
+    int8_t total_injections{0};                 // total number of injections at buses
     bool is_possibly_ill_conditioned{false};
 };
 enum class ConnectivityStatus : std::int8_t {
@@ -75,7 +76,7 @@ ObservabilitySensorsResult scan_network_sensors(MeasuredValues<sym> const& measu
         // diagonal for bus injection measurement
         if (measured_values.has_bus_injection(bus)) {
             result.bus_injections[bus] = 1;
-            result.bus_injections.back() += 1;
+            result.total_injections += 1;
             result.flow_sensors[current_bus_entry] = 1;
             has_at_least_one_sensor = true;
             bus_neighbourhood_info[bus].status = ConnectivityStatus::node_measured; // only treat power/0 injection
@@ -351,13 +352,12 @@ inline bool find_spanning_tree_from_node(Idx start_bus, Idx n_bus,
         return false;
     };
 
-    auto try_general_connection_rules = [&local_neighbour_list, &visited, &discovered_edges, &edge_track, &current_bus,
-                                         &downwind](bool& step_success, bool current_bus_no_measurement) {
+    auto try_general_connection_rules = [&local_neighbour_list, &visited, &discovered_edges, &edge_track,
+                                         &current_bus](bool& step_success, bool current_bus_no_measurement) {
         // Helper lambda to handle common edge processing logic
-        auto process_edge = [&local_neighbour_list, &visited, &discovered_edges, &edge_track, &current_bus, &downwind,
+        auto process_edge = [&local_neighbour_list, &visited, &discovered_edges, &edge_track, &current_bus,
                              &step_success](auto& neighbour, ConnectivityStatus neighbour_status,
-                                            ConnectivityStatus reverse_status, bool use_current_node,
-                                            bool set_upwind = false) {
+                                            ConnectivityStatus reverse_status, bool use_current_node) {
             discovered_edges.emplace_back(current_bus, neighbour.bus);
             edge_track.emplace_back(current_bus, neighbour.bus);
             visited[current_bus] = BusVisited::Visited;
@@ -380,11 +380,6 @@ inline bool find_spanning_tree_from_node(Idx start_bus, Idx n_bus,
                 local_neighbour_list[neighbour.bus].status = ConnectivityStatus::has_no_measurement;
             }
 
-            // Set direction flag if needed
-            if (set_upwind) {
-                downwind = false; // upwind
-            }
-
             current_bus = neighbour.bus;
             step_success = true;
             return true;
@@ -398,20 +393,15 @@ inline bool find_spanning_tree_from_node(Idx start_bus, Idx n_bus,
             bool const neighbour_bus_has_no_measurement =
                 local_neighbour_list[neighbour.bus].status == ConnectivityStatus::has_no_measurement;
 
-            if (!current_bus_no_measurement && (downwind && neighbour_bus_has_no_measurement)) {
-                // Case: current has measurement, downwind, neighbour empty
-                return process_edge(neighbour, ConnectivityStatus::branch_discovered_with_to_node_sensor,
-                                    ConnectivityStatus::branch_discovered_with_from_node_sensor, true);
-            }
-            if (!current_bus_no_measurement && (downwind || neighbour_bus_has_no_measurement)) {
-                // Case: current has measurement, (downwind OR neighbour empty)
+            if (!current_bus_no_measurement && neighbour_bus_has_no_measurement) {
+                // Case: current has measurement, neighbour empty (not in downwind mode)
                 return process_edge(neighbour, ConnectivityStatus::branch_discovered_with_from_node_sensor,
                                     ConnectivityStatus::branch_discovered_with_to_node_sensor, true);
             }
             if (!neighbour_bus_has_no_measurement) {
                 // Case: neighbour has measurement
                 return process_edge(neighbour, ConnectivityStatus::branch_discovered_with_from_node_sensor,
-                                    ConnectivityStatus::branch_discovered_with_to_node_sensor, false, true);
+                                    ConnectivityStatus::branch_discovered_with_to_node_sensor, false);
             }
         }
         return false;
@@ -420,10 +410,10 @@ inline bool find_spanning_tree_from_node(Idx start_bus, Idx n_bus,
     // Helper function to reassign nodal measurement between two connected nodes
     auto reassign_nodal_measurement = [&local_neighbour_list](Idx from_node, Idx to_node) {
         // no reassignment possible if reached via edge measurement
-        auto const branch_it =
-            std::ranges::find_if(local_neighbour_list[from_node].direct_neighbours,
-                                 [to_node](auto const& neighbour) { return neighbour.bus == to_node; });
-        if (branch_it != local_neighbour_list[from_node].direct_neighbours.end() &&
+        if (auto const branch_it =
+                std::ranges::find_if(local_neighbour_list[from_node].direct_neighbours,
+                                     [to_node](auto const& neighbour) { return neighbour.bus == to_node; });
+            branch_it != local_neighbour_list[from_node].direct_neighbours.end() &&
             branch_it->status == ConnectivityStatus::branch_native_measurement_consumed) {
             return;
         }
@@ -447,7 +437,7 @@ inline bool find_spanning_tree_from_node(Idx start_bus, Idx n_bus,
         // Find and update the reverse connection from to_node to from_node
         for (auto& neighbour : local_neighbour_list[to_node].direct_neighbours) {
             if (neighbour.bus == from_node) {
-                // Change to downstream connection (from from_node to to_node perspective)
+                // Change from from_node to to_node perspective
                 neighbour.status = ConnectivityStatus::branch_discovered_with_from_node_sensor;
                 break;
             }
@@ -509,10 +499,7 @@ inline bool find_spanning_tree_from_node(Idx start_bus, Idx n_bus,
 }
 
 inline bool
-sufficient_condition_meshed_without_voltage_phasor(std::vector<detail::BusNeighbourhoodInfo> const& _neighbour_list) {
-    // make a copy of the neighbour list
-    std::vector<detail::BusNeighbourhoodInfo> const neighbour_list = _neighbour_list;
-
+sufficient_condition_meshed_without_voltage_phasor(std::vector<detail::BusNeighbourhoodInfo> const& neighbour_list) {
     auto const n_bus = static_cast<Idx>(neighbour_list.size());
     std::vector<Idx> starting_candidates;
     prepare_starting_nodes(neighbour_list, n_bus, starting_candidates);
@@ -563,7 +550,7 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     }
 
     //  Sufficient early out, enough nodal measurement equals observable
-    if (observability_sensors.bus_injections.back() > n_bus - 2) {
+    if (observability_sensors.total_injections > n_bus - 2) {
         return ObservabilityResult{.is_observable = true,
                                    .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
     }


tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -2,12 +2,29 @@
...
+using power_grid_model::math_solver::detail::ConnectivityStatus;
+using power_grid_model::math_solver::detail::expand_neighbour_list;
+using power_grid_model::math_solver::detail::necessary_condition;
+using power_grid_model::math_solver::detail::ObservabilityNNResult;
+using power_grid_model::math_solver::detail::ObservabilitySensorsResult;
+using power_grid_model::math_solver::detail::prepare_starting_nodes;
+using power_grid_model::math_solver::detail::scan_network_sensors;
+using power_grid_model::math_solver::detail::starting_from_node;
+using power_grid_model::math_solver::detail::sufficient_condition_meshed_without_voltage_phasor;
+using power_grid_model::math_solver::detail::sufficient_condition_radial_with_voltage_phasor;
```
mgovers: please put `using` declarations inside an anonymous `namespace` and after the last `#include` statement or you may run into unexpected linking errors

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +57,1330 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[3].direct_neighbours = {{.bus = 1, .status = ConnectivityStatus::has_no_measurement}};
+
+        std::vector<Idx> starting_candidates;
+        prepare_starting_nodes(neighbour_list, 4, starting_candidates);
+
+        // Should find buses 1 and 3 as starting candidates
+        // (nodes without measurements and all edges have no edge measurements)
+        CHECK(starting_candidates.size() == 2);
+        CHECK(std::find(starting_candidates.begin(), starting_candidates.end(), 1) != starting_candidates.end());
+        CHECK(std::find(starting_candidates.begin(), starting_candidates.end(), 3) != starting_candidates.end());
```
mgovers: here and elsewhere: shorter is to use

```suggestion
        CHECK(std::ranges::find(starting_candidates, 1) != starting_candidates.end());
        CHECK(std::ranges::find(starting_candidates, 3) != starting_candidates.end());
```

hopefully, we will have `contains` in the near future (currently not yet implemented by some compilers even though it's super simple 😢 )

```suggestion
        CHECK(std::ranges::contains(starting_candidates, 1));
        CHECK(std::ranges::contains(starting_candidates, 3));
```

Jerry-Jinfeng-Guo:
`do-not-merge`. Only merge after https://github.com/PowerGridModel/power-grid-model/pull/1136

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1492 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        CHECK(neighbour_results.size() == 3);
+        for (size_t i = 0; i < neighbour_results.size(); ++i) {
+            CHECK(neighbour_results[i].bus == static_cast<Idx>(i));
+        }
+
+        // Bus 2 should have node_measured status due to injection sensor
+        CHECK(neighbour_results[2].status == ConnectivityStatus::node_measured);
+    }
+
+    SUBCASE("Meshed network") {
```
Jerry-Jinfeng-Guo, line 159: @figueroa1395 Please check this test case to verify the `break` logic in the core PR

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
-TEST_CASE("Necessary observability check") {
+// Original integration tests
+TEST_CASE("Observable voltage sensor - basic integration test") {
+    MathModelTopology topo;
+    topo.slack_bus = 0;
+    topo.phase_shift = {0.0, 0.0, 0.0};
+    topo.branch_bus_idx = {{0, 1}, {1, 2}};
+    topo.sources_per_bus = {from_sparse, {0, 1, 1, 1}};
+    topo.shunts_per_bus = {from_sparse, {0, 0, 0, 0}};
+    topo.load_gens_per_bus = {from_sparse, {0, 0, 0, 0}};
```
figueroa1395, line 57: Can you add `load_gens` everywhere to “replicate” mathematical behavior? That way we don’t have “hidden” injection sensors that are harder to keep track of, but we have everything we need or don’t need explicitly. In addition, since there are no loads, I believe this grid should be observable even without a single sensor in place.

This may apply to other tests as well where `check_observable` or `check_not_observable` is used.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        CHECK(result.bus_injections.size() == 4);         // n_bus + 1
+
+        // Verify voltage phasor sensors
+        CHECK(result.voltage_phasor_sensors[0] == 1); // Bus 0 has voltage phasor (complex measurement)
+        CHECK(result.voltage_phasor_sensors[1] == 0); // Bus 1 has only magnitude (no angle)
+        CHECK(result.voltage_phasor_sensors[2] == 0); // Bus 2 has no voltage sensor
+
+        // Verify bus injections - should count the bus injection sensor at bus 2
+        CHECK(result.bus_injections[2] == 1);     // Bus 2 has injection sensor
+        CHECK(result.bus_injections.back() >= 1); // Total count should be at least 1
```
figueroa1395: I would prefer we check here against the exact number of sensors including zero injection ones.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
```
figueroa1395, line 147: Should we also test for `possibly_ill_conditioned` in `result`?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        // Verify neighbour results structure
+        CHECK(neighbour_results.size() == 3);
+        for (size_t i = 0; i < neighbour_results.size(); ++i) {
+            CHECK(neighbour_results[i].bus == static_cast<Idx>(i));
+        }
+
+        // Bus 2 should have node_measured status due to injection sensor
+        CHECK(neighbour_results[2].status == ConnectivityStatus::node_measured);
```
figueroa1395: Can you also check for `ConnectivityStatus::branch_native_measurement_unused` for branch 0?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        // branch 3: bus2-bus4, branch 4: bus3-bus5, branch 5: bus4-bus5
+        topo.branch_bus_idx = {{0, 2}, {1, 2}, {2, 3}, {2, 4}, {3, 5}, {4, 5}};
+
+        topo.sources_per_bus = {from_sparse, {0, 1, 1, 1, 1, 1, 1}};
+        topo.shunts_per_bus = {from_sparse, {0, 0, 0, 0, 0, 0, 0}};
+        topo.load_gens_per_bus = {from_sparse, {0, 0, 0, 0, 0, 0, 0}}; // No load_gens for simplicity
+
+        // Power sensors: bus 0, bus 4 have injection sensors (2 total sensors)
+        // Format: bus0 has sensors [0:1), bus1 has [1:1), bus2 has [1:1), bus3 has [1:1), bus4 has [1:2), bus5 has
+        // [2:2)
```
figueroa1395, line 187: I think this ("Format:...") comment is very confusing. The above one is sufficient. I would prefer to remove it; same for some similar ones below.

This one may be stylistic though, so if it helps you go through things more quickly, feel free to leave and resolve this.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
+        // Basic size checks
+        CHECK(result.flow_sensors.size() > 0);
+        CHECK(result.voltage_phasor_sensors.size() == 6);
+        CHECK(result.bus_injections.size() == 7);
+
+        // Check that we have the expected sensor arrays
+        CHECK(result.flow_sensors.size() == y_bus.y_bus_structure().row_indptr.back());
+        CHECK(result.voltage_phasor_sensors.size() == 6); // n_bus
+        CHECK(result.bus_injections.size() == 7);         // n_bus + 1
```
figueroa1395: Why the first 3 if you inmediately after have the second 3?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        // Initialize all measurement vectors to correct sizes first
+        se_input.measured_voltage.resize(topo.voltage_sensors_per_bus.element_size());
+        se_input.measured_bus_injection.resize(topo.power_sensors_per_bus.element_size());
+        se_input.measured_branch_from_power.resize(topo.power_sensors_per_branch_from.element_size());
+        se_input.measured_branch_to_power.resize(topo.power_sensors_per_branch_to.element_size());
+        se_input.measured_branch_from_current.resize(topo.current_sensors_per_branch_from.element_size());
+        se_input.measured_branch_to_current.resize(topo.current_sensors_per_branch_to.element_size());
+        se_input.measured_shunt_power.resize(topo.power_sensors_per_shunt.element_size());
+        se_input.measured_load_gen_power.resize(topo.power_sensors_per_load_gen.element_size());
+        se_input.measured_source_power.resize(topo.power_sensors_per_source.element_size());
```
figueroa1395, line 222: Nitpick: Seeing this pattern multiple times, maybe we can make it into a helper function.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: You are right, the `break` in the logic PR indeed doesn't affect this case. I tried modifying the case locally as I believed I could break it, but I wasn't able to do so. The logic is indeed sound.

Nevertheless, could you add an extra branch sensor between buses 2 and 4? This is so it's easy to see from the test that the inner loop (containing the `break`) doesn't affect two neighboring branches that are measured, as the `if` conditional in the logic can make it confusing since the traversing logic of the y-bus is highly non-trivial.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        CHECK(neighbour_results[0].status == ConnectivityStatus::node_measured); // bus 0 has injection sensor
+        CHECK(neighbour_results[1].status ==
+              ConnectivityStatus::node_measured); // bus 1 has pseudo measurement (zero injection)
+        CHECK(neighbour_results[2].status == ConnectivityStatus::node_measured); // bus 2 has voltage sensor
+        CHECK(neighbour_results[2].direct_neighbours.size() == 2);               // bus 2 has 2 neighbours
+        CHECK(neighbour_results[3].status ==
+              ConnectivityStatus::node_measured); // bus 3 has pseudo measurement (zero injection)
+        CHECK(neighbour_results[4].status == ConnectivityStatus::node_measured); // bus 4 has injection sensor
+        CHECK(neighbour_results[5].status ==
+              ConnectivityStatus::node_measured); // bus 5 has pseudo measurement (zero injection)
```
figueroa1395: Seeing you didn't test explicitly for `ConnectivityStatus::branch_native_measurement_unused` in the buses' neighbors here either, is there a reason why? Or was it just forgotten? Maybe I'm overlooking something.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
+        // All sensor vectors should be initialized but empty/zero
+        CHECK(result.flow_sensors.size() == y_bus.y_bus_structure().row_indptr.back());
+        CHECK(result.voltage_phasor_sensors.size() == 1);
+        CHECK(result.bus_injections.size() == 2);
+
+        // All sensors should be zero
+        CHECK(std::ranges::all_of(result.flow_sensors, [](int8_t val) { return val == 0; }));
+        CHECK(std::ranges::all_of(result.voltage_phasor_sensors, [](int8_t val) { return val == 0; }));
+        CHECK(result.bus_injections.back() == 0); // No bus injections
```
figueroa1395, line 334: Nitpick: Maybe I'd make sense to add one additional "empty" bus and check that there is one bus injection coming from the zero injection constraint?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
+        // Should be marked as possibly ill-conditioned due to no sensors
+        CHECK(result.is_possibly_ill_conditioned == false);
+    }
+
+    SUBCASE("Mixed sensor types") {
```
figueroa1395, line 340: Can you extend this test a bit and add one sensor of each type? And then make sure that all are being treaten correctly? Also at different from/to positions. In addition, maybe here as well, could you check that two sensors in the same branch just count as one flow sensor?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[1].bus = 1;
+        neighbour_list[1].status = ConnectivityStatus::has_no_measurement;
+        neighbour_list[1].direct_neighbours = {{.bus = 0, .status = ConnectivityStatus::has_no_measurement},
+                                               {.bus = 3, .status = ConnectivityStatus::has_no_measurement}};
+
+        // Bus 2: has measurement
+        neighbour_list[2].bus = 2;
+        neighbour_list[2].status = ConnectivityStatus::node_measured;
+        neighbour_list[2].direct_neighbours = {
+            {.bus = 0, .status = ConnectivityStatus::branch_native_measurement_unused}};
```
figueroa1395: Nitpick: Could you connect nodes 2 and 3 and make it meshed? Shouldn't matter now, but to keep meshed in mind if we change the logic later for whatever reason.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1516 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+    SUBCASE("Empty network") {
+        // Edge case: empty network
+        std::vector<BusNeighbourhoodInfo> neighbour_list;
+        std::vector<Idx> starting_candidates;
+
+        prepare_starting_nodes(neighbour_list, 0, starting_candidates);
+
+        // Should fallback to first node (0) even with empty network
+        CHECK(starting_candidates.size() == 1);
+        CHECK(starting_candidates[0] == 0);
```
figueroa1395, line 517: Is this desired? Shouldn't we `throw` here?

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: It explains the sparse list used below.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: I don't get your question here, could you elaborate?

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: There is a dedicated unit test with meshed topo.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: It's allowed. It's out of jurisdiction here to throw error.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: Zero injections are verified in other tests with no `load_gen`s

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: This pattern happens only once here

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: This can go easily out of hand. Will require some EE input to proceed.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: In theory I would agree. However, making the whole topo thing to work in this unit tests is unreasonably complex. I would deem an addition to have after the refactoring. Namely, when measured value is workable.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: I know it is allowed, my question is, should we think about disallowing it? Not necessarily here, but this test case made me raise this question. What is the purpose of having an empty grid?

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: For `scan_network_sensors` there is. There isn't any for `prepare_starting_nodes`.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Why? From the observability perspective a sensor is a sensor in spite of its type, right? And we know which sensors can (not) be on each component; in addition, the measured values are just "arbitrary", so we don't need input there.

There can be a power sensor in terminal A of a branch, then a current sensor at the other terminal, a zero injection in one node, a power sensor in another node, a phasor sensor in another node, and we are done. I don't mean testing all the combinatory possibilities, but just that all are parsed correctly.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: But if you add `topo.load_gens_per_bus` then you only have to add the respective `topo.load_gen_type` and `se_input.load_gen_status`; the rest should remain the same including the `measured_values`. See `TEST_CASE("Test Observability - Necessary check end to end test")`.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: You first do the checks
```
CHECK(result.flow_sensors.size() > 0);
        CHECK(result.voltage_phasor_sensors.size() == 6);
        CHECK(result.bus_injections.size() == 7);
```

and then
```
 CHECK(result.flow_sensors.size() == y_bus.y_bus_structure().row_indptr.back());
        CHECK(result.voltage_phasor_sensors.size() == 6); // n_bus
        CHECK(result.bus_injections.size() == 7);         // n_bus + 1
```
Doesn't the second block already covers the first?

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: I confused things. Resolving.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Resolved in [05d126f](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/05d126f532bb9f7b0b5786bab8df23d3f52aa098)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Resolved in [05d126f](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/05d126f532bb9f7b0b5786bab8df23d3f52aa098)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Resolved in [05d126f](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/05d126f532bb9f7b0b5786bab8df23d3f52aa098)

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1485 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[0].status = has_no_measurement;
+        neighbour_list[0].direct_neighbours = {{1, has_no_measurement}, {2, node_measured}};
+
+        neighbour_list[1].bus = 1;
+        neighbour_list[1].status = node_measured;
+        neighbour_list[1].direct_neighbours = {{0, has_no_measurement}};
+
+        neighbour_list[2].bus = 2;
+        neighbour_list[2].status = node_measured;
+        neighbour_list[2].direct_neighbours = {{0, has_no_measurement}};
```
figueroa1395: I think we should leave `neighbour_list[1].direct_neighbours` and `neighbour_list[2].direct_neighbours` initially empty and test that `complete_bidirectional_neighbourhood_info()` actually completes it; this because nodes 1 and 2 are both downstream. It may be a good idea to also connect 1 and 2 so we have a mesh to test this functionality for this case.
```suggestion
        neighbour_list[1].bus = 1;
        neighbour_list[1].status = node_measured;

        neighbour_list[2].bus = 2;
        neighbour_list[2].status = node_measured;
```

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1485 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
+        // Test the function
+        complete_bidirectional_neighbourhood_info(neighbour_list);
+
+        // Basic verification - structure should be maintained
+        CHECK(neighbour_list.size() == 3);
+        CHECK(neighbour_list[0].bus == 0);
+        CHECK(neighbour_list[1].bus == 1);
+        CHECK(neighbour_list[2].bus == 2);
```
figueroa1395, line 586: Can you extend the `CHECK`s to something along the lines of:

```suggestion
        CHECK(neighbour_list.size() == 3);
        CHECK(neighbour_list[0].bus == 0);
        CHECK(neighbour_list[1].bus == 1);
        CHECK(neighbour_list[2].bus == 2);
        CHECK(neighbour_list[1].direct_neighbours[0].bus == 0);
        CHECK(neighbour_list[2].direct_neighbours[0].bus == 0);
        CHECK(neighbour_list[1].direct_neighbours[0].status == has_no_measurement);
        CHECK(neighbour_list[2].direct_neighbours[0].status == has_no_measurement);
        ...
```

That to check that the function actually extended the neighbour list bidirectionally.

Note: I wrote the suggestion quickly, it may contain errors.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: I don't think an error is granted for an empty network anywhere, it just aligns with the rest of the tests (in other places) where an empty input is provided.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: > Shouldn't matter now

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: > From the observability perspective a sensor is a sensor in spite of its type, right? And we know which sensors can (not) be on each component; in addition, the measured values are just "arbitrary", so we don't need input there.

I don't get what you are saying here

As mentioned in other comments, the foundations on which the functionalities here depend on are not in the most easy workable state.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: Copy - paste error

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+    }
+
+    SUBCASE("Empty neighbour list") {
+        std::vector<BusNeighbourhoodInfo> empty_list;
+        CHECK_NOTHROW(complete_bidirectional_neighbourhood_info(empty_list));
+        CHECK(empty_list.empty());
+    }
+}
+
+TEST_CASE("Test Observability - assign_independent_sensors_radial") {
```
figueroa1395, line 742: Even though I don't like this unit test, I fully understand why you did the testing this way and I have to agree with it; the y-bus coupling makes it a horrible task to try to test this well to the point it's not worth it if we are refactoring it. But let's not forget to address this later.

Could you please add a `// TODO: properly clean up after y-bus access refactoring` or something like that?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        CHECK_NOTHROW(assign_independent_sensors_radial(y_bus_struct, flow_sensors, voltage_phasor_sensors));
+
+        // Last bus injection should be removed if there are buses
+        if (n_bus > 0) {
+            CHECK(flow_sensors[y_bus_struct.bus_entry[n_bus - 1]] == 0);
+        }
+    }
+}
+
+TEST_CASE("Test Observability - find_spanning_tree_from_node") {
```
figueroa1395, line 849: Since `find_spanning_tree_from_node` is only meant to be used in meshed grids (at least for now), can you make the all the tests grid under this test case meshed? Keep the logic the same, but add extra connections to make a mesh

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+    using power_grid_model::math_solver::detail::BusNeighbourhoodInfo;
+    using enum power_grid_model::math_solver::detail::ConnectivityStatus;
+    using power_grid_model::math_solver::detail::find_spanning_tree_from_node;
+
+    SUBCASE("Simple spanning tree with native edge measurements") {
```
figueroa1395, line 854: I don't know whether this will happen in the current logic PR, at a later refactor or not at all, but good to keep in mind that these tests may be split into testing the different parts that make up `find_spanning_tree_from_node` if we decide to remove the `lambda`s and make them standalone functions (my preference). The same testing logic should still hold though.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[1].status = has_no_measurement;
+        neighbour_list[1].direct_neighbours = {{.bus = 0, .status = has_no_measurement},
+                                               {.bus = 2, .status = has_no_measurement}};
+
+        // Bus 2: has measurement
+        neighbour_list[2].bus = 2;
+        neighbour_list[2].status = node_measured;
+        neighbour_list[2].direct_neighbours = {{.bus = 1, .status = has_no_measurement}};
+
+        Idx const start_bus = 0;
```
figueroa1395: According to the logic in `prepare_starting_nodes`, shouldn't the starting node be 1? Or else we would be testing against an "impossible" starting node, wouldn't we?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        Idx const n_bus = 3;
+
+        bool const result = find_spanning_tree_from_node(start_bus, n_bus, neighbour_list);
+
+        // Should work with measurements at both ends of the chain
+        CHECK((result == true || result == false)); // Algorithm may not always find spanning tree
```
figueroa1395: This is observable, and radial even, why wouldn't the algorithm always find the spanning tree if it is deterministic? If that is the case, then something needs revisiting. Why not just `CHECK(result == true);?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[2].status = has_no_measurement;
+        neighbour_list[2].direct_neighbours = {{.bus = 1, .status = has_no_measurement},
+                                               {.bus = 3, .status = has_no_measurement}};
+
+        // Bus 3: has measurement
+        neighbour_list[3].bus = 3;
+        neighbour_list[3].status = node_measured;
+        neighbour_list[3].direct_neighbours = {{.bus = 2, .status = has_no_measurement}};
+
+        Idx const start_bus = 0;
```
figueroa1395: Shouldn't the starting bus be 2 as it is the first in the list according to `prepare_starting_nodes`?

Moreover, should we test in all this subcases against all possible starting nodes just to go over robustness of the core of the algorithm?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[2].status = has_no_measurement;
+        neighbour_list[2].direct_neighbours = {}; // Isolated
+
+        Idx const start_bus = 0;
+        Idx const n_bus = 3;
+
+        bool const result = find_spanning_tree_from_node(start_bus, n_bus, neighbour_list);
+
+        // Should fail because bus 2 is isolated and cannot be reached
+        CHECK(result == false);
```
figueroa1395, line 974: `result == false` is true even if I comment out bus 2 and change to `n_bus = 2`. It fails because the first two aren't observable on their own.

In addition, would we reach such a case? I might be wrong, but I believe that disconnected networks are treated as different networks from the topology perspective and will probably reach this separately. Perhaps @nitbharambe knows this by heart? In any case, this can still be tested, but making sure the first two buses constitute an observable grid themselves.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[3].status = node_measured;
+        neighbour_list[3].direct_neighbours = {{.bus = 2, .status = has_no_measurement}};
+
+        Idx const start_bus = 0;
+        Idx const n_bus = 4;
+
+        bool const result = find_spanning_tree_from_node(start_bus, n_bus, neighbour_list);
+
+        // Should successfully build spanning tree using combination of edge and node measurements
+        CHECK((result == true || result == false)); // Algorithm may not always find spanning tree
```
figueroa1395: Why isn't it just `CHECK(result == true)`? I see this in multiple places. Why aren't you sure of these results?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[0].direct_neighbours = {}; // No neighbours
+
+        Idx const start_bus = 0;
+        Idx const n_bus = 1;
+
+        // Just test that the function executes without crashing
+        bool const result = find_spanning_tree_from_node(start_bus, n_bus, neighbour_list);
+
+        // Don't make assumptions about the result - just verify it returns a boolean
+        CHECK((result == true || result == false));
```
figueroa1395: If we are only checking that we can handle this
```suggestion
        // Just test that the function executes without crashing
        // Don't make assumptions about the result - just verify it returns a boolean
        CHECK_NO_THROW(find_spanning_tree_from_node(start_bus, n_bus, neighbour_list);
```

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        neighbour_list[3].direct_neighbours = {{.bus = 1, .status = has_no_measurement}};
+
+        Idx const start_bus = 0;
+        Idx const n_bus = 4;
+
+        // Test that function executes and returns a boolean result
+        bool const result = find_spanning_tree_from_node(start_bus, n_bus, neighbour_list);
+
+        // Verify function completes and returns valid boolean
+        CHECK((result == true || result == false));
```
figueroa1395: Asserting for the right result also verifies the flow, right? Why not just do that?

tests/cpp_unit_tests/test_observability.cpp
figueroa1395, line 849: I didn't check in this detail, but can you make sure that all the tested (meshed) grids cover that we are indeed using `try_native_edge_measurements`, `try_downwind_measurement`, `try_general_connection_rules`, `reassign_nodal_measurement` and `try_backtrack`?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
+TEST_CASE("Test Observability - necessary_condition") {
+    using power_grid_model::math_solver::detail::necessary_condition;
+    using power_grid_model::math_solver::detail::ObservabilitySensorsResult;
+
+    SUBCASE("Sufficient measurements") {
+        ObservabilitySensorsResult sensors;
+        sensors.flow_sensors = {1, 1, 0, 1};
+        sensors.voltage_phasor_sensors = {1, 0, 1};
+        sensors.bus_injections = {2, 2, 3}; // cumulative count ending at 3
```
figueroa1395: I'm confused about this. Size should be `n_bus + 1` and it should only have `0`s or `1`s except for the last entry.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+    }
+}
+
+TEST_CASE("Test Observability - necessary_condition") {
+    using power_grid_model::math_solver::detail::necessary_condition;
+    using power_grid_model::math_solver::detail::ObservabilitySensorsResult;
+
+    SUBCASE("Sufficient measurements") {
+        ObservabilitySensorsResult sensors;
+        sensors.flow_sensors = {1, 1, 0, 1};
```
figueroa1395, line 1139: The size of `flow_sensors` should be (at least) the `n_bus` + `n_branch`.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        sensors.voltage_phasor_sensors = {1, 0, 1};
+        sensors.bus_injections = {2, 2, 3}; // cumulative count ending at 3
+        sensors.is_possibly_ill_conditioned = false;
+
+        Idx const n_bus = 3;
+        Idx n_voltage_phasor = 2;
```
figueroa1395: This should be correctly assigned inside the function.
```suggestion
        Idx n_voltage_phasor{};
```

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        CHECK_THROWS_AS(necessary_condition(sensors, n_bus, n_voltage_phasor, false), NotObservableError);
+    }
+
+    SUBCASE("Empty sensors") {
+        ObservabilitySensorsResult const sensors;
+        // All vectors empty - should not be observable
+
+        Idx n_voltage_phasor = 0;
+        CHECK_NOTHROW(necessary_condition(sensors, 0, n_voltage_phasor, false));
+        // Edge case: no buses means trivially observable
```
figueroa1395: The first comment seems contradictory with the last one. Can you just have one coment saying this trivially passes necessary condition because it is a bus-less grid?

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        // Create a simple 4-bus radial network: bus0--bus1--bus2--bus3
+        MathModelTopology topo;
+        topo.slack_bus = 0;
+        topo.is_radial = true;
+        topo.phase_shift = {0.0, 0.0, 0.0, 0.0};
+        topo.branch_bus_idx = {{0, 1}, {1, 2}, {2, 3}};
+        topo.sources_per_bus = {from_sparse, {0, 1, 1, 1, 1}};
+        topo.shunts_per_bus = {from_sparse, {0, 0, 0, 0, 0}};
+        topo.load_gens_per_bus = {from_sparse, {0, 0, 0, 0, 0}};
+        topo.power_sensors_per_bus = {from_sparse, {0, 1, 1, 2, 2}}; // Injection sensors at bus 0 and 1
```
figueroa1395: This sensors are at buses 0 and 2

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+            std::ranges::fold_left(observability_sensors.voltage_phasor_sensors, 0, std::plus<>{});
+
+        // Test sufficient_condition_radial_with_voltage_phasor
+        CHECK_NOTHROW(sufficient_condition_radial_with_voltage_phasor(y_bus.y_bus_structure(), observability_sensors,
+                                                                      n_voltage_phasor_sensors));
+
+        // Verify that it returns true (no exception thrown means observable)
+        bool const result = sufficient_condition_radial_with_voltage_phasor(
+            y_bus.y_bus_structure(), observability_sensors, n_voltage_phasor_sensors);
+        CHECK(result == true);
```
figueroa1395:
```suggestion
        // Verify that it returns true (no exception thrown means observable)
        bool const result = sufficient_condition_radial_with_voltage_phasor(
            y_bus.y_bus_structure(), observability_sensors, n_voltage_phasor_sensors);
        CHECK(result == true);
```

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        Idx const initial_voltage_sensors =
+            std::ranges::fold_left(observability_sensors.voltage_phasor_sensors, 0, std::plus<>{});
+
+        // Count voltage phasor sensors for the function
+        Idx const n_voltage_phasor_sensors = initial_voltage_sensors;
+
+        // Test that the function works and modifies the sensor vectors
+        bool const result = sufficient_condition_radial_with_voltage_phasor(
+            y_bus.y_bus_structure(), observability_sensors, n_voltage_phasor_sensors);
+        CHECK(result == true);
```
figueroa1395, line 1301: Without loads this will always return true as every node will have an inherent zero injection measurement.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        Idx const final_flow_sensors = std::ranges::fold_left(observability_sensors.flow_sensors, 0, std::plus<>{});
+        Idx const final_voltage_sensors =
+            std::ranges::fold_left(observability_sensors.voltage_phasor_sensors, 0, std::plus<>{});
+
+        // For a 3-bus radial network, should have 2 independent flow sensors
+        CHECK(final_flow_sensors >= 2);
+
+        // Should retain at least 1 voltage phasor sensor as reference if we started with any
+        if (n_voltage_phasor_sensors > 0) {
+            CHECK(final_voltage_sensors >= 1);
```
figueroa1395, line 1313: Let's test against the exact number once loads are added.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
+        // Count voltage phasor sensors (should be 0)
+        Idx const n_voltage_phasor_sensors =
+            std::ranges::fold_left(observability_sensors.voltage_phasor_sensors, 0, std::plus<>{});
+        CHECK(n_voltage_phasor_sensors == 0);
+
+        // Should pass with sufficient flow sensors even without voltage phasor sensors
+        bool const result = sufficient_condition_radial_with_voltage_phasor(
+            y_bus.y_bus_structure(), observability_sensors, n_voltage_phasor_sensors);
+        CHECK(result == true);
```
figueroa1395, line 1368: Same as above.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+        // Bus 0: has measurement
+        neighbour_list[0].bus = 0;
+        neighbour_list[0].status = node_measured;
+        neighbour_list[0].direct_neighbours = {{.bus = 1, .status = has_no_measurement},
+                                               {.bus = 3, .status = has_no_measurement}};
+
+        // Bus 1: no measurement, connected to measured nodes
+        neighbour_list[1].bus = 1;
+        neighbour_list[1].status = has_no_measurement;
+        neighbour_list[1].direct_neighbours = {{.bus = 0, .status = has_no_measurement},
```
figueroa1395, line 1440: I believe this has gone over the scope of `scan_network_sensors()` and is already doing what `complete_bidirectional_neighbourhood_info()` should do. Can you confirm? If so, let's remove the "downstream" direction to see that `complete_bidirectional_neighbourhood_info()` works or remove the use of `complete_bidirectional_neighbourhood_info()`.

This potentially applies to other subcases below.

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -40,7 +46,1488 @@ void check_not_observable(MathModelTopology const& topo, MathModelParam<symmetri
...
+
+        bool const result = sufficient_condition_meshed_without_voltage_phasor(neighbour_list);
+
+        // Highly connected network with multiple measurements should be observable
+        CHECK(result == true);
+    }
+
+    SUBCASE("Performance test with larger network") {
+        // Test with a larger meshed network to verify algorithm doesn't hang
+        constexpr Idx n_bus = 8;
```
figueroa1395, line 1694: Should we go larger? Following the nice "generic" rules below we should be able to, right?

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: Not necessarily. But one can be added indeed.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: Not in this PR.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: This one was missed

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: It won't matter actually here (the starting is meant for something else). But can indeed be changed nonetheless.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: This is test case "sufficient_condition_meshed_without_voltage_phasor"

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: No.

figueroa1395: Why is 8 large enough?

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Can you elaborate with the answer? I know what test case this is, the suggestion still applies.

tests/cpp_unit_tests/test_observability.cpp
Jerry-Jinfeng-Guo: See subcase ""Performance test with larger network""

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Offline discussion: complexity grows combinatorially, this is complex and "large" enough. If a really large grid is to be tested, let's add as a validation test in a follow up (probably not needed).

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Addressed in [acdc421](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/acdc421ce843941e04fb6831a04840b4a189d8e1). Resolving.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [77e9ec1](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/77e9ec1a42bf95acf49a73a732bdff60607e20a8). Resolving

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [77e9ec1](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/77e9ec1a42bf95acf49a73a732bdff60607e20a8)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [77e9ec1](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/77e9ec1a42bf95acf49a73a732bdff60607e20a8)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [77e9ec1](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/77e9ec1a42bf95acf49a73a732bdff60607e20a8)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [77e9ec1](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/77e9ec1a42bf95acf49a73a732bdff60607e20a8)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Resolved in [05d126f](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/05d126f532bb9f7b0b5786bab8df23d3f52aa098)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Resolved in [05d126f](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/05d126f532bb9f7b0b5786bab8df23d3f52aa098)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Considering we will decouple the sensor logic in the future from this test, let's skip such details then as it is not easily workable. Resolving.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Seems like I commented in the wrong location.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Very well done with what you did in [79824f0](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/79824f00bfa975145cc4492b1195b7fa8fa48b23). Resolving.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: I see you addressed it anyways, thanks. Resolving.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Idem. Resolving

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: [06710f4](https://github.com/PowerGridModel/power-grid-model/pull/1146/commits/06710f4f0e20027b6130432af9ad5c7d03c58bfb)

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Offline: These were added in specific test cases. Adding them everywhere doesn't add anything to the test logic. Resolving.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: This was taken care of as subcases. Resolving.

tests/cpp_unit_tests/test_observability.cpp
figueroa1395: Addressed.

figueroa1395 (approved):
No further comments from me. Ready to be merged when the parent is.

Jerry-Jinfeng-Guo:
`do-not-merge` removed and pending merge to main once https://github.com/PowerGridModel/power-grid-model/pull/1136 is merged.

PR #1148, opened 2025-10-03
Modernize internal implementation: iterator facade deducing this

mgovers:
This makes sure that iterator facade no longer uses CRTP but instead the more modern "deducing this" pattern, which does not require complicated two-way-friend derivation patterns and templated inheritance. Constraints on the facadeable type are enforced by the (now public) explicit constructor.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
@@ -45,7 +45,7 @@ class IteratorFacade {
         return (self <=> other) == std::strong_ordering::equivalent;
     }
 
-    // NOLINTNEXTLINE(cert-dcl21-cpp) // pre-increment but clang-tidy incorrectly sees this as post-increment
+    // NOLINTNEXTLINE(cert-dcl21-cpp) // pre-decrement but clang-tidy incorrectly sees this as post-decrement
     template <typename Self> constexpr std::add_lvalue_reference_t<Self> operator++(this Self& self) {
         if constexpr (requires { self.increment(); }) { // NOTE: IteratorFacade should be a friend class
             self.increment();
@@ -64,14 +64,14 @@ class IteratorFacade {
         return self;
     }
     template <typename Self>
-    constexpr std::add_const_t<std::remove_reference_t<Self>> operator++(this Self& self, std::integral auto /*idx*/) {
+    constexpr std::remove_cvref_t<Self> operator++(this Self& self, std::integral auto /*idx*/) {
         using Result = std::remove_cvref_t<Self>;
         Result result{self};
         ++self;
         return result;
     }
     template <typename Self>
-    constexpr std::add_const_t<std::remove_reference_t<Self>> operator--(this Self& self, std::integral auto /*idx*/) {
+    constexpr std::remove_cvref_t<Self> operator--(this Self& self, std::integral auto /*idx*/) {
         using Result = std::remove_cvref_t<Self>;
         Result result{self};
         --self;


power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
```diff
@@ -8,94 +8,111 @@
...
-    constexpr auto operator*() -> reference
-        requires requires(iterator it) {
-            { it.dereference() } -> std::same_as<reference>;
-        }
-    {
-        return static_cast<iterator*>(this)->dereference();
-    }
-    constexpr auto operator->() const -> decltype(auto) { return &(*(*this)); }
-    constexpr auto operator->() -> decltype(auto) { return &(*(*this)); }
+    constexpr auto operator->(this auto&& self) -> decltype(auto) { return &(*std::forward<decltype(self)>(self)); }
```
figueroa1395: Nitpick:
```suggestion
    constexpr decltype(auto) operator->(this auto&& self) { return &(*std::forward<decltype(self)>(self)); }
```
I prefer the leading return type in this case because it is then obvious that the `const` version is being handled right away. I feel this is more readable for such cases and I believe equivalent in practice.

Edit: Perhaps this is stylistic only, but if you agree, take a look at other places to change it.

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
```diff
@@ -8,94 +8,111 @@
...
+    template <typename Self, typename Other>
+        requires std::same_as<std::remove_cvref_t<Self>, std::remove_cvref_t<Other>>
+    constexpr bool operator==(this Self const& self, Other const& other) {
+        return (self <=> other) == std::strong_ordering::equivalent;
     }
 
-    constexpr auto operator++() -> iterator& {
-        if constexpr (requires(iterator it) { it.increment(); }) {
-            static_cast<iterator*>(this)->increment();
+    constexpr auto operator++(this auto& self) -> std::add_lvalue_reference_t<decltype(self)> {
```
figueroa1395:
```suggestion
    constexpr decltype(auto) operator++(this auto& self) {
```
Since the signature is defined with `(this auto& self)` isn't this an lvalue already making `std::add_lvalue_reference_t` redundant?

Edit: this applies to several operator overloads below as well.

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
```diff
@@ -8,94 +8,111 @@
...
+    constexpr auto operator--(this Self& self, std::integral auto /*idx*/) -> std::remove_cvref_t<Self> {
+        using Result = std::remove_cvref_t<Self>;
+        Result result{self};
+        --self;
         return result;
     }
-    constexpr auto operator+=(std::integral auto offset) -> iterator& {
-        static_cast<iterator*>(this)->advance(offset);
-        return *static_cast<iterator*>(this);
+    constexpr auto operator+=(this auto& self,
```
figueroa1395: Nitpick: Can you move this overload higher up? `++` and `--` defined above may depend on this one, and it breaks the flow a bit when reading and is defined after those.

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
```diff
@@ -8,94 +8,111 @@
...
+        return std::forward<Other>(other).distance_to(std::forward<Self>(self));
     }
+
+    template <typename Self>
+    constexpr auto operator[](this Self const& self, typename Self::difference_type idx) -> decltype(auto) {
+        return *(self + idx);
+    }
+
+    // prevent construction by non-derived and non-iterator-facadeable types
+    IteratorFacade() = delete;
```
figueroa1395, line 111: What is the advantage of deleting it over making it `protected`?

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
```diff
@@ -8,94 +8,111 @@
...
         return result;
     }
-    friend constexpr auto operator+(difference_type offset, iterator it) -> iterator { return (it += offset); }
-    friend constexpr auto operator-(iterator const& it, difference_type idx) -> iterator { return it + (-idx); }
-    friend constexpr auto operator-(IteratorFacade const& first, IteratorFacade const& second) -> difference_type {
-        return second.distance_to(first);
+    template <typename Self>
+        requires std::derived_from<std::remove_cvref_t<Self>, IteratorFacade> &&
+                     detail::iterator_facadeable_c<std::remove_cvref_t<Self>>
+    friend constexpr auto operator+(std::integral auto offset, Self&& self) -> std::remove_cvref_t<Self> {
```
figueroa1395: I suppose this is standard in definind overloads for `+` operators, but I hadn't payed attention before, very interesting. Would you quickly ellaborate in case I am missing something? Or is this just to take care of `integral + iterator` instead of `iterator + integral` (taken care of by the first overload)?

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
```diff
@@ -94,26 +94,22 @@ constexpr auto from_dense = from_dense_t{};
 
 class SparseGroupedIdxVector {
   private:
-    class GroupIterator : public IteratorFacade<GroupIterator, IdxRange const, Idx> {
-        friend class IteratorFacade<GroupIterator, IdxRange const, Idx>;
-        using Base = IteratorFacade<GroupIterator, IdxRange const, Idx>;
+    class GroupIterator : public IteratorFacade {
+        friend class IteratorFacade;
```
figueroa1395: There aren't `private` members in `IteratorFacade` and the constructor is currently deleted, this shouldn't be necessary then, right?

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
```diff
@@ -8,94 +8,111 @@
...
     }
+
+    template <typename Self>
+    constexpr auto operator[](this Self const& self, typename Self::difference_type idx) -> decltype(auto) {
+        return *(self + idx);
+    }
+
+    // prevent construction by non-derived and non-iterator-facadeable types
+    IteratorFacade() = delete;
+    template <typename Self> constexpr explicit IteratorFacade(Self&& /*self*/) {
```
figueroa1395: Tied to the deletion of the above constructor, this should make it possible to only instantiate from the derived class perspective when `Self` is determined by `this` in the derived scope, right?

tests/cpp_unit_tests/test_iterator_facade.cpp
```diff
@@ -0,0 +1,101 @@
...
+using power_grid_model::IdxRange;
+using power_grid_model::IdxVector;
+using power_grid_model::IteratorFacade;
+using power_grid_model::detail::iterator_facadeable_c;
+
+template <typename UnderlyingType> class BaseTestIterator : public IteratorFacade {
+    friend class IteratorFacade;
+
+  public:
+    using underlying = UnderlyingType;
```
figueroa1395, line 39: Why are this and the templated class needed?

tests/cpp_unit_tests/test_iterator_facade.cpp
```diff
@@ -0,0 +1,101 @@
...
+        CHECK(*it == 30);
+        it = it - 5;
+        CHECK(*it == 25);
+
+        auto const it2 = TestIterator{vec.begin() + 25};
+        CHECK(it == it2);
+        CHECK((it <=> it2) == std::strong_ordering::equivalent);
+        auto const it3 = TestIterator{vec.begin() + 30};
+        CHECK((it <=> it3) == std::strong_ordering::less);
+        CHECK((it3 <=> it) == std::strong_ordering::greater);
```
figueroa1395, line 155: Should we also test individual `<`, `>`, `<=`, etc? Or if `<=>` works, the other should as well by "default"?

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
mgovers: it's indeed stylistic only. trailing return types have the benefit of keeping the function names on the left side of the page, while pushing implementation details back. in addition, they also keep the equivalence between lambda functions and regular functions. however, since it's equivalent, I will change to leading

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
mgovers: nope because the deduced type is either `Self` or `Self const` (and maybe volatile but let's not go into those details). Therefore, we need to re-add the reference for the return type

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
mgovers: every `delete`d constructor must be public (otherwise, it is not known to other types that it is in fact `delete`d. `protected` alone still does not guarantee that `iterator_facade` is implemented. Only when forcing construction via another code path can the `static_assert` be hit.

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
mgovers: the latter, indeed 😬

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
mgovers: It's the other way around: `IteratorFacade` needs to have access to `private`/`protected` members in this class. By design, if `class A` declares `class B` as a `friend`, then `class B` has `private` access to `class A`, but not the other way around, as that would allow code injection ("I now have access to your private members, and there's nothing you can do about it" sounds like something The Joker would say).

In CRTP, the `friend` needs to be declared both ways, but in `deducing this`, only one way is necessary.

figueroa1395: I allegedly deleted this comment after I realized I was reading it the wrong way 😆 . I guess it wasn't deleted then. Resolving.

tests/cpp_unit_tests/test_iterator_facade.cpp
mgovers: cfr. C++20 standard: implementation of `operator<=>` implies handling of all other inequality comparison operators. Equality comparison operator was actually a C++23 feature that was marked as a defect report to C++20 (DR20: The Equality Operator You Are Looking For: [P2468R2](https://wg21.link/P2468R2)), meaning that all C++23 compilers are required to have that, and if they support C++20 mode, then they also are required to support that.

tests/cpp_unit_tests/test_iterator_facade.cpp
mgovers: i want to test both const iterator and non-const iterator using only one implementation

power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp
mgovers: `this` does not exist in constructors (unfortunately), so we have to do old-school templating with static-asserts

power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp
mgovers: you're right. fixed.

mgovers:
sonar cloud issues are false-positives

PR #1149, opened 2025-10-09
CI: Prebuilt binaries for python3.14 in CIBW

mgovers:
No longer skip python3.14 in CIBW as it was released on Tuesday


figueroa1395:
@mgovers Let's double check in `conda-forge` that we don't have any pin there either. I suppose that if the build there passed is fine, but good to take a Quick Look.

mgovers:
> @mgovers Let's double check in `conda-forge` that we don't have any pin there either. I suppose that if the build there passed is fine, but good to take a Quick Look.

conda-forge manages python versions. We only set minimum version

PR #1150, opened 2025-10-09
Current sensor statistics: to polar to decomposed statistics

### LINKED ISSUE #1068: [BUG] Handle current sensor for i_measured=0 and i_angle_measured=npi/2
**Describe the bug**

In #1056 we mention that the current sensor is invalid for this specific case of measured value. This issue is to handle such a case. Although unrealistic, it is more neat to support the whole range.

@mgovers suggested a nice idea to to extend the 1st order approximation of polar -> rectangular variance to 2nd order order approximation to cover this gap.

mgovers:
Fixes #1068

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
@@ -127,15 +127,29 @@ template <symmetry_tag sym_type> struct DecomposedComplexRandVar {
 };
 
 namespace detail {
+// Var(f_i) ≈ \Sum_j Var(x_j) * ||df_i/dx_j||^2                              // first order
+//              + 1/2 * \Sum_{jk} Var(x_j) * ||d^2 f_i/dx_jdx_k||^2 Var(x_k) // second order
+//
+// Re{I} = I cos(θ)
+// Im{I} = I sin(θ)
+//
+// dRe{I}/dI = cos(θ)            dIm{I}/dI = sin(θ)
+// dRe{I}/dθ = -I sin(θ)         dIm{I}/dθ = I cos(θ)
+// d^2 Re{I}/dI^2  = 0               d^2 Im{I}/dI^2  = 0
+// d^2 Re{I}/dθ^2  = -I cos(θ)       d^2 Im{I}/dθ^2  = -I sin(θ)
+// d^2 Re{I}/dIdθ = -sin(θ)        d^2 Im{I}/dIdθ = cos(θ)
+//
+// Var(Re{I}) = Var(I cos(θ))
+//            ≈ Var(I) * cos^2(θ) + Var(θ) * I^2  * sin^2(θ)                      // first order
+//               + 1/2 * Var(θ)^2  * I^2  * cos^2(θ) + Var(I) * Var(θ) * sin^2(θ) // second order
+// Var(Im{I}) = Var(I sin(θ))
+//            ≈ Var(I) * sin^2(θ) + Var(θ) * I^2  * cos^2(θ)                      // first order
+//               + 1/2 * Var(θ)^2  * I^2  * sin^2(θ) + Var(I) * Var(θ) * cos^2(θ) // second order
 inline auto compute_decomposed_variance_from_polar(auto const& magnitude, auto const& cos_angle, auto const& sin_angle,
                                                    auto const& magnitude_variance, auto const& angle_variance) {
     auto const cos2_angle = cos_angle * cos_angle;
     auto const sin2_angle = sin_angle * sin_angle;
     auto const magnitude2 = magnitude * magnitude;
-
-    // // first order approximation
-    // return std::make_pair(magnitude_variance * cos2_angle + magnitude2 * angle_variance * sin2_angle,
-    //                       magnitude_variance * sin2_angle + magnitude2 * angle_variance * cos2_angle);
 
     // second order approximation
     return std::make_pair(magnitude_variance * cos2_angle + magnitude2 * angle_variance * sin2_angle +
@@ -248,7 +262,7 @@ template <symmetry_tag sym> inline auto conj(PolarComplexRandVar<sym> var) {
 }
 
 namespace statistics {
-// Var(s x) ≈ Var(x) * ||s||²
+// Var(s x) ≈ Var(x) * ||s||^2
 template <symmetry_tag sym, template <symmetry_tag> typename RandVarType>
     requires is_in_list_c<RandVarType<sym>, UniformRealRandVar<sym>, IndependentRealRandVar<sym>,
                           UniformComplexRandVar<sym>, IndependentComplexRandVar<sym>>


tests/cpp_unit_tests/test_statistics.cpp
```diff
@@ -444,11 +452,34 @@ TEST_CASE("Test statistics") {
 
                     CHECK(decomposed.real_component.value == doctest::Approx(real(uniform.value)));
                     CHECK(decomposed.imag_component.value == doctest::Approx(imag(uniform.value)));
-                    CHECK(decomposed.real_component.variance == doctest::Approx(uniform.variance / 2));
+                    // the following terms break the symmetry between
+                    // - polar -> decomposed -> uniform
+                    // - polar -> independent -> uniform
+                    // note that that is OK because decomposed is a 2D transformation of polar, while
+                    // independent is a radial-only transformation of polar, so it is more stable.
```
mgovers, line 459: This also means that because we use `DecomposedComplexRandomVariable` in our `CurrentSensorCalcParams`, our `IterativeLinear` solver actually uses a slightly larger variance than strictly necessary (because it uses `UniformComplexRandomVariable`, not `DecomposedComplexRandomVariable`).

This can be resolved by using `PolarComplexRandomVariable` in our `CurrentSensorCalcParams` and deferring the conversion either to `DecomposedComplexRandomVariable` (NRSE) or directly to `UniformComplexRandomVariable` (ILSE) where needed. `MeasuredValues` may then require some updates as well, as it would require different accumulations for the ILSE and the NRSE.

Out of scope of this PR, though.

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -126,6 +126,27 @@ template <symmetry_tag sym_type> struct DecomposedComplexRandVar {
...
+namespace detail {
+inline auto compute_decomposed_variance_from_polar(auto const& magnitude, auto const& cos_angle, auto const& sin_angle,
+                                                   auto const& magnitude_variance, auto const& angle_variance) {
+    auto const cos2_angle = cos_angle * cos_angle;
+    auto const sin2_angle = sin_angle * sin_angle;
+    auto const magnitude2 = magnitude * magnitude;
+
+    // // first order approximation
+    // return std::make_pair(magnitude_variance * cos2_angle + magnitude2 * angle_variance * sin2_angle,
+    //                       magnitude_variance * sin2_angle + magnitude2 * angle_variance * cos2_angle);
```
mgovers: do we want to remove this?

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -145,65 +166,66 @@ template <symmetry_tag sym_type> struct PolarComplexRandVar {
     }
 
     // For sym to sym conversion:
-    // Var(I_Re) ≈ Var(I) * cos^2(θ) + Var(θ) * I^2 * sin^2(θ)
-    // Var(I_Im) ≈ Var(I) * sin^2(θ) + Var(θ) * I^2 * cos^2(θ)
+    // Var(I_Re) ≈ Var(I) * cos^2(θ) + Var(θ) * I^2 * sin^2(θ)                    // first order
+    //             + 1/2 * Var(θ)^2 * I^2 * sin^2(θ) + Var(I) * Var(θ) * cos^2(θ) // second order
+    // Var(I_Im) ≈ Var(I) * sin^2(θ) + Var(θ) * I^2 * cos^2(θ)                    // first order
+    //             + 1/2 * Var(θ)^2 * I^2 * cos^2(θ) + Var(I) * Var(θ) * sin^2(θ) // second order
```
mgovers:
```suggestion
    //             + 1/2 * Var(θ)^2 * I^2 * cos^2(θ) + Var(I) * Var(θ) * sin^2(θ) // second order
    // Var(I_Im) ≈ Var(I) * sin^2(θ) + Var(θ) * I^2 * cos^2(θ)                    // first order
    //             + 1/2 * Var(θ)^2 * I^2 * sin^2(θ) + Var(I) * Var(θ) * cos^2(θ) // second order
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -126,6 +126,27 @@ template <symmetry_tag sym_type> struct DecomposedComplexRandVar {
     }
 };
 
+namespace detail {
+inline auto compute_decomposed_variance_from_polar(auto const& magnitude, auto const& cos_angle, auto const& sin_angle,
```
mgovers, line 148: add docstring to explain how first order + second order are obtained

power_grid_model_c/power_grid_model/include/power_grid_model/common/statistics.hpp
```diff
@@ -126,6 +126,41 @@ template <symmetry_tag sym_type> struct DecomposedComplexRandVar {
...
+//              + 1/2 * \Sum_{jk} Var(x_j) * ||∂²f_i/∂x_j∂x_k||^2 Var(x_k) // second order
+//
+// Re{I} = I cos(θ)
+// Im{I} = I sin(θ)
+//
+// ∂Re{I}/∂I = cos(θ)            ∂Im{I}/∂I = sin(θ)
+// ∂Re{I}/∂θ = -I sin(θ)         ∂Im{I}/∂θ = I cos(θ)
+// ∂²Re{I}/∂I² = 0               ∂²Im{I}/∂I² = 0
+// ∂²Re{I}/∂θ² = -I cos(θ)       ∂²Im{I}/∂θ² = -I sin(θ)
+// ∂²Re{I}/∂I∂θ = -sin(θ)        ∂²Im{I}/∂I∂θ = cos(θ)
```
TonyXiang8787: @mgovers  I have removed the PR from merge queue to make sure we understand this. Are you using a non-ascii character here?

I have hesitation about this. To use non-ascii character in the code.

mgovers: i can rewrite to just use only `d` directly

TonyXiang8787: I think that's a good idea

TonyXiang8787: Also avoid the superscript square, that's not ascii either

mgovers: fixed + removed from history

TonyXiang8787: the superscript `²` should also be removed.

mgovers: that is ascii though (https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.ascii-code.com/character/%25C2%25B2&ved=2ahUKEwiU6YnVzKOQAxV2_7sIHeugCWAQFnoECBcQAQ&usg=AOvVaw0yd-yM2ve2Tjh05zJ2gPRr)

we also use it in other locations in our docs

TonyXiang8787: It's encoding is `178`, outside of ascii.

mgovers: done

PR #1157, opened 2025-10-16
Docs: add homebrew installation to docs

mgovers:
C API can be installed using brew. We should add it to the docs

### DIFF
--- a/docs/advanced_documentation/c-api.md
+++ b/docs/advanced_documentation/c-api.md
@@ -24,21 +24,6 @@ program to use this C API.
 
 In this documentation, the main design choices and concepts of the C API are presented.
 
-## Installation
-
-To install this library, please follow the steps to [build from source](./build-guide.md).
-You can also install via Homebrew:
-
-```bash
-brew install powergridmodel/pgm/power-grid-model
-```
-
-Note that you will need to satisfy the basic minimum [build requirements](./build-guide.md#build-requirements).
-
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -48,6 +48,18 @@ If you are using `conda`, you can directly install the package from `conda-forge
...
+If you are using the C API with `brew`, you can install the C API (and experimental C++ wrapper) directly:
+
+```sh
+brew install powergridmodel/pgm/power-grid-model
+```
+
+If you are interested in having this package in [`homebrew-core`](https://github.com/Homebrew/homebrew-core), please
+consider contributing the [formula](https://github.com/PowerGridModel/homebrew-pgm), in conformance to
+[their documentation](https://docs.brew.sh/Acceptable-Formulae).
+
```
mgovers, line 62: do we have to add this to the main README as well?

TonyXiang8787: I don't think so. In C-API is enough

TonyXiang8787: I mean it's duplicated to have the same docs twice. I think keep this or the other in the `c-api.md`

mgovers: i'll keep this one then. It's good to publicly show the different ways of supporting PGM in one location

mgovers: that's also why i asked about the main repo README, btw. i think it's good to publicly show that we support both there as well

PR #1160, opened 2025-10-20
fix JSON deserializer message dump when token is extremely long

mgovers:
When there is an error during JSON deserialization, the error handling generates a message that contains the last parsed token.

However, if that token is extremely large, or if the end of the token is just never found in the first place (e.g. when the file to parse is extremely large and exceeds contiguous RAM limits, e.g. 20 GiB), then the error message may be equally large (also limited by contiguous RAM limitations).

This PR resolves that.

## Repro case
One such message dump case was reproduced using a 10 GiB JSON file. The data was generated for 1 billion nodes using the following code
```py
{
    component_type: initialize_array(
        DatasetType.update, component_type, shape=1e9
    )
    for component_type in [ComponentType.node]
}
```

### Current `main`
~2 GiB was written to STDERR, effectively crashing the terminal.
### This branch
```
power_grid_model._core.errors.PowerGridSerializationError: Parse error in JSON. Position: 2147483648, Last token: "node": [<U+000A>      {},<U+000A>      {},<U+000A>      {},<U+000A>      {},<U+000A>      {},<U+000...[truncated]. Exception message: [json.exception.parse_error.101] parse error at line 214748361, column 9: syntax error while parsing array - unexpected end of input; expected ']'
```

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
@@ -48,6 +48,8 @@ struct JsonMapArrayData {
 };
 
 struct JsonSAXVisitor {
+    static constexpr size_t max_error_message_token_length = 100;
+
     msgpack::packer<msgpack::sbuffer> top_packer() {
         if (data_buffers.empty()) {
             throw SerializationError{"Json root should be a map!\n"};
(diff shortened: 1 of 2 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -122,12 +124,13 @@ struct JsonSAXVisitor {
         return true;
     }
 
-    [[noreturn]] static bool parse_error(std::size_t position, std::string const& last_token,
-                                         json::exception const& ex) {
-        std::stringstream ss;
-        ss << "Parse error in JSON. Position: " << position << ", last token: " << last_token
-           << ". Exception message: " << ex.what() << '\n';
-        throw SerializationError{ss.str()};
+    [[noreturn]] static bool parse_error(std::size_t position, std::string_view last_token, json::exception const& ex) {
```
TonyXiang8787, line 127: Are you sure string view will work? This does not comply with the SAX parse from nlohmann json.

mgovers: any interface that uses `std::string const&` (specifically the reference part) always has an input lifetime that exceeds the lifetime of the function that is called. Therefore, by design, `std::string_view` is a plug-and-play replacement for `std::string const&`

NOTE: it would be different if `std::move(str)` was used with `std::string` as an interface instead of `std::string const&`. then, the only function argument would have to be `std::string`.

If you're not sure, i can use `std::string const&` as an intermediate type in the interface

TonyXiang8787 (approved):

PR #1161, opened 2025-10-20
Dev Environment: setup and the documentation for using the container

Laurynas-Jagutis:
Setup of the Dev Container on the Python side only.
Also has the documentation on how to set it up.

The gitattributes file is for standardizing line endings to LF instead of CRLF

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -50,13 +50,13 @@ jobs:
           python -m build --sdist --outdir wheelhouse .
 
       - name: Keep version file
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: version
           path: VERSION
 
       - name: Keep SDist
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: wheelhouse
           path: ./wheelhouse/*.tar.gz
@@ -198,7 +198,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - uses: actions/download-artifact@v5
+      - uses: actions/download-artifact@v6
         with:
           name: version
           path: .
@@ -221,7 +221,7 @@ jobs:
           CIBW_BUILD: ${{ matrix.cibw_build }}
 
       - name: Keep wheel files
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: wheelhouse-${{ matrix.platform }}
           path: ./wheelhouse/*.whl
@@ -314,9 +314,9 @@ jobs:
           python-version: '3.12'
           architecture: x64
 
-      - uses: actions/download-artifact@v5
+      - uses: actions/download-artifact@v6
 
-      - uses: actions/download-artifact@v5
+      - uses: actions/download-artifact@v6
         with:
           name: version
           path: .


docs/working_with_containers/container-setup.md
```diff
@@ -0,0 +1,53 @@
...
+the Python side.
+
+## Setup Requirements
+
+To begin using the container there are two open-source applications that must be installed.
+
+* Podman (<https://podman.io/>) - a container engine that lets developers build, run, and manage containers and pods
+without needing a background service.
+* DevPod (<https://devpod.sh/>) - a tool that instantiates reproducible, disposable development environments allowing
+developers to code inside the containers with their usual editors.
```
TonyXiang8787, line 19: Does the developer need to install two additional software (besides Docker) to use the Dev Container? I am hesitating about whether this actually brings convenience to the developers.

mgovers: just having docker desktop would also work (that's one tool instead of 2, but it's a paid service). Many python-only developers will have these tools installed anyways.

Devcontainers are especially useful for those python-only developers that have never worked with a C++ compiler but come from the IT industry in which containers are extremely common

TonyXiang8787: In that case, please take [`container`](https://formulae.brew.sh/cask/container) into account for macOS.

https://github.com/apple/container

It is the preferred container engine.

figueroa1395:
I was able to install everything and launch the container through VScode, but I'm having trouble with `python` and `pip`. My hunch is that something went wrong because of MacOS. @Laurynas-Jagutis Let's see if we can resolve the issue together.

figueroa1395:
> I was able to install everything and launch the container through VScode, but I'm having trouble with `python` and `pip`. My hunch is that something went wrong because of MacOS. @Laurynas-Jagutis Let's see if we can resolve the issue together.

After diving into the issue together with @Laurynas-Jagutis, we found that there were two issues:

1. One must set the container from the correct branch (this branch and not main - my mistake).
2. On MacOS (and perhaps other machines) `podman` has the default RAM of the containers set to a very small value, not enough to build PGM. This is very easy to solve by increasing the RAM from `podman` itself.

Both points will be added to the documentation, but I can now confirm that this approach works well for MacOS as well.

TonyXiang8787:
> > I was able to install everything and launch the container through VScode, but I'm having trouble with `python` and `pip`. My hunch is that something went wrong because of MacOS. @Laurynas-Jagutis Let's see if we can resolve the issue together.
> 
> After diving into the issue together with @Laurynas-Jagutis, we found that there were two issues:
> 
> 1. One must set the container from the correct branch (this branch and not main - my mistake).
> 2. On MacOS (and perhaps other machines) `podman` has the default RAM of the containers set to a very small value, not enough to build PGM. This is very easy to solve by increasing the RAM from `podman` itself.
> 
> Both points will be added to the documentation, but I can now confirm that this approach works well for MacOS as well.

@figueroa1395 please try `container` app from Apple for dev container in MacOS, instead of podman/devpod

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+* text=auto
```
mgovers: please do not blindly set `* text=auto`. You can refer to https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file

What's the reason you needed this in the first place?

figueroa1395:
> > > I was able to install everything and launch the container through VScode, but I'm having trouble with `python` and `pip`. My hunch is that something went wrong because of MacOS. @Laurynas-Jagutis Let's see if we can resolve the issue together.
> > 
> > 
> > After diving into the issue together with @Laurynas-Jagutis, we found that there were two issues:
> > 
> > 1. One must set the container from the correct branch (this branch and not main - my mistake).
> > 2. On MacOS (and perhaps other machines) `podman` has the default RAM of the containers set to a very small value, not enough to build PGM. This is very easy to solve by increasing the RAM from `podman` itself.
> > 
> > Both points will be added to the documentation, but I can now confirm that this approach works well for MacOS as well.
> 
> @figueroa1395 please try `container` app from Apple for dev container in MacOS, instead of podman/devpod

I will try as soon as I have the time and report back.

.gitattributes
Laurynas-Jagutis: I needed it because I would get a lot of messages for the changes in line endings whenever I went in to the container from my Windows system. But I am looking into how to make the formatting manual and not let git guess which file to format and which to not

figueroa1395:
@TonyXiang8787 Due to time, I haven't tried out `container`. However, since this workflow works as is, we have decided within the team to try `container` as a followup instead. I should be able to pick it up tomorrow or the day after.

Issue #1159, opened 2025-10-16
[BUG] Observability in radial grid with phasor voltage measurement is not correct


TonyXiang8787:
# Background

In #914 and #1024 we have implemented the observability check in radial grid with phasor voltage measurement. However, the mathematical workout is not correct. We assumed that the phasor voltage sensor can be arbitrarily assigned to a branch, just as an injection measurement. This assumption does not hold.

# Rationale

We should not assign phasor voltage sensor to any branches. Instead, we should solely rely on branch and injection sensor to connect the graph. After trying to make maximum connected graph, we check if the graph is fully connected. If yes, there should be at least one voltage sensor (phasor or not) for the grid to be observable. If no, each island should have at least one voltage phasor sensor for the grid to be observable.

PR #1163, opened 2025-10-22
BugFix: temporarily drop voltage phasor sensor handling in observability check

Jerry-Jinfeng-Guo:
This PR temporarily drops the handling of the voltage phasor sensor in the context of observability check.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -557,8 +557,13 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
 
     // check the sufficient condition for observability
     // the check is currently only implemented for grids without voltage phasor sensor
-    if (n_voltage_phasor_sensors == 0) {
+    if (n_voltage_phasor_sensors != 0) {
         // Only handle networks without voltage phasor
+        is_sufficient_condition_met = false;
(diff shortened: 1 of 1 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -556,14 +556,9 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
...
-    if (topo.is_radial) {
-        // Temporary path, ideally this is only called when
-        // n_voltage_phasor_sensors > 0, regardless of network type
-        is_sufficient_condition_met = detail::sufficient_condition_radial_with_voltage_phasor(
-            y_bus_structure, observability_sensors, n_voltage_phasor_sensors);
-    } else {
-        // Temporary path, to be refined later
+    // the check is currently only implemented for grids without voltage phasor sensor
+    if (n_voltage_phasor_sensors == 0) {
+        // Only handle networks without voltage phasor
```
figueroa1395, line 561: Whether we "disable" voltage phasor sensors or not should be independent of enabling the general meshed check for both radial and meshed grids. We need to benchmark and make such decision first.

Jerry-Jinfeng-Guo:
After discussing this with in the team, we proceed with a sustainable solution to this bug https://github.com/PowerGridModel/power-grid-model/issues/1159 and dropping is not needed.

PR #1166, opened 2025-10-23
Performance: reduce copying during batch error handling

mgovers:
A serious performance regression was introduced in #1018 (Release [`v1.11.33`](https://github.com/PowerGridModel/power-grid-model/releases/tag/v1.11.33)) to batch error handling that caused the combined batch error message to be copied completely for every scenario, resulting in a quadratic blowup in the amount of failing scenarios.

This fixes that.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
@@ -184,7 +184,7 @@ class JobDispatch {
     }
 
     static void handle_batch_exceptions(std::vector<std::string> const& exceptions) {
-        std::stringstream combined_error_message;
+        std::ostringstream combined_error_message;
         IdxVector failed_scenarios;
         std::vector<std::string> err_msgs;
         for (Idx batch = 0; batch < static_cast<Idx>(exceptions.size()); ++batch) {


power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -184,20 +184,20 @@ class JobDispatch {
     }
 
     static void handle_batch_exceptions(std::vector<std::string> const& exceptions) {
-        std::string combined_error_message;
+        std::stringstream combined_error_message;
```
figueroa1395: Why do we need `std::stringstream` instead of `std::string` here?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -184,20 +184,20 @@ class JobDispatch {
...
         IdxVector failed_scenarios;
         std::vector<std::string> err_msgs;
         for (Idx batch = 0; batch < static_cast<Idx>(exceptions.size()); ++batch) {
             // append exception if it is not empty
             if (!exceptions[batch].empty()) {
-                combined_error_message =
-                    std::format("{}Error in batch #{}: {}\n", combined_error_message, batch, exceptions[batch]);
+                combined_error_message << std::format("Error in batch #{}: {}\n", batch, exceptions[batch]);
```
figueroa1395, line 193: This was the bug, right? Combining each time the previous `combined_error_message` for each scenario which blows up rather fast.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -184,20 +184,20 @@ class JobDispatch {
...
                 failed_scenarios.push_back(batch);
                 err_msgs.push_back(exceptions[batch]);
             }
         }
-        if (!combined_error_message.empty()) {
-            throw BatchCalculationError(combined_error_message, failed_scenarios, err_msgs);
+        if (!combined_error_message.view().empty()) {
+            throw BatchCalculationError(std::move(combined_error_message).str(), std::move(failed_scenarios),
+                                        std::move(err_msgs));
```
figueroa1395, line 200: This is just additional optimization, right?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers:
```suggestion
        std::ostringstream combined_error_message;
```

mgovers: Intentionality: we explicitly keep appending to it until it's finished, then we're done and gove it back

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
mgovers: Yes

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
figueroa1395: Offline: It is, we should have done this in the first place.

figueroa1395 (approved):

PR #1167, opened 2025-10-23
LU solver: ill-condition (SparseMatrixError) defined by matrix norm - cleanup

mgovers:
Cleanup to #1125 .

This PR is just to store the knowledge and will be closed immediately after opening in accordance to the decisions based on the investigation in #1125 and its follow-ups.


mgovers:
in accordance to the decisions based on the investigation in https://github.com/PowerGridModel/power-grid-model/pull/1125 and its follow-ups, we decided to abandon this path in favor of #35

PR #1168, opened 2025-10-24
Clean up main model: migrate out "calculation preparation" related stuff

figueroa1395:
This  PR is aimed to remove all the stuff related to preparing the calculation/solvers out of the main model. This PR should be an enabler to unit test the touched logic more easily,  keep trimming the main model, and get it ready to remove everything related to the calculation logic in a followup.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_preparation.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_preparation.hpp
@@ -94,14 +94,12 @@ class SolversCacheStatus {
             return;
         }
 
+        set_topology_status(true);
         if (changes.param) {
-            set_topology_status(true);
             set_parameter_status<symmetric_t>(false);
             set_parameter_status<asymmetric_t>(false);
             return;
         }
-
-        set_topology_status(true);
         set_parameter_status<symmetric_t>(is_parameter_valid<symmetric_t>());
         set_parameter_status<asymmetric_t>(is_parameter_valid<asymmetric_t>());
     }
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -304,6 +304,7 @@ class MainModelImpl {
             [this, &sequence_idx]<typename CompType>() { this->restore_component<CompType>(sequence_idx); });
 
         solvers_cache_status_.update(cached_state_changes_);
+        // we can probably reset topo, solvers and ybus here
         cached_state_changes_ = {};
     }
 
--- a/tests/cpp_unit_tests/test_calculation_preparation.cpp
+++ b/tests/cpp_unit_tests/test_calculation_preparation.cpp
@@ -15,7 +15,7 @@ using MainModelType = main_core::MainModelType<AllExtraRetrievableTypes, AllComp
 
 TEST_CASE("Test SolversCacheStatus") {
     SUBCASE("Default construction") {
-        SolversCacheStatus<MainModelType> cache_status{};
+        SolversCacheStatus<MainModelType> const cache_status{};
 
         CHECK_FALSE(cache_status.is_topology_valid());
         CHECK_FALSE(cache_status.is_parameter_valid<symmetric_t>());
@@ -71,7 +71,7 @@ TEST_CASE("Test SolversCacheStatus") {
 
 TEST_CASE("Test SolverPreparationContext") {
     SUBCASE("Default construction") {
-        SolverPreparationContext context{};
+        SolverPreparationContext const context{};
 
         CHECK(context.math_solver_dispatcher == nullptr);
         CHECK(context.math_state.y_bus_vec_sym.empty());
@@ -81,8 +81,8 @@ TEST_CASE("Test SolverPreparationContext") {
     }
 
     SUBCASE("Dummy construction") {
-        MathSolverDispatcher dispatcher{math_solver::math_solver_tag<MathSolver>{}};
-        SolverPreparationContext context{.math_state = {}, .math_solver_dispatcher = &dispatcher};
+        MathSolverDispatcher const dispatcher{math_solver::math_solver_tag<MathSolver>{}};
+        SolverPreparationContext const context{.math_state = {}, .math_solver_dispatcher = &dispatcher};
 
         CHECK(context.math_solver_dispatcher == &dispatcher);
         CHECK(context.math_state.y_bus_vec_sym.empty());


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -33,11 +34,9 @@
 #include "main_core/calculation_input_preparation.hpp"
 #include "main_core/input.hpp"
 #include "main_core/main_model_type.hpp"
-#include "main_core/math_state.hpp"
 #include "main_core/output.hpp"
 #include "main_core/topology.hpp"
 #include "main_core/update.hpp"
-#include "main_core/y_bus.hpp"
```
figueroa1395, line 40: I made sure to only remove imports which aren't used at all (directly or indirectly) within this file anymore.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -542,89 +528,27 @@ class MainModelImpl {
...
+    // math_state
+    // n_math_solvers
+    // math_solver_dispatcher
+
+    StatusCheckingContext<ImplType> state_status_context_{};
+    // is_topology_up_to_date
+    // is_sym_parameter_up_to_date
+    // is_asym_parameter_up_to_date
+    // last_updated_calculation_symmetry_mode
+    // parameter_changed_components
```
figueroa1395: These comments can be useful at least for the reviewing part. Whether they stay or not is up to the input I receive.

mgovers: I think they can be removed indeed

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -131,22 +129,17 @@ class MainModelImpl {
...
-    using MathState = main_core::MathState;
     using MetaData = meta_data::MetaData;
 
     // constructor with data
     explicit MainModelImpl(double system_frequency, ConstDataset const& input_data,
                            MathSolverDispatcher const& math_solver_dispatcher, Idx pos = 0)
         : system_frequency_{system_frequency},
           meta_data_{&input_data.meta_data()},
-          math_solver_dispatcher_{&math_solver_dispatcher} {
+          solver_preparation_context_{.math_solver_dispatcher = &math_solver_dispatcher} {
```
mgovers: This construction can IMO go to `MainModel` instead of `MainModelImpl` just like `JobDispatcher`

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include "main_core/main_model_type.hpp"
+#include "main_core/math_state.hpp"
+#include "main_core/topology.hpp"
+#include "main_core/y_bus.hpp"
+
+namespace power_grid_model {
+struct SolverPreparationContext {
+    main_core::MathState math_state;
+    Idx n_math_solvers{0};
```
nitbharambe: Is it required to keep n_math_solvers?
Is it ever not equivalent to `math_state.topology.size()`?

figueroa1395: I removed it (in [4f0d589](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/4f0d5898788b196742a060e233b7a29545a01834)) in favor of a helper function that does the calling from `state.math_topology.size()` indeed. It should be fine and
```
assert(n_math_solvers == static_cast<Idx>(main_core::get_y_bus<sym>(solver_context.math_state).size()));|
```
still remains as a sanity check.

Let me know what you think.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: I addressed this in [66e5449](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/66e5449fc348e4a39428bf439f603f06677632ed). There are now some additional dependencies to `prepare_calculate.hpp` which I'm not a fan of, but I suppose that should be fine because it's only temporal until we really get rid of `MainModelImpl`. Thoughts?

mgovers: should be fine. this touches the topic of dependency inversion, which is part of @nitbharambe 's PR

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
mgovers: i think this is indeed legacy from the era when we would re-use a math solver when our amount of connected components in the grid remained the same, even when our topology changed (i.e. we would reuse preallocated math solver memory, but it doesn't make sense because the elephant in the room is not the math solver but the y bus structure/topology. therefore, we changed that)

as long as our cache invalidation works correctly, with the current state of the code base, that should work as intended

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
```diff
@@ -0,0 +1,119 @@
...
+struct StatusCheckingContext {
+    bool is_topology_up_to_date{false};
+    bool last_updated_calculation_symmetry_mode{false};
+    typename ModelType::SequenceIdx parameter_changed_components{};
+    struct IsParameterUpToDateHelper {
+        bool sym{false};
+        bool asym{false};
+    };
+    IsParameterUpToDateHelper is_parameter_up_to_date{};
+};
```
mgovers: although not strictly necessary, for the sake of readability:


```suggestion
struct StatusCheckingContext {
    struct IsParameterUpToDateHelper {
        bool sym{false};
        bool asym{false};
    };
    
    using SequenceIdx = typename ModelType::SequenceIdx;

    bool is_topology_up_to_date{false};
    bool last_updated_calculation_symmetry_mode{false};
    SequenceIdx parameter_changed_components{};
    IsParameterUpToDateHelper is_parameter_up_to_date{};
};
```

alternatively, if you really want to do it inline, you can also do


```suggestion
struct StatusCheckingContext {
    bool is_topology_up_to_date{false};
    bool last_updated_calculation_symmetry_mode{false};
    typename ModelType::SequenceIdx parameter_changed_components{};
    struct IsParameterUpToDateHelper {
        bool sym{false};
        bool asym{false};
    } is_parameter_up_to_date{};
};
```

but i'm not a fan of that personally

mgovers: definitively out of scope but i have also been thinking about doing a class that can only invalidate, never validate (which plain bool flags can), that is, something along the lines of

```suggestion
class StatusCheckingContext {
  public:
    struct invalid_t {};
    struct valid_t {};

    static constexpr invalid_t invalid_tag{};
    static constexpr valid_t valid_tag{};

    StatusCheckingContext() = default;
    StatusCheckingContext(invalid_t /*tag*/): StatusCheckingContext() {}
    StatusCheckingContext(valid_t /*tag*/):
      is_topology_up_to_date_{true},
      is_parameter_up_to_date_{{.sym=true, .asym=true}},
      last_updated_calculation_symmetry_mode_{true} {}

    void set_topo_invalid() {
        is_topology_up_to_date_= false;
        set_parameter_invalid_<symmetric_t>();
        set_parameter_invalid_<asymmetric_t>();
    }
    template <symmetry_c sym> void set_parameter_invalid() {
        if constexpr (is_symmetric<sym>) {
            is_parameter_up_to_date.sym = false;
        } else {
            is_parameter_up_to_date.asym = false;
        }
    }
    bool is_topo_up_to_date() { return is_topology_up_to_date_; }
    template <symmetry_c sym>

  private:
    struct IsParameterUpToDateHelper {
        bool sym{false};
        bool asym{false};
    };

    bool is_topology_up_to_date_{false};
    IsParameterUpToDateHelper is_parameter_up_to_date_{};
    bool last_updated_calculation_symmetry_mode_{false};
    typename ModelType::SequenceIdx parameter_changed_components_{}; // we need to figure out what to do with this.
};
```

while it looks like overkill, it ensures the invariant in which any invalidated topology will always also invalidate the parameters. It cannot be accidentally forgotten.

the only way to reset is to create a new object, e.g.

```cpp
StatusCheckingContext validity; // all invalid / default
foo(validity);

update(validity);
validity = StatusCheckingContext{StatusCheckingContext::valid}; // all valid

bar_invalidates_params(validity); // now, topo is still up to date

update(validity); // now, we make the bunch valid again
validity = StatusCheckingContext{StatusCheckingContext::valid}; // tada
```

alternatively, we can also create either an enum or extend the set of tags to reflect this stuff

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
```diff
@@ -0,0 +1,119 @@
...
+    main_core::MathState math_state;
+    MathSolverDispatcher const* math_solver_dispatcher;
+};
+
+template <class ModelType>
+    requires(main_core::is_main_model_type_v<ModelType>)
+struct StatusCheckingContext {
+    bool is_topology_up_to_date{false};
+    bool last_updated_calculation_symmetry_mode{false};
+    typename ModelType::SequenceIdx parameter_changed_components{};
```
mgovers: it feels a little bit confusing to me that the status and the cache are part of the same struct called `StatusCheckingContext`.

Maybe the name is just not clear enough. Maybe something along the lines of `CacheStatus` or `CacheValidity`?

That said, maybe the confusing part is not the name of the type but the fact that one thing tells something about the cache validity while the other about the context in which it is valid? maybe that means that they shouldn't be part of the same class in the first place?

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
```diff
@@ -0,0 +1,119 @@
...
+struct SolverPreparationContext {
+    main_core::MathState math_state;
+    MathSolverDispatcher const* math_solver_dispatcher;
+};
+
+template <class ModelType>
+    requires(main_core::is_main_model_type_v<ModelType>)
+struct StatusCheckingContext {
+    bool is_topology_up_to_date{false};
+    bool last_updated_calculation_symmetry_mode{false};
```
mgovers: i don't like this name with the fact that it's a bool. Maybe you can make it an enum? or otherwise rename to something along the lines of `last_updated_was_symmetric`?

this is basically an invariant, namely a mapping of the previous calculation type to the next. maybe you can solve it using a setter (instead of doing the checking at the user side), e.g.

```cpp
class StatusCheckingContext {
    template <symmetry_tag sym> set_most_recent_calculation_symmetry() {
        last_updated_calculation_was_symmetric_ = is_symmetric_v<sym>;
    }
    template <symmetry_tag sym> is_same_calculation_symmetry() {
        return last_updated_calculation_was_symmetric_ == is_symmetric_v<sym>;
    }
};
```

You can also use `IsParameterUpToDateHelper` for caching (although slightly less memory efficient, but if we really want to go down that direction, we also don't need a separate bool for topology; we could just use bitflags or enums

note that using `std::variant` would be overkill.

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
```diff
@@ -0,0 +1,119 @@
...
+    Topology topology{*state.comp_topo, comp_conn};
+    std::tie(state.math_topology, state.topo_comp_coup) = topology.build_topology();
+    status_context.is_topology_up_to_date = true;
+    status_context.is_parameter_up_to_date.sym = false;
+    status_context.is_parameter_up_to_date.asym = false;
+}
+} // namespace detail
+
+template <symmetry_tag sym, class ModelType>
+bool& is_parameter_up_to_date(
```
mgovers: i prefer to make this a member function

figueroa1395: Take a look at [52ea34d](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/52ea34d987e61230ab03578fbd5e256f2f9fb8a8)

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
figueroa1395: Take a look at [52ea34d](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/52ea34d987e61230ab03578fbd5e256f2f9fb8a8)

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
figueroa1395: Take a look at [52ea34d](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/52ea34d987e61230ab03578fbd5e256f2f9fb8a8)

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
figueroa1395: Take a look at [52ea34d](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/52ea34d987e61230ab03578fbd5e256f2f9fb8a8)

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
```diff
@@ -0,0 +1,159 @@
...
+template <class ModelType>
+    requires(main_core::is_main_model_type_v<ModelType>)
+class SolversCacheStatus {
+  private:
+    struct YBusParameterCacheValidity {
+        bool sym{false};
+        bool asym{false};
+    };
+
+    enum class SymmetryMode { symmetric, asymmetric, not_set };
```
mgovers:
```suggestion
    enum class SymmetryMode : IntS { symmetric = 0, asymmetric = 1, not_set = na_IntS };
```

power_grid_model_c/power_grid_model/include/power_grid_model/prepare_calculate.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+    bool topology_cache_validity{false};
+    YBusParameterCacheValidity parameter_cache_validity{};
+    SymmetryMode previous_symmetry_mode{SymmetryMode::not_set};
+
+  public:
+    using SequenceIdx = typename ModelType::SequenceIdx;
+    SequenceIdx changed_components_indices{};
```
mgovers: should not be a public member variable

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_preparation.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+#include "math_solver/math_solver_dispatch.hpp"
+
+#include "main_core/main_model_type.hpp"
+#include "main_core/math_state.hpp"
+#include "main_core/topology.hpp"
+#include "main_core/y_bus.hpp"
+
+namespace power_grid_model {
+struct SolverPreparationContext {
```
mgovers, line 19: probably good to fully unit test this. it's simple yet valuable

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -424,7 +409,8 @@ class MainModelImpl {
 
             return calculate_<ShortCircuitSolverOutput<sym>, MathSolverProxy<sym>, YBus<sym>, ShortCircuitInput>(
                 [this, voltage_scaling](Idx n_math_solvers) {
-                    assert(is_topology_up_to_date_ && is_parameter_up_to_date<sym>());
+                    assert((solvers_cache_status.is_topology_valid() &&
+                            solvers_cache_status.template is_parameter_valid<sym>()));
```
mgovers:
```suggestion
                    assert(solvers_cache_status.is_topology_valid());
                    assert(solvers_cache_status.template is_parameter_valid<sym>());
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -363,18 +346,20 @@ class MainModelImpl {
         // prepare
         auto const& input = [this, &logger, prepare_input_ = std::forward<PrepareInputFn>(prepare_input)] {
             Timer const timer{logger, LogEvent::prepare};
-            prepare_solvers<sym>();
-            assert(is_topology_up_to_date_ && is_parameter_up_to_date<sym>());
-            return prepare_input_(n_math_solvers_);
+            assert(construction_complete_);
+            prepare_solvers<sym>(state_, solver_preparation_context_, solvers_cache_status);
+            assert(
+                (solvers_cache_status.is_topology_valid() && solvers_cache_status.template is_parameter_valid<sym>()));
```
mgovers: please split up the assert

```suggestion
            assert(solvers_cache_status.is_topology_valid());
            assert(solvers_cache_status.template is_parameter_valid<sym>());
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -307,9 +288,11 @@ class MainModelImpl {
...
         // if topology changed, everything is not up to date
         // if only param changed, set param to not up to date
-        is_topology_up_to_date_ = is_topology_up_to_date_ && !changes.topo;
-        is_sym_parameter_up_to_date_ = is_sym_parameter_up_to_date_ && !changes.topo && !changes.param;
-        is_asym_parameter_up_to_date_ = is_asym_parameter_up_to_date_ && !changes.topo && !changes.param;
+        solvers_cache_status.set_topology_status(solvers_cache_status.is_topology_valid() && !changes.topo);
+        solvers_cache_status.template set_parameter_status<symmetric_t>(
+            solvers_cache_status.template is_parameter_valid<symmetric_t>() && !changes.topo && !changes.param);
+        solvers_cache_status.template set_parameter_status<asymmetric_t>(
+            solvers_cache_status.template is_parameter_valid<asymmetric_t>() && !changes.topo && !changes.param);
```
mgovers: this function feels less readable in its current form. would it be possible to refactor?

to me this seems like adding separate `set` and `reset` functions actually makes sense... That way, we don't accidentally set anything to `true` here. e.g.

```suggestion
        auto const invalid_topo = changes.topo;
        auto const invalid_params = invalid_topo || changes.param;

        solvers_cache_status.invalidate_topology(invalid_topo);
        solvers_cache_status.template invalidate_parameter<symmetric_t>(invalid_params);
        solvers_cache_status.template invalidate_parameter<asymmetric_t>(invalid_params);
```

or

```suggestion
        solvers_cache_status.change_topology(changes.topo);
        solvers_cache_status.template parameters_changed<symmetric_t>(changes.topo || changes.param);
        solvers_cache_status.template parameters_changed<asymmetric_t>(changes.topo || changes.param);
```

and separately, maybe this function is not in the correct location in the first place

```suggestion
        solvers_cache_status.apply(changes);
```

still not perfect but better than what it is now...

figueroa1395: I went in a different direction. I went for readability. [a2771a9](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/a2771a9ba2efaa2f48c0dae37005ee2a19d50ec0)

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_preparation.hpp
figueroa1395: [e40f2b3](https://github.com/PowerGridModel/power-grid-model/pull/1168/commits/e40f2b37ad248bdddcae58027c17512db70557ce) Let me know if more detailed unit tests are needed.

On the other hand, tomorrow I'll check how easy/hard is to unit test `prepare_solvers` and that should finish this.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_preparation.hpp
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        if (changes.param) {
+            set_topology_status(true);
+            set_parameter_status<symmetric_t>(false);
+            set_parameter_status<asymmetric_t>(false);
+            return;
+        }
+
+        set_topology_status(true);
```
mgovers:
```suggestion
        set_topology_status(true);

        if (changes.param) {
            set_topology_status(true);
            set_parameter_status<symmetric_t>(false);
            set_parameter_status<asymmetric_t>(false);
            return;
        }
```

figueroa1395:
@mgovers Addressed all. I believe this is ready to be merged. Adding more unit tests becomes hard because of all the topology, components and such dependencies. I can try, but I don't think it adds much values at this point. If you agree, feel free to approve and send to merge queue.

PR #1169, opened 2025-10-24
BugFix: unify `fast_any_tap` behavior when handling small `u_band` (smaller than `tap_size`)

Jerry-Jinfeng-Guo:
The binary search logic in any mode can not handle edge case when `u_band` is significantly smaller than `tap_size` due to the lack of back-forth jumping behavior. This PR fixes that and `fast_any_tap` strategy now correctly throws a `MaxIterationReached` error in equivalent cases as other strategies do.
- [x] The fix
- [x] Tests

Note: the edge case is in itself not a valid input, as unrealistic `u_band` shouldn't be allowed, this PR aligns the behavior across different strategies.

This PR relates to https://github.com/PowerGridModel/power-grid-model/pull/887

### DIFF
--- a/tests/data/power_flow/automatic-tap-regulator/achievable-voltage-above-band-fast-any/params.json
+++ b/tests/data/power_flow/automatic-tap-regulator/achievable-voltage-above-band-fast-any/params.json
@@ -7,7 +7,7 @@
     "u": 100.0
   },
    "xfail": {
-    "raises": "AssertionError",
+    "raises": "MaxIterationReached",
     "reason": "Bug in fast any tap changing strategy in this edge case: https://github.com/PowerGridModel/power-grid-model/issues/1215"
   }
 }


tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-any-max-iter/params.json
```diff
@@ -1,6 +1,6 @@
 {
   "calculation_method": "newton_raphson",
-  "tap_changing_strategy": "any_valid_tap",
+  "tap_changing_strategy": "fast_any_tap",
```
nitbharambe: Do we support the multiple tap_changing_strategy? 
in that case
```suggestion
  "tap_changing_strategy": ["any_valid_tap", "fast_any_tap"],
```

Jerry-Jinfeng-Guo: I don't think we do.

mgovers: I think both should be tested. Since the case name literally contains `any` as a strategy, either only `any_valid_tap` or both `any_valid_tap` and `fast_any` should be tested.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -1134,7 +1138,7 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
                 return;
             }
 
-            if (strategy_ == OptimizerStrategy::fast_any) {
+            if (strategy_ == OptimizerStrategy::fast_any && cmp == 0) { // NOLINT(modernize-use-nullptr)
```
mgovers: I know there is a false positive regarding cmp, but please use ==equivalent rather than ==0` if possible. I think it's something along the lines of

```suggestion
            if (strategy_ == OptimizerStrategy::fast_any && cmp == std::partial_ordering::equivalent) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -1143,6 +1147,16 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
             current_bs.recalibrate(strategy_max);
 
             IntS const tap_pos = current_bs.repropose_tap(strategy_max, previous_down, tap_changed);
+            // The new same tap pos is only valid in a cmp == equivalent situation
+            // in other words, a same tap pos in a non equivalent situation means
+            // the binary search cannot find a valid tap position
+            // _tap_changed_ is an aggregated flag with _inevitable_run_, so we can not use it here
+            if (tap_pos == transformer.tap_pos() && cmp != 0 && !current_bs.get_end_of_bs()) {
+                current_bs.rewind(tap_pos, transformer.tap_min(), transformer.tap_max());
+                throw MaxIterationReached{std::format(
```
mgovers, line 1161: I still am not 100% convinced that this really is undesirable behavior: just like `fast_int` in gcc: it works in cases it's designed to work in, but if not, the behavior is still well-behaved, and it is opt-in (you chose to go with fast_any in the first place).

if it couldn't find a solution, it should return either the lower or the upper bound that was found, but not `MaxIterationReached` because it did not exceed the amount of iterations. If you really want to raise an error, then maybe `NoValidTapFound` or something

Also, in real-world cases where the intention is not to get the exact state but to get an approximate one, it's undesirable to throw a hard error. Users may in fact be OK with an approximate solution

tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-any-max-iter/params.json
Jerry-Jinfeng-Guo: Please have a complete look of the PR first. Most related: `tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-small-u-band-max-iter-*-any/*`

Jerry-Jinfeng-Guo:
```suggestion
  "tap_changing_strategy": "any_valid_tap",
```
This test should not be touched. Single purpose only.

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
mgovers: my bad, my (still open) comment was apparently from before the discussion happened... Sorry for the confusion. Closing this ticket

tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-fast-any-max-iter/input.json
mgovers, line 1: I would also like some tests added here for the other edge case discussed for this topic in which all valid tap positions lie outside the voltage band (should settle at either `tap_min` or `tap_max`, whichever results in a voltage that is closest to the selected voltage band)

mgovers:
Please rename the PR to follow the PR naming guidelines

tests/data/power_flow/automatic-tap-regulator/auto-tap-changer-meshed-fast-any-max-iter/input.json
Jerry-Jinfeng-Guo: See https://github.com/PowerGridModel/power-grid-model/pull/1214

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model/pull/1169/commits/1bf0beebe353731f519f5fb1ad0bcc283091651f

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -1081,10 +1086,10 @@ class TapPositionOptimizerImpl<std::tuple<TransformerTypes...>, StateCalculator,
 
             auto const cmp = node_state <=> param;
             auto new_tap_pos = [&transformer, &cmp, &control_at_tap_side] {
-                if (cmp > 0) { // NOLINT(modernize-use-nullptr)
+                if (std::is_gt(cmp)) {
```
mgovers, line 1093: nice :+1:

mgovers:
for the readthedocs build, you're probably missing a `#include <cassert>` somewhere that was never an issue until this PR (maybe because you now include `#include <compare>`?)

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
```diff
@@ -62,10 +63,10 @@ struct TrafoGraphEdge {
     } // thanks boost
 
     auto constexpr operator<=>(TrafoGraphEdge const& other) const {
-        if (auto cmp = weight <=> other.weight; cmp != 0) { // NOLINT(modernize-use-nullptr)
+        if (auto cmp = weight <=> other.weight; !std::is_eq(cmp)) {
```
figueroa1395: Is there a difference between `!std::is_eq` and `std:is_neq`? If not, minor nitpick to use `std:is_neq` instead :)

Jerry-Jinfeng-Guo:
I deem most of the comments are complimentary and or out of scope. Please see whether a final review is possible? @mgovers @figueroa1395

mgovers:
> I deem most of the comments are complimentary and or out of scope. Please see whether a final review is possible? @mgovers @figueroa1395

I think @figueroa1395's comment is the only one remaining and would consider that in-scope: https://github.com/PowerGridModel/power-grid-model/pull/1169#discussion_r2602564930 . After that it's ready for final review

power_grid_model_c/power_grid_model/include/power_grid_model/optimizer/tap_position_optimizer.hpp
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model/pull/1169/commits/eddf6f821e3493d243a021e1d93e3881e79846b7

PR #1174, opened 2025-10-28
Observability: enable error for meshed observability check

Jerry-Jinfeng-Guo:
Enable exception `NotObservableError` from meshed observability check.

- [x] Exception thrown in sub check, align with radial check behavior
- [x] Adjust test error type accordingly

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -510,7 +510,7 @@ sufficient_condition_meshed_without_voltage_phasor(std::vector<detail::BusNeighb
     });
 
     if (!found_spanning_tree) {
-        throw NotObservableError{"No full-spanning tree can be constructed from this meshed network. Unobservable.\n"};
+        throw NotObservableError{"Meshed observability check fail. Network unobservable.\n"};
     }
 
     return true;


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -505,9 +505,15 @@ sufficient_condition_meshed_without_voltage_phasor(std::vector<detail::BusNeighb
     prepare_starting_nodes(neighbour_list, n_bus, starting_candidates);
 
     // Try each starting candidate
-    return std::ranges::any_of(starting_candidates, [&](Idx const start_bus) {
+    bool const found_spanning_tree = std::ranges::any_of(starting_candidates, [&](Idx const start_bus) {
         return find_spanning_tree_from_node(start_bus, n_bus, neighbour_list);
     });
+
+    if (!found_spanning_tree) {
+        throw NotObservableError{"No full-spanning tree can be constructed from this meshed network. Unobservable.\n"};
```
figueroa1395: I think this is too low level of a message for the average user. Can you re-write it in terms of sensors in a similar way to the one used for the radial check?

figueroa1395: So currently, if the meshed grid is not observable, we still get `SparseMatrixError`, right?

Jerry-Jinfeng-Guo: Indeed. Noticed during the demo and this should be now updated.

Jerry-Jinfeng-Guo: Suggestions?

Jerry-Jinfeng-Guo:
```suggestion
        throw NotObservableError{"Meshed observability check fail. Network unobservable.\n"};
```
Better? @figueroa1395

figueroa1395: Very nice catch! This is nasty one.

figueroa1395: I would make it even higher level. What about something like 
```suggestion
        throw NotObservableError{"Meshed observability check fail. Network unobservable. The number of sensors appears sufficient, but they are not independent enough.\n"};
```
Why? So it's VERY explicit to the users what is the problem.

mgovers (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: I don' think independence is necessarily inferred here as the cause.

PR #1178, opened 2025-11-04
Type hinting: Fix + add overloads in multiple often-used functions

mgovers:
* [x] Fix: `PowerGridModel.calculate*` functions had the output type still specified to row-based data sets. It did not provide the correct output type hint for columnar or mixed output data
* [x] Quality of life improvements:
  * [x] `initialize_array` now provides type type hinting specific to whether single or batch data was requested. This is the biggest quality of life improvement
  * [x] `PowerGridModel.calculate*` functions are now type-hint whether they return a single output dataset or a batch output dataset
  * [x] `create_output_data` now provides specific output 

#### Before
```py
node_input: ComponentData = initialize_array(DatasetType.input, ComponentType.node, 1)
line_update: ComponentData = initialize_array(DatasetType.input, ComponentType.line, (3, 4))

# nodes: SingleArray = node_input      # mypy error
nodes = cast(SingleArray, node_input)  # ugly
lines = cast(SingleArray, line_update) # oops, but allowed
```

#### After
```py
node_input: SingleArray = initialize_array(DatasetType.input, ComponentType.node, 1)
line_update: BatchArray = initialize_array(DatasetType.input, ComponentType.line, (3, 4))

# no need for casting here
# less error-prone
```

#### Before
```py
model = PowerGridModel(...)
result = model.calculate_power_flow()  # type hints SingleRowBasedDataset

update_batch: BatchDataset
result = model.calculate_power_flow(update_data=update_batch)  # type hints DenseBatchRowBasedOutputDataset
```

### DIFF
--- a/src/power_grid_model/_core/power_grid_meta.py
+++ b/src/power_grid_model/_core/power_grid_meta.py
@@ -176,20 +176,20 @@ The data types for all dataset types and components used by the Power Grid Model
 def initialize_array(
     data_type: DatasetTypeLike,
     component_type: ComponentTypeLike,
-    shape: int,
+    shape: int | tuple[int],
     empty: bool = False,
 ) -> SingleArray: ...
 @overload
 def initialize_array(
     data_type: DatasetTypeLike,
     component_type: ComponentTypeLike,
-    shape: tuple,
+    shape: tuple[int, int],
     empty: bool = False,
 ) -> DenseBatchArray: ...
 def initialize_array(
     data_type: DatasetTypeLike,
     component_type: ComponentTypeLike,
-    shape: tuple | int,
+    shape: int | tuple[int] | tuple[int, int],
     empty: bool = False,
 ) -> SingleArray | DenseBatchArray:
     """
--- a/src/power_grid_model/_core/power_grid_meta.py
+++ b/src/power_grid_model/_core/power_grid_meta.py
@@ -8,7 +8,7 @@ Load meta data from C core and define numpy structured array
 
 from dataclasses import dataclass
 from enum import IntEnum
-from typing import Any
+from typing import Any, overload
 
 import numpy as np
 
@@ -172,6 +172,20 @@ The data types for all dataset types and components used by the Power Grid Model
 """
 
 
+@overload
+def initialize_array(
+    data_type: DatasetTypeLike,
+    component_type: ComponentTypeLike,
+    shape: int,
+    empty: bool = False,
+) -> SingleArray: ...
+@overload
+def initialize_array(
+    data_type: DatasetTypeLike,
+    component_type: ComponentTypeLike,
+    shape: tuple,
+    empty: bool = False,
+) -> DenseBatchArray: ...
 def initialize_array(
     data_type: DatasetTypeLike,
     component_type: ComponentTypeLike,
--- a/src/power_grid_model/_core/power_grid_model.py
+++ b/src/power_grid_model/_core/power_grid_model.py
@@ -7,6 +7,7 @@ Main power grid model class
 """
 
 from enum import IntEnum
+from typing import overload
 
 import numpy as np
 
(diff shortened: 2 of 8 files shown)


src/power_grid_model/_core/power_grid_meta.py
```diff
@@ -172,6 +172,20 @@ def _generate_meta_attributes(component: ComponentPtr) -> dict:
...
+    data_type: DatasetTypeLike,
+    component_type: ComponentTypeLike,
+    shape: int,
+    empty: bool = False,
+) -> SingleArray: ...
+@overload
+def initialize_array(
+    data_type: DatasetTypeLike,
+    component_type: ComponentTypeLike,
+    shape: tuple,
```
nitbharambe:
```suggestion
    shape: int | tuple[int],
    empty: bool = False,
) -> SingleArray: ...
@overload
def initialize_array(
    data_type: DatasetTypeLike,
    component_type: ComponentTypeLike,
    shape: tuple[int, int],
```
Maybe its not an issue but just to be precise

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -405,20 +419,128 @@ def _calculate_short_circuit(  # noqa: PLR0913
...
+        *,
+        symmetric: bool = ...,
+        error_tolerance: float = ...,
+        max_iterations: int = ...,
+        calculation_method: CalculationMethod | str = ...,
+        update_data: None = ...,
+        threading: int = ...,
+        output_component_types: dict[
+            ComponentType, set[str] | list[str] | None | ComponentAttributeFilterOptions
+        ] = ...,
```
nitbharambe:
```suggestion
        output_component_types: ComponentAttributeMappingDict = ...,
```

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -405,20 +419,128 @@ def _calculate_short_circuit(  # noqa: PLR0913
...
+        *,
+        symmetric: bool = ...,
+        error_tolerance: float = ...,
+        max_iterations: int = ...,
+        calculation_method: CalculationMethod | str = ...,
+        update_data: BatchDataset = ...,
+        threading: int = ...,
+        output_component_types: dict[
+            ComponentType, set[str] | list[str] | None | ComponentAttributeFilterOptions
+        ] = ...,
```
nitbharambe:
```suggestion
        output_component_types: ComponentAttributeMappingDict = ...,
```

nitbharambe:
Rest LGTM

src/power_grid_model/_core/power_grid_model.py
mgovers: we need to double-check which types we want to officially expose

nitbharambe: As per our discussion now: None.

nitbharambe (approved):

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -405,20 +419,124 @@ def _calculate_short_circuit(  # noqa: PLR0913
...
+        symmetric: bool = ...,
+        error_tolerance: float = ...,
+        max_iterations: int = ...,
+        calculation_method: CalculationMethod | str = ...,
+        threading: int = ...,
+        output_component_types: None | set[ComponentTypeVar] | list[ComponentTypeVar] = ...,
+        continue_on_batch_error: bool = ...,
+        decode_error: bool = ...,
+        tap_changing_strategy: TapChangingStrategy | str = ...,
+    ) -> SingleRowBasedDataset: ...
```
furqan463, line 435: @mgovers why is this overload required?
Isn't this equivalent to overload below?
Isn't `update_data: None` equivalent to omitting it?

figueroa1395: @furqan463 I believe you are correct and the first overload is indeed redundant as it's already covered in the second case. @mgovers can you confirm?

If that's the case, feel free to open a PR with the fix :)

furqan463: @mgovers can you please confirm.

nitbharambe: @figueroa1395  @furqan463 Do we mean to compare one from L422 with one on L436?
Then I see that the latter one takes in an update_data keyword argument additionally.

figueroa1395: @nitbharambe Indeed, but in the second overload we have `update_data: None = ...`, which I believe is equivalent to not having that keyword at all because `update_data` is a kwarg in the second one.

furqan463: > equivalent to not having that keyword

Or Otherwise most of the arguments are optional, so we'll have to create an overload for all, or even combinations.

mgovers: a user may do something along the lines of

```py
model.calculate_powerflow()
model.calculate_powerflow(update_data=None)  # not equivalent according to overload decorator
```

this is also made explicit by the fact that `update_data` is a strict keyword argument (comes after `*`), so specifying it explicitly vs not specifying it are 2 distinct overloads

mgovers: oh wait, i see your point. the ellipsis should take care of it. what you can do is make a code snippet, remove the overload without the `update_data` and then see what happens. if it works both in your IDE autocomplete and in mypy, then feel free to create a PR

PR #1182, opened 2025-11-05
relax numerical boundary for two validation tests

TonyXiang8787:
The `eigen` update to `5.0.0` causes some small numerical value changes, which are normal. This just cause the validation test just outside the tolerence boundary. So the boundary is relaxed.


mgovers (approved):
oh interesting. good that it's something small

PR #1183, opened 2025-11-09
Improvement: avoid string lookup when getting scenario from batch dataset

TonyXiang8787:
The current implementation of getting scenario from a batch dataset involves string lookups. This is not needed and also executed in tight loops.

The PR directly manipulates the internal data to avoid string lookup.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
@@ -108,6 +108,13 @@ struct MetaAttribute {
         using CharType = std::conditional_t<std::is_const_v<T>, char const*, char*>;
         return *reinterpret_cast<T*>(reinterpret_cast<CharType>(ptr) + offset);
     }
+
+    RawDataPtr advance_ptr(RawDataPtr ptr, Idx difference) const {
+        return reinterpret_cast<char*>(ptr) + difference * size;
+    }
+    RawDataConstPtr advance_ptr(RawDataConstPtr ptr, Idx difference) const {
+        return reinterpret_cast<char const*>(ptr) + difference * size;
+    }
 };
 
 // meta component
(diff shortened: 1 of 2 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -108,6 +108,13 @@ struct MetaAttribute {
         using CharType = std::conditional_t<std::is_const_v<T>, char const*, char*>;
         return *reinterpret_cast<T*>(reinterpret_cast<CharType>(ptr) + offset);
     }
+
+    RawDataPtr advance_ptr(RawDataPtr ptr, Idx difference) const {
+        return reinterpret_cast<char*>(ptr) + difference * size;
```
Jerry-Jinfeng-Guo, line 113: Would there be potential issue if `difference*size` is too big? Not sure if an assert is necessary here.

TonyXiang8787: We can't do `assert` here because the buffer is provided by the user (via C-API). From our side we cannot possibly know if user allocates enough buffer room.

TonyXiang8787:
Benchmark shows that there is no significant deviation on performance. The difference shown below is merely a run-to-run deviation.

**main**

```
*****Run with batch calculation*****
Number of nodes: 1804
0000.Total: 1.05045
0064.Total single calculation in thread: 1.00978
0128.Total batch calculation in thread: 1.01077
1100.		Copy model: 0.0009555
1200.		Update model: 0.0550138
1201.		Restore model: 0.0145357
2100.		Prepare: 0.0203751
2200.		Math Calculation: 0.771786
2220.			Math solver: 0.713862
2221.				Initialize calculation: 0.0958138
2225.				Solve sparse linear equation: 0.384615
2226.				Iterate unknown: 0.0387758
2227.				Calculate math result: 0.057919
2242.				Prepare the matrices: 0.136414
2246.				Max number of iterations: 4
3000.	Produce output: 0.141665
```


**PR**

```
*****Run with batch calculation*****
Number of nodes: 1804
0000.Total: 1.04518
0064.Total single calculation in thread: 0.997395
0128.Total batch calculation in thread: 0.998456
1100.		Copy model: 0.00103767
1200.		Update model: 0.0541637
1201.		Restore model: 0.0143279
2100.		Prepare: 0.0200385
2200.		Math Calculation: 0.760915
2220.			Math solver: 0.700028
2221.				Initialize calculation: 0.0941314
2225.				Solve sparse linear equation: 0.377153
2226.				Iterate unknown: 0.0376174
2227.				Calculate math result: 0.0585373
2242.				Prepare the matrices: 0.132269
2246.				Max number of iterations: 4
3000.	Produce output: 0.141883
```

PR #1186, opened 2025-11-11
Feature: support multi-dimensional batch calculation[experimental]

TonyXiang8787:
This PR experiments an idea of multi-dimensional batch calculation in the C-API without introducing breaking changes. The Python API is also adjusted without breaking changes.

### DIFF
--- a/.gersemirc
+++ b/.gersemirc
@@ -1,0 +1,12 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+definitions: [./cmake/stubs/]
+disable_formatting: false
+extensions: []
+indent: 4
+line_length: 80
+list_expansion: favour-inlining
+unsafe: false
+warn_about_unknown_commands: true
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -31,7 +31,7 @@ jobs:
     name: Build sdist and set version
     runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: actions/setup-python@v6
         with:
@@ -77,7 +77,7 @@ jobs:
       HOMEBREW_FORCE_BREWED_CURL: 1
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Install packages
         run: |
           sudo apt-get update
@@ -108,7 +108,7 @@ jobs:
       PRESET: ${{ matrix.compiler }}-${{ matrix.build-option }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Activate conda
         uses: conda-incubator/setup-miniconda@v3  # install miniforge instead
@@ -161,7 +161,7 @@ jobs:
       PRESET: ci-clang-${{ matrix.build-option }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up XCode
         uses: maxim-lobanov/setup-xcode@v1
@@ -198,7 +198,7 @@ jobs:
     needs: [acquire-python-version-build-sdist]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: actions/download-artifact@v6
         with:
@@ -215,7 +215,7 @@ jobs:
           xcode-version: latest-stable
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v3.2
+        uses: pypa/cibuildwheel@v3.3
         # GitHub Actions specific build parameters
         env:
           # pass GitHub runner info into Linux container
@@ -247,7 +247,7 @@ jobs:
       run:
         shell: ${{ matrix.shell }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: conda-incubator/setup-miniconda@v3
         with:
@@ -308,7 +308,7 @@ jobs:
           echo "Previous jobs have failures!"
           exit 1
 
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Setup python
         uses: actions/setup-python@v6
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -25,7 +25,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -51,6 +51,11 @@ jobs:
         run: |
           sudo apt-get install -y clang-format-18
           find . -regex '.*\.\(h\|c\|cpp\|hpp\|cc\|cxx\)' -exec clang-format-18 -style=file -i {} \;
+
+      - name: Install and run cmake-linter
+        run: |
+          pip install gersemi
+          gersemi -i --warnings-as-errors src/ tests/ cmake/
 
       - name: Install and run markdownlint
         run: |
@@ -95,4 +100,3 @@ jobs:
           else
             echo "The test-workflow.yml file matches main."
           fi
-
--- a/.github/workflows/citations.yml
+++ b/.github/workflows/citations.yml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
     - name: checkout
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
     - name: Install R
       run: |
         sudo apt-get update && sudo apt-get install -y r-base
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -48,7 +48,7 @@ jobs:
       HOMEBREW_FORCE_BREWED_CURL: 1
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Install packages
         run: |
--- a/.github/workflows/reuse-compliance.yml
+++ b/.github/workflows/reuse-compliance.yml
@@ -19,6 +19,6 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
     - name: checkout
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
     - name: REUSE Compliance Check
       uses: fsfe/reuse-action@v6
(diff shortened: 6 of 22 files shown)


power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -309,6 +309,36 @@ PGM_API void PGM_dataset_mutable_add_attribute_buffer(PGM_Handle* handle, PGM_Mu
...
+PGM_get_array_pointer_from_multidimensional(PGM_Handle* handle,
+                                            PGM_MultiDimensionalDataset const* multidimensional_dataset);
+
+/**
+ * @brief destroy the multidimensional dataset object
+ *
+ * @param multidimensional_dataset
+ */
+PGM_API void PGM_destroy_multidimensional_dataset(PGM_MultiDimensionalDataset* multidimensional_dataset);
+
```
TonyXiang8787, line 341: @mgovers This is one option to pass a MD datasets into the C-API. You need to have additional helper functions to create an array of datasets. We need a new opaque struct.

Another interesting approach would be using chaining. So define a function called `dataset_const_chain_another`, which store a pointer to the next dimension dataset into this one. Internnaly we can check if the `next_dataset` pointer is `nullptr` or not to traverse dimensions. This way is less intuitive, but still used in many other APIs. And you only need one extra function. We don't even need to change the options.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -485,7 +488,7 @@ template <dataset_type_tag dataset_type_> class Dataset {
     Dataset get_individual_scenario(Idx scenario) const
         requires(!is_indptr_mutable_v<dataset_type>)
     {
-        assert(0 <= scenario && scenario < batch_size());
+        assert(0 <= scenario && scenario <= batch_size());
```
mgovers: i think this will crash because `buffer.indptr[scenario + 1]` is called. In python, it's also not possible to call `[1, 2, 3][3]` (`IndexOutOfRange`)

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -510,6 +513,27 @@ template <dataset_type_tag dataset_type_> class Dataset {
         return result;
     }
 
+    // get slice dataset from batch
+    Dataset get_slice_scenario(Idx begin, Idx end) const
```
mgovers, line 517: really useful :+1: If we add an additional `increment` here, we can actually also use this in the dispatching of multithreaded batch calculations

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
mgovers: i prefer this separate ND approach, but how would we use ND-datasets, e.g. a 3-dimensional one instead of only 2-dimensional

power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
```diff
@@ -36,3 +38,4 @@ using PGM_ConstDataset = power_grid_model::meta_data::Dataset<power_grid_model::
 using PGM_MutableDataset = power_grid_model::meta_data::Dataset<power_grid_model::mutable_dataset_t>;
 using PGM_WritableDataset = power_grid_model::meta_data::Dataset<power_grid_model::writable_dataset_t>;
 using PGM_DatasetInfo = power_grid_model::meta_data::DatasetInfo;
+using PGM_MultiDimensionalDataset = std::vector<PGM_ConstDataset>;
```
mgovers, line 41: i feel like this should be able to use arbitrary dimensions

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -309,6 +309,36 @@ PGM_API void PGM_dataset_mutable_add_attribute_buffer(PGM_Handle* handle, PGM_Mu
...
+PGM_API PGM_ConstDataset const*
+PGM_get_array_pointer_from_multidimensional(PGM_Handle* handle,
+                                            PGM_MultiDimensionalDataset const* multidimensional_dataset);
+
+/**
+ * @brief destroy the multidimensional dataset object
+ *
+ * @param multidimensional_dataset
+ */
+PGM_API void PGM_destroy_multidimensional_dataset(PGM_MultiDimensionalDataset* multidimensional_dataset);
```
mgovers, line 340: inconsistent `PGM_dataset_create_*` vs `PGM_destroy_*`

power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
mgovers: C++23 has `std::mdspan`, which would probably solve this

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
```diff
@@ -137,6 +137,12 @@ typedef struct PGM_WritableDataset PGM_WritableDataset;
  * @brief Opaque struct for the information of the dataset.
  */
 typedef struct PGM_DatasetInfo PGM_DatasetInfo;
+
+/**
+ * @brief Opaque struct for the multi dimensional dataset class.
+ *
+ */
+typedef struct PGM_MultiDimensionalDataset PGM_MultiDimensionalDataset;
```
mgovers, line 145: maybe following the conventions of `std::mdspan`:


```suggestion
typedef struct PGM_MDDataset PGM_MDDataset;
```

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -309,6 +309,36 @@ PGM_API void PGM_dataset_mutable_add_attribute_buffer(PGM_Handle* handle, PGM_Mu
...
+/**
+ * @brief Get the array pointer from a PGM_MultiDimensionalDataset
+ *
+ * @param handle
+ * @param multidimensional_dataset
+ * @return
+ */
+PGM_API PGM_ConstDataset const*
+PGM_get_array_pointer_from_multidimensional(PGM_Handle* handle,
+                                            PGM_MultiDimensionalDataset const* multidimensional_dataset);
```
mgovers, line 333: maybe we can do a multi-indexing approach instead? e.g.

```cpp
PGM_mddataset_get_data(PGM_Handle* handle, PGM_MDDataset const* mddataset);

PGM_mddataset_get_flat_element(PGM_Handle* handle, PGM_MDDataset const* mddataset, PGM_Idx flattened_index);

PGM_mddataset_get_scenario(PGM_Handle* handle, PGM_MDDataset const* mddataset, PGM_Idx** multi_index, PGM_Idx n_dimensions);
```

(we can even omit `n_dimensions`, although it's probably safer to keep it)

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787: It is not really a ND, but a list of datasets. The core will make a cross-join ND calculation of combinations in the list.

The questions, do we create separate opaque struct to store an array of datasets? This requires a lot of new functions. Or do we do the chaining, this only needs one new function.

mgovers: i do believe under the current implementation, sure, but it also restricts us going forward. i feel like if we are going to support 1D and have a separate object for 2D, then we might as well go for ND. otherwise, we'll fall into the same rabbit hole that C++ has been in (watch any video on `std::mdspan` and you'll know, they are better at explaining what i mean than i am)

TonyXiang8787: The current experiment is about ND, not just 2D.

The ND calculation is created by a list of datasets. Each dataset in the list represents the mutation of that dimension. The core creates a cross-join on the mutations.

power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
TonyXiang8787: This has nothing todo with `mdspan`. We are talking about list of 1D datasets, which will be treated as cross-join for MD calculations.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787: Again, the update batch dataset collection is not MD, it is a list of 1D datasets.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
TonyXiang8787: I am thinking more like the chaining method, and omit the idea of `MDDataset` entirely, because it is not really MD.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: What do you mean exactly?

mgovers: i mean that we can do something like

```cpp
vector<thread> threads{
    ranges::transform(
        IdxView{n_threads},
        [](Idx thread_idx) {
            return dataset.get_slice_scenario(begin + thread_idx, end, n_threads);
        });
```

TonyXiang8787: I don't think that will work. The multi-threading is not partitioning the output in the way you described. The output buffer is actually divided interwined: 0 for thread 0, 1 for thread 2, 2 for thread 2, N-1 for thread N-1. Then N for thread 0 again.

mgovers: that's exactly how slicing with a stride works, e.g. (in python):

```py
>>> a = [0, 1, 2, 3, 4, 5, 6]
>>> a[0:5:3]
[0, 3]
>>> a[0:5:2]
[0, 2, 4]
```

`stride` in this case would be `n_threads`, `start` would be `0` and `end` would be `n_scenarios`

nitbharambe:
Agreed on the part that this is not "MD" and that name seems misleading. I see `KroneckerProductDatasets` can be representative. or `FlatOuterProductDatasets`.

I wonder if we can direct the call via `JobDisptach`. That can also give us better control over thread dispatch and logs in future. This can be a refactoring task.

TonyXiang8787:
@nitbharambe  @mgovers please have a look at the other PR #1201. I personally prefer that one since the interface is much cleaner.

Issue #1185, opened 2025-11-11
[BUG] `deepcopy` a `PowerGridModel` segfaults


kornerc:
**Describe the bug**

When copying a `PowerGridModel` with `deepcopy`, the code segfaults.
This is in my opinion the case because `PowerGridModel` does not implement a method `__depcopy__` and has an internal pointer to a C++ object.

**To Reproduce**

Steps to reproduce the behavior, including some calling code.

```python
from copy import copy, deepcopy

from power_grid_model import PowerGridModel

model = PowerGridModel({})

# Ok
model_pgm_copy = model.copy()

# Ok
model_copy = copy(model)

# --> Segfault <--
model_deepcopy = deepcopy(model)
```

**Expected behavior**
`deepcopy` a `PowerGridModel` succeeds.

kornerc:
I can open a PR to fix this issue.

TonyXiang8787:
@kornerc Thank you for raising this issue. I am looking forward to your PR. Note that the normal `copy` in `PowerGridModel` class behaves already like a `deepcopy`. We do not do shadow copy, it does not make sense. It is good to implement `deepcopy` explicitely.

PR #1187, opened 2025-11-11
implement `__deepcopy__`

kornerc:
**Fixes issue**: #1185 

- Implement `__deepcopy__` for `PowerGridModel`
- Write unit test for this

### DIFF
--- a/tests/unit/test_power_grid_model.py
+++ b/tests/unit/test_power_grid_model.py
@@ -176,9 +176,34 @@ def test_copy_model(model: PowerGridModel, sym_output):
     compare_result(result, sym_output, rtol=0.0, atol=1e-8)
 
 
-def test_deepcopy_model(model: PowerGridModel, sym_output):
-    model_2 = deepcopy(model)
-    result = model_2.calculate_power_flow()
+def test_deepcopy_model(model: PowerGridModel, sym_output, update_batch, sym_output_batch):
+    model_other = PowerGridModel({})
+
+    # list containing different models twice
+    model_list = [model, model_other, model, model_other]
+
+    new_model_list = deepcopy(model_list)
+
+    # check if identities are as expected
+    assert id(new_model_list[0]) != id(model_list[0])
+    assert id(new_model_list[1]) != id(model_list[1])
+    assert id(new_model_list[0]) != id(new_model_list[1])
+    assert id(new_model_list[0]) == id(new_model_list[2])
+    assert id(new_model_list[1]) == id(new_model_list[3])
+
+    # check if the deepcopied objects are really independent from the original ones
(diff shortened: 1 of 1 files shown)


src/power_grid_model/_core/power_grid_model.py
```diff
@@ -108,9 +108,18 @@ def copy(self) -> "PowerGridModel":
         new_model._all_component_count = self._all_component_count
         return new_model
 
-    def __copy__(self):
+    def __copy__(self) -> "PowerGridModel":
         return self.copy()
 
+    def __deepcopy__(self, memo: dict[int, Any]) -> "PowerGridModel":
+        # PowerGridModel.copy makes already a deepcopy
```
TonyXiang8787, line 115: Do you need to check if `id(self)` is already in the `memo`? Otherwise we do a lot of copies if this object is referenced multiple times in upper layer object?

tests/unit/test_power_grid_model.py
```diff
@@ -176,6 +176,12 @@ def test_copy_model(model: PowerGridModel, sym_output):
     compare_result(result, sym_output, rtol=0.0, atol=1e-8)
 
 
+def test_deepcopy_model(model: PowerGridModel, sym_output):
+    model_2 = deepcopy(model)
+    result = model_2.calculate_power_flow()
+    compare_result(result, sym_output, rtol=0.0, atol=1e-8)
```
TonyXiang8787, line 207: Please try to test a `deepcopy` of a `list` with two references to the same model to make sure that the `memo` works as intended.

```python
model_list = [model, model]
new_model_list = deepcopy(model_list)
assert id(new_model_list[0]) == id(new_model_list[1])
assert id(new_model_list[0]) != id(model)
```

kornerc: I've changed the test to cover more edge cases.
Please let me know if this is okay.

src/power_grid_model/_core/power_grid_model.py
kornerc: This is not needed since deepcopy checks [this already](https://github.com/python/cpython/blob/6f988b08d122e44848e89c04ad1e10c25d072cc7/Lib/copy.py#L126).
In our case it would even be possible to avoid setting memo entirely without any negative impact since deeopcopy is [taking care about that](https://github.com/python/cpython/blob/6f988b08d122e44848e89c04ad1e10c25d072cc7/Lib/copy.py#L161) too.
However, to the best of my knowledge it is recommended to set memo in custom `__deepcopy__` implementations.

TonyXiang8787 (approved):
@kornerc Thanks for your contribution. After CI check passes, the PR will be merged.

PR #1175, opened 2025-10-31
SparseMatrixErrors: power flow and short circuit special case

nitbharambe:
**Relates to issue**: #1162

### Changes proposed in this PR include:

* Added a small admittance at zero sequence where unavailable.
* Test for all combinations of windings for 2 grids:
1. source -- trafo
2. source -- trafo --- asym_load
Only DYn and YnD  delta combination is added as the behaviour is same as Y for other windings.

Note that:
* All output is calculated from current main branch. 
* If the results were failing for these cases earlier, the result validation is skipped.
* Certain test cases for NRPF are marked as xfail for IterationDiverge. These are YNd, YYn, YY, YnY.

Issue #1162, opened 2025-10-21
[BUG] *Sparse matrix error when calculating the transformer 63/10*


WinApiMan:
**Hi!I encountered a problem. When calculating the asymmetrical short circuit (A) on the 0.4 kV busbar of a transformer with Id=1503, a sparse matrix error occurs.Moreover, when calculating an asymmetrical short circuit on the buses of a similar transformer ID=1426, such an error does not occur. We reconnected the objects. We replaced ID=1426 with ID=1503, and set 1426 instead of 1503. However, the error still occurs precisely where ID=1503 is set. Perhaps this isn't a bug, but my own mistake, and you can help me solve this problem.**

**Input Data Validity**
[calculate_model 1503.json](https://github.com/user-attachments/files/23022103/calculate_model.1503.json)

[calculate_model ID1426.json](https://github.com/user-attachments/files/23022104/calculate_model.ID1426.json)

[calculate_output_model ID1426.json](https://github.com/user-attachments/files/23022114/calculate_output_model.ID1426.json)

**Screenshots**

<img width="700" height="218" alt="Image" src="https://github.com/user-attachments/assets/285882f7-4aea-4236-a1ac-be64d123a8a6" />

mgovers:
hi @WinApiMan, Thank you for reporting this. 

Can you try to load the data into Python and use the data validator (https://power-grid-model.readthedocs.io/en/stable/user_manual/data-validator.html)? This may expose some basic usage problems that may not be caught in the C API.

Can you please also provide the exact PGM version you are running (either release tag or Git SHA) and, ideally, an example script of how you run the PGM? With that, we can try to have a look at this tomorrow.

mgovers:
FYI, I've been able to reproduce the `SparseMatrixError` and I also confirmed that our data validator does not signify any errors. We will dive into this now.

WinApiMan:
Hi @mgovers, Thanks for the quick reply.

We are using version 1.12.54. I'm attaching a test example of the call script and errors.

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/cbb3a622-1476-4dfc-bb02-d6aa3c706544" />

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/0f346fb6-7855-4eec-a0ee-ce283909252e" />

mgovers:
[...]

PR #1189, opened 2025-11-16
Fix zero sequence floating in some scenarios

TonyXiang8787:
Relates to #1162.

Replaces #1175 for simpicity

* [x] Add a small admittance at zero sequence where unavailable.
* [x] Adjust unit test
* [x] Add some validation tests
* [x] Relax some tolerance of short circuit validation tests for voltage and current the `atol=1e-3`

### DIFF
--- a/tests/data/short_circuit/two_phase_to_ground_c_minimum/params.json
+++ b/tests/data/short_circuit/two_phase_to_ground_c_minimum/params.json
@@ -3,7 +3,7 @@
   "rtol": 1e-7,
   "atol": {
     "default": 1e-8,
-    "i.*": 1e-3,
+    "i(_.+)?": 1e-3,
     "u_pu": 1e-7,
     "u": 1e-3,
     "u_angle": 1e-3
--- a/tests/data/short_circuit/single_phase_to_ground_c_maximum/params.json
+++ b/tests/data/short_circuit/single_phase_to_ground_c_maximum/params.json
@@ -3,7 +3,7 @@
   "rtol": 1e-7,
   "atol": {
     "default": 1e-8,
-    "i.*": 1e-3,
+    "i(_.+)?": 1e-3,
     "u_pu": 1e-7,
     "u": 1e-3,
     "u_angle": 1e-3
--- a/tests/data/short_circuit/single_phase_to_ground_c_minimum/params.json
+++ b/tests/data/short_circuit/single_phase_to_ground_c_minimum/params.json
@@ -3,7 +3,7 @@
   "rtol": 1e-7,
   "atol": {
     "default": 1e-8,
-    "i.*": 1e-3,
+    "i(_.+)?": 1e-3,
     "u_pu": 1e-7,
     "u": 1e-3,
     "u_angle": 1e-3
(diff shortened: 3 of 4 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -258,6 +258,21 @@ class Transformer : public Branch {
...
         }
 
+        auto const low_susceptance = -transformer_low_susceptance_ratio * sn_ / base_power_3p / uk_;
+        bool const zero_seq_available_from =
+            (winding_from_ == WindingType::wye_n && winding_to_ == WindingType::wye_n) ||
+            (winding_from_ == WindingType::wye_n && winding_to_ == WindingType::delta) ||
+            winding_from_ == WindingType::zigzag_n;
+        bool const zero_seq_available_to = (winding_to_ == WindingType::wye_n && winding_from_ == WindingType::wye_n) ||
+                                           (winding_to_ == WindingType::wye_n && winding_from_ == WindingType::delta) ||
+                                           winding_to_ == WindingType::zigzag_n;
```
mgovers: maybe more readable (optionally move `zero_seq_available` to separate function):

```suggestion
        auto const zero_seq_available = [](WindingType this_side, WindingType other_side) {
            using enum WindingType;
            switch (winding_from_) {
            case wye_n:
                return winding_to_ == wye_n || winding_to_ == delta;
            case zigzag_n:
                return true;
            default:
                return false;
            }
        }();
        bool const zero_seq_available_from = zero_seq_available(winding_from_, winding_to_);
        bool const zero_seq_available_to = zero_seq_available(winding_to_, winding_from_);
```

tests/data/short_circuit/two_phase_to_ground_c_minimum/params.json
```diff
@@ -1,6 +1,12 @@
...
   "calculation_method": "iec60909",
-  "rtol": 1e-08,
-  "atol": 1e-08,
+  "rtol": 1e-7,
+  "atol": {
+    "default": 1e-8,
+    "i.*": 1e-3,
+    "u_pu": 1e-7,
+    "u": 1e-3,
+    "u_angle": 1e-3
```
mgovers, line 9: this indeed looks much better :+1: It shows intentionality

tests/data/short_circuit/two_phase_to_ground_c_minimum/params.json
```diff
@@ -1,6 +1,12 @@
 {
   "calculation_method": "iec60909",
-  "rtol": 1e-08,
-  "atol": 1e-08,
+  "rtol": 1e-7,
+  "atol": {
+    "default": 1e-8,
+    "i.*": 1e-3,
```
mgovers: it's a little bit dangerous if we ever come up with other attributes that start with `i` (like `ifoo`), but i do not expect those to be added anytime soon as we're talking about output, which is more stable than input parameters, but i want to call it out anyways. 

maybe we can do something like this? (untested, so i don't know whether that is the correct regex by heart)

```suggestion
    "i(_.*)?": 1e-3,
```

TonyXiang8787: it should be `i(_.+)?`. After underscore there should be at least one character

TonyXiang8787: done

PR #1191, opened 2025-11-17
Dev environment: add support and documentation for Docker

Laurynas-Jagutis:
Made sure the setup works with Docker instead of Podman. Added documentation on Docker's setup.
Mention Apple's container support

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -55,8 +55,8 @@ template <symmetry_tag sym_type> struct BranchShortCircuitSolverOutput {
 // fault math calculation parameters and math output
 struct FaultCalcParam {
     DoubleComplex y_fault;
-    FaultType fault_type{};
-    FaultPhase fault_phase{};
+    FaultType fault_type{FaultType::nan};
+    FaultPhase fault_phase{FaultPhase::nan};
 };
 
 template <symmetry_tag sym_type> struct FaultShortCircuitSolverOutput {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/calculation_input_preparation.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/calculation_input_preparation.hpp
@@ -274,7 +274,10 @@ std::vector<ShortCircuitInput> prepare_short_circuit_input(main_model_state_c au
         }
 
         sc_input[i].fault_buses = {from_dense, std::move(map.indvector), state.math_topology[i]->n_bus()};
+
+        // we need to reserve enough space because fault_object is updateable
         sc_input[i].faults.resize(state.components.template size<Fault>());
+
         sc_input[i].source.resize(state.math_topology[i]->n_source());
     }
 
@@ -288,6 +291,16 @@ std::vector<ShortCircuitInput> prepare_short_circuit_input(main_model_state_c au
         state, state.topo_comp_coup->source, sc_input, [&state, voltage_scaling](Source const& source) {
             return std::pair{state.components.template get_item<Node>(source.node()).u_rated(), voltage_scaling};
         });
+
+    // only now do we know where the fault objects are placed => resize to used only
+    for (auto& input : sc_input) {
+        auto const is_unused = [](FaultCalcParam const& param) {
+            return param.fault_type == FaultType::nan && param.fault_phase == FaultPhase::nan;
+        };
+        auto const first_non_used = std::ranges::find_if(input.faults, is_unused);
+        assert(std::all_of(first_non_used, input.faults.end(), is_unused));
+        input.faults.resize(std::distance(input.faults.begin(), first_non_used));
+    }
 
     return sc_input;
 }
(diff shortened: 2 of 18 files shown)


docs/working_with_containers/container-setup.md
```diff
@@ -11,8 +11,11 @@ the Python side.
 
 ## Setup Requirements
 
-To begin using the container there are two open-source applications that must be installed.
+To begin using the container, you have several options. First, you can use Docker together with DevPod. Alternatively,
+you can use the open-source combination of Podman and DevPod.
```
mgovers: by default, free tools should be recommended


```suggestion
To begin using the container, you have several options, e.g.:

* Podman + DevPod (both are free, open-source tools)
* Docker + DevPod (NOTE: Docker Desktop requires a subscription that is not free for commercial usage)
```

docs/working_with_containers/container-setup.md
```diff
@@ -11,8 +11,11 @@ the Python side.
...
 
+* Docker (<https://www.docker.com/>) - a popular container platform that uses a background service (daemon) to build,
+run, and manage containerized applications.
```
mgovers: idem: should be the last tool in the list

mgovers: please explicitly mention that Docker Desktop requires a paid subscription for commercial usage.

docs/working_with_containers/container-setup.md
```diff
@@ -11,8 +11,11 @@ the Python side.
 
 ## Setup Requirements
 
-To begin using the container there are two open-source applications that must be installed.
+To begin using the container, you have several options. First, you can use Docker together with DevPod. Alternatively,
```
mgovers: do you need Docker or Docker Desktop? The former is free, the latter is not (for commercial usage)

docs/working_with_containers/container-setup.md
```diff
@@ -28,20 +31,46 @@ Firstly, setup the repository by cloning it into the desired destination.
 git clone <https>
 ```
 
-Make sure that you are in the right branch, the branch should have ".devcontainer" folder in it.
+Make sure that you are in the right branch, the branch should have a ".devcontainer" folder in it.
```
mgovers: redundant because it is already present in `main`
```suggestion
Make sure that you are on the right branch.
```

docs/working_with_containers/container-setup.md
```diff
@@ -28,20 +31,46 @@ Firstly, setup the repository by cloning it into the desired destination.
...
+### Docker and DevPod Setup
+
+#### Docker Setup
+
+Before you begin, install Docker Desktop and open it. It will initialize automatically when launched.
+
+#### DevPod Setup
 
 Secondly, open DevPod and go into the "Providers" section. There click on "+ Add", select the docker icon, give it a
+name and click on save. DevPod should find the PATH to the docker.exe itself, however if you think that there are issues
```
mgovers:
```suggestion
name and click on save. DevPod should find the PATH to the `docker` executable itself, however if you think that there are issues
```

`.exe` extensions are windows-specific. unix does not use extensions for executables.

docs/working_with_containers/container-setup.md
figueroa1395: I believe that for Mac OS you must use `docker desktop`, as (unless this has recently changed) there isn't a CLI version. Nevertheless, for Linux and Windows, using the `docker` CLI should be possible, but the workflow would be different.

With that said, given that the "universal" way is using the desktop version, I would only explain it that way, as someone familiar with the CLI version would now how to "translate". Hence, we should be very explicit with the commercial licensing notice.

docs/working_with_containers/container-setup.md
```diff
@@ -59,3 +88,8 @@ Finally, click "Create Workspace"
 * If you are facing issues while building the container then try and allocate more memory in Podman. Go to
 Podman -> Settings (Bottom left corner) -> Resources -> Edit.
 * Try and make a fresh clone of the repository and redo the setup on it.
+
+## Support for Apple Container
+
+We have tested a container setup with Apple's container. However, currently there is not a simplistic way to open that
+container within VSCode. With future plugin developments we expect to add support for Apple's container.
```
figueroa1395: Nitpick:
```suggestion
We have tested a container setup with Apple's `container` app. However, currently there is not a simplistic way to open that
container within VSCode. With future plugin developments we expect to add support for Apple's container.
```

docs/working_with_containers/container-setup.md
```diff
@@ -28,20 +31,46 @@ Firstly, setup the repository by cloning it into the desired destination.
...
+Finally, to initialize and open the container everything will be done through DevPod.
+
+* Upon opening DevPod go to "Workspaces" and click on "Create Workspace".
+* Under the "Enter Workspace Source", select the "Folder" option and navigate to the destination where you cloned the
+repository to.
+* Under the "Default IDE" select VSCode.
+* Under "Provider" select the provider that you have setup previously (it should have the docker icon)
+
+Finally, click "Create Workspace"
+
```
figueroa1395: This section is duplicated from the `podman` section. Can you then remove the `podman` section and just hyperlink to this? Or better yet, if you switch the order around as suggested in https://github.com/PowerGridModel/power-grid-model/pull/1191/files#r2536649482, so first the setup is for `Podman` and then for `docker`, you can point things in the right order.

Keep an eye open to avoid duplication.

PR #1193, opened 2025-11-19
Bugfix: correctly handle short circuit faults in multiple subgrids

### LINKED ISSUE #1192: Validation error when validation is successful
Hi! When modeling 4 network regions (4SC.json), adding an asymmetrical short circuit in each region (4SC.png, 4SC_py.png) results in the following error during calculation. Validation proceeds without error (validation.png). When combining the districts into a single network, the error disappears. We don't know whether this is an error or not, so we're sending you the error code for analysis and correction, or for validation adjustments.

Files

[4SC.json](https://github.com/user-attachments/files/23621215/4SC.json)

Image

Image

Image

mgovers:
Fixes #1192 

The underlying issue was that in the calculation preparation, for all subgrids, space was reserved (and initialized) for all faults.

The rationale was that fault objects are updateable, and to re-use existing code, we had to do so. However, downstream in the logic, this turned out to be problematic, because we assume that the vector of faults for a subgrid is always complete.

Considered alternative: intialize faults to `nan` and update the limitation handling regarding identical fault types+phases to handle `nan` types. I did not go for this, because it is not known where else in the code this may lead to problems.

### DIFF
--- a/docs/user_manual/components.md
+++ b/docs/user_manual/components.md
@@ -128,17 +128,17 @@ If `i_n` is not provided, `loading` of line will be a `nan` value.
 
 #### Input
 
-| name   | data type | unit       | description                                |                 required                  |  update  |            valid values            |
-| ------ | --------- | ---------- | ------------------------------------------ | :---------------------------------------: | :------: | :--------------------------------: |
-| `r1`   | `double`  | ohm (Ω)    | positive-sequence serial resistance        |                 &#10004;                  | &#10060; | `r1` and `x1` cannot be both `0.0` |
-| `x1`   | `double`  | ohm (Ω)    | positive-sequence serial reactance         |                 &#10004;                  | &#10060; | `r1` and `x1` cannot be both `0.0` |
-| `c1`   | `double`  | farad (F)  | positive-sequence shunt capacitance        |                 &#10004;                  | &#10060; |                                    |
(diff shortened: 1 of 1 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/main_core/calculation_input_preparation.hpp
```diff
@@ -289,6 +292,16 @@ std::vector<ShortCircuitInput> prepare_short_circuit_input(main_model_state_c au
             return std::pair{state.components.template get_item<Node>(source.node()).u_rated(), voltage_scaling};
         });
 
+    // only now do we know where the fault objects are placed => resize to used only
+    for (auto& input : sc_input) {
+        auto const is_unused = [](FaultCalcParam const& param) {
+            return param.fault_type == FaultType::nan && param.fault_phase == FaultPhase::nan;
+        };
+        auto const first_non_used = std::ranges::find_if(input.faults, is_unused);
+        assert(std::all_of(first_non_used, input.faults.end(), is_unused));
```
Jerry-Jinfeng-Guo, line 301: Why not `ranges::all_of` here?

mgovers: had that first, but it would require something along the lines of
```suggestion
        assert(std::ranges::all_of(std::ranges::subrange(first_non_used, input.faults.end()), is_unused));
```
which is more verbose and has the same downsides as `std::all_of` has

PR #1196, opened 2025-11-21
CI: add cmake linter

mgovers:
* Add a CMake linter
  * to pre-commit
  * to CI
  * add stubs for `doctest_discover_tests`
* out of scope:
  * update pre-commit versions of other hooks
  * using versions to ensure that the version in CI is the same as in pre-commit

### DIFF
--- a/.gersemirc
+++ b/.gersemirc
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
 definitions: [./cmake/stubs/]
 disable_formatting: false
 extensions: []


.pre-commit-config.yaml
```diff
@@ -37,6 +37,12 @@ repos:
     hooks:
       - id: clang-format
         types_or: [ c++, c ]
+  - repo: https://github.com/BlankSpruce/gersemi
+    rev: 0.23.1
+    hooks:
+    - id: gersemi
+      name: cmake-linter
+      alias: cmake-linter
```
mgovers, line 45: i went with this alias because it is clearer than `gersemi`. happy to hear suggestions

TonyXiang8787: make sense

TonyXiang8787:
@mgovers do we need to adjust vscode extension recommendation to include the plugin for the linter?

mgovers:
> @mgovers do we need to adjust vscode extension recommendation to include the plugin for the linter?

i don't think so. i only see `vscode-gersemi` and it only has ~200 downloads, which is a little bit too low to my liking to make it an official recommendation.

PR #1197, opened 2025-11-21
Fix zero-sequence model of Yyn transformer

TonyXiang8787:
Relates to #1162 

Allow magnetization path flow of zero sequence in the `yn` side.

# Physical modelling

More attention is needed to correctly (enough) to model the zero-sequence magnetization impedance, espacially for 3-leg core-type transformer. 

## OpenDSS

See refereces from OpenDSS

https://opendss.epri.com/3-PhaseTransformerModeling.html
https://opendss.epri.com/ModelingTransformerCoreEffectsin.html

## pandapower

`pandapower` uses the following two attributes to represent zero-sequence magnetization:

https://pandapower.readthedocs.io/en/latest/elements/trafo.html

variable name | description
-- | -- 
`mag0_percent` | z_mag0 / z0 ratio between magnetizing and short circuit impedance (zero sequence)
`mag0_rx` | zero sequence magnetizing r/x ratio

But this definition has some problem. If the user does not wish to model zero-sequence magnetization impedance at all. He/she has to give infinity to the `mag0_percent`. Also, it depends on the specification of leakage impedance of zero-sequence, which the user might not want to specify. 

## proposal in PGM

We can adopt `pandapower` attribute but allow user to direct specify `i0` in zero sequence.

variable name | description | default
-- | -- | --|
`i0_zero_sequence` | zero-sequence magnetizing current in per unit | `i0`
`p0_zero_sequence` | zero-sequence magnetizing core loss in W | `p0 + pk * (i0_zero_sequence^2 - i0^2)`

The values defaults to the positive sequence variant to allow user to leave them empty. We can add in the documentation to emphasize that if user has a 3-leg core-type transformer with `Yyn` winding, `i0_zero_sequence` should be given, and a good guess of the value is `1.0`. We cannot default `i0_zero_sequence` to `1.0` because for tranformers which use separated iron core (usually for large transformers), the magnetization flow is the same for positive and zero sequence.

**Note:** The reason we name these `*_zero_sequence` is due to naming conflicts and the precedence of full naming like we have done with `r_grounding_from`.

# Check list

* [x] fix core
* [x] fix existing test
* [x] add new validation test
* [x] documentation on new attributes
* [x] documentation on explanation

### DIFF
--- a/docs/user_manual/components.md
+++ b/docs/user_manual/components.md
@@ -17,7 +17,7 @@ The base type for all power-grid-model components.
 #### Input
 
 | name | data type | unit | description                                                                                                                                                       | required |        update        |
-|------|-----------|------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|:--------:|:--------------------:|
+| ---- | --------- | ---- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- | :------: | :------------------: |
 | `id` | `int32_t` | -    | ID of a component. The ID should be unique across all components within the same scenario, e.g., you cannot have a node with `id=5` and another line with `id=5`. | &#10004; | &#10060; (see below) |
 
 If a component update is uniform and is updating all the elements with the same component type, IDs can be omitted or
@@ -33,7 +33,7 @@ An example of the usage of optional IDs is given in [Power Flow Example](./Power
 #### Steady state output and Short circuit output
 
 | name        | data type | unit | description                                                                                                                        |
-|-------------|-----------|------|------------------------------------------------------------------------------------------------------------------------------------|
+| ----------- | --------- | ---- | ---------------------------------------------------------------------------------------------------------------------------------- |
 | `id`        | `int32_t` | -    | ID of a component, the ID should be unique across all components, e.g., you cannot have a node with `id=5` and a line with `id=5`. |
 | `energized` | `int8_t`  | -    | Indicates if a component is energized, i.e. connected to a source                                                                  |
 
@@ -48,13 +48,13 @@ Physically a node can be a busbar, a joint, or other similar component.
 #### Input
 
 | name      | data type | unit     | description             | required |  update  | valid values |
-|-----------|-----------|----------|-------------------------|:--------:|:--------:|:------------:|
+| --------- | --------- | -------- | ----------------------- | :------: | :------: | :----------: |
 | `u_rated` | `double`  | volt (V) | rated line-line voltage | &#10004; | &#10060; |    `> 0`     |
 
 #### Steady state output
 
 | name      | data type         | unit                       | description                                                                                     |
-|-----------|-------------------|----------------------------|-------------------------------------------------------------------------------------------------|
+| --------- | ----------------- | -------------------------- | ----------------------------------------------------------------------------------------------- |
 | `u_pu`    | `RealValueOutput` | -                          | per-unit voltage magnitude                                                                      |
 | `u_angle` | `RealValueOutput` | rad                        | voltage angle                                                                                   |
 | `u`       | `RealValueOutput` | volt (V)                   | voltage magnitude, line-line for symmetric calculation, line-neutral for asymmetric calculation |
@@ -69,7 +69,7 @@ The `p` and `q` output of injection follows the `generator` reference direction 
 #### Short circuit output
 
 | name      | data type         | unit     | description                      |
-|-----------|-------------------|----------|----------------------------------|
+| --------- | ----------------- | -------- | -------------------------------- |
 | `u_pu`    | `RealValueOutput` | -        | per-unit voltage magnitude       |
 | `u_angle` | `RealValueOutput` | rad      | voltage angle                    |
 | `u`       | `RealValueOutput` | volt (V) | voltage magnitude (line-neutral) |
@@ -87,17 +87,17 @@ In this case, the attribute `from_status` and `to_status` is always 1.
 
 #### Input
 
-| name          | data type | unit | description                    | required |  update  |  valid values   |
-|---------------|-----------|------|--------------------------------|:--------:|:--------:|:---------------:|
+| name          | data type | unit | description                    | required |  update  |   valid values  |
+| ------------- | --------- | ---- | ------------------------------ | :------: | :------: | :-------------: |
 | `from_node`   | `int32_t` | -    | ID of node at from-side        | &#10004; | &#10060; | a valid node ID |
 | `to_node`     | `int32_t` | -    | ID of node at to-side          | &#10004; | &#10060; | a valid node ID |
-| `from_status` | `int8_t`  | -    | connection status at from-side | &#10004; | &#10004; |   `0` or `1`    |
-| `to_status`   | `int8_t`  | -    | connection status at to-side   | &#10004; | &#10004; |   `0` or `1`    |
+| `from_status` | `int8_t`  | -    | connection status at from-side | &#10004; | &#10004; |    `0` or `1`   |
+| `to_status`   | `int8_t`  | -    | connection status at to-side   | &#10004; | &#10004; |    `0` or `1`   |
 
 #### Steady state output
 
 | name      | data type         | unit                       | description                                                |
-|-----------|-------------------|----------------------------|------------------------------------------------------------|
+| --------- | ----------------- | -------------------------- | ---------------------------------------------------------- |
 | `p_from`  | `RealValueOutput` | watt (W)                   | active power flowing into the branch at from-side          |
 | `q_from`  | `RealValueOutput` | volt-ampere-reactive (var) | reactive power flowing into the branch at from-side        |
 | `i_from`  | `RealValueOutput` | ampere (A)                 | magnitude of current at from-side                          |
@@ -111,7 +111,7 @@ In this case, the attribute `from_status` and `to_status` is always 1.
 #### Short circuit output
 
 | name           | data type         | unit       | description                       |
-|----------------|-------------------|------------|-----------------------------------|
+| -------------- | ----------------- | ---------- | --------------------------------- |
 | `i_from`       | `RealValueOutput` | ampere (A) | magnitude of current at from-side |
 | `i_from_angle` | `RealValueOutput` | rad        | current angle at from-side        |
 | `i_to`         | `RealValueOutput` | ampere (A) | magnitude of current at to-side   |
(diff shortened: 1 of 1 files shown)


docs/user_manual/components.md
```diff
@@ -219,6 +219,16 @@ It can happen that `tap_min > tap_max`.
...
 
+**Note:**
+By default, PGM uses the same magnetization current (and thus impedance) for positive- and zero-sequence circuit.
+This is typically not the case for 3-leg core-type transformers.
+Due to lack of iron-core pass for zero-sequence flux,
+the zero-sequence magnetization current is usually significantly higher than positive sequence.
+If you want to do asymmetrical calculation with 3-leg core-type transformers,
+please set the attribute `i0_zero_sequence`.
+If the transformer specificaiton does not provide such an attribute,
+a good guess will be `i0_zero_sequence = 1.0`.
```
mgovers, line 230: Do you think it's a good idea to explicitly reference the magnetic permeability of air?

TonyXiang8787: That the guess value of `i0_zero_sequence = 1.0` has nothing to with the magnetic permeability of air. It's just coincidence.

mgovers: Oh i thought i heard you mention thet but maybe i'm mistaken. Then maybe you can give a hint to the reason why 1.0 is a good guess?

petersalemink95:
Added the `do-not-merge` tag, so we don't accidentally merge without positive confirmation from the user side.

code_generation/data/attribute_classes/input.json
```diff
@@ -223,7 +223,7 @@
                 },
                 {
                     "data_type": "double",
-                    "names": ["uk", "pk", "i0", "p0"],
+                    "names": ["uk", "pk", "i0", "p0", "i0_zero_sequence", "p0_zero_sequence"],
```
petersalemink95, line 226: Not 100% sure if `i0_zero_sequence` / `p0_zero_sequence` would be the best naming here - also not sure if there would be something better, but let's have a chat about it, since it is in the API

petersalemink95:
Adding a comment so we won't forget to update three-winding trafo's - can be done in a follow-up PR. This will just be updating the input - and probably tests, since it exists of 3 `Transformer`s

docs/user_manual/components.md
```diff
@@ -185,38 +185,53 @@ An example of usage of transformer is given in [Transformer Examples](../example
...
+| `pk`               | `double`                                                    | watt (W)         | short circuit (copper) loss                                                                                                                                                                                                 |                         &#10004;                         | &#10060; |                                 `>= 0`                                 |
+| `i0`               | `double`                                                    | -                | relative no-load current                                                                                                                                                                                                    |                         &#10004;                         | &#10060; |                         `>= p0 / sn` and `< 1`                         |
```
petersalemink95: Let's add `(magnetization)` behind `no-load`, just to be explicit - so, for example: `relative no-load (magnetization) current`.
Here and below for `io`, `p0`, `i0_zero_sequence` and `po_zero_sequence`

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -192,29 +206,39 @@ class Transformer : public Branch {
...
-
         auto const y_shunt_imag_squared = y_shunt_abs * y_shunt_abs - y_shunt.real() * y_shunt.real();
         y_shunt.imag(y_shunt_imag_squared > 0.0 ? -std::sqrt(y_shunt_imag_squared) : 0.0);
-
-        // y shunt
         y_shunt = y_shunt / base_y_to;
+
+        // shunt zero sequence
+        DoubleComplex y0_shunt;
+        // Y0 = I0_0 / (U2/sqrt3) = i0_zero_sequence_ * (S / sqrt3 / U2) / (U2/sqrt3) = i0_zero_sequence_ * S * / U2 /
```
petersalemink95:
```suggestion
        // Y0 = I0_0 / (U2/sqrt3) = i0_zero_sequence_ * (S / sqrt3 / U2) / (U2/sqrt3) = i0_zero_sequence_ * S / U2 /
```
Typo

code_generation/data/attribute_classes/input.json
TonyXiang8787: After discussion we decide to keep the name, but add explanation in docs.

mgovers: @TonyXiang8787 , @figueroa1395 and I discussed that in the office last Thursday. This was the best approach. This aligns with how we have `(r|x)_grounding_(from|to)`, in which we also have written out the full name to avoid naming conflicts with other use cases for the number `0`.

Considered alternatives were:
* `i0_0` and `p0_0`
* `i0_zero` and `p0_zero`

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -192,29 +206,39 @@ class Transformer : public Branch {
...
+
+        // shunt zero sequence
+        DoubleComplex y0_shunt;
+        // Y0 = I0_0 / (U2/sqrt3) = i0_zero_sequence_ * (S / sqrt3 / U2) / (U2/sqrt3) = i0_zero_sequence_ * S / U2 /
+        // U2
+        double const y0_shunt_abs = i0_zero_sequence_ * sn_ / u2 / u2;
+        // G0 = P0_0 / (U2^2)
+        y0_shunt.real(p0_zero_sequence_ / u2 / u2);
+        auto const y0_shunt_imag_squared = y0_shunt_abs * y0_shunt_abs - y0_shunt.real() * y0_shunt.real();
+        y0_shunt.imag(y0_shunt_imag_squared > 0.0 ? -std::sqrt(y0_shunt_imag_squared) : 0.0);
```
mgovers, line 228: out of scope of this PR but we should make a utility function for this. this happens plenty of times in our code and it has caused issues in the past when we forgot the fallback logic

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
```diff
@@ -230,19 +254,37 @@ class Transformer : public Branch {
...
+            // ground path always possible via magnetization branch
+            DoubleComplex y0 = y0_shunt;
+            if (winding_to_ == WindingType::delta) {
+                // additional path via zk
+                y0 += y_series;
+            }
+            if (y0 != DoubleComplex{0.0, 0.0}) {
+                // avoid division by zero
+                DoubleComplex const z0 = 1.0 / y0 + 3.0 * z_grounding_from_ / k / k;
+                y0 = 1.0 / z0;
```
mgovers, line 270: i think we should also do a check for `z0` OR check that `real(z_grounding_from_) \geq 0`. Otherwise, if a user specifies `z_grounding_from_ = -k*k / (3.0 * y0)`, that my result in UB.

relating to that: we almost no-where in our docs specify that `r >= 0`, only that `r` and `x` cannot be both `0.0`. I think we may have more places of UB in our code in which we do not check for negative values of `r`.

The only places where we do check that `r >= 0` is for source (via `rx_ratio` and `z01_ratio`), the asymmetric line, in which case we also require that `x > 0` and `c > 0` (is that correct?) and the line drop compensation for automatic tap

TonyXiang8787: Please create a seperate issue to address this.

TonyXiang8787: The chance of `z_grounding_from_ = -k*k / (3.0 y0)` is almost zero. I would leave it as UB.

mgovers: > The chance of z_grounding_from_ = -k*k / (3.0 y0) is almost zero. I would leave it as UB.

alright, especially if we're going to add the `r >= 0` check where needed.

Created #1203 for that

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
petersalemink95: Since out of scope - can I resolve the conversation?

petersalemink95 (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/component/transformer.hpp
mgovers: yes

mgovers (approved):
LGTM. Also approving given the positive feedback on real data as reported in https://github.com/PowerGridModel/power-grid-model/issues/1162#issuecomment-3597851023

petersalemink95:
FYI: for three winding transformers the issue still needs to be fixed. @TonyXiang8787 will make an issue for it

furqan463:
@nitbharambe @mgovers is this change only for Yyn or all 2-W trafos. Should corresponding change be made in pandapower_converter of PGM-IO to read zero seq parameters specified in PP.

TonyXiang8787:
> @nitbharambe @mgovers is this change only for Yyn or all 2-W trafos. Should corresponding change be made in pandapower_converter of PGM-IO to read zero seq parameters specified in PP.

Hi @furqan463,

Theoretically this change affects all 2W trafos. But only `Yyn` will see significant improvements. You are right, the PGM-IO should be adjusted by fetch `pandapower` zero sequence parameters. Can you create an issue in PGM-IO issue board?

furqan463:
> > @nitbharambe @mgovers is this change only for Yyn or all 2-W trafos. Should corresponding change be made in pandapower_converter of PGM-IO to read zero seq parameters specified in PP.
> 
> Hi @furqan463,
> 
> Theoretically this change affects all 2W trafos. But only `Yyn` will see significant improvements. You are right, the PGM-IO should be adjusted by fetch `pandapower` zero sequence parameters. Can you create an issue in PGM-IO issue board?

I'll create a PR.

furqan463:
> > @nitbharambe @mgovers is this change only for Yyn or all 2-W trafos. Should corresponding change be made in pandapower_converter of PGM-IO to read zero seq parameters specified in PP.
> 
> Hi @furqan463,
> 
> Theoretically this change affects all 2W trafos. But only `Yyn` will see significant improvements. You are right, the PGM-IO should be adjusted by fetch `pandapower` zero sequence parameters. Can you create an issue in PGM-IO issue board?

@TonyXiang8787 just to be sure, `power-grid-model` is still not getting `uk_zero_sequence` and `pk_zero_sequence` from the user, as `pandapower` does take these inputs as well from the user as `vk0_percent`, `vkr0_percent`.

TonyXiang8787:
> > > @nitbharambe @mgovers is this change only for Yyn or all 2-W trafos. Should corresponding change be made in pandapower_converter of PGM-IO to read zero seq parameters specified in PP.
> > 
> > 
> > Hi @furqan463,
> > Theoretically this change affects all 2W trafos. But only `Yyn` will see significant improvements. You are right, the PGM-IO should be adjusted by fetch `pandapower` zero sequence parameters. Can you create an issue in PGM-IO issue board?
> 
> @TonyXiang8787 just to be sure, `power-grid-model` is still not getting `uk_zero_sequence` and `pk_zero_sequence` from the user, as `pandapower` does take these inputs as well from the user as `vk0_percent`, `vkr0_percent`.

@furqan463, yes, that's correct. `power-grid-model` still uses the same `uk` and `pk` for zero-sequence (adjusted for winding configuration). 

In general, we are very careful to add new attributes, because removing them will be a breaking change. If users have a strong use-case to include `uk_zero_sequence` and `pk_zero_sequence`, we might consider it.

Issue #796, opened 2024-10-21
[BUG] Permanent update with batch datasets in Mac yields error


figueroa1395:
**Describe the bug**

As found in the investigation #790, when doing a permanent simple update `model.update()` with a batch dataset (batch size > 1), there is a bug that only happens on Mac and on certain architectures that made the [feedstock CI crash](https://github.com/conda-forge/power-grid-model-feedstock/pull/272) (not our own).

**To Reproduce**

A detailed explanation with investigation can be found in #790, but to simply reproduce, make a branch from main and revert https://github.com/PowerGridModel/power-grid-model/pull/790/commits/795b7e9c1c57d50d270b55ab12266a2e8f217f14, then build and run using the same [configuration ](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1054810&view=results)on Mac as the failed azure pipelines.

**Expected behavior**

Feedstock CI should not break as this feature is [properly tested](https://github.com/PowerGridModel/power-grid-model/pull/790) and [passes in our CI](https://github.com/PowerGridModel/power-grid-model/pull/778).

**Additional context**

After #790, it was found that permanent updates with batch datasets were not properly tested, and the necesary tests were added in the core in the same PR, showing that the core indeeds support it. Additionally, the root of this issue maybe related or similar to https://github.com/PowerGridModel/power-grid-model/pull/769.

mgovers:
FYI: renamed `Simple update` to `Permanent update` because it carries the meaning better. This is not a big problem, just a minor thing. "Simple update" may appear more severe to a user than it actually is.

figueroa1395:
> [@figueroa1395](https://github.com/figueroa1395) is this still an issue?

I'll try locally (now that I can) and report back later. I ad forgotten about this, thanks for the reminder.

figueroa1395:
> [@figueroa1395](https://github.com/figueroa1395) is this still an issue?

@mgovers I tried to reproduce locally on the Mac M3 by reverting https://github.com/PowerGridModel/power-grid-model/commit/795b7e9c1c57d50d270b55ab12266a2e8f217f14 without success. Since I can't reproduce locally, it still only fails on the [feedstock CI for Mac](https://github.com/conda-forge/power-grid-model-feedstock/pull/272) (Runners perhaps are of an older architecture?).

[...]

PR #1201, opened 2025-11-25
Multi-dimensional batch calculation: list of update batch dataset as a cartesian product

### LINKED ISSUE #1176: [FEATURE] 2-D or cross-join of batch updates
We need to do an hourly calculation for a full year of multiple N-1 scenarios.

You now have to create a new model for each N-1 situation and then do a batch update for the timeseries.

It would be convenient if this can be done in one step which will in effect be a cross-join between the two update sets.

n_min_1_and_timeseries = {
    ComponentType.sym_load: load_profile, # Contains an hourly value for each load
    ComponentType.line: line_status, # Contains a network mutation for a cable
}
output_data = model.calculate_power_flow(update_data=n_min_1_and_timeseries)

TonyXiang8787:
Closes #1176 

Let user to provide a list of update batch datasets, which are used as cartesian product to produce a multi-dimentional batch calculation.

The implementation is a cartesian product over a linked list.

# check list
* [x] C-API and core
* [x] C-API test
* [x] C-API in-code documentation
* [x] Python wrapper
* [x] Python test
* [x] Python in-code documentation
* [x] Adjustment on general documentation and examples

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cassert>
 #include <cmath>
 #include <complex>
 #include <concepts>


TonyXiang8787:
@mgovers this is the alternative chaining approach compared to #1186. The API is much cleaner.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -510,10 +513,43 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+
+        // empty slice
+        if (begin == end) {
+            Dataset result{true, 0, dataset_info_.dataset->name, *meta_data_};
+            result.add_buffer("node", 0, 0, nullptr, nullptr);
+            return result;
+        }
+
+        // start with begin
+        Dataset result = get_individual_scenario(begin);
```
mgovers, line 534: If i remember correctly, there may be a small bug somewhere in the `get_individual_scenario(idx)` when `idx != 0` for columnar. This will probably be blocked by that fix.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -510,10 +513,43 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+        Idx const batch_size = end - begin;
+        result.dataset_info_.is_batch = true;
+        result.dataset_info_.batch_size = batch_size;
+        for (auto&& [buffer, component_info] : std::views::zip(result.buffers_, result.dataset_info_.component_info)) {
```
mgovers: i don't think capturing by rvalue is correct here, right?

```suggestion
        for (auto& [buffer, component_info] : std::views::zip(result.buffers_, result.dataset_info_.component_info)) {
```

or is it correct because of how `zip` works?

mgovers: you're not using `buffer` but you are zipping it. is that intended?

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -212,6 +212,16 @@ PGM_API void PGM_dataset_const_add_attribute_buffer(PGM_Handle* handle, PGM_Cons
...
+/**
+ * @brief Set next const dataset for chaining for MD batch
+ *
+ * @param handle
+ * @param dataset
+ * @param next_dataset The next dataset in the chain.
+ */
+PGM_API void PGM_dataset_const_set_next(PGM_Handle* handle, PGM_ConstDataset* dataset,
+                                        PGM_ConstDataset const* next_dataset);
+
```
mgovers, line 230: do we need to add an explicit chaining function? e.g. `PGM_API void PGM_dataset_const_chain(PGM_Handle* handle, PGM_ConstDataset** datasets, Idx count);`

power_grid_model_c/power_grid_model_c/src/dataset.cpp
```diff
@@ -101,6 +101,11 @@ void PGM_dataset_const_add_attribute_buffer(PGM_Handle* handle, PGM_ConstDataset
         PGM_regular_error);
 }
 
+void PGM_dataset_const_set_next(PGM_Handle* /*unused*/, PGM_ConstDataset* dataset,
+                                PGM_ConstDataset const* next_dataset) {
+    dataset->set_next(next_dataset);
+}
```
mgovers: we should probably prevent infinite loops

```suggestion
void PGM_dataset_const_set_next(PGM_Handle* handle, PGM_ConstDataset* dataset,
                                PGM_ConstDataset const* next_dataset) {
    call_with_catch(handle, [dataset, next_dataset] {
        dataset->set_next(next_dataset);
    }
}
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -510,10 +513,43 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+            Idx const size = component_info.elements_per_scenario * batch_size;
+            component_info.total_elements = size;
+        }
+        return result;
+    }
+
+    void set_next(Dataset const* next) { next_ = next; }
```
mgovers: we should probably prevent infinite loops (not sure if this is the right syntax but you get the picture)

```suggestion
    void set_next(Dataset const* next) {
        if (this == next) {
            throw DatasetError{};
        }
        next_ = next;
    }
```

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,87 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
+        batch_dataset = batch_dataset->get_next();
+    }
+    return dimension;
+}
+
+} // namespace
+
+// run calculation
+void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options const* opt,
+                   PGM_MutableDataset const* output_dataset, PGM_ConstDataset const* batch_dataset) {
```
mgovers, line 230: here and elsewhere: shouldn't we also check for `nullptr` (either explicitly as input sanitization or as `assert`)

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,87 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
+}
+
+Idx get_batch_dimension(PGM_ConstDataset const* batch_dataset) {
+    Idx dimension = 0;
+    while (batch_dataset != nullptr) {
+        ++dimension;
+        batch_dataset = batch_dataset->get_next();
+    }
+    return dimension;
+}
```
mgovers, line 214: isn't this more like `chain` than a product?


```suggestion
Idx get_batch_dimension(PGM_ConstDataset const* batch_dataset) {
    if (batch_dataset == nullptr) {
        return 0;
    }
    if (batch_dataset->get_next() == nullptr) {
        return 1;
    }
    return 2;
}
```

tests/native_api_tests/test_api_model_multi_dimension.cpp
```diff
@@ -0,0 +1,107 @@
...
+    Model model{50.0, input_dataset};
+
+    // 3-D batch update
+    double const u_rated = 10e3;
+    std::vector<double> const u_ref{0.9, 1.0, 1.1};
+    std::vector<double> const p_specified{1e6, 2e6, 3e6, 4e6};
+    std::vector<double> const q_specified{0.1e6, 0.2e6, 0.3e6, 0.4e6, 0.5e6};
+    Idx const size_u_ref = static_cast<Idx>(u_ref.size());
+    Idx const size_p_specified = static_cast<Idx>(p_specified.size());
+    Idx const size_q_specified = static_cast<Idx>(q_specified.size());
```
mgovers: if you use `ssize` then you don't need to `static_cast`

```suggestion
    Idx const size_u_ref = std::ssize(u_ref);
    Idx const size_p_specified = std::ssize(p_specified);
    Idx const size_q_specified = std::ssize(q_specified);
```
or even
```suggestion
    auto const size_u_ref = std::ssize(u_ref);
    auto const size_p_specified = std::ssize(p_specified);
    auto const size_q_specified = std::ssize(q_specified);
```

tests/native_api_tests/test_api_model_multi_dimension.cpp
```diff
@@ -0,0 +1,107 @@
...
+    DatasetConst batch_p_specified{"update", true, size_p_specified};
+    batch_p_specified.add_buffer("sym_load", 1, size_p_specified, nullptr, nullptr);
+    batch_p_specified.add_attribute_buffer("sym_load", "p_specified", p_specified.data());
+    DatasetConst batch_q_specified{"update", true, size_q_specified};
+    batch_q_specified.add_buffer("sym_load", 1, size_q_specified, nullptr, nullptr);
+    batch_q_specified.add_attribute_buffer("sym_load", "q_specified", q_specified.data());
+
+    // chaining
+    batch_u_ref.set_next(batch_p_specified);
+    batch_p_specified.set_next(batch_q_specified);
```
mgovers: it's a little bit confusing to me because chaining is more like

```py
assert chain([1, 2], [3], [], [4]) == [1, 2, 3, 4])
```
but this is more like cartesian product, e.g. like Python `itertools.product`:
```py
assert product([1, 2], [3], [4, 5]) == [[[1, 3, 4], [1, 3, 5]], [[2, 3, 4], [2, 3, 5]]]
```

mgovers (commented):
i like the approach.

however, if i interpret the code correctly, i think the naming is wrong: rather than a `chain`-like method, it's more a `cartesian_product`-like method, is that correct?

if that is correct, please ignore the comments i made before i realized that but please do update the naming conventions

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: I'll review as soon as I can and I will get back to you on this. Because I believe I found this bug before. But give me a bit to confirm first. Just leaving this comment in the meantime to give support to this.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: You have to use `&&` because `zip` itseft returns a temperory proxy object.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787: I don't think that's needed. At least not for now until a C-API user ask for it. Let's keep the number of added C-API symbol as minimum.

tests/native_api_tests/test_api_model_multi_dimension.cpp
TonyXiang8787: I can change the C-API name to something like `PGM_dataset_const_set_next_dimension_dataset_as_cartesian_product`. 

It is long, but this name can be more descriptive for C-API user. In the description of the function I can explain the way we a allow user to provide cartesian product is through chaining (in the progrma API sense).

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: What do you mean check `nullptr`? Put `nullptr` as  `batch_dataset` is a defined behavior to indicate a single calculation.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: Something is lost during refactor. Now `buffer` is gone.

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: Your proposal will not work. The whole thing is a chained dataset through `next` pointers. You traverse down until you find `nullptr`. It can have arbitrary number of dimensions.

power_grid_model_c/power_grid_model_c/src/model.cpp
mgovers: i was also thinking about the `handle`, `opt`, `output_dataset`, etc. Probably out of scope of this PR, though

tests/native_api_tests/test_api_model_multi_dimension.cpp
mgovers: maybe `PGM_dataset_const_set_next_cartesian_product_dimension`? it's slightly shorter but carries the same meaning.

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: I don't think that's needed. C-API is normal to have UB when you provide a `nullptr` where it should not.

tests/native_api_tests/test_api_model_multi_dimension.cpp
TonyXiang8787: Let's try that name.

power_grid_model_c/power_grid_model_c/src/model.cpp
mgovers: exploitable UB and unexploitable UB/erroneous behavior are two different things, though. See e.g. https://www.youtube.com/watch?v=GtYD-AIXBHk&list=PLHTh1InhhwT57vblPGsVag5MkTm_Z9-uq&index=4 .

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -510,10 +513,48 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
 
+    // get slice dataset from batch
+    Dataset get_slice_scenario(Idx begin, Idx end) const
+        requires(!is_indptr_mutable_v<dataset_type>)
+    {
+        assert(begin <= end);
+        assert(0 <= begin);
+        assert(end <= batch_size());
+        assert(is_batch());
+        assert(is_dense());
```
figueroa1395, line 524: Shouldn't both of these also be valid checks in release mode? Specially assert(is_dense)? So `throw` here instead?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395: I did a bit of searching and what I think rings a bell for you are https://github.com/PowerGridModel/power-grid-model/issues/796 and https://github.com/PowerGridModel/power-grid-model/issues/800. However, those are not relevant anymore. So as long as this is unit tested we should be fine.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -510,10 +513,48 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
+        }
+        next_ = next;
+    }
+    Dataset const* get_next() const { return next_; }
+
   private:
     MetaData const* meta_data_;
     DatasetInfo dataset_info_;
     std::vector<Buffer> buffers_;
+    Dataset const* next_{};
```
figueroa1395: Maybe we can use more expressive naming? Something along the lines of:
- `set_next`-> `chain_next_dataset`
- `get_next`-> `get_next_chained_dataset`
- `next_`-> `next_chained_dataset_`

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -212,6 +212,16 @@ PGM_API void PGM_dataset_const_add_attribute_buffer(PGM_Handle* handle, PGM_Cons
...
 PGM_API PGM_DatasetInfo const* PGM_dataset_const_get_info(PGM_Handle* handle, PGM_ConstDataset const* dataset);
 
+/**
+ * @brief Set next const dataset for chaining for MD batch
+ *
+ * @param handle
+ * @param dataset
+ * @param next_dataset The next dataset in the chain.
+ */
+PGM_API void PGM_dataset_const_set_next(PGM_Handle* handle, PGM_ConstDataset* dataset,
```
figueroa1395: In the same spirit of https://github.com/PowerGridModel/power-grid-model/pull/1201/changes#r2614201530, can we rename to something like `PGM_dataset_const_chain_next`? Similar suggestion in other relevant places (I don't want to comment the same in multiple instances).

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -55,6 +58,7 @@ void PGM_get_indexer(PGM_Handle* handle, PGM_PowerGridModel const* model, char c
         PGM_regular_error);
 }
 
+// helper functions
 namespace {
```
figueroa1395, line 62: Should we do 

```suggestion
// helper functions
namespace detail {
```
instead?

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -142,9 +146,11 @@ constexpr auto extract_calculation_options(PGM_Options const& opt) {
 }
 } // namespace
 
-// run calculation
-void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options const* opt,
-                   PGM_MutableDataset const* output_dataset, PGM_ConstDataset const* batch_dataset) {
+// calculation implementation
+namespace {
```
figueroa1395, line 150:
```suggestion
namespace calculation_impl {
```

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,87 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
     }
 }
 
+void merge_batch_error_msgs(PGM_Handle* handle, PGM_Handle const& local_handle, Idx scenario_offset, Idx stride_size) {
+    if (local_handle.err_code == PGM_no_error) {
+        return;
+    }
```
figueroa1395, line 192: After this `if` statement, can you 
```
assert(local_handle.err_code != PGM_regular_error)
assert(local_handle.err_code != PGM_serialization_error)
```
?

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,87 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
+    if (local_handle.err_code == PGM_batch_error) {
+        for (auto&& [idx, err_msg] : std::views::zip(local_handle.failed_scenarios, local_handle.batch_errs)) {
+            handle->failed_scenarios.push_back(idx + scenario_offset);
+            handle->batch_errs.push_back(err_msg);
+        }
+    } else {
+        for (Idx i = 0; i < stride_size; ++i) {
+            handle->failed_scenarios.push_back(scenario_offset + i);
+            handle->batch_errs.push_back(local_handle.err_msg);
+        }
```
figueroa1395, line 203: Isn't this `if-else` block redundant as at this point we are expecting a `local_handle.err_code == PGM_batch_error`to always hold? When aren't we?

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,87 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
+    Idx const first_batch_size = batch_dataset->batch_size();
+    Idx const stride_size = [batch_dataset]() {
+        Idx size = 1;
+        PGM_ConstDataset const* current = batch_dataset->get_next();
+        while (current != nullptr) {
+            size *= current->batch_size();
+            current = current->get_next();
+        }
+        return size;
+    }();
```
figueroa1395: Nitpick: Can you make this a separate function like you did with `get_batch_dimension`?

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: `detail` or `impl` namespace is only needed for header files (in header-only libraries). In `cpp` file, anonymous namespace is the way to go.

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: same as above

power_grid_model_c/power_grid_model_c/src/model.cpp
figueroa1395: Edit: Typos

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: See above. @mgovers and I have a new name proposal, similar to what you said.

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: I do not see the added value here. Why do we need these asserts?

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: No, not always. If things go wrong in the update phase, it will not throw `PGM_batch_error`.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
TonyXiang8787: See comment below, @mgovers  and I have a proposal for new name.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: `throw` should only be used when we expect that user accidentally (and indirectly) put some invalid input. 

This function is not called by the user but only by our internal code. Thus only `assert`. `throw` should not be used for reporting bugs.

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: done

tests/native_api_tests/test_api_model_multi_dimension.cpp
TonyXiang8787: Naming changed

TonyXiang8787:
@mgovers @figueroa1395 @nitbharambe,

The main part is almost finished. We still need to add general documentation and some notebook examples. I propose one of you to add this. This make sure we actually have the same understanding about how cartesian product of MD batch calculation works.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -543,8 +543,12 @@ template <dataset_type_tag dataset_type_> class Dataset {
...
 
     void set_next_cartesian_product_dimension(Dataset const* next) {
-        if (this == next) {
-            throw DatasetError{"Cannot chain dataset to itself!\n"};
+        Dataset const* current = next;
+        while (current != nullptr) {
+            if (this == current) {
+                throw DatasetError{"Cannot create cyclic cartesian product dimension chain!\n"};
+            }
+            current = current->get_next_cartesian_product_dimension();
```
mgovers, line 551: Nice. This is indeed required, because it would be unrecoverable unless you explicitly send a `SIGKILL` or power cycle. That would actually be a vulnerability. Good catch!

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/model.h
```diff
@@ -91,6 +91,11 @@ PGM_API void PGM_get_indexer(PGM_Handle* handle, PGM_PowerGridModel const* model
  * If batch_dataset == NULL, it is a one-time calculation.
  * If batch_dataset != NULL, it is a batch calculation with batch update in the batch_dataset.
  *
+ * The user can use the function set_next_cartesian_product_dimension() to chain multiple batch datasets
+ * to create a multi-dimension batch calculation. The calculation core will interpret the chained datasets
+ * as a cartesian product of all the scenarios.
```
mgovers: i would explicitly not mention the word `chain` because it's easily confused with the word `chain` in other programming languages including Python, in which case it's reserved for flattening.

You can use the word/phrasing `linked list`, which is the design pattern used here, e.g.

```suggestion
 * The user can use the function set_next_cartesian_product_dimension() to combine multiple batch datasets
 * to create a multi-dimension batch calculation using a linked list pattern. The calculation core will
 * interpret the combined dataset as a cartesian product on a linked list of all the scenarios.
```

figueroa1395:
> @mgovers @figueroa1395 @nitbharambe,
> 
> The main part is almost finished. We still need to add general documentation and some notebook examples. I propose one of you to add this. This make sure we actually have the same understanding about how cartesian product of MD batch calculation works.

Good idea. I can probably pick this up. I'll bring it to Peter today so we allocate some time for it.

Edit: No further comments from me. It looks good to me at this stage besides the additional stuff you pointed out. I would only also silence the Sonar Cloud warnings as I deem them false positives.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/dataset.h
```diff
@@ -212,6 +212,22 @@ PGM_API void PGM_dataset_const_add_attribute_buffer(PGM_Handle* handle, PGM_Cons
  */
 PGM_API PGM_DatasetInfo const* PGM_dataset_const_get_info(PGM_Handle* handle, PGM_ConstDataset const* dataset);
 
+/**
+ * @brief Set the next const dataset as cartesian product dimension.
+ *
+ * This function allows user to run a batch calculation with multiple dimensions of scenarios.
+ * The way user can archive this is to combine multiple batch datasets
```
mgovers: Typo
```suggestion
 * This function allows users to run a batch calculation with multiple dimensions of scenarios.
 * The way users can archive this is to combine multiple batch datasets
```

power_grid_model_c/power_grid_model_c/src/model.cpp
figueroa1395: Before https://github.com/PowerGridModel/power-grid-model/pull/1201#discussion_r2614488331 I thought these would be valuable. Now they don't even make sense as you pointed out in that comment. Resolving.

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,88 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
+        }
+    } else {
+        for (Idx i = 0; i < stride_size; ++i) {
+            handle->failed_scenarios.push_back(scenario_offset + i);
+            handle->batch_errs.push_back(local_handle.err_msg);
+        }
+    }
+}
+
+Idx get_batch_dimension(PGM_ConstDataset const* batch_dataset) {
```
mgovers, line 207: does it make sense to expose this and the total number of scenarios to the user? so that they can (if they want to) do a sanity check whether they - for instance - don't accidentally create a dimension too many (a check that they actually may want to do since the amount of scenarios explodes combinatorically)

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,88 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
+    if (auto const batch_dimension = get_batch_dimension(batch_dataset); batch_dimension < 2) {
+        calculate_impl(handle, model, opt, output_dataset, batch_dataset);
+        return;
+    }
+
+    // get stride size of the rest of dimensions
+    Idx const first_batch_size = batch_dataset->batch_size();
+    Idx const stride_size = get_stride_size(batch_dataset);
+
+    // loop over the first dimension batche
```
mgovers: typo

```suggestion
    // loop over the first dimension batch
```

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -180,5 +186,88 @@ void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options co
...
+        size *= current->batch_size();
+        current = current->get_next_cartesian_product_dimension();
+    }
+    return size;
+}
+
+} // namespace
+
+// run calculation
+void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options const* opt,
```
mgovers, line 229: are we sure that we want to do this logic here and not in the C++ job dispatch where it's easier to test and also better multi-thread-able? E.g. we could only create separate threads for the outermost dimensions.

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -281,7 +283,7 @@ def _calculate_impl(  # noqa: PLR0913
         self,
         calculation_type: CalculationType,
         symmetric: bool,
-        update_data: Dataset | None,
+        update_data: Dataset | list[Dataset] | None,
```
mgovers, line 292: maybe make a separate type alias for this

mgovers, line 292: the return type should also be list[Dataset], right? If not, I would prefer to extend the Python Dataset class so that it is actually able to handle multi-dimensional arrays, but that's probably quite tricky because we unfortunately have many locations where we hardcode the amount of dimensions for batch data to 2.

Or we can reshape the output to the input batch shape to more clearly provide this to the user - although i think we would need to update the (de)serializer to do that... but it would be a breaking change to do so later.

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -303,15 +305,19 @@ def _calculate_impl(  # noqa: PLR0913
...
-            batch_size = 1
+            is_batch = True
+        if not isinstance(update_data, list):
+            update_data = [update_data]
+        update_data = [_map_to_component_types(x) for x in update_data]
+        prepared_update = [prepare_update_view(x) for x in update_data]
+        for this_dataset, next_dataset in itertools.pairwise(prepared_update):
+            this_dataset.set_next_cartesian_product_dimension(next_dataset)
+        update_ptr: ConstDatasetPtr = prepared_update[0].get_dataset_ptr() if prepared_update else ConstDatasetPtr()
+        batch_size = prod([x.get_info().batch_size() for x in prepared_update])
```
mgovers: `prod` can take in a generator expression, right?
```suggestion
        batch_size = prod(x.get_info().batch_size() for x in prepared_update)
```

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
mgovers, line 1: do we need to update the (de)serializer as well? Or maybe throw a hard error if the user specifies the next dimension?

tests/native_api_tests/test_api_model_multi_dimension.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    // check results
+    for (Idx idx = 0; idx < total_batch_size; ++idx) {
+        CHECK(i_source_result[idx] == doctest::Approx(i_source_ref[idx]));
+    }
+
+    // check cannot set next to itself
+    CHECK_THROWS_AS(batch_u_ref.set_next_cartesian_product_dimension(batch_u_ref), PowerGridRegularError);
+    // check cannot create cyclic linked list
+    CHECK_THROWS_AS(batch_q_specified.set_next_cartesian_product_dimension(batch_u_ref), PowerGridRegularError);
```
mgovers:
```suggestion
    SUBCASE("Correct cartesian product usage") {
        batch_u_ref.set_next_cartesian_product_dimension(batch_p_specified);
        batch_p_specified.set_next_cartesian_product_dimension(batch_q_specified);

        // output dataset
        std::vector<double> i_source_result(total_batch_size);
        DatasetMutable batch_output_dataset{"sym_output", true, total_batch_size};
        batch_output_dataset.add_buffer("source", 1, total_batch_size, nullptr, nullptr);
        batch_output_dataset.add_attribute_buffer("source", "i", i_source_result.data());

        // options
        Options const options{};

        // calculate
        model.calculate(options, batch_output_dataset, batch_u_ref);

        // check results
        for (Idx idx = 0; idx < total_batch_size; ++idx) {
            CHECK(i_source_result[idx] == doctest::Approx(i_source_ref[idx]));
        }
    }
    SUBCASE("Linked list item referring to itself is not allowed") {
        CHECK_THROWS_AS(batch_u_ref.set_next_cartesian_product_dimension(batch_u_ref), PowerGridRegularError);
    }
    SUBCASE("Cyclic linked list is not allowed") {
        CHECK_THROWS_AS(batch_q_specified.set_next_cartesian_product_dimension(batch_u_ref), PowerGridRegularError);
    }
```

src/power_grid_model/_core/power_grid_model.py
mgovers, line 292: i think we also need to update the outer `assert_valid_batch_data` and `validate_batch_data` functions

TonyXiang8787: Return a `list[Dataset]` will not make sense. The output is really ND.

The current way is flattened for all dimensions, so no adjustment is needed.

TonyXiang8787: I think this is already pretty clear. Let's not make unnessesary type alias.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: No. The serializer, deserializer just handles one dataset. Not list of dataset.

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: We can think about this kind of refactor/optimization in later stage. The C-API will not be affected.

src/power_grid_model/_core/power_grid_model.py
TonyXiang8787: We don't. The `assert_valid_batch_data` checks a single input data and a single batch dataset. Not a list of batch dataset.

power_grid_model_c/power_grid_model_c/src/model.cpp
TonyXiang8787: We can see if the user ask for this. But let's make number of added C-API functions as little as possible unless user want that.

figueroa1395:
> > @mgovers @figueroa1395 @nitbharambe,
> > The main part is almost finished. We still need to add general documentation and some notebook examples. I propose one of you to add this. This make sure we actually have the same understanding about how cartesian product of MD batch calculation works.
> 
> Good idea. I can probably pick this up. I'll bring it to Peter today so we allocate some time for it.
> 
> Edit: No further comments from me. It looks good to me at this stage besides the additional stuff you pointed out. I would only also silence the Sonar Cloud warnings as I deem them false positives.

@TonyXiang8787 I can pick this up whenever you give me the green light and finish it off.

TonyXiang8787:
> > > @mgovers @figueroa1395 @nitbharambe,
> > > The main part is almost finished. We still need to add general documentation and some notebook examples. I propose one of you to add this. This make sure we actually have the same understanding about how cartesian product of MD batch calculation works.
> > 
> > 
> > Good idea. I can probably pick this up. I'll bring it to Peter today so we allocate some time for it.
> > Edit: No further comments from me. It looks good to me at this stage besides the additional stuff you pointed out. I would only also silence the Sonar Cloud warnings as I deem them false positives.
> 
> @TonyXiang8787 I can pick this up whenever you give me the green light and finish it off.

@figueroa1395 it's ready now. So go ahead.

power_grid_model_c/power_grid_model_c/src/model.cpp
mgovers: however, this is one of those things in which it's hard to refactor at a later stage. i'd much rather refactor now to ensure that we design for change.

TonyXiang8787: I do not think it is difficult to refactor in later stage. The current implementation is pretty separated in the C-API files. If we decide to migrate this into the core, we can just reset this change to the old C-API file and implement the MD batch in the core.

mgovers: then isn't it easier to do the migration now? we of course don't need to fully support the multi threading thing i thought of, but why not put it in the right location from the start?

TonyXiang8787: We have a working implementation now and well-defined user-API. I think the priority should be releasing this fast to the user and hear user experience feedback, before we do another round of internal refactoring.

At a later stage, if and how we are going to refactor this will depend on user business case.

TonyXiang8787:
@figueroa1395 @nitbharambe @mgovers, the examples and documentations are merged here. Feel free to give green light to merge.

power_grid_model_c/power_grid_model_c/src/model.cpp
figueroa1395: Offline discussion: We'll merge this implementation and refactor/optimize as a follow up. Hence enabling users and their feedback as soon as possible.

This thread can be resolved.

PR #1206, opened 2025-12-04
Enable UV in build and CI

TonyXiang8787:
This PR tries to enable `uv` in the configuration and CI. The feature is basically opt-in. Developer can still choose the `pip` way of doing things.

# list of changes

* [x] configure `uv`.
* [x] adjust CI
* [x] adjust readthedocs
* [x] manual update pre-commit once
* [x] configure kron job to auto-update `uv.lock` and pre-commit
  * Test in CI:https://github.com/PowerGridModel/power-grid-model/actions/runs/19963543459/job/57249548521
  * Test in a fake PR: https://github.com/PowerGridModel/power-grid-model/pull/1211
* [x] modify dev container
* [x] modify documentation including build guide
* [x] fix new issues in mypy

# Testing

## Testers check list

Those have tested please check the box with your name

* [x] Windows
  * [x] Tested under the assumption that `[uv.tool] environments = [...]` is removed (https://github.com/PowerGridModel/power-grid-model/pull/1206#discussion_r2602470671, see also https://github.com/PowerGridModel/power-grid-model/pull/1206#discussion_r2602434973) (by @mgovers)
  * [x] TODO: redo testing on final version of PR
* [x] Linux (WSL) (by @mgovers)
* [x] macOS (by @figueroa1395)
* [x] Devcontainer
  * [x] Windows (by @mgovers)
  * [x] MacOS (by @figueroa1395)

### Test local

* Clone and check-out in a clean folder
* Open vscode, try to build `cmake` target, it should work as before.
* In console, run `uv sync`. Then `uv run pytest`.
* In console, run `uv run pre-commit install`. Then `uv run pre-commit`

### Test dev container

* Clone and check-out in a clean folder
* Open vscode, start dev container
* In **container** console, run `uv sync`. Then `uv run pytest`.
* In **container** console, run `uv run pre-commit install`. Then `uv run pre-commit`

### DIFF
--- a/tests/unit/utils.py
+++ b/tests/unit/utils.py
@@ -38,7 +38,9 @@ from power_grid_model.utils import json_deserialize, json_deserialize_from_file,
 
 _Failed: type[BaseException] | None
 try:
-    from _pytest.outcomes import Failed as _Failed
+    from _pytest.outcomes import Failed as _PytestFailedOutcome
+
+    _Failed = _PytestFailedOutcome
 except ImportError:
     import warnings
 
@@ -57,7 +59,7 @@ EXPORT_OUTPUT = ("POWER_GRID_MODEL_VALIDATION_TEST_EXPORT" in os.environ) and (
     os.environ["POWER_GRID_MODEL_VALIDATION_TEST_EXPORT"] == "ON"
 )
 
-KNOWN_EXCEPTIONS: dict[str, type[BaseException] | None] = {
+KNOWN_EXCEPTIONS: dict[str, type[BaseException] | None] = {  # blas
     ex.__name__: ex
     for ex in (
         PowerGridBatchError,


uv.lock.license
mgovers, line 1: can we put this under our own license? or do we need a separate license for this?

pyproject.toml
```diff
@@ -33,16 +33,21 @@ requires-python = ">=3.11"
...
-    "pytest-cov",
     "ruff",
     "mypy",
-    "gersemi",  # CMake linter
+    "numpy",  # needed for mypy
+    "gersemi",              # CMake linter
+    "clang-format ~= 18.0", # C++ formatter
 ]
+
+dev = [{ include-group = "test" }, { include-group = "lint" }]
```
mgovers, line 49: nice

pyproject.toml
```diff
@@ -33,16 +33,21 @@ requires-python = ">=3.11"
...
+
+lint = [
     "pre-commit",
-    "pytest",
-    "pytest-cov",
     "ruff",
     "mypy",
-    "gersemi",  # CMake linter
+    "numpy",  # needed for mypy
+    "gersemi",              # CMake linter
```
mgovers, line 45: nitpick: consistent formatting
```suggestion
    "numpy",                # needed for mypy
    "gersemi",              # CMake linter
```

.python-version
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+3.14
```
mgovers, line 5: 😮

.github/workflows/check-code-quality.yml
```diff
@@ -27,35 +27,29 @@ jobs:
...
-        run: pip install --upgrade pip
+      - name: Install most linters from uv
+        run: uv sync --only-group lint
 
-      - name: Install and run mypy
+      - name: Run mypy
         run: |
-          pip install mypy numpy
-          mypy .
+          uv run mypy .
```
mgovers: looks like if you use `uv run` without `--no-project`, then it actually tries to install the project.

There is also `uv tool` but they explicitly discourage that for `mypy` (https://docs.astral.sh/uv/concepts/tools/#relationship-to-uv-run):

> If the tool should not be isolated from the project, e.g., when running pytest or mypy, then uv run should be used instead of uv tool run.

So i guess we cannot do without doing a full installation -- unless we can find a way to run `uv sync` without compiling the DLL, because I do not think the linters and type checkers need the DLL

mgovers: or would it work to do `--no-binary` (i.e. `uv sync --no-binary --only-group lint`)

TonyXiang8787: Let's just then install the project. Do not complicate things.

pyproject.toml
TonyXiang8787: It's very hard. I don't think it is needed for `pyproject.toml`.

uv.lock.license
TonyXiang8787: Auto generated file can use the same license as project.

mgovers:
docs for using `uv` in readthedocs: https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv
```yml
      pre_create_environment:
         - asdf plugin add uv
         - asdf install uv latest
         - asdf global uv latest
      create_environment:
         - uv venv "${READTHEDOCS_VIRTUALENV_PATH}"
      install:
         - UV_PROJECT_ENVIRONMENT="${READTHEDOCS_VIRTUALENV_PATH}" uv sync --frozen --group docs
```

.github/workflows/refresh-lock-and-linter-dependencies.yml
```diff
@@ -0,0 +1,94 @@
...
+            local version
+            version=$(uv pip freeze | grep -i "^$pkg==" | head -n1 | cut -d= -f3)
+            echo "${pkg}=$version" >> "$GITHUB_OUTPUT"
+          }
+          grab_installed_version "ruff"
+          grab_installed_version "clang-format"
+          grab_installed_version "mypy"
+          grab_installed_version "gersemi"
+
+      - name: Update .pre-commit-config.yaml
```
mgovers, line 67: I also just found this: https://pre-commit.com/#updating-hooks-automatically

But this PR is probably the better path going forward since otherwise the versions may diverge again.

One risk that this PR has is that the version on precommit may be lagging by a couple days, however, I don't think that's an issue because it means that the `uv` version should probably not be synced yet anyways.

TonyXiang8787: I can add a pre-commit install before creating the PR. If that fails, the refresh will fail. And it will wait until next time.

TonyXiang8787: That auto-update will not garantee the same version as in the dependency group of uv.

mgovers: Yup that's also what I meant, hence my statement that this PR is probably the way to go anyways

.devcontainer/devcontainer.json
```diff
@@ -13,10 +9,10 @@
 
   // cache pip installs
   "mounts": [
-    "type=volume,source=dev-pip-cache,target=/root/.cache/pip"
+    "type=volume,source=dev-pip-cache,target=/root/.cache/uv"
   ],
 
-  "postCreateCommand": "mkdir -p src/power_grid_model/_core/power_grid_model_c/lib && ln -sf /usr/local/lib/python3.13/site-packages/power_grid_model/_core/power_grid_model_c/lib/libpower_grid_model_c.so src/power_grid_model/_core/power_grid_model_c/lib/ && pre-commit install || true",
+  "postCreateCommand": "uv sync && uv run pre-commit install",
```
mgovers: is there any particular reason to also remove `|| true`?

.devcontainer/Dockerfile
mgovers, line 1: is this no longer needed?

.github/workflows/ci.yml
```diff
@@ -55,9 +55,18 @@ jobs:
...
       target: "all"
 
+  refresh-lock-and-linter-dependencies:
+    name: Refresh lock and linter Dependencies
+    uses: "./.github/workflows/refresh-lock-and-linter-dependencies.yml"
+    permissions:
+      contents: write
+      pull-requests: write
+    with:
+      create-pr: false
```
mgovers: do we want to do this on every single PR? it may be quite common for a 3rd party package to update e.g. between our CI run and our merge queue run. I also don't know whether there will be many merge conflicts.

maybe we only need to run it in CI if one of the `pyproject.toml`, `.pre-commit-config.yaml` and `.pre-commit-config.yaml.jinja`

TonyXiang8787: I do not intent to run it in PR. It's just for now to test the workflow. Before merge that will be removed.

.devcontainer/Dockerfile
TonyXiang8787: No I suppose. But that needs test.

.devcontainer/devcontainer.json
TonyXiang8787: I don't know what original `|| true` is doing. I will ask for testers.

.github/workflows/ci.yml
TonyXiang8787: Let's leave this conversation open until I have removed that run from PR.

.devcontainer/devcontainer.json
mgovers: `|| true` ensures continue on failure. that is: if the `postCreateCommand` fails, we still want to be able to enter the container rather than hard crash

TonyXiang8787: Then it makes sense to add back.

.github/workflows/refresh-lock-and-linter-dependencies.yml
```diff
@@ -0,0 +1,100 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-linter-dependencies
+  cancel-in-progress: true
+
+
+jobs:
+  refresh-lock-and-linter-dependencies:
+    name: Refresh lock and linter Dependencies
+    runs-on: ubuntu-24.04
```
figueroa1395: Nitpick:
```suggestion
    runs-on: ubuntu-latest
```

I see this isn't homogenous throughout the codebase though.

pyproject.toml
```diff
@@ -68,6 +73,14 @@ Discussion = "https://github.com/orgs/PowerGridModel/discussions"
 [project.entry-points."cmake.root"]
 power_grid_model = "power_grid_model._core.power_grid_model_c"
 
+[tool.uv]
+
+environments = [
+    "implementation_name == 'cpython' and sys_platform == 'darwin'",
+    "implementation_name == 'cpython' and sys_platform == 'linux'",
+    "implementation_name == 'cpython' and sys_platform == 'win32' and platform_machine == 'amd64'",
```
mgovers: not sure what's up here but 
```bash
> uv sync
Using CPython 3.14.2
Creating virtual environment at: .venv
Resolved 118 packages in 1ms
error: The current Python platform is not compatible with the lockfile's supported environments: `implementation_name == 'cpython' and sys_platform == 'darwin'`, `implementation_name == 'cpython' and sys_platform == 'linux'`, `implementation_name == 'cpython' and platform_machine == 'amd64' and sys_platform == 'win32'`
```

NOTE: I just ran `uv self update` and am fully up-to-date here

After removing `platform_machine == 'amd64'`, it was fine



```suggestion
    "implementation_name == 'cpython' and sys_platform == 'win32'",
```

and this also worked 
see also https://github.com/astral-sh/uv/issues/11664#issuecomment-2675114731

```suggestion
    "implementation_name == 'cpython' and sys_platform == 'win32' and platform_machine == 'AMD64'",
```

However, this suggests that maybe adding this explicit requirement causes us to not be able to be built for other platforms... Looking at https://docs.astral.sh/uv/concepts/projects/config/#limited-resolution-environments, we may not actually be needing this section cfr. https://github.com/PowerGridModel/power-grid-model/pull/1206/files#r2602470671

pyproject.toml
```diff
@@ -68,6 +73,14 @@ Discussion = "https://github.com/orgs/PowerGridModel/discussions"
...
+]
+
```
mgovers: Cfr. https://github.com/PowerGridModel/power-grid-model/pull/1206/files#r2602434973

```suggestion
```

.devcontainer/devcontainer.json
mgovers, line 1: tested on Windows following the docs in https://power-grid-model.readthedocs.io/en/stable/working_with_containers/container-setup.html#container-setup-guide

tests/unit/test_0Z_model_validation.py
```diff
@@ -97,7 +97,9 @@ def get_kwargs(sym: bool, calculation_type: str, calculation_method: str, params
 def maybe_raises(params: dict):
     raises = params.get("raises")
     if raises:
-        return pytest.raises(KNOWN_EXCEPTIONS[raises.get("raises", "AssertionError")])
+        raise_type = KNOWN_EXCEPTIONS[raises.get("raises", "AssertionError")]
+        if raise_type is not None:
+            return pytest.raises(raise_type)
```
mgovers, line 102: what was the problem?

tests/unit/utils.py
```diff
@@ -36,10 +36,11 @@
 )
 from power_grid_model.utils import json_deserialize, json_deserialize_from_file, json_serialize_to_file
 
+_Failed: type[BaseException] | None
 try:
-    from _pytest.outcomes import Failed as _Failed  # pylint: disable=import-outside-toplevel
+    from _pytest.outcomes import Failed as _Failed
 except ImportError:
-    import warnings  # pylint: disable=import-outside-toplevel
+    import warnings
```
mgovers, line 45: pre-commit triggers on this:
```
tests/unit/utils.py:41:5: error: Name "_Failed" already defined on line 39  [no-redef]
```
(or was that the intention? 😛 )

.devcontainer/devcontainer.json
```diff
@@ -1,10 +1,6 @@
 {
   "name": "power-grid-model",
-  "build": {
-    "dockerfile": "Dockerfile",
-    "context": ".."
-  },
-
+  "image": "ghcr.io/astral-sh/uv:debian",
```
figueroa1395, line 3: So the `Dockerfile` is no longer needed because we use this pre-built image, right? Is the license on this image compliant? Double checking.

TonyXiang8787: The license is fine. The old way also used a pre-built image and added our own stuff.

tests/unit/utils.py
TonyXiang8787: I do not have any error in pre-commit. Where is the error coming from.

tests/unit/test_0Z_model_validation.py
TonyXiang8787: `pytest.raises` does not accept `None`, at least in their signature. `mypy` complains.

pyproject.toml
TonyXiang8787: The limited resolution environments only applies for `uv.lock` so we can limit the size of `uv.lock` to certain environment. This is a dev only stuff. Final user is not seeting this.

tests/unit/utils.py
mgovers: `mypy`.

mgovers: I can reproduce it on Windows, on WSL and in the devcontainer. Only MacOS seems to behave differently 🤔

pyproject.toml
mgovers: but what is the benefit of that? if I rerun without the environment set, it is actually not that much larger (only about 2%):

* this branch: 357 KiB
* after running removing the environment + running `uv lock`: 363 KiB
  * although it contains -198 + 580 = 382 more lines, many of the lines are much shorter because the environment resolution doesn't need to be added to each entry

pyproject.toml
```diff
@@ -33,16 +33,21 @@ requires-python = ">=3.11"
...
+lint = [
     "pre-commit",
-    "pytest",
-    "pytest-cov",
     "ruff",
     "mypy",
-    "gersemi",  # CMake linter
+    "numpy",  # needed for mypy
+    "gersemi",              # CMake linter
+    "clang-format ~= 18.0", # C++ formatter
```
mgovers, line 46: I am getting this weird error from `pre-commit` now if I accidentally forget to load the environment that i did not get before.

> ```
> ==> File C:\Users\MartijnGovers\.cache\pre-commit\repojyv_np3h\.pre-commit-hooks.yaml
> ==> At Hook(id='clang-format')
> ==> At key: types_or
> ==> At index 10
> =====> Type tag 'metal' is not recognized.  Try upgrading identify and pre-commit?
> ```

Seems like it's caused by https://github.com/pre-commit/mirrors-clang-format/compare/v18.1.3...v18.1.4

Upgrading pre-commit + identify for the user environment did not solve it. But I guess that will not be a blocker - it just means that we shouldn't forget to load the correct environment cause then it will work. Maybe it's good to add that to the trouble-shooting

pyproject.toml
mgovers: in addition, if you limit environments, devs on a different environment will actually not be able to develop in that environment using `uv sync` at all and it forces them to manually install their environment. If you do not limit, then they will be able to without any issues. It adds a restriction that shouldn't be there in the first place.

tests/unit/utils.py
TonyXiang8787: @mgovers I am running into wall here, can you try to solve it?

pyproject.toml
TonyXiang8787: Do we need action point?

pyproject.toml
TonyXiang8787: removed.

pyproject.toml
mgovers: i don't think so... but if more people run into this, maybe we need to add it to the docs. for now, let's not

tests/unit/utils.py
mgovers: let me have a look :eyes:

mgovers: fixed

PR #1208, opened 2025-12-05
fix 2 blocker-level sonar cloud issues

mgovers:
During reviewing of #1207, I found out that there were some previously accepted Sonar Cloud issues that were marked severity level `Blocker` that were very easy to fix (missing `noexcept` for `move` operations), so I went ahead and did just that.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
@@ -45,6 +45,13 @@ using nlohmann::json;
 struct JsonMapArrayData {
     size_t size{};
     msgpack::sbuffer buffer;
+
+    JsonMapArrayData() = default;
+    JsonMapArrayData(JsonMapArrayData const&) = delete;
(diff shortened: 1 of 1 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -45,6 +45,13 @@ using nlohmann::json;
 struct JsonMapArrayData {
     size_t size{};
     msgpack::sbuffer buffer;
+
+    JsonMapArrayData() = default;
+    JsonMapArrayData(JsonMapArrayData const&) = delete;
+    JsonMapArrayData& operator=(JsonMapArrayData const&) = delete;
+    JsonMapArrayData(JsonMapArrayData&&) noexcept = default;
+    JsonMapArrayData& operator=(JsonMapArrayData&&) noexcept = default;
+    ~JsonMapArrayData() = default;
```
mgovers, line 54: we have to explicitly declare these because `msgpack::sbuffer` does not have a move constructor marked `noexcept`, so this struct is not `noexcept` move constructible by default

figueroa1395 (approved):

PR #1212, opened 2025-12-08
Code quality: MSVC with warning level /W4

mgovers:
This finally brings the warning level for MSVC at the same severity level as for GCC, Clang and Apple Clang.

NOTE: this should also (slightly?) reduce the amount of times CI/clang-tidy/Sonar complain in CI on things that were passing while developing locally on MSVC, as MSVC mostly complains about similar things

PR #1213, opened 2025-12-09
Sonar cloud: fix after #1212

mgovers:
#1212 introduced some changes that sonar cloud did not like. It seemed like they were warnings on `main` already but apparently, the actual contents of the warning differed slightly. This PR seeks to improve that.

~Note: this PR does not point to `main` yet because this is the way to test. Please point to `main` before merging.~ EDIT: pointed to `main`

### DIFF
--- a/tests/cpp_unit_tests/test_deserializer.cpp
+++ b/tests/cpp_unit_tests/test_deserializer.cpp
@@ -282,7 +282,7 @@ class OverloadErrorVisitorImpl : public detail::DefaultErrorVisitor {
 
     static constexpr auto static_err_msg = "Overload error message.\n"sv;
 
-    bool visit_nil() {
+    [[noreturn]] bool visit_nil() {
         visited_nil_ = true;
         throw_error();
     }


mgovers:
this PR does not point to main yet because this is the way to test. Please point to main before merging. `do-not-merge` label added for this reason.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/deserializer.hpp
```diff
@@ -170,36 +176,35 @@ struct CheckHasMap : DefaultNullVisitor {
...
-    bool start_map_value() { return throw_error(); }
-    bool end_map_value() { return throw_error(); }
-    bool end_map() { return throw_error(); }
-
-    bool throw_error() { throw SerializationError{(static_cast<T&>(*this)).get_err_msg()}; }
-
-    std::string get_err_msg() { return std::string{T::static_err_msg}; }
-
-  private:
+    [[noreturn]] bool visit_nil(this auto const& self) { self.throw_error(); }
```
mgovers, line 182: the `this auto const& self` was actually required because this was an actual regression

figueroa1395: Why wasn't this watched earlier in the CI and only in the nightly?

figueroa1395: Answering my self. We simply didn't have tests in place for this

PR #1221, opened 2025-12-14
CI: remove pre-commit before run pull request

TonyXiang8787:
The refresh workflow has a problem. When PR is automatically created, pre-commit kicks in which might give error and block the PR from created. So pre-commit is now removed before PR is created.

### DIFF
--- a/.github/workflows/refresh-lock-and-linter-dependencies.yml
+++ b/.github/workflows/refresh-lock-and-linter-dependencies.yml
@@ -85,6 +85,7 @@ jobs:
           source .venv/bin/activate
           pre-commit install
           pre-commit install-hooks
+          pre-commit uninstall
 
       - name: create pull request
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh-lock-and-linter-dependencies.yml
```diff
@@ -85,6 +85,7 @@ jobs:
           source .venv/bin/activate
           pre-commit install
           pre-commit install-hooks
+          pre-commit uninstall
```
mgovers, line 88: So then what is the point of having `pre-commit install`? Is it just to check whether the hooksbexist?

TonyXiang8787: Yes, to check if the newly refreshed version also exist in the hooks.

Issue #1190, opened 2025-11-17
[IMPROVEMENT] OpenSSF Best Practice badge at the ‘Silver’ Level


petersalemink95:
Receive OpenSSF Best Practice badge at the [‘Silver’ Level](https://bestpractices.coreinfrastructure.org/en/criteria/1)

TonyXiang8787:
@petersalemink95 @mgovers do we consider that this is done?

mgovers:
> [@petersalemink95](https://github.com/petersalemink95) [@mgovers](https://github.com/mgovers) do we consider that this is done?

IMO, yes because we now have the silver badge. I do think it's still good to keep the PR name of the hardening one (#1241) including the 'OpenSSF silver badge' but i think that this issue can be closed

PR #1228, opened 2025-12-16
Docs: add things that relate to OpenSSF

mgovers:
* [x] Add OpenSSF badge to README
* [x] Add documentation how to build the Python library with debug symbols for the native library

Relates to #1190

### DIFF
--- a/src/power_grid_model/utils.py
+++ b/src/power_grid_model/utils.py
@@ -6,12 +6,13 @@
 This module contains functions that may be useful when working with the power-grid-model library.
 """
 
+import io
 import json
 import math
 import tempfile
 import warnings
 from pathlib import Path
-from typing import cast as cast_type
+from typing import IO, Any, cast as cast_type
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


docs/advanced_documentation/build-guide.md
```diff
@@ -386,6 +386,22 @@ uv sync
 uv run pytest
 ```
 
+#### Build Python Library from Source with Native Debug Symbols
+
+For debugging purposes, it may be useful to build the Power Grid Model with debug symbols in the shared native library.
+To do so, the following command can be used to override the default build settings.
+
+```shell
+uv sync --config-settings=cmake.build-type="RelWithDebInfo"
```
TonyXiang8787, line 395: Why does it have to do with OpenSSF?

mgovers: Cfr. OpenSSF Silver Badge criterion:

> The build and installation system SHOULD preserve debugging information if they are requested in the relevant flags (e.g., "install -s" is not used). If there is no build or installation system (e.g., typical JavaScript libraries), select "not applicable" (N/A). [build_preserve_debug]
> E.G., setting CFLAGS (C) or CXXFLAGS (C++) should create the relevant debugging information if those languages are used, and they should not be stripped during installation. Debugging information is needed for support and analysis, and also useful for measuring the presence of hardening features in the compiled binaries. 

Note that this PR is still in draft because we will likely add more things that are still missing for the OpenSSF silver security badge to this PR.

mgovers: in addition, this is also useful for us because it actually enables us to debug C++ code from Python

power_grid_model_c/power_grid_model_c/CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ set_target_properties(
         SOVERSION ${PGM_VERSION}
         INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE
         INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO TRUE
+        BUILD_RPATH_USE_ORIGIN TRUE
```
TonyXiang8787: Set `RPATH` will not make sense for this target? It does not depend on other runtime target.

mgovers:
```suggestion
```

tests/package_tests/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ target_link_libraries(
 )
 set_target_properties(
     power_grid_model_package_test
-    PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE
+    PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE BUILD_RPATH_USE_ORIGIN TRUE
```
petersalemink95: This should not be in this PR

mgovers:
```suggestion
    PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE
```

mgovers:
```suggestion
    PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE
```

petersalemink95 (approved):

PR #1233, opened 2025-12-17
use github apps to create pr

TonyXiang8787:
This should fix the issue for CI check not triggering.

Check

PR successfully created in #1235, the CI is successfully triggered.
Refresh CI run in https://github.com/PowerGridModel/power-grid-model/actions/runs/20298182362/job/58296847394

### DIFF
--- a/.github/workflows/refresh-lock-and-linter-dependencies.yml
+++ b/.github/workflows/refresh-lock-and-linter-dependencies.yml
@@ -87,7 +87,8 @@ jobs:
           pre-commit install-hooks
           pre-commit uninstall
 
-      - uses: actions/create-github-app-token@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh-lock-and-linter-dependencies.yml
```diff
@@ -87,6 +87,12 @@ jobs:
           pre-commit install-hooks
           pre-commit uninstall
 
+      - uses: actions/create-github-app-token@v2
```
figueroa1395: Nitpick: Can you add `name` to keep it consistent with the rest?

TonyXiang8787: done

PR #1241, opened 2025-12-18
OpenSSF gold badge: hardening

mgovers:
Cherry-picked from #1228 

Investigate the possibilities of hardening

Relates to #1190


tests/native_api_tests/CMakeLists.txt
```diff
@@ -31,4 +31,9 @@ target_compile_definitions(
     PRIVATE PGM_ENABLE_EXPERIMENTAL
 )
 
+set_target_properties(
+    power_grid_model_api_tests
+    PROPERTIES BUILD_RPATH $<TARGET_FILE_DIR:power_grid_model_c>
+)
+
```
mgovers, line 44: to be removed

CMakePresets.json
```diff
@@ -137,6 +137,9 @@
     {
       "name": "msvc-debug",
       "displayName": "Debug (MSVC)",
+      "environment": {
+        "SANITIZER_FLAGS": "/RTC1"
+      },
```
mgovers, line 153: can we move this also to the CMakeLists?

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
```diff
@@ -356,7 +356,7 @@ class Topology {
         for (auto&& [idx, branch_node_idx, branch_connected] :
              std::views::zip(std::views::iota(0), std::as_const(comp_topo_.branch_node_idx),
                              std::as_const(comp_conn_.branch_connected))) {
-            assert(std::ssize(branch_connected) == 2); // NOSONAR(R354)
+            assert(std::ssize(branch_connected) == 2);
```
mgovers:
```suggestion
            assert(std::ssize(branch_connected) == 2); // NOSONAR(R354)
```

CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ include("cmake/pgm_version.cmake")
 project(power_grid_model VERSION ${PGM_VERSION})
 
 option(PGM_ENABLE_DEV_BUILD "Enable developer build (e.g.: tests)" OFF)
+option(PGM_ENABLE_HARDENING "Enable compile and link time hardening options" ON)
```
TonyXiang8787, line 15: I don't think enable it by default is logical. Disable by default and only enable in the our dev preset is logical.

In the Python build we do not enable it.

pyproject.toml
```diff
@@ -209,6 +211,8 @@ skip = ["pp*", "*-musllinux_aarch64"]
 
 [tool.cibuildwheel.linux]
 archs = ["x86_64", "aarch64"]
+before-all = """yum install -y gcc-toolset-14 && \
+    source /opt/rh/gcc-toolset-14/enable"""
```
TonyXiang8787, line 216: This should not be needed if we disable build hardening by default.

CMakeLists.txt
mgovers: Hardening is intended to be used in production

TonyXiang8787: The hardened code self (so the adjustment to make compilation to pass with hardening check flag) is intended to be used in production. The compilation flags to check hardening are not intended to be used in production.

mgovers: From the GCC command line options (https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fhardened)

> This option is intended to be used in production builds, not merely in debug builds.

See also https://www.youtube.com/watch?v=GtYD-AIXBHk&list=PLHTh1InhhwT57vblPGsVag5MkTm_Z9-uq&index=2

TonyXiang8787: I still has huge doubt for the default. This goes against the other golden rule of open-source cmake project: make as little as needed for extra compile and link flag in the default cmake build.

Also, enabling address sanitizing for production is questionable.

We need more in-depth consideration and discussion.

mgovers: We can also let CIBW set the compiler flags for hardening

TonyXiang8787:
@mgovers, whether we enable hardening by default still needs extensive discussions. Can we park that discussions first, and change this PR to enable hardening only in developer build in develop and CI cmake presets? We do not enable anything in `CMakeLists.txt` and `pyproject.toml`.

mgovers:
Cfr. offline discussions, closing for the time being in favor of #1481

PR #1242, opened 2025-12-19
Feature for more io message pack targets

SteffanWullems:
This update introduces support for additional targets with read/write I/O capabilities, 
including local storage, remote file systems, I/O buffers, and similar resources.

For example (using https://pypi.org/project/smart-open/):
`from smart_open import open as open_s3_file

TARGET_FILENAME="results.mgp"
S3_PATH = f"s3://some-aws-bucket/{TARGET_FILENAME}"

with open_s3_file( S3_PATH, "wb", transport_params={"client": s3client}) as fp:
    msgpack_serialize_to_fileobj(fp, results)

with open_s3_file( S3_PATH, "rb", transport_params={"client": s3client}) as fp:
    loaded_s3 = msgpack_deserialize_from_fileobj(fp)`


src/power_grid_model/utils.py
```diff
@@ -471,3 +472,45 @@ def _make_test_case(  # noqa: PLR0913
...
+
+def msgpack_deserialize_from_fileobj(
+    file: mp_IO[mp_Any],
+    data_filter: ComponentAttributeMapping = None,
+) -> Dataset:
+    """
+    Load and deserialize a msgpack file to a new dataset.
+
+    Args:
+        file: the ByteIO to the file to load and deserialize.
```
mgovers:
```suggestion
        file: the ByteIO stream to load and deserialize.
```

src/power_grid_model/utils.py
```diff
@@ -471,3 +472,45 @@ def _make_test_case(  # noqa: PLR0913
...
+    file: mp_IO[mp_Any],
+    data: Dataset,
+    dataset_type: DatasetType | None = None,
+    use_compact_list: bool = False,
+):
+    """
+    Export msgpack data in most recent format.
+
+    Args:
+        file: the ByteIO to the file to load and deserialize.
```
mgovers:
```suggestion
        file: the ByteIO stream to load and deserialize.
```

src/power_grid_model/utils.py
```diff
@@ -11,7 +12,7 @@
 import tempfile
 import warnings
 from pathlib import Path
-from typing import cast as cast_type
+from typing import IO as mp_IO, Any as mp_Any, cast as cast_type
```
mgovers: what does the prefix `mp` mean? Since you're naming it `ByteIO` later on, why not alias as `ByteIO` here? or just keep it as `IO`

src/power_grid_model/utils.py
```diff
@@ -471,3 +472,45 @@ def _make_test_case(  # noqa: PLR0913
...
+    """
+    Export msgpack data in most recent format.
+
+    Args:
+        file: the ByteIO to the file to load and deserialize.
+        data: a single or batch dataset for power-grid-model.
+        use_compact_list: write components on a single line.
+
+    Returns:
+        Save to BytesIO file.
```
mgovers, line 518:
```suggestion
        Save to BytesIO stream.
```

src/power_grid_model/utils.py
```diff
@@ -471,3 +472,45 @@ def _make_test_case(  # noqa: PLR0913
     params_json = json.dumps(params, indent=2)
     (output_path / "params.json").write_text(data=params_json, encoding="utf-8")
     (output_path / "params.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
+
+
+def msgpack_deserialize_from_fileobj(
```
mgovers: please add a unit test

src/power_grid_model/utils.py
```diff
@@ -471,3 +472,45 @@ def _make_test_case(  # noqa: PLR0913
...
+        ValueError: if the data is inconsistent with the rest of the dataset or a component is unknown.
+        PowerGridError: if there was an internal error.
+
+    Returns:
+        The deserialized dataset in Power grid model input format.
+    """
+    return msgpack_deserialize(file.read(), data_filter=data_filter)
+
+
+def msgpack_serialize_to_fileobj(
```
mgovers: please add a unit test

src/power_grid_model/utils.py
SteffanWullems: reason is IO and Any can clash

SteffanWullems: Also IO[Any] allows all IO streams

mgovers: but what does `mp_` mean?

if name clashes are the only reason, please instead use module-private prefixes

```suggestion
from typing import IO as _IO, Any as _Any, cast as cast_type
```

src/power_grid_model/utils.py
mgovers: thinking whether we should explicitly rename to `msgpack_deserialize_from_stream`. @figueroa1395 @nitbharambe do you have any opinion?

tests/unit/test_serialization.py
```diff
@@ -790,3 +798,14 @@ def test_serialize_deserialize_double_round_trip(deserialize, serialize, seriali
 
             np.testing.assert_array_equal(nan_a, nan_b)
             np.testing.assert_allclose(field_result_a[~nan_a], field_result_b[~nan_b], rtol=1e-15)
+
+
+def test_messagepack_round_trip_to_io(serialized_data):
+    data = to_msgpack(serialized_data)
+    input_data: Dataset = msgpack_deserialize(data)
+
+    io_buffer_data = BytesIO()
```
mgovers: since the input type is `IO[Any]`, we should test both `BytesIO` and `TextIO`.

EDIT: I implemented this and it fails (exception is raised during deserialization)

SteffanWullems: The issue is if you set BytesIO it will not work for all bytes based streams.

mgovers: i'd still prefer that over things that we say we support but in reality don't

figueroa1395:
@SteffanWullems Can you please sign your commits? I'll make a full review first thing on Monday morning.

src/power_grid_model/utils.py
```diff
@@ -1,3 +1,4 @@
+# SPDX-FileCopyrightText: 2025 Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
```
figueroa1395: Can you remove this line? Where did it come from?

src/power_grid_model/utils.py
```diff
@@ -471,3 +473,58 @@ def _make_test_case(  # noqa: PLR0913
     params_json = json.dumps(params, indent=2)
     (output_path / "params.json").write_text(data=params_json, encoding="utf-8")
     (output_path / "params.json.license").write_text(data=LICENSE_TEXT, encoding="utf-8")
+
+
+def msgpack_deserialize_from_stream(
+    stream: IO[bytes],
```
figueroa1395, line 478: Nitpick: Here and elsewhere
```suggestion
    stream: BinaryIO,
```

src/power_grid_model/utils.py
```diff
@@ -471,3 +473,58 @@ def _make_test_case(  # noqa: PLR0913
...
+
+    Raises:
+        ValueError: if the data is inconsistent with the rest of the dataset or a component is unknown.
+        PowerGridError: if there was an internal error.
+
+    Returns:
+        The deserialized dataset in Power grid model input format.
+    """
+    if stream is IO[Any]:
+        raise TypeError("Expected a stream.")
```
figueroa1395, line 495: I believe this is broken and does nothing. `typing.IO` is to be used only for type hinting and this check should always be `false`. I think it can be removed.

src/power_grid_model/utils.py
```diff
@@ -471,3 +473,58 @@ def _make_test_case(  # noqa: PLR0913
...
+    Args:
+        stream: the Binary IO stream to the file to load and deserialize.
+        data: a single or batch dataset for power-grid-model.
+        use_compact_list: write components on a single line.
+
+    Returns:
+        Save to BytesIO file.
+    """
+    if stream is IO[Any]:
+        raise TypeError("Expected a stream.")
```
figueroa1395, line 521: Same as above.

nitbharambe:
LGTM after resolution of comments

src/power_grid_model/utils.py
mgovers: @SteffanWullems can you please reply?

```suggestion
```

tests/unit/test_serialization.py
```diff
@@ -790,3 +817,48 @@ def test_serialize_deserialize_double_round_trip(deserialize, serialize, seriali
...
+
+
+def test_messagepack_to_stream_text_type_error(serialized_data):
+    data = to_msgpack(serialized_data)
+    input_data: Dataset = msgpack_deserialize(data)
+
+    io_buffer_data = TextIOBase()
+    with pytest.raises(TypeError) as excinfo:
+        msgpack_serialize_to_stream(io_buffer_data, input_data, dataset_type=serialized_data["type"])
+    assert "Expected a binary stream." in str(excinfo.value)
```
mgovers:
```suggestion
    with pytest.raises(TypeError, match="Expected a binary stream."):
        msgpack_serialize_to_stream(io_buffer_data, input_data, dataset_type=serialized_data["type"])
```

tests/unit/test_serialization.py
```diff
@@ -790,3 +817,48 @@ def test_serialize_deserialize_double_round_trip(deserialize, serialize, seriali
...
+
+
+def test_messagepack_from_stream_text_type_error():
+    io_buffer_data = TextIOBase()
+    with pytest.raises(TypeError) as excinfo:
+        _ = msgpack_deserialize_from_stream(io_buffer_data)
+    assert "Expected a binary stream." in str(excinfo.value)
```
mgovers:
```suggestion
    with pytest.raises(TypeError, match="Expected a binary stream."):
        _ = msgpack_deserialize_from_stream(io_buffer_data)
```

tests/unit/test_serialization.py
```diff
@@ -790,3 +817,48 @@ def test_serialize_deserialize_double_round_trip(deserialize, serialize, seriali
...
+
+
+def test_messagepack_from_stream_readable_error():
+    io_buffer_data = FakeRawIO(initial_bytes=b"bla")
+    with pytest.raises(UnsupportedOperation) as excinfo:
+        _ = msgpack_deserialize_from_stream(io_buffer_data)
+    assert "Stream is not readable." in str(excinfo.value)
```
mgovers:
```suggestion
    with pytest.raises(UnsupportedOperation, match="Stream is not readable."):
        _ = msgpack_deserialize_from_stream(io_buffer_data)
```

tests/unit/test_serialization.py
```diff
@@ -790,3 +817,48 @@ def test_serialize_deserialize_double_round_trip(deserialize, serialize, seriali
...
+
+
+def test_messagepack_to_stream_writable_error(serialized_data):
+    data = to_msgpack(serialized_data)
+    input_data: Dataset = msgpack_deserialize(data)
+
+    io_buffer_data = FakeRawIO(initial_bytes=b"bla")
+    with pytest.raises(UnsupportedOperation) as excinfo:
+        _ = msgpack_serialize_to_stream(io_buffer_data, input_data, dataset_type=serialized_data["type"])
+    assert "Stream is not writable." in str(excinfo.value)
```
mgovers:
```suggestion
    with pytest.raises(UnsupportedOperation, match="Stream is not writable."):
        _ = msgpack_serialize_to_stream(io_buffer_data, input_data, dataset_type=serialized_data["type"])
```

src/power_grid_model/utils.py
```diff
@@ -471,3 +472,58 @@ def _make_test_case(  # noqa: PLR0913
...
+    if isinstance(stream, io.TextIOBase):
+        raise TypeError("Expected a binary stream.")
+    if not stream.writable():
+        raise io.UnsupportedOperation("Stream is not writable.")
```
mgovers, line 525: question: should we instead throw a `SerializationError` rather than a generic error? or do something equivalent to (but less hacky than)

```suggestion
    try:
        if isinstance(stream, io.TextIOBase):
            raise TypeError("Expected a binary stream.")
        if not stream.writable():
            raise io.UnsupportedOperation("Stream is not writable.")
    except Exception as e:
        raise SerializationError from e
```

src/power_grid_model/utils.py
SteffanWullems: This is outdated please close

tests/unit/test_serialization.py
SteffanWullems: This is outdated please close

src/power_grid_model/utils.py
SteffanWullems: IO[bytes] works and BinaryIO does not work with all streams unfortunately

src/power_grid_model/utils.py
SteffanWullems: No it actually works in some stream types it will cause true.

src/power_grid_model/utils.py
SteffanWullems: No it actually works in some stream types it will cause true.

src/power_grid_model/utils.py
SteffanWullems: this suggestion is not correct. try except should not include more then 1 statement.

SteffanWullems: also i do not see the generic error here. If the UnsupportedOperation is incorrect or the TypeError is incorrect then provide the proper errors and not a Exception that is very unclear.

src/power_grid_model/utils.py
figueroa1395: Can you provide an example? See https://docs.python.org/3/library/typing.html:
> The Python runtime does not enforce function and variable type annotations. They can be used by third party tools such as [type checkers](https://docs.python.org/3/glossary.html#term-static-type-checker), IDEs, linters, etc.

src/power_grid_model/utils.py
figueroa1395: Can you provide an example?

src/power_grid_model/utils.py
SteffanWullems: well some http and nfs stream objects can identify as IO[any]. I do not remember for which packages  that was true. But incorrectly created aws botoClient streams have the same issue

src/power_grid_model/utils.py
mgovers: > this suggestion is not correct. try except should not include more then 1 statement.

Try-block and Except-blocks can contain an arbitrary amount of statements. Why would it only contain a single statement? It's a design choice, not a language rule. Also, this suggestion was just a hacky example (as I mentioned - I also do not like it in its current form, but it's about the rest of the statement).

That said, I understand your sentiment. Let's go with TypeError and UnsupportedOperation

src/power_grid_model/utils.py
figueroa1395: See https://stackoverflow.com/a/59712897:
>BinaryIO and TextIO directly subclass IO[bytes] and IO[str] respectively, and add on a few extra methods -- see the [definitions in typeshed](https://github.com/python/typeshed/blob/64bc0590a6615266180b5640c730f62f8ad84d04/stdlib/typing.pyi#L686) for the specifics.

`BinaryIO` should be almost equivalent to `IO[bytes]` but with added methods and explicit intention. Why wouldn't it work then?

src/power_grid_model/utils.py
figueroa1395: That's a good example. Resolving.

src/power_grid_model/utils.py
figueroa1395: Resolving per https://github.com/PowerGridModel/power-grid-model/pull/1242#discussion_r2661662536

src/power_grid_model/utils.py
SteffanWullems: Because it does not work. Not every stream is a BinaryIO while it is an IO[bytes]

SteffanWullems: And yes it should be an alias

mgovers (dismissed):

figueroa1395 (dismissed):

mgovers:
i'll resolve the code quality issues.

mgovers (approved):

PR #1245, opened 2025-12-29
[BUG] handle corruption in python multi-threads

TonyXiang8787:
The current Python wrapper for PGM core is not thread safe. It uses a global handle for the C-API core. If user tries to run PGM in multiple Python threads, the handle can be corrupted.

This is already the case when GIL in in place. It will be worse if the no-GIL version of Python is used, e.g. `cp314t`.

We should fix this by creating per thread a separate handle.

* [x] Reproduce error
* [x] Create separate handle per thread

### DIFF
--- a/src/power_grid_model/_core/power_grid_core.py
+++ b/src/power_grid_model/_core/power_grid_core.py
@@ -564,6 +564,8 @@ class PowerGridCore:
 
 # make one instance
 def get_power_grid_core() -> PowerGridCore:
-    if not hasattr(_thread_local_data, "power_grid_core"):
+    try:
+        return _thread_local_data.power_grid_core
+    except AttributeError:
         _thread_local_data.power_grid_core = PowerGridCore()
-    return _thread_local_data.power_grid_core
+        return _thread_local_data.power_grid_core


tests/unit/test_python_threading.py
Jerry-Jinfeng-Guo, line 1: Shall we enhance the test to cover:
- more threads, instead of just 2 start-join pairs per loop
- more function calls, instead of just the (de-)serializations

The rest of the fix looks good to me.

src/power_grid_model/_core/power_grid_core.py
```diff
@@ -560,4 +563,7 @@ def destroy_serializer(self, serializer: SerializerPtr) -> None:  # type: ignore
 
 
 # make one instance
-power_grid_core = PowerGridCore()
+def get_power_grid_core() -> PowerGridCore:
+    if not hasattr(_thread_local_data, "power_grid_core"):
+        _thread_local_data.power_grid_core = PowerGridCore()
+    return _thread_local_data.power_grid_core
```
Jerry-Jinfeng-Guo: How about the following to have less look-ups.
```
def get_power_grid_core() -> PowerGridCore:
    try:
        return _thread_local_data.power_grid_core
    except AttributeError:
        _thread_local_data.power_grid_core = PowerGridCore()
        return _thread_local_data.power_grid_core
```

tests/unit/test_python_threading.py
TonyXiang8787: The problem is the shared global handle. Using one test is enough to re-produce and fix the issue.

src/power_grid_model/_core/power_grid_core.py
TonyXiang8787: done

mgovers:
Good catch!

I've been thinking about something related to this but didn't realize that this was also broken.

I myself have been thinking about (for instance) process cancellation (e.g.: `KeyboardInterrupt` / `SIGINT` handling). E.g. if the user expects a calculation to take milliseconds but then it takes a couple minutes, then there's probably something wrong, and maybe they want to cancel the calculation.

Should I create a separate issue for that?

TonyXiang8787:
> Good catch!
> 
> 
> 
> I've been thinking about something related to this but didn't realize that this was also broken.
> 
> 
> 
> I myself have been thinking about (for instance) process cancellation (e.g.: `KeyboardInterrupt` / `SIGINT` handling). E.g. if the user expects a calculation to take milliseconds but then it takes a couple minutes, then there's probably something wrong, and maybe they want to cancel the calculation.
> 
> 
> 
> Should I create a separate issue for that?

Feel free to create a new issue. @mgovers

PR #1246, opened 2025-12-31
Multi-dimensional batch calculation: documentation

nitbharambe:
Batch chaining was introduced in #1201.

This PR contains its documentation.


TonyXiang8787:
@nitbharambe, @mgovers proposed to not to use the word "chaning" in any places to avoid confusion. But use the wording: "a list of batch datasets which are intepreted as cartesian product"

docs/examples/Power Flow Example.ipynb
mgovers, line 1: does it make sense to do this in the same PF example? especially given our intention to split up the examples. Maybe this is the right moment to do this splitting (ofc. first in main, not here)

docs/user_manual/calculations.md
```diff
@@ -1065,6 +1065,26 @@ for component_idx, scenario in enumerate(line_update):
 independent_update_data = {'line': line_update}
 ```
 
+### Cartesian product of Batch Datasets
+
+Consider an example of running a contingency analysis with a timeseries data.
+Or maybe probablistic data along with timeseries data.
+In such simulations, it is required to perform a loadflow on a cartesian product of situations.
+This is possible to do via providing the `update_data` with a list of multiple batch datasets.
+The output of such calculation would be flattened with dimension &scenarios * components$.
```
mgovers: i don't think we use the `&...$` pattern anywhere in the documentation. is that intended?

docs/user_manual/data-validator.md
```diff
@@ -13,14 +13,6 @@ An alternative approach would be to validate only when an exception is raised, b
 will raise exceptions, most of them wil just yield invalid results without warning.
 The main validation functions and classes can be included from `power_grid_model.validation`.
 
-Two helper type definitions are used throughout the validation functions, `InputData` and `UpdateData`.
-They are not special types or classes, but merely type hinting aliases:
-
-```python
-InputData = dict[str, np.ndarray]
-UpdateData = dict[str, np.ndarray | dict[str, np.ndarray]]
-```
```
mgovers, line 22: oof... this is ancient 😬

docs/user_manual/calculations.md
mgovers: does not render correctly in readthedocs
image
https://power-grid-model--1246.org.readthedocs.build/en/1246/user_manual/calculations.html#cartesian-product-of-batch-datasets

nitbharambe:
```suggestion
The output of such calculation would be flattened with dimension $scenarios * components$.
```

docs/examples/Power Flow Example.ipynb
nitbharambe: From our discussion: we wish to demonstrate this feature to all type of users. Its content is small enough and builds on previous code cells. Hence it does belong here even after we split up the examples.

docs/user_manual/calculations.md
```diff
@@ -1065,6 +1065,26 @@ for component_idx, scenario in enumerate(line_update):
...
 
+### Cartesian product of Batch Datasets
+
+Consider an example of running a contingency analysis with a timeseries data.
+Or maybe probablistic data along with timeseries data.
+In such simulations, it is required to perform a loadflow on a cartesian product of situations.
+This is possible to do via providing the `update_data` with a list of multiple batch datasets.
+The output of such calculation would be flattened with dimension $scenarios * components$.
+
+#### Example: Chaining datasets
```
figueroa1395: https://github.com/PowerGridModel/power-grid-model/pull/1246#issuecomment-3705054849

Same below.

docs/user_manual/calculations.md
```diff
@@ -1065,6 +1065,26 @@ for component_idx, scenario in enumerate(line_update):
...
+
+```py
+# 5 scenarios of timeseries
+load_update = initialize_array('update', 'sym_load', (5, 1))
+# Fil load_update
```
figueroa1395: Typo:
```suggestion
# Fill load_update
```

docs/user_manual/calculations.md
figueroa1395: I think we should also demonstrate this with an example using columnar data format. Why? Because we can do cartesian batch calculations on attributes of the same component with different sizes, which I imagine should be useful and I feel it is also more intuitive. For instance a `sym_load` with `p_specified` of `d=3` and `q_specified` of `d=5` (see https://github.com/PowerGridModel/power-grid-model/blob/fb9486c7fb97921b1075a1dd62953836f9d4d590/tests/native_api_tests/test_api_model_multi_dimension.cpp).

docs/user_manual/data-validator.md
```diff
@@ -68,6 +62,10 @@ In such cases, the latter is leading when only running batch calculations.
 Running single calculations on an incomplete input data set is, of course, unsupported.
 ```
 
+Validating a cartesian product of datasets used in PGM's `update_data` via providing it with `list[BatchDataset]`is
+not straightforward.
+User should convert such multiple dataset into a single flat batch dataset.
+
```
figueroa1395, line 68: As I read this, my understanding is that we don't provide support for direct validation of batch datasets that come as a cartesian product or `list[batch_dataset}` and that the user should flatten them first to validate individually. Is that what was intended here? If so, let's make sure (I am reading the example last, so it may be there already) an example on how to do this is added.

docs/user_manual/performance-guide.md
```diff
@@ -53,6 +53,13 @@ If you are running on a system where memory is the bottle-neck, using a columnar
 footprint.
 This may or may not induce a slight computational overhead during calculations.
 
+Some simulations might require a cartesian product of scenarios of two batch datasets.
+This can be done by passing them to `update_data` as a list.
```
figueroa1395: Maybe add a link here to where this is explained in more detail?

docs/examples/Power Flow Example.ipynb
```diff
@@ -1146,6 +1146,64 @@
...
+   "cell_type": "markdown",
+   "id": "344efbe8",
+   "metadata": {},
+   "source": [
+    "#### Cartesian product of batch datasets\n",
+    "\n",
+    "It is possible to conduct a batch calculation of multiple datasets in form of a cartesian product of their scenarios.\n",
+    "Assume certain batch datasets with N1, N2, N3, ... scenarios. \n",
+    "This would give us $N1 * N2 * N3 * ...$ possible combinations via the cartesian product.\n",
+    "The resultant output data is in the flat form and it has dimension of N1 * N2 * N3 with first dataset being the highest \n",
```
figueroa1395:
```suggestion
    "The resulting output data is flattened and it has dimension of N1 * N2 * N3 with first dataset being the highest \n",
```


> with first dataset being the highest

Why is this relevant? Or, what do you mean with this?

docs/examples/Power Flow Example.ipynb
```diff
@@ -1146,6 +1146,64 @@
...
+    "dimension. \n",
+    "This can be beneficial in reducing complexity of implementation of such batch calculation \n",
+    "along with keeping the size of such resultant update_data to a minimum.\n",
+    "\n",
+    "```note\n",
+    "The data validation module mentioned in #validation-(optional) does not support this chaining. \n",
+    "It is advised to chain them manually via numpy to use the validator.\n",
```
figueroa1395: I think we should provide a small example for this.

docs/examples/Validation Examples.ipynb
```diff
@@ -267,11 +275,839 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1d82cfc5",
+   "metadata": {},
+   "source": [
+    "### Validating cartesian product of batch datasets\n",
+    "\n",
+    "[Cartesian product of  batch datasets](../user_manual/calculations.md#cartesian-product-of-batch-datasets) are not supported in the `validate_batch_data`. \n",
+    "The user should faltten the dataset manually to use the validator.\n",
```
figueroa1395: Typo:
```suggestion
    "The user should flatten the dataset manually to use the validator.\n",
```

Also, I see that you indeed provided an example on the flattening. You can resolve the comments relevant to this.

docs/user_manual/calculations.md
nitbharambe: Mentioned that either of formats are allowed.
Its probably not important to demonstrate this specially since there can be many applications.

docs/user_manual/data-validator.md
nitbharambe: Yes, an example is in the notebook.

docs/examples/Power Flow Example.ipynb
nitbharambe: Didnt mention flattened directly since we don't specially flatten it but its already in that form.

> Why is this relevant? Or, what do you mean with this?

Extra clarification to know which update gets applied first and last. 
Is it the first in list or last in list. ie. N1 * N2 * N3 vs N3 * N2 * N1.

TonyXiang8787 (approved):

PR #1251, opened 2026-01-06
use pgm build for cpp ci

TonyXiang8787:
To ensure the CI build CPP and Python package using the same dependencies

### DIFF
--- a/tests/native_api_tests/CMakeLists.txt
+++ b/tests/native_api_tests/CMakeLists.txt
@@ -16,7 +16,12 @@ add_executable(power_grid_model_api_tests ${PROJECT_SOURCES})
 
 target_link_libraries(
     power_grid_model_api_tests
-    PRIVATE power_grid_model_c power_grid_model_cpp doctest::doctest
+    PRIVATE
+        power_grid_model_c
+        power_grid_model_cpp
+        doctest::doctest
+        nlohmann_json
+        nlohmann_json::nlohmann_json
 )
 
 # TODO: remove this when the API becomes stable


.github/workflows/build-test-release.yml
```diff
@@ -85,13 +85,14 @@ jobs:
...
-      - name: Install C++ dependencies
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+      
+      - name: Install pgm-build-dependencies
         run: |
-          brew update
-          brew install boost eigen nlohmann-json msgpack-cxx doctest
+          uv tool install https://github.com/PowerGridModel/pgm-build-dependencies/releases/latest/download/pgm_build_dependencies-0.1.0-py3-none-any.whl
+          pgm-build-setup-ga-ci
```
mgovers, line 94: it may be worth making a separate action for this since it's common across the different jobs

tests/native_api_tests/CMakeLists.txt
```diff
@@ -16,7 +16,12 @@ add_executable(power_grid_model_api_tests ${PROJECT_SOURCES})
 
 target_link_libraries(
     power_grid_model_api_tests
-    PRIVATE power_grid_model_c power_grid_model_cpp doctest::doctest
+    PRIVATE
+        power_grid_model_c
+        power_grid_model_cpp
+        doctest::doctest
+        nlohmann_json
+        nlohmann_json::nlohmann_json
```
mgovers, line 24: 😮  was this broken? but then why didn't it cause issues?

TonyXiang8787: Yes this was broken. Previously all the header files are in the same include folder, so you did not know.

.github/workflows/build-test-release.yml
TonyXiang8787: I am hesitating. We also did not do that before. I feel a bit over engineered.

mgovers: ok :+1:

PR #1253, opened 2026-01-06
[BUG] add cassert in common include

TonyXiang8787:
The header include order of the core has a problem with `cassert`. When building benchmark target in release in macOS, it fails to load `assert` macro.

Now the inclusion is added in the common.


figueroa1395:
> I tried building the benchmark target locally on my Mac with clang-release and it was successful. This isn't a problem for me.

@TonyXiang8787  See above. Can you share a bit more about the issue? It seems fine for me.

mgovers (approved):

PR #1254, opened 2026-01-07
OpenSSF silver badge: input sanitization in C API

mgovers:
* Input sanitization:
  * Check user-provided pointers for `nullptr`
  * Clearer distinction between "safe" and "unsafe" pointers
* Refactor the error handling in the C API
  * Explicit illegal operation and other C API exceptions
  * All error handling goes via the same `call_with_catch` function
  * Custom exception handlers can be provided
  * Fix small bug where the `PGM_calculate` function was not able to handle `nullptr` handles
  * NOSONAR the instances of catching generic exceptions and new/delete where they really were intentional

Relates to #1190

### DIFF
--- a/power_grid_model_c/power_grid_model_c/src/model.cpp
+++ b/power_grid_model_c/power_grid_model_c/src/model.cpp
@@ -201,11 +201,7 @@ constexpr BatchExceptionHandler batch_exception_handler{};
 template <typename T, std::ranges::input_range R>
     requires std::convertible_to<std::ranges::range_value_t<R>, T>
 void append_range(std::vector<T>& vec, R&& range) {
-#if __cpp_lib_containers_ranges >= 202202L
-    vec.append_range(std::forward<R>(range));
-#else
-    vec.insert(vec.end(), std::ranges::begin(range), std::ranges::end(range));
-#endif
+    std::ranges::move(std::forward<R>(range), std::back_inserter(vec));
 }
 
 class MDBatchExceptionHandler : public power_grid_model_c::DefaultExceptionHandler {


power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+template <typename T> constexpr T* safe_ptr_maybe_nullptr(T* ptr) { return ptr; }
+
+template <typename T> constexpr T* safe_ptr(T* ptr) {
+    if (ptr) {
+        return safe_ptr_maybe_nullptr(ptr);
+    }
+    throw IllegalOperationError{"Received null pointer when not allowed"};
+}
+
+template <typename T> constexpr T& safe_ptr_get(T* ptr) { return *safe_ptr(ptr); }
```
mgovers, line 91: should i rename this to `safe_ref`, `safe_deref` or `safe_ptr_deref`?

power_grid_model_c/power_grid_model_c/src/handle.cpp
```diff
@@ -8,26 +8,47 @@
...
+}
 
 // error handling
-PGM_Idx PGM_error_code(PGM_Handle const* handle) { return handle->err_code; }
-char const* PGM_error_message(PGM_Handle const* handle) { return handle->err_msg.c_str(); }
-PGM_Idx PGM_n_failed_scenarios(PGM_Handle const* handle) { return static_cast<Idx>(handle->failed_scenarios.size()); }
-PGM_Idx const* PGM_failed_scenarios(PGM_Handle const* handle) { return handle->failed_scenarios.data(); }
+PGM_Idx PGM_error_code(PGM_Handle const* handle) {
+    return handle != nullptr ? compile_time_safe_cast<PGM_Idx>(handle->err_code) : PGM_Idx{0};
+}
```
mgovers, line 34: here and below: the handle is inherently unsafe, so we shouldn't do something like `safe_ptr_maybe_nullptr` as it may give off the wrong signal

power_grid_model_c/power_grid_model_c/src/handle.hpp
```diff
@@ -27,21 +28,65 @@ struct PGM_Handle {
...
-template <class Exception = std::exception, class Functor>
-inline auto call_with_catch(PGM_Handle* handle, Functor func, PGM_Idx error_code,
-                            std::string_view extra_msg = {}) -> std::invoke_result_t<Functor> {
-    if (handle) {
-        PGM_clear_error(handle);
+namespace power_grid_model_c {
+constexpr void clear_error(PGM_Handle* handle) {
+    if (handle != nullptr) {
+        *handle = PGM_Handle{};
     }
```
mgovers, line 35: we should not go via the C API here so i replaced it

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
```diff
@@ -38,68 +57,94 @@ power_grid_model::meta_data::MetaData const& get_meta_data() {
...
-        }
-        return &get_meta_data().datasets[idx];
-    });
+    return call_with_catch(
+        handle,
+        [idx] {
+            if (idx < 0 || idx >= get_meta_data().n_datasets()) {
+                throw std::out_of_range{"Index out of range!\n"};
+            }
+            return &get_meta_data().datasets[idx];
```
mgovers: should i refactor this to simply this?
```suggestion
            return &get_meta_data().datasets.at(idx);
```

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
```diff
@@ -38,68 +57,94 @@ power_grid_model::meta_data::MetaData const& get_meta_data() {
...
-        return &dataset->components[idx];
-    });
+    return call_with_catch(
+        handle,
+        [idx, dataset] {
+            auto const& safe_dataset = safe_ptr_get(dataset);
+            if (idx < 0 || idx >= safe_dataset.n_components()) {
+                throw std::out_of_range{"Index out of range!\n"};
+            }
+            return &safe_dataset.components[idx];
```
mgovers, line 90: should i refactor this to simply this?
```suggestion
            return &safe_ptr_get(dataset).components.at(idx);
```

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
```diff
@@ -38,68 +57,94 @@ power_grid_model::meta_data::MetaData const& get_meta_data() {
...
-        return &component->attributes[idx];
-    });
+    return call_with_catch(
+        handle,
+        [idx, component] {
+            auto const& safe_component = safe_ptr_get(component);
+            if (idx < 0 || idx >= safe_component.n_attributes()) {
+                throw std::out_of_range{"Index out of range!\n"};
+            }
+            return &safe_component.attributes[idx];
```
mgovers: should i refactor this to be simply this?
```suggestion
            return &safe_ptr_get(component).attributes.at(idx);
```

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -140,45 +146,67 @@ constexpr auto extract_calculation_options(PGM_Options const& opt) {
                               .threading = opt.threading,
                               .short_circuit_voltage_scaling = get_short_circuit_voltage_scaling(opt)};
 }
-} // namespace
 
-// run calculation
-void PGM_calculate(PGM_Handle* handle, PGM_PowerGridModel* model, PGM_Options const* opt,
-                   PGM_MutableDataset const* output_dataset, PGM_ConstDataset const* batch_dataset) {
-    PGM_clear_error(handle);
+class BadCalculationRequest : public PowerGridError {
```
mgovers, line 155: it's not an illegal operation, so i decided to go with `PowerGridError` here.

Interesting note: the python wrapper cannot detect one (it will output `"Unknown Error!"`) this but it should not in the first place, because this would imply a bug in the python code, so it's fine as-is.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
TonyXiang8787: The name is okay

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+#pragma once
+
+#include <exception>
+#include <format>
+#include <string>
+#include <string_view>
+#include <utility>
+
+namespace power_grid_model_c {
```
figueroa1395, line 13: So the point of putting this piece of logic in this newly introduced `namespace`, is that it covers input sanitation of the C API but from C++, Is that correct? Hence the `*_c` at the end even though most things (all before this introduction) under this folder went under `namespace power_grid_model`?

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+class IllegalOperationError : public std::exception {
```
figueroa1395, line 14: The point of defining this new error type here is that it will only pop up when interacting with the C API and never directly from the CPP core, right? And this is because of the scope of this PR is sanitation on the C API side only, correct?

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+  public:
+    explicit IllegalOperationError(std::string_view message) : msg_{std::format("Illegal operation: {}\n", message)} {}
+
+    char const* what() const noexcept final { return msg_.c_str(); }
+
+  private:
+    std::string msg_;
+};
+
+template <typename T, typename U> constexpr T compile_time_safe_cast(U&& value) noexcept {
```
figueroa1395: Nitpick:
```suggestion
template <typename U, typename T> constexpr T compile_time_safe_cast(U&& value) noexcept {
```
because it follows the order in `std::convertible_to<U, T`, which makes it more intuitive and less error prone when developing. This is a matter of taste though.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+    static_assert(std::convertible_to<U, T>,
+                  "Type U is not convertible to type T; compile-time safe cast not possible");
+    static_assert(std::same_as<T, std::common_type_t<T, U>>,
+                  "Loss of conversion possible; common type differs from target type");
```
figueroa1395: Can you write the message more clearly? It seems a bit cryptic to me. Maybe something along the lines of

```suggestion
                  "Only narrowing conversion possible; common type differs from target type");
```

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    if constexpr (std::same_as<T, U>) {
+        return std::forward<U>(value);
+    } else {
+        return static_cast<T>(std::forward<U>(value));
+    }
+}
+
+template <std::integral T, std::integral U>
```
figueroa1395, line 37: Nitpick, similar to above:
```suggestion
template <std::integral U, std::integral T>
```

Also, why restrain this to `std::integral` types?

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+    requires std::convertible_to<U, T>
+constexpr auto safe_cast(U value) {
+    if constexpr (std::same_as<T, U>) {
+        return value;
+    } else {
+        if (std::in_range<T>(value)) {
```
figueroa1395, line 43: Isn't this redundant after `requires std::convertible_to<U, T>`?

Also, wouldn't this constrain this function? I quote: _This function cannot be used with [enums](https://en.cppreference.com/w/cpp/language/enum.html) (including [std::byte](https://en.cppreference.com/w/cpp/types/byte.html)), char, char8_t, char16_t, char32_t, wchar_t and bool._ (see [cpp reference](https://en.cppreference.com/w/cpp/utility/in_range.html))

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+    if constexpr (std::same_as<T, U>) {
+        return std::forward<U>(value);
+    } else {
+        return static_cast<T>(std::forward<U>(value));
+    }
+}
+
+template <std::integral T, std::integral U>
+    requires std::convertible_to<U, T>
+constexpr auto safe_cast(U value) {
```
figueroa1395, line 39: Wouldn't this function also be compile time? Isn't it achieving the same as the above, but above only does some checks at debug mode? Shouldn't we change the names to `safe_cast_debug` and `safe_cast`?

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+    throw IllegalOperationError{"Received negative value for size"};
+}
+template <std::integral T> constexpr PGM_Idx to_c_size(T value) {
+    assert(value >= 0);
+    return safe_cast<PGM_Idx>(value);
+}
+
+constexpr bool safe_bool(PGM_Idx value) { return value != 0; }
+template <std::integral T> constexpr T to_c_bool(bool value) { return value ? 1 : 0; }
```
figueroa1395: Can't this one be an overload of `safe_bool`? Why the explicit distinction?

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+template <typename T>
+    requires std::is_enum_v<T>
+constexpr T safe_enum(PGM_Idx value) {
+    return static_cast<T>(safe_cast<std::underlying_type_t<T>>(value));
+}
+template <typename T>
+    requires std::is_enum_v<T>
+constexpr PGM_Idx to_c_enum(T value) {
+    return safe_cast<PGM_Idx>(std::to_underlying(value));
+}
```
figueroa1395, line 79: Don't these clash with what I mentioned in https://github.com/PowerGridModel/power-grid-model/pull/1254/changes#r2668984618

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+// no-op sanitization for clarity that we did think about input sanitization but we deliberately allow nullptr
+template <typename T> constexpr T* safe_ptr_maybe_nullptr(T* ptr) { return ptr; }
+
+template <typename T> constexpr T* safe_ptr(T* ptr) {
+    if (ptr) {
+        return safe_ptr_maybe_nullptr(ptr);
+    }
```
figueroa1395, line 87: Since this would never be `nullptr` shouldn't we just `return ptr;` here?

src/power_grid_model/_core/error_handling.py
```diff
@@ -44,10 +46,20 @@
...
+    NO_ERROR = 0
+    REGULAR_ERROR = 1
+    BATCH_ERROR = 2
+    SERIALIZATION_ERROR = 3
+
+
+PGM_NO_ERROR = _PgmCErrorCode.NO_ERROR
+PGM_REGULAR_ERROR = _PgmCErrorCode.REGULAR_ERROR
+PGM_BATCH_ERROR = _PgmCErrorCode.BATCH_ERROR
+PGM_SERIALIZATION_ERROR = _PgmCErrorCode.SERIALIZATION_ERROR
```
figueroa1395: Why is this needed then? Why not just go via above `Enum` directly?

power_grid_model_c/power_grid_model_c/src/handle.hpp
figueroa1395: Is the reason that C API calls are slow as the cross system layers? And since we have `PGM_Handle{}` available here directly, we can just use it?

power_grid_model_c/power_grid_model_c/src/handle.hpp
```diff
@@ -27,21 +28,65 @@ struct PGM_Handle {
...
+    void operator()(PGM_Handle& handle) const noexcept { handle_all_errors(handle, PGM_regular_error); }
+
+    static void handle_all_errors(PGM_Handle& handle, PGM_Idx error_code,
+                                  std::string_view extra_message = "") noexcept {
+        std::exception_ptr const ex_ptr = std::current_exception();
+        try {
+            std::rethrow_exception(ex_ptr);
+        } catch (std::exception const& ex) { // NOSONAR(S1181)
+            handle_regular_error(handle, ex, error_code, extra_message);
+        } catch (...) {
```
figueroa1395: I suppose you should also add
```suggestion
        } catch (...) {  // NOSONAR(S2738)
```

power_grid_model_c/power_grid_model_c/src/handle.hpp
```diff
@@ -27,21 +28,65 @@ struct PGM_Handle {
...
-        PGM_clear_error(handle);
+namespace power_grid_model_c {
+constexpr void clear_error(PGM_Handle* handle) {
+    if (handle != nullptr) {
+        *handle = PGM_Handle{};
     }
+}
+
+struct DefaultExceptionHandler {
+    void operator()(PGM_Handle& handle) const noexcept { handle_all_errors(handle, PGM_regular_error); }
```
figueroa1395, line 39: Why do you need a functor? I don't see the `struct` holds any data. Why not free functions and group them in a `namespace` if needed?

power_grid_model_c/power_grid_model_c/src/handle.hpp
```diff
@@ -27,21 +28,65 @@ struct PGM_Handle {
...
+        if (!extra_message.empty()) {
+            handle.err_msg += extra_message;
+        }
+    }
+
+    static void handle_unkown_error(PGM_Handle& handle) noexcept {
+        handle.err_code = PGM_regular_error;
+        handle.err_msg = "Unknown error!\n";
+    }
+};
```
figueroa1395, line 66: I haven't finished reviewing, but if I recall correctly the logic above was somewhat implemented elsewhere. Was that removed somewhere else? Or maybe I'm wrong? Just pointing it out to avoid duplication if possible.

power_grid_model_c/power_grid_model_c/src/handle.hpp
```diff
@@ -27,21 +28,65 @@ struct PGM_Handle {
...
     using ReturnValueType = std::remove_cvref_t<std::invoke_result_t<Functor>>;
-    static std::conditional_t<std::is_void_v<ReturnValueType>, int, ReturnValueType> const empty{};
+
+    static constexpr std::conditional_t<std::is_void_v<ReturnValueType>, int, ReturnValueType> empty{};
+
     try {
+        if (handle != nullptr) {
+            clear_error(handle);
+        }
         return func();
```
figueroa1395, line 84: Why `clear_error(handle)` at this point?
Shouldn't it be
```
if (handle != nullptr) {
            return func();
        }
```

figueroa1395 (commented):
I'll finish the rest tomorrow morning, but here is some input already.

power_grid_model_c/power_grid_model_c/src/handle.hpp
mgovers: yes exacly

mgovers: if you're lucky, the C API call would be optimized away, but there's no guarantee that it will be. Doing it this way makes it inline (`constexpr` implies `inline`) so the optimizer has hints as to the intention

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: this directory is `power_grid_model_c/power_grid_model_c`; it's the C API implementation, so it should have its own namespace, separate from the interface library `power_grid_model_c/power_grid_model` (which is `namespace power_grid_model`).

I do see that `power_grid_model_c\power_grid_model_c\src\math_solver.hpp` and `power_grid_model_c\power_grid_model_c\src\math_solver.cpp` do contain `namespace power_grid_model`, but those files are minimal. I can change them to `namespace power_grid_model_c` as well. your choice.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: exactly

power_grid_model_c/power_grid_model_c/src/handle.cpp
```diff
@@ -8,26 +8,47 @@
...
+    delete handle; // NOSONAR(S5025)
+}
 
 // error handling
-PGM_Idx PGM_error_code(PGM_Handle const* handle) { return handle->err_code; }
-char const* PGM_error_message(PGM_Handle const* handle) { return handle->err_msg.c_str(); }
-PGM_Idx PGM_n_failed_scenarios(PGM_Handle const* handle) { return static_cast<Idx>(handle->failed_scenarios.size()); }
-PGM_Idx const* PGM_failed_scenarios(PGM_Handle const* handle) { return handle->failed_scenarios.data(); }
+PGM_Idx PGM_error_code(PGM_Handle const* handle) {
+    return handle != nullptr ? compile_time_safe_cast<PGM_Idx>(handle->err_code) : PGM_Idx{0};
```
TonyXiang8787, line 33: Here and all other places:

I find this `handle != nullptr ? ...` is very unreadable. I know many guidelines prefer this way. But I do think the old way is better `handle ? ...`. Just treat pointer as `bool` for judging `nullptr` is a well-understood stuff and very concise and readable. I never understand why some guidelines are agaist it.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+};
+
+template <typename T, typename U> constexpr T compile_time_safe_cast(U&& value) noexcept {
+    static_assert(std::convertible_to<U, T>,
+                  "Type U is not convertible to type T; compile-time safe cast not possible");
+    static_assert(std::same_as<T, std::common_type_t<T, U>>,
+                  "Loss of conversion possible; common type differs from target type");
+
+    if constexpr (std::same_as<T, U>) {
+        return std::forward<U>(value);
```
TonyXiang8787: Why are we using `std::forward` here? I feel we just use it for satisfy linters, instead of having meaningful intention.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
TonyXiang8787: Should it be just `U const& value`? I do not see the need to have a universal reference here. There is no point to bind a rvalue reference.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
```diff
@@ -0,0 +1,94 @@
...
+    assert(value >= 0);
+    return safe_cast<PGM_Idx>(value);
+}
+
+constexpr bool safe_bool(PGM_Idx value) { return value != 0; }
+template <std::integral T> constexpr T to_c_bool(bool value) { return value ? 1 : 0; }
+
+// safe enum conversion from C integer type to C++ enum type.
+//
+// It does NOT check that the value is a valid enumerator of the enum type T.
```
TonyXiang8787: The comment is confusing. Any integer within the range of underlying type of the enum **is** a valid value of that enum. Enum does not require a value to be one of the pre-defined names in the list to be valid. A value which is not defined in the name list is still a value enum.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: no, it's actually the opposite, it follows the `std::constructible_from<T, U>` pattern

* `std::convertible_to<U, T>` is a concept in which the first argument is the "relevant part", namely "type U is allowed here if and only if it is `std::convertible_to<T>`.
  * Therefore, when specifying a concept, you use the order `template <typename Relevant, typename... Args>`. Since `T` is an argument, it comes second.
  * So: we specify second-argument-first `T` and the compiler does not deduce `U` (like in a function), but uses it as an implicit argument that is prefixed to our user-specified arguments.
* However, when **using** the concept (e.g., here), the relevant object is `T`, followed by `... Args`:
  * "This function is requested for `typename T`, and it is instantiated for (the) other argument(s) `typename U`, which must be `std::convertible_to<T>`"
  * Therefore, it follows the pattern that tells you `template <typename T, typename... Args>`, or even (using concepts to restrict the other arguments) `template <typename T, std::convertible_to<T> U>`

Therefore, we actually are following the order of `std::constructible_from`. Now as for the reason why we use `std::convertible_to` instead of `std::constructible_from` in the body of this function, that is because `std::convertible_to` is less restrictive (because it uses `static_cast`) than `std::constructible_from` (which uses only the constructor)

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
figueroa1395: The question was inclined towards the fact that in this folder, which handles the C API as you mentioned, there are indeed some files that are in the `namespace power_grid_model`, so I do think we should homogenize them then, as you suggested. Let's keep homogeneity.

power_grid_model_c/power_grid_model_c/src/buffer.cpp
```diff
@@ -16,58 +19,92 @@ using namespace power_grid_model;
...
 #else
     std::free(ptr); // NOLINT(hicpp-no-malloc)
 #endif
 }
-void PGM_buffer_set_nan(PGM_Handle* /* handle */, PGM_MetaComponent const* component, void* ptr, PGM_Idx buffer_offset,
+} // namespace
+
+// buffer control
+RawDataPtr PGM_create_buffer(PGM_Handle* handle, PGM_MetaComponent const* component, PGM_Idx size) {
+    return call_with_catch(handle, [component, size] { return create_buffer_impl(safe_ptr_get(component), size); });
```
TonyXiang8787: Why do you put a separate implementation function? I do not see the need of extra abstraction.

mgovers: went with this because i thought the macros made it less readable. i can revert if you prefer

TonyXiang8787: Why do we need macro here?

mgovers: the `#ifdef _WIN32 ... #else ... #endif` things, i mean

figueroa1395: I think keeping this one as well as `PGM_destroy_buffer` like the originals is fine. The macros logic is small enough.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: changed to `"Loss of precision during conversion possible; common type differs from target type"`

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: because `std::in_range` requires integral types. floating point casting would require a different approach

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: indeed, for enums, we should use `safe_enum`/`to_c_enum` (which go via the `underlying_type`). `numeric_limits` has overloads for `std::uint8_t`/`std::int8_t`, so it works on `IntS` even though their underlying types are `(unsigned) char`

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: `safe_bool` extracts from the C API version into the C++ `bool` but `to_c_bool` works in the other direction. I did not want to use `from_c_bool` because theoretically, one can also use `safe_bool` to go from C++ to C++ without going via the `PGM_Idx`. That said, I will explicitly template `safe-bool` as well.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: no because i go via `underlying_type`

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: there may be other sanitizing safety checks. just to be explicit, I went with `safe_ptr_maybe_nullptr`. even if there are no additional sanity safety checks, that shows the right intention.

An example of such an extended sanity check is whether the pointer is in a valid address within the available memory (platform-dependent, but may be possible via e.g. `sysconf` or `Windows.h`)

src/power_grid_model/_core/error_handling.py
mgovers: backwards compatibility, but since it's in `_core` anyways, will remove

power_grid_model_c/power_grid_model_c/src/handle.hpp
mgovers: sonar does not complain about this, even on `main` but will add anyways.

power_grid_model_c/power_grid_model_c/src/handle.hpp
mgovers: 2 reasons:

1. being explicit because handles are very specific, and I do not want to accidentally use free functions anywhere where I do not want them.
2. just like when specifying custom destructors, you want the callable to be storable without weird magic like function pointers. using a type like this makes it much easier to manage

willing to change, though, if needed

power_grid_model_c/power_grid_model_c/src/handle.hpp
mgovers: the other location now goes via this `handle_unknown_error`. This is the only location where this is left

power_grid_model_c/power_grid_model_c/src/handle.hpp
mgovers: no, the function should be called regardless of whether a handle was specified or not. the handle is an explicit handle (yes, that is why it's called that way) to ensure that the correct API version and ABI are used, as well as to handle (yes, there it is again) things like errors. If you do not specify the handle, the functionality will still work, but you do not get error handling or ABI checks.

power_grid_model_c/power_grid_model_c/src/handle.cpp
mgovers: i actually agree with the guidelines as it is more explicit. since we disagree, are you OK with following the guidelines?

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: this allows the compiler to choose whether to pass by reference or by value, depending on whether the type is small or large, so that the user has the choice to obtain a reference or by value. however, i have been thinking about it here as well, so I will not use a reference but instead restrict it to value-types

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: it also enables capture-by-value for small types, but i'll try to find a different solution

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: how do you want me to rephrase it? i do still want to keep the intentionality: it does not check whether the actual names were declared

TonyXiang8787: Just like that. It does not (and cannot) check if the value is one of the pre-defined names.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
TonyXiang8787: Even with that in mind, forward will not make sense. You are returning by value.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: maybe
```cpp
// safe enum conversion from C integer type to C++ enum type.
//
// It does not check against the named values listed in the enum definition.
// It only ensures that there is no unexpected behavior like wrap-arounds during the conversion.
// E.g.: for an enum class with underlying type uint8_t, passing 256 would wrap around to 0 without this check.
```

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
figueroa1395: Do we get any penalty or potentially allow any unintended behavior  if we keep universal reference? Specially if capture-by-value is "valuable" for small types?

mgovers: stuff like this can be optimized for types with `sizeof(T) <= sizeof(T&)`. A reference results in more registers being used and at least one additional operation

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
TonyXiang8787: that's fine

power_grid_model_c/power_grid_model_c/src/buffer.cpp
mgovers: done

power_grid_model_c/power_grid_model_c/src/handle.hpp
figueroa1395: Number 2 makes it worth for me. Resolving.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: changed to `const&`

power_grid_model_c/power_grid_model_c/src/handle.hpp
figueroa1395: Weird... Maybe don't add it and let's let Sonar complain about it if it does. It should be clear for us that this is needed from the code itself.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
mgovers: no because there is a range check that needs to happen at runtime if the function is called at runtime. E.g., if the user specifies `int64_t{5}` then casting to `uint8_t` will work, but if they specify `int64_t{-1}`, then it will throw. Since it depends on runtime user input, that usage is never compile-time.

power_grid_model_c/power_grid_model_c/src/handle.hpp
mgovers: nah, it's fine, we can keep it in

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
figueroa1395: So `std::in_range` is to constraint this implementation outside of `enum`s basically? or is there another reason?

mgovers: idd, because sometimes we use `PGM_Idx` for something that is a size-type by default or vice versa. This adds a runtime guarantee that the bounds are satisfied

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
figueroa1395: I had miss-read cpp reference and thought it was a compile time check ([see](https://en.cppreference.com/w/cpp/utility/in_range.html) _It is a compile-time error if either T or U is a non-[integer](https://en.cppreference.com/w/cpp/language/types.html#Integer_types) type, a character type, or bool._). This makes it clear. Thanks, resolving.

power_grid_model_c/power_grid_model_c/src/input_sanitization.hpp
figueroa1395: Yup, I just read the other comment. Thanks for clarifying.

power_grid_model_c/power_grid_model_c/src/handle.cpp
TonyXiang8787: Let's keep with the guidelines then.

power_grid_model_c/power_grid_model_c/src/handle.hpp
figueroa1395: User's responsibility philosophy when working with the C API directly. Got it!

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
```diff
@@ -18,16 +19,34 @@
...
+        using namespace std::string_literals;
 
-auto const meta_catch = [](PGM_Handle* handle, auto func) -> decltype(auto) {
-    return call_with_catch<std::out_of_range>(handle, func, PGM_regular_error, bound_error_msg);
+        std::exception_ptr const ex_ptr = std::current_exception();
+        try {
+            std::rethrow_exception(ex_ptr);
+        } catch (std::out_of_range const& ex) {
+            handle_regular_error(handle, ex, PGM_regular_error, "\n You supplied wrong name and/or index!\n");
+        } catch (std::exception const& ex) {
```
figueroa1395, line 41: Weird that Sonar doesn't complain about this one either 😆. It's funny how it seems to complain about the same error in multiple instances in an arbitrary manner...

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
figueroa1395: Yes! It also aligns with what you have done elsewhere in this PR.

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
figueroa1395: Yes!

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
figueroa1395: Yes!

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -27,32 +36,32 @@ struct PGM_PowerGridModel : public MainModel {
...
                                      PGM_ConstDataset const* input_dataset) {
-    return call_with_catch(
-        handle,
-        [system_frequency, input_dataset] {
-            return new PGM_PowerGridModel{system_frequency, *input_dataset, get_math_solver_dispatcher(), 0};
-        },
-        PGM_regular_error);
+    return call_with_catch(handle, [system_frequency, input_dataset] {
+        return new PGM_PowerGridModel{// NOSONAR(S5025)
+                                      system_frequency, safe_ptr_get(input_dataset), get_math_solver_dispatcher(), 0};
```
figueroa1395, line 44: Shouldn't we also use a similar pattern like `safe_ptr_get` but with `get_math_solver_dispatcher`?

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
```diff
@@ -38,68 +57,94 @@ power_grid_model::meta_data::MetaData const& get_meta_data() {
...
+                throw std::out_of_range{"Index out of range!\n"};
+            }
+            return &get_meta_data().datasets[idx];
+        },
+        ranged_exception_handler);
 }
 PGM_MetaDataset const* PGM_meta_get_dataset_by_name(PGM_Handle* handle, char const* dataset) {
-    return meta_catch(handle, [dataset]() { return &get_meta_data().get_dataset(dataset); });
+    return call_with_catch(
+        handle, [dataset] { return &get_meta_data().get_dataset(safe_str_view(dataset)); }, ranged_exception_handler);
```
figueroa1395, line 72: Why would `std::out_of_range` be thrown here?

power_grid_model_c/power_grid_model_c/src/serialization.cpp
```diff
@@ -5,85 +5,111 @@
...
+using power_grid_model_c::call_with_catch;
+using power_grid_model_c::safe_bool;
+using power_grid_model_c::safe_enum;
+using power_grid_model_c::safe_ptr;
+using power_grid_model_c::safe_ptr_get;
+using power_grid_model_c::safe_size;
+using power_grid_model_c::to_c_size;
+
+struct SerializationExceptionHandler : public power_grid_model_c::DefaultExceptionHandler {
+    void operator()(PGM_Handle& handle) const noexcept { handle_all_errors(handle, PGM_serialization_error); }
```
figueroa1395, line 32: This was introduced only to homogenize behavior, right?

mgovers: yes

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
mgovers: if a non-existent dataset name is thrown, then it will be out-of-range (in the old code at least, i need to double-check for new code)

power_grid_model_c/power_grid_model_c/src/model.cpp
mgovers: why? it is managed by us and statically linked. it should never be compromised

power_grid_model_c/power_grid_model_c/src/meta_data.cpp
mgovers: oh it's a C++26 feature 😢

mgovers:
there are still more cleanups that are possible, but i have a feeling that the PR is in a decent state as of now. it was fairly hard to merge with main do to conflicts with #1176 but i think the current state is the best of both worlds. can you please do another round of review @figueroa1395 ?

PR #1256, opened 2026-01-07
pre-commit hook for formatting JSON files

kornerc:
* pre-commit hook for auto-formatting JSON files
* only JSON files in the folders `tests` and `code_generation` are formatted
* `pre-commit run --all-files pretty-format-json` has been executed to re-format all JSON files.
  This has been done in a _separate commit_ so that these changes can be easily _reverted_ if formatting options need to be adjusted
* the motivation behind this PR is to harmonize the formatting of the JSON files since some are not consistently formatted

I am not sure if this pre-commit hook is wanted.
Feel free to close this PR if this is behavior is not desired. 

* A lot of files have been reformatted in this commit.
  I hope that this does not cause any unforseen (numerical) issues.
  E.g.
  * `10.5e3` -> `10500.0`
  * `1e-5` -> `1e-05`


tests/data/power_flow/asym-line/rmatrix-xmatrix-c1-c0/input.json
```diff
@@ -33,19 +33,19 @@
...
-        "id": 4, 
-        "node": 1, 
-        "status": 1, 
+        "id": 4,
+        "node": 1,
+        "status": 1,
         "u_ref": 1.0,
         "rx_ratio": 6.0,
-        "sk": 1e15
+        "sk": 1000000000000000.0
```
kornerc: This change in the format reduces the readability in my opinion a bit.
But I am not sure if this is a problem.

.pre-commit-config.yaml
```diff
@@ -48,3 +48,18 @@ repos:
...
+      - --autofix   # automatically format json files
+      - --indent=2  # 2 spaces for indentation
+      - --no-sort-keys  # when autofixing, retain the original key ordering (instead of sorting the keys)
+      - --no-ensure-ascii  # preserve unicode characters instead of converting to escape sequences
+      files: >
+        (?x)^(
+          # fix JSON formatting in the following top-level folders:
+          code_generation|
+          tests
+        )/.*\.json$
```
kornerc: This is the only change in this PR which is not coming from the re-formatting.

If you prefer I can also revert the commit which reformats all the JSON files and you can run `pre-commit run --all-files pretty-format-json` and commit the changes yourself.

mgovers:
Hi @kornerc,

Thank you for the contribution. We actually have been thinking about doing this in the past. For some configuration files (e.g. `code_generator/*`, `CMakePresets.json` and `tests/**/params.json`), this is a good improvement.

However, for the data files (`input.json`, `update.json`, `sym_output.json`, `asym_output.json` and `sc_output.json`), this is not necessarily an improvement. The sheer amount of added data as a direct consequence of nesting is quite large. E.g.:

* the amount of added lines (about 170k; obtained from the summary `+215,469 −44,959`); every line contains a number of whitespaces due to nesting. 
* your finding that `1e15` is expanded to `1000000000000000.0`

This exact reason is why we deliberately use a compact notation with a maximum indentation level (see [this formatter](https://github.com/PowerGridModel/power-grid-model/blob/feature/input-sanitization/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp#L90-L105) and [its configuration](https://github.com/PowerGridModel/power-grid-model/blob/feature/input-sanitization/power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/serialization/serializer.hpp#L389-L390)).

That said, having consistent formatting helps with both code quality and code reviews. Here are a couple potential paths going forward:

* [ ] make it so that the formatter also uses a max indentation; or
* [ ] if you can use the PGM (de)serializer to reformat (e.g. by deserializing and then serializing again); or
* [ ] if neither is possible, then maybe we can restrict the formatting to only include files that are not PGM datasets.

One last remark: to keep the git commit history clean, can you please either squash the commits after you're done reverting (at the latest when sending it to the merge queue) or open a new PR?

.pre-commit-config.yaml
```diff
@@ -48,3 +48,18 @@ repos:
     hooks:
     - id: markdownlint
       args: ["--fix"]
+  - repo: https://github.com/pre-commit/pre-commit-hooks
```
mgovers: can you please also add this to the code quality github actions check?

TonyXiang8787:
Hi @kornerc, thanks for proposing this new improvement.

In addition of what @mgovers said, I would say ignore all PGM data json files. These files already have a custom indent pattern which is produced by the serializer. Also, user/contributor might want to a specific custom indent for a certain data file for some reason (usually modelling or readability). Enforcing a rule on PGM data json files does not make sense.

So I would go for option 3 of @mgovers proposed:

* we can restrict the formatting to only include files that are not PGM datasets.

kornerc:
Hi @mgovers and @TonyXiang8787 thank you for your review!

Based on your feedback I made some changes:
* I reverted the last commits and force pushed and followed as suggested option 3
  > we can restrict the formatting to only include files that are not PGM datasets.
* Initially I used the pre-commit hook [pretty-format-json](https://github.com/pre-commit/pre-commit-hooks/blob/f1dff44d3a9ae852957f34def96390f28719c232/pre_commit_hooks/pretty_format_json.py).
  Even though it is quite simple, its configuration is rather limited and I was not too happy that it made major changes on numbers e.g. `1e15` -> `1000000000000000.0`
* Thus, I searched for an alternative and found [Biome](https://biomejs.dev/) which convinced me:
  * Empty lines are kept
  * Numbers are only slightly changed e.g. `1e-08` -> `1e-8`
  * A [pre-commit hook](https://github.com/biomejs/pre-commit) is available
  * An [extension](https://github.com/biomejs/biome-vscode) for VS Code exists
  * Standalone [executeables](https://biomejs.dev/guides/manual-installation/#using-a-published-binary) are available
  * It's possible to overwrite the configuration for certain files e.g. allow comments in .json files located in `.vscode`
* I again split the changes into 2 commit:
  * 7652dce65360c5b6aaae4a3db2824a204ea29e44 with the relevant changes
  * c0e5e33ec6974b9c0c19e27b5a46f7ed36ea8977 where the formatter is applied on the repo
* Instead of whitelistening the files to format, I blacklisted the PGM dataset files
* I added a step in the `check-code-quality` action to run the formatter
* I didn't manage to set up the VS Code extension in the devcontainer properly and instead adjusted the default formatter in `.vscode/settings.json` to function similar to Biome. I can configure the devcontainer for Biome properly in another PR.
* I've noticed in the last run of the actions that Biome formats irrelevant files e.g. files in the [.mypy_cache](https://github.com/PowerGridModel/power-grid-model/actions/runs/20969763911/job/60269814919?pr=1256#step:10:16).
  I can add these folders to the blacklisted files after the rest of the PR is approved so that I can still easily revert the commit which formats all the files.

I am happy to hear your thoughts!

.vscode/extensions.json
```diff
@@ -1,29 +1,29 @@
...
-		"fredericbonnet.cmake-test-adapter",
-		"sonarsource.sonarlint-vscode",
-		"davidanson.vscode-markdownlint"
-	]
-}
\ No newline at end of file
+  // See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+  // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+  // List of extensions which should be recommended for users of this workspace.
+  "recommendations": [
```
mgovers, line 5: please add biome here as well

.vscode/settings.json
```diff
@@ -1,43 +1,46 @@
...
+    "editor.rulers": [120]
+  },
+
+  // can be removed when Biome is used in VS Code as formatter
+  "json.format.keepLines": true,
+  "[json][jsonc]": {
+    "editor.tabSize": 2,
+    "editor.insertSpaces": true,
+    "editor.detectIndentation": false
+  }
```
mgovers: if you add biome to vscode workspace recommendations, then you can configure that as the default formatter

code_generation/data/attribute_classes/input.json
```diff
@@ -1,583 +1,538 @@
...
+        }
+      ]
+    },
+    {
+      "name": "BranchInput",
+      "base": "BaseInput",
+      "attributes": [
+        {
+          "data_type": "ID",
+          "names": ["from_node", "to_node"],
```
mgovers, line 32: i think this makes sense as a default configuration for these files :+1:

code_generation/data/attribute_classes/input.json
```diff
@@ -1,583 +1,538 @@
...
+          "names": ["i_measured", "i_angle_measured"],
+          "description": "measured current and current angle"
         }
-    ]
-}
\ No newline at end of file
+      ]
+    }
+  ]
+}
```
mgovers, line 538: surprised to see newline at EOF because i thought it's not officially part of the JSON standard (but it is for JSONC) but as long as the code generator + CMake etc. work fine, it's alright, and it seems that they do.

.pre-commit-config.yaml.jinja
```diff
@@ -48,3 +48,9 @@ repos:
     hooks:
     - id: markdownlint
       args: ["--fix"]
+  # format .json and .jsonc files
+  - repo: https://github.com/biomejs/pre-commit  # outside pypi
```
mgovers, line 52: :+1:

biome.jsonc
```diff
@@ -0,0 +1,36 @@
...
+{
+  "$schema": "https://biomejs.dev/schemas/2.3.11/schema.json",
+  "files": {
+    "includes": [
+      "**/*.json",
+      "**/*.jsonc",
+      "!docs/**/*.json",
+      "!tests/**/input.json",
+      "!tests/**/*_output*.json",
+      "!tests/**/update_batch.json"
```
mgovers: probably desirable to restrict to files in the `data` subfolder

```suggestion
      "!tests/data/**/input.json",
      "!tests/data/**/*_output*.json",
      "!tests/data/**/update_batch.json"
```

CMakePresets.json
mgovers, line 1: I expect that recent changes will cause a merge conflict on this file. Should be easy to resolve, though, by resetting only this file to main during the merge, and then re-applying the reformatting (e.g. by running `pre-commit run --all-files`)

mgovers (commented):
Great job, very nice reformatting. I agree with your choice and the changes. Thank you for taking the time for researching the various possibilities.

> I didn't manage to set up the VS Code extension in the devcontainer properly and instead adjusted the default formatter in .vscode/settings.json to function similar to Biome. I can configure the devcontainer for Biome properly in another PR.

i think that that is alright, but with the existing configuration, i do get suggestions along the lines of 

```diff
 {
-  "calculation_method": ["linear", "newton_raphson", "iterative_current", "linear_current"],
+  "calculation_method": [ "linear", "newton_raphson", "iterative_current", "linear_current" ],
   "rtol": 1e-5,
   "atol": 1e-5
 }
```

If you want to postpone recommendation for Biome, then please try to see whether you can turn off those extra whitespace suggestions, or otherwise whether you can configure either Biome to add them. If neither of the options are feasible, it's fine to keep it as-is for now.

> I've noticed in the last run of the actions that Biome formats irrelevant files e.g. files in the [.mypy_cache](https://github.com/PowerGridModel/power-grid-model/actions/runs/20969763911/job/60269814919?pr=1256#step:10:16).
> I can add these folders to the blacklisted files after the rest of the PR is approved so that I can still easily revert the commit which formats all the files.

good choice. Please indeed blacklist them.

TonyXiang8787:
Hi @kornerc,

We would prefer a formatter which can be installed from `PyPI`. In that way, our automatic dependency refresh can make sure that the dev environment, pre-commit, and CI all use the same version of the formatter.

Your proposal of `biome` does not provide  a `PyPI` package. So we need to sync the formatter versions manually. Do you see other possibilities from `PyPI`? Or do you see a way to also automate the formatter/linter versions from nodejs?

TonyXiang8787:
Hi @kornerc,

After some digging. I think we can add `npm` structure into the repo to keep track of versioning. 

So first create a `package.json`

```json
{
  "private": true,
  "devDependencies": {
    "biomejs/biome": "*",
    "markdownlint-cli": "*"
  }
}
```

Add `node_modules/` to `.gitignore`.

Then run `npm install` locally once so we can also get `package-lock.json` to be check in to VCS.

Then we can change CI code quality, to install `npm` packages locally without `-g` and change the `markdownlint` and `biome` call with prefix of `npm run`. 

Then we modify the refresh dependency job to also refresh `npm` packages including adjusting `pre-commit.yaml`.

Finally change the build guide to instruct developers use `npm` to install the formatter.

code_generation/data/attribute_classes/input.json
kornerc: I looked into RFC 8259 (The JavaScript Object Notation (JSON) Data Interchange Format) and [section 2](https://datatracker.ietf.org/doc/html/rfc8259#section-2) explicitely allows _insignificant_ whitespaces like new lines at the end:
> A JSON text is a sequence of tokens.  The set of tokens includes six
>   structural characters, strings, numbers, and three literal names.
>
>   A JSON text is a serialized value.  Note that certain previous
>   specifications of JSON constrained a JSON text to be an object or an
>   array.  Implementations that generate only objects or arrays where a
>   JSON text is called for will be interoperable in the sense that all
>   implementations will accept these as conforming JSON texts.
>
>      JSON-text = ws value ws
>
>   These are the six structural characters:
>
>      begin-array     = ws %x5B ws  ; [ left square bracket
>      begin-object    = ws %x7B ws  ; { left curly bracket
>      end-array       = ws %x5D ws  ; ] right square bracket
>      end-object      = ws %x7D ws  ; } right curly bracket
>      name-separator  = ws %x3A ws  ; : colon
>      value-separator = ws %x2C ws  ; , comma
>
>   **Insignificant whitespace is allowed before or after any of the six
>   structural characters.**
>
>      ws = *(
>              %x20 /              ; Space
>              %x09 /              ; Horizontal tab
>              %x0A /              ; Line feed or New line
>              %x0D )              ; Carriage return

and accordingly to the [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) standard a line must end with a new line:
> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

So my own interpretation is: it does not hurt to have a newline at the EOF for a JSON file and it makes such a file a valid POSIX text file.
There exists is also a related ticket for [VS Code](https://github.com/microsoft/vscode/issues/176246)

kornerc:
> > I didn't manage to set up the VS Code extension in the devcontainer properly and instead adjusted the default formatter in .vscode/settings.json to function similar to Biome. I can configure the devcontainer for Biome properly in another PR.
> 
> i think that that is alright, but with the existing configuration, i do get suggestions along the lines of
> 
> ```diff
>  {
> -  "calculation_method": ["linear", "newton_raphson", "iterative_current", "linear_current"],
> +  "calculation_method": [ "linear", "newton_raphson", "iterative_current", "linear_current" ],
>    "rtol": 1e-5,
>    "atol": 1e-5
>  }
> ```
> 
> If you want to postpone recommendation for Biome, then please try to see whether you can turn off those extra whitespace suggestions, or otherwise whether you can configure either Biome to add them. If neither of the options are feasible, it's fine to keep it as-is for now.

~Unfortunately this is giving me some headaches:~
* `VS Code Biome Extension:` Somehow I am having problems formatting .json files with the VS Code extension
  https://github.com/biomejs/biome-vscode/issues/908 
* `Biome missing configuration option:` currently Biome has no option to add spaces into arrays
  https://github.com/biomejs/biome/issues/4607
* `VS Code missing option:` currently the .json formatter of VS Code does not have an option to disable the spaces in arrays
  https://github.com/microsoft/vscode/issues/247441

~I would suggest to 1) get this PR merged where the VS Code formatter is used in the devcontainer and 2) write an issue so that I can tackle this as soon the issue with the Biome extension is clarified.~

_Update:_ I managed to setup the VS-Code extension (Biome needs to be installed).
I will integrate it in the devcontainer. 

> > I've noticed in the last run of the actions that Biome formats irrelevant files e.g. files in the [.mypy_cache](https://github.com/PowerGridModel/power-grid-model/actions/runs/20969763911/job/60269814919?pr=1256#step:10:16).
> >    I can add these folders to the blacklisted files after the rest of the PR is approved so that I can still easily revert the commit which formats all the files.
>
> good choice. Please indeed blacklist them.

Biome now parses the `.gitignore` file and ignores the files and folders specified in it.

> After some digging. I think we can add npm structure into the repo to keep track of versioning.

Yes I like the idea to also automatically update the node dependencies.
Would it be okay if I try to implement this for Biome and markdownlint in a different PR, since I don't think that I find the time in the next week to address this?

> Finally change the build guide to instruct developers use npm to install the formatter.

~This should not be needed.~
`pre-commit` takes care about setting everything up for the Biome formatter hook.
~The VS Code Biome extension ships with everything which is needed to run Biome (I guess it is the Biome binary (programmed in Rust)).
Thus, the formatting should work with pre-commit and with the VS Code extension without installing node.~

_Update:_ The VS Code extension needs Biome to be installed I will have a look if it is able to use the Biome binary installed with precommit.

CMakePresets.json
kornerc: Interestingly the merge succeeded without any conflicts.
Also `uv run pre-commit run biome-format --all-files` does not complain.

mgovers: awesome :+1:

mgovers:
Hi @kornerc, Thank you again for the extensive investigation and update. Let us know if there's anything you need.

package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "devDependencies": {
+    "@biomejs/biome": "2.3.11",
+    "markdownlint-cli": "0.47.0"
```
TonyXiang8787: You need to keep the versions here as `*`, means the latest version.

The pinned version is in the lock file.

Then you can mofity the workflow to regularly update the lock file and pre-commit as what we do for `pypi` packages.

https://github.com/PowerGridModel/power-grid-model/blob/main/.github/workflows/refresh-lock-and-linter-dependencies.yml

.vscode/extensions.json
kornerc: ✅

.vscode/settings.json
kornerc: ✅

package.json
kornerc: ✅

kornerc:
@mgovers @TonyXiang8787 thank you for your reviews.
I've tried to address all your comments in this PR.

Overall, this PR was significantly more complex than I had initially anticipated but I am happy with the outcome.

The summary of the changes:
* [x] Use Biome for formatting .json and .jsonc files
  * Configuration is loacted in `biome.jsonc`
  * This configuration blacklists also files which should be ignored.
    In principal [biome is able](https://biomejs.dev/guides/integrate-in-vcs/#use-the-ignore-file) to parse `.gitignore` files correctly.
    However, since `.vscode` files are listed in the `.gitignore` I could not use this feature since the files in this folder should be formatted.
* [x] Install the VS Code extension in the devcontainer
  * I had to pin the extension to version `3.3.0` because I ran into this issue: https://github.com/biomejs/biome-vscode/issues/891
  * We can write an issue to remove the version pin later on.
* [x] Add Biome as pre-commit hook
* [x] Introduce an ignore file `.markdownlintignore` to not waste time formatting files which are ignored by git
* [x] Node.js dependencies
  * Track the version for the npm package dependencies in `package.json` and `package-lock.json`
  * From a user point of view it is not strictly necessary to install node for formatting .json and .jsonc files since the pre-commit hook takes care to obtain an isolated Node.js and npm binary if necessary to run node pre-commit hooks.
  * If the user wants to use the VS Code extension for Biome, having Node.js installed is mandatory
  * Futhermore, the devcontainer comes now with an installed Node.js and npm binary (enabled via `features` in the devcontainer)
* [x] Update npm lock files
  * I've updated the GitHub Action to not only update uv but also npm packages.
  * !!! This is **not tested** since I was not sure how to best test it !!!
* [x] All relevant .json and .jsonc files have been formatted with Biome
* [x] Docs
  * @TonyXiang8787 you mentioned to update the docs regarding the dependency to Node.js.
    Where is the best place to do that?
    As mentioned before it is not stricty needed to install Node.js since the pre-commit hook for formatting .json and .jsonc files will also work when Node.js is not installed.
* [x] Update branch: another branch update is necessary before merging

Side note: I am not a JS developer, therefore, I don't know Node.js that well.
Please have a critical look at the `npm` and `npx` commands which have been introduced.

PS I hope it's fine that I sticked to Biome even though it cannot be installed via pypi.
I wasn't able to find a Python formatter for json which is flexible enough to keep (single) empty lines within documents, can be configured to have a max line length, can format .json files in .vscode and .decontainer which can contain comments and trailing commas, etc.
Since markdownlint-cli uses also Node.js I don't see a big problem.

kornerc:
~Update: I just noticed: changing the versions in `package.json` to `*`, destroys the correct extraction of the version:
`npm pkg get devDependencies.markdownlint-cli` -> `"*"`~

~I guess I will use `npm ls --json --package-lock-only <package>` which lists the version of the _installed_ package.
E.g.~
```bash
npm ls --json --package-lock-only markdownlint-cli
```
```json
{
  "name": "power-grid-model",
  "dependencies": {
    "markdownlint-cli": {
      "version": "0.47.0",
      "resolved": "https://registry.npmjs.org/markdownlint-cli/-/markdownlint-cli-0.47.0.tgz",
      "overridden": false
    }
  }
}
```

~Then I can get the installed version with the field `version`.~

~I'll let you knew when I have solved this issue!~

Update: I've adjusted `refresh-lock-and-linter-dependencies.yml` and tested the individual commands locally.
Is there an easy possibility to test the correct functioning of the action?

.devcontainer/devcontainer.json
```diff
@@ -30,7 +35,9 @@
         "ms-vscode.test-adapter-converter",
         "njpwerner.autodocstring",
         "cschlosser.doxdocgen",
-        "davidanson.vscode-markdownlint"
+        "davidanson.vscode-markdownlint",
+        // version pinned (for now) since an old version (1.6.2) is installed otherwise
```
mgovers, line 39: 😮

mgovers:
> Is there an easy possibility to test the correct functioning of the action?

In your fork, you can go to https://github.com/kornerc/power-grid-model/actions/workflows/refresh-lock-and-linter-dependencies.yml (make sure you enable running of GitHub Actions in the settings) and then you can manually run the workflow on a certain branch
image
image

hope this helps

.markdownlintignore
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+.git
+.mypy_cache
+.pytest_cache
+.ruff_cache
+.venv
+build
```
mgovers, line 11: do we need to add other ones as well? e.g. `cpp_build`? or do we not expect any issues there?

package-lock.json
mgovers, line 1: was this intended to be checked in?

package-lock.json.license
mgovers, line 1: was this file intended to be checked in?

package.json
mgovers, line 1: was this file intended to be checked in?

package.json.license
mgovers, line 1: was this file intended to be checked in?

TonyXiang8787:
Hi @kornerc,

I notice that we also don't have this kind of things now in the build guide docs. Let's skip that part.

package.json.license
TonyXiang8787: Yes, it is intended to be checked-in. Just like `uv.lock`. So developers install same version of markdown linter and json linter, the same version is used also in pre-commit and CI.

.markdownlintignore
TonyXiang8787: I would say add `cpp_build` as well.

Maybe we want to investigate if we can just ignore all files which are already specified in gitignore. Instead of make another ignore list.

package.json
TonyXiang8787: Yes, it is intended to be checked-in. Just like `uv.lock`. So developers install same version of markdown linter and json linter, the same version is used also in pre-commit and CI.

.devcontainer/devcontainer.json
kornerc: Somehow I triggered this issue: https://github.com/biomejs/biome-vscode/issues/891

.markdownlintignore
kornerc: I added `cpp_build` as well to the ignored folders.

Regarding parsing the `.gitignore` file:
* `Biome`: [biome support this](https://biomejs.dev/guides/integrate-in-vcs/#use-the-ignore-file). However, I've disabled this feature since the folder `.vscode` is listed in `.gitignore`. I found no way to enable parsing `.gitignore` _and_ to force Biome to parse files even though they are listed in `.gitignore`.
  If you prefer it I can enable the parsing of `.gitignore` again with the downside that VS Code .json files don't get formatted. 
* `markdownlint-cli:` I haven't found such an option. I've found it only for `markdown-cli2` [supports this](https://github.com/DavidAnson/markdownlint-cli2?tab=readme-ov-file#markdownlint-cli2jsonc)

kornerc:
`Refresh lock and linter Dependencies` [seems to work](https://github.com/kornerc/power-grid-model/actions/runs/21183816893/job/60932803068).
I downgraded the versions of the npm packages and checked if the Action would correctly update the packages.
(Only an irrelevant part is failing at the end since I haven't set the options `owner` and `repositories` on my side)

.markdownlintignore
mgovers: i think it's fine as-is for now. @TonyXiang8787 do you have a strong opinion?

.vscode/extensions.json
```diff
@@ -1,29 +1,30 @@
...
+    "ms-vscode.cpptools-themes",
+    "ms-vscode.test-adapter-converter",
+    "njpwerner.autodocstring",
+    "jeff-hykin.better-cpp-syntax",
+    "matepek.vscode-catch2-test-adapter",
+    "cschlosser.doxdocgen",
+    "fredericbonnet.cmake-test-adapter",
+    "sonarsource.sonarlint-vscode",
+    "davidanson.vscode-markdownlint",
+    "biomejs.biome"
```
mgovers, line 28: I am getting this error in VSCode if I do not open it in a devcontainer.

Image

I even selected installing specifically version `3.3.0`

How did you manage to make it work?

mgovers (commented):
@kornerc thanks for testing. You're the real MVP!

The only question I have is how to make it work on VSCode - the rest LGTM

.markdownlintignore
TonyXiang8787: Let's keep it as is then.

TonyXiang8787:
Hi @kornerc,

All my comments have been resolved. If @mgovers gives some final check, I think we are good to go. Thanks again for your contribution.

.vscode/extensions.json
TonyXiang8787: Does this error only happen when you try to format a json with comment file? Or it will always show up even for a normal json file?

mgovers: any JSON file; here's the error when trying to format `CMakePresets.json`
image

kornerc: Did you install Biome by running `npm install` in the root folder of the project?

mgovers: not able to, but it seems to be an issue on my PC: https://stackoverflow.com/questions/24721903/npm-npm-cli-js-not-found-when-running-npm ... investigating

kornerc: I have installed Node.js via [scoop](https://scoop.sh/) under Windows.

mgovers: reinstalling Node.js worked 🎉 .

I will add this to the build guide.

mgovers: also verified to work under WSL2 (Ubuntu 24.04). Took a couple tries but (again) manually updating worked. Since I added that to the build guide, I think this suffices.

TonyXiang8787 (approved):

mgovers (approved):
LGTM; tested on Windows + WSL Ubuntu 24.04 + MacOS (done by @TonyXiang8787 )

.vscode/extensions.json
mgovers: tony verified macos. resolving.

kornerc:
@mgovers @TonyXiang8787 thanks a lot for your feedback and for the fruitful discussions leading to the final form of the PR!

mgovers:
> @mgovers @TonyXiang8787 thanks a lot for your feedback and for the fruitful discussions leading to the final form of the PR!

no problem! thank you for the extensive investigation and great end result!

mgovers:
Hi @kornerc,

Thank you once more for your nice presentation during the last community meeting and suggestions regarding additional improvements for the PGM project. It was very interesting to see your side of the story. How was the community meeting for you?

We would also like to ask you for your experience and maybe also some input regarding the documentation and in particular the examples.

PR #1264, opened 2026-01-15
OpenSSF silver badge: fix the issues on hardening

mgovers:
in #1241, hardening was enabled. Clang actually raised some valid undefined behavior issues:

```txt
/power-grid-model/power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp:95:27: runtime error: call to function PGM_dataset_foo through pointer to incorrect function type 'const PGM_ConstDataset *(*)(PGM_Handle *, long)'
/power-grid-model/power_grid_model_c/power_grid_model_c/src/dataset.cpp: note: PGM_dataset_foo defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /power-grid-model/power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp:95:27
```

The underlying reason was that we do

```cpp
#ifndef PGM_DLL_EXPORTS
typedef struct PGM_ConstDataset PGM_ConstDataset;
// ...
#endif

// ...
PGM_ConstDataset* PGM_create_dataset_const (/*...*/);
void              PGM_destroy_dataset_const(PGM_ConstDataset* /*...*/);
PGM_Foo           PGM_dataset_foo          (PGM_ConstDataset* /*...*/);
```

and the C API implementation then contains

```cpp
#define PGM_DLL_EXPORTS
#include "power_grid_model_c/basics.h"

#include <power_grid_model/auxiliary/dataset.hpp>

using PGM_ConstDataset = power_grid_model::meta_data::ConstDataset;
// ...

PGM_ConstDataset* PGM_create_dataset_const(/*...*/) { /* ... */ }
PGM_destroy_dataset_const(PGM_ConstDataset* /*...*/) { /* ... */ }
PGM_Foo* PGM_dataset_foo(PGM_ConstDataset /*...*/) { /* ... */ }
```

but the user will import like

```cpp
#include "power_grid_model_c/basics.h"

PGM_ConstDataset dataset = PGM_create_dataset_const(/*...*/);
PGM_dataset_foo(dataset);
PGM_destroy_dataset_const(dataset);
```

This effectively means that the C API implementation will see a type

```cpp
using PGM_ConstDataset = power_grid_model::meta_data::ConstDataset;
using PGM_DatasetInfo = power_grid_model::meta_data::DatasetInfo;
// ...

power_grid_model::meta_data::ConstDataset* PGM_create_dataset_const(/*...*/) { /* ... */ }
PGM_destroy_dataset_const(power_grid_model::meta_data::ConstDataset* /*...*/) { /* ... */ }
PGM_Foo PGM_dataset_foo(power_grid_model::meta_data::ConstDataset* /*...*/) { /* ... */ }
```

while the user will see a type

```cpp
typedef struct PGM_ConstDataset PGM_ConstDataset;

PGM_ConstDataset* PGM_create_dataset_const(/*...*/);
PGM_destroy_dataset_const(PGM_ConstDataset* /*...*/);

// usage
PGM_ConstDataset dataset = PGM_create_dataset_const(/*...*/);
PGM_dataset_foo(dataset);
PGM_destroy_dataset_const(dataset);
```

Or, in short, the user will see a type `struct PGM_ConstDataset` that is its own type, but the implementation will see the type `power_grid_model::meta_data::ConstDatast`.

After digging into it, it seemed very similar to something that was found in the CPython implementation (https://github.com/python/cpython/issues/111178). I then found that we do not have such issues for `PGM_PowerGridModel`, the implementation of which is defined as `struct PGM_PowerGridModel: public power_grid_model::MainModel {};`, which IS, in fact, its own type, rather than a type **alias** that comes from a `using` declaration.

The reason a `using` declaration was used in the first place is so that we could apply forward declarations, but now it seems that this was not the right solution. Note as well that the `#ifndef PGM_DLL_EXPORT` guard was used in the first place to suppress the exact same compiler errors, but we did not realize that the same undefined behavior was still being invoked on the user side.

To solve the issue in a way that still allows support for both features, I ended up following a similar approach as to main model, by using a derived type implementation, but also added forward declared, statically linked `wrap` and `unwrap` functions to handle the type conversions for files that use the forward declared version.

### DIFF
--- a/power_grid_model_c/power_grid_model_c/src/dataset.cpp
+++ b/power_grid_model_c/power_grid_model_c/src/dataset.cpp
@@ -30,35 +30,6 @@ using power_grid_model_c::to_c_bool;
 using power_grid_model_c::to_c_size;
 } // namespace
 
-struct PGM_ConstDataset : public power_grid_model::meta_data::ConstDataset {
-    using Dataset::Dataset;
-};
-struct PGM_MutableDataset : public power_grid_model::meta_data::MutableDataset {
-    using Dataset::Dataset;
-};
-struct PGM_WritableDataset : public power_grid_model::meta_data::WritableDataset {
-    using Dataset::Dataset;
-};
-struct PGM_DatasetInfo : public power_grid_model::meta_data::DatasetInfo {};
-
-namespace power_grid_model_c {
-ConstDataset const* unwrap(PGM_ConstDataset const* instance) {
-    return static_cast<ConstDataset const*>(instance); // may invoke undefined behavior
-}
-ConstDataset const& unwrap(PGM_ConstDataset const& instance) {
-    return static_cast<ConstDataset const&>(instance); // may invoke undefined behavior
-}
-MutableDataset const& unwrap(PGM_MutableDataset const& instance) {
-    return static_cast<MutableDataset const&>(instance); // may invoke undefined behavior
-}
-PGM_WritableDataset& wrap(WritableDataset& instance) {
-    return static_cast<PGM_WritableDataset&>(instance); // may invoke undefined behavior
-}
-PGM_DatasetInfo const& wrap(DatasetInfo const& instance) {
-    return static_cast<PGM_DatasetInfo const&>(instance); // may invoke undefined behavior
-}
-} // namespace power_grid_model_c
-
 // dataset info
 
 char const* PGM_dataset_info_name(PGM_Handle* handle, PGM_DatasetInfo const* info) {
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
@@ -356,7 +356,7 @@ class Topology {
         for (auto&& [idx, branch_node_idx, branch_connected] :
              std::views::zip(std::views::iota(0), std::as_const(comp_topo_.branch_node_idx),
                              std::as_const(comp_conn_.branch_connected))) {
-            assert(std::ssize(branch_connected) == 2);
+            assert(std::ssize(branch_connected) == 2); // NOSONAR(R354)
 
             auto const [i, j] = branch_node_idx;
             IntS const i_status = branch_connected[0];
--- a/power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
+++ b/power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
@@ -26,9 +26,6 @@ class Serializer;
 class Deserializer;
 
 template <dataset_type_tag dataset_type> class Dataset;
-using ConstDataset = Dataset<const_dataset_t>;
-using MutableDataset = Dataset<mutable_dataset_t>;
-using WritableDataset = Dataset<writable_dataset_t>;
 
 struct DatasetInfo;
 


TonyXiang8787:
What undefined behavior we are talking about? What we do is a well-known practice for many libraries, to declare an opaque type in C-API user side. But in C-API implementation side, we surpress that opaque type, and instead directly have a using directive to refer to a concrete type inside namespace.

TonyXiang8787:
Hi @mgovers, after some digging I see this is indeed UB. Not because of pointer case, which is well-defined if you do round-trip opaque pointer cast and never dereference the opaque pointer (you can't anyway). But because the function signature has different types between translation units.

But I do not think using wrapper type (even for `MainModel`) is an appropriate solution. A `reintepret_cast` should be more justified here because that's exactly what we are doing: define opaque type at user side, and `reintepret_cast` at implementation side to concrete types. Note in C++, cast from `CPPType*` to `CAPIType*`, and cast back `CAPIType*` to `CPPType*` and dereference it is well-defined as long as you never dereference `CAPIType*`, which we don't because the type is opaque.

So I would propose to just pre-define a `reintepret_cast` list, like this:

```c++
template<class c_type, class cpp_type>
struct type_matching;

template<class...>
struct type_matching_list_impl;

using type_matching_list = typematching_list<
    type_matching<PGM_MetaComponent, power_grid_model::meta_data::MetaComponent>,
    // list all opaque to concrete type matching
>;

template<class c_type_pointer>
using cpp_type_pointer_t = // some template magic to find the correct cpp type pointer, also keep const correctness

template<class c_type_pointer>
auto cast_to_cpp(c_type_pointer ptr) { 
return reintepret_cast<cpp_type_pointer_t<c_type_pointer>>(ptr); 
}

template<class cpp_type_pointer>
using c_type_pointer_t = // some template magic to find the correct c type pointer, also keep const correctness

template<class cpp_type_pointer>
auto cast_to_c(cpp_type_pointer ptr) { 
return reintepret_cast<c_type_pointer_t<cpp_type_pointer>>(ptr); 
}
```

After this, we can

1. remove the macro guard of those opaque times. Then in translation unit of our C-API compilation and user side they see the same funciton signature.
2. In cpp side of C-API implementation, call `cast_to_cpp` when you receive a C opaque type, and call `cast_to_c` when you want to return as C opaque type.

mgovers:
> So I would propose to just pre-define a reintepret_cast list, like this

I thought about something along those lines as well, but then we still have the `typedef struct PGM_Foo PGM_Foo;` forward declaration that needs to be defined somewhere. What would that definition be? Would you change it to `typedef void PGM_Foo`? Wouldn't that be a breaking change?

TonyXiang8787:
> > So I would propose to just pre-define a reintepret_cast list, like this
> 
> I thought about something along those lines as well, but then we still have the `typedef struct PGM_Foo PGM_Foo;` forward declaration that needs to be defined somewhere. What would that definition be? Would you change it to `typedef void PGM_Foo`? Wouldn't that be a breaking change?

That's not needed. We do not need to define a forward opaque type. We can just cast between `PGM_Foo*` and `power_grid_model::Foo*`. The `PGM_Foo` is always opaque type, never defined.

mgovers:
> > > So I would propose to just pre-define a reintepret_cast list, like this
> > 
> > 
> > I thought about something along those lines as well, but then we still have the `typedef struct PGM_Foo PGM_Foo;` forward declaration that needs to be defined somewhere. What would that definition be? Would you change it to `typedef void PGM_Foo`? Wouldn't that be a breaking change?
> 
> That's not needed. We do not need to define a forward opaque type. We can just cast between `PGM_Foo*` and `power_grid_model::Foo*`. The `PGM_Foo` is always opaque type, never defined.

let me try

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -240,31 +235,35 @@ class MDBatchExceptionHandler : public power_grid_model_c::DefaultExceptionHandl
...
+        size *= safe_current.batch_size();
+        current = safe_current.get_next_cartesian_product_dimension();
     }
     return size;
 }
 
 // run calculation
-void calculate_multi_dimensional_impl(PGM_PowerGridModel& model, PGM_Options const& opt,
-                                      PGM_MutableDataset const& output_dataset, PGM_ConstDataset const* batch_dataset) {
+void calculate_multi_dimensional_impl(MainModel& model, MainModel::Options const& options,
```
mgovers: this is still missing from #1266

power_grid_model_c/power_grid_model_c/src/dataset.cpp
```diff
@@ -27,6 +28,36 @@ using power_grid_model_c::safe_ptr_maybe_nullptr;
...
+MutableDataset const& unwrap(PGM_MutableDataset const& instance) {
+    return static_cast<MutableDataset const&>(instance); // may invoke undefined behavior
+}
+PGM_WritableDataset& wrap(WritableDataset& instance) {
+    return static_cast<PGM_WritableDataset&>(instance); // may invoke undefined behavior
+}
+PGM_DatasetInfo const& wrap(DatasetInfo const& instance) {
+    return static_cast<PGM_DatasetInfo const&>(instance); // may invoke undefined behavior
+}
+} // namespace power_grid_model_c
```
TonyXiang8787: There is no need to do this anymore.

mgovers:
Closed in favor of #1269

PR #1266, opened 2026-01-18
OpenSSF silver badge: Fix UB of different type names for C-API opaque types

TonyXiang8787:
As discovered by #1241, this PR is a new proposal to fix the C-API opaque types issue by explicit casting, instead of defining child class as in #1264.

This is the standard way of avoiding the different type violation between user compilation and library compilation, used by many open-source projects.

### DIFF
--- a/power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
+++ b/power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
@@ -35,9 +35,12 @@ struct DatasetInfo;
 
 namespace power_grid_model_c {
 
+namespace detail {
+
 template <class c_type_input, class cpp_type_input> struct c_cpp_type_map {
     using c_type = c_type_input;
     using cpp_type = cpp_type_input;
+    static_assert(!std::is_same_v<c_type, cpp_type>, "C and C++ types in type map cannot be the same!");
 };
 
 template <class... type_maps> struct type_mapping_list_impl {
@@ -94,5 +97,10 @@ template <class CTypePtr> auto cast_to_cpp(CTypePtr ptr) {
 template <class CPPTypePtr> auto cast_to_c(CPPTypePtr ptr) {
     return reinterpret_cast<convert_ptr_to_c_t<CPPTypePtr>>(ptr);
 }
+
+} // namespace detail
+
+using detail::cast_to_c;
+using detail::cast_to_cpp;
 
 } // namespace power_grid_model_c


power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
```diff
@@ -25,14 +29,70 @@ template <dataset_type_tag dataset_type> class Dataset;
...
+template <class CTypePtr> using convert_ptr_to_cpp_t = typename convert_ptr_to_cpp<CTypePtr>::type;
+
+template <class CPPTypePtr> struct convert_ptr_to_c {
+    static constexpr bool is_const = std::is_const_v<std::remove_pointer_t<CPPTypePtr>>;
+    using base_cpp_type = std::remove_const_t<std::remove_pointer_t<CPPTypePtr>>;
+    using mapped_c_raw_type = type_mapping_list::get_c_type_t<base_cpp_type>;
+    using mapped_c_type = std::conditional_t<is_const, std::add_const_t<mapped_c_raw_type>, mapped_c_raw_type>;
+    using type = mapped_c_type*;
+};
+template <class CPPTypePtr> using convert_ptr_to_c_t = typename convert_ptr_to_c<CPPTypePtr>::type;
```
mgovers, line 91: optionally, we can put this in `namespace detail`. i don't have a strong opinion here, but it feels like a good idea to prevent accidentally picking the wrong type

power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
```diff
@@ -25,14 +29,70 @@ template <dataset_type_tag dataset_type> class Dataset;
...
+
+    template <class c_type>
+    using get_cpp_type_t =
+        std::conditional_t<std::is_same_v<typename first_map::c_type, c_type>, typename first_map::cpp_type,
+                           typename type_mapping_list_impl<rest_maps...>::template get_cpp_type_t<c_type>>;
+
+    template <class cpp_type>
+    using get_c_type_t =
+        std::conditional_t<std::is_same_v<typename first_map::cpp_type, cpp_type>, typename first_map::c_type,
+                           typename type_mapping_list_impl<rest_maps...>::template get_c_type_t<cpp_type>>;
```
mgovers, line 60: there is a theoretical risk that this breaks if someone provides 2 mappings from/to the same type. should there also be a check that ensures uniqueness in the input parameters? if not, then that's all the more reason to put this into `namespace detail` to prevent incorrect overloading.

mgovers (commented):
code review is OK. i will do some local checks to ensure:
* [x] that the issues found in #1241 are resolved in this branch
* [x] that there are no other changes in #1264 that are missing here

power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model_c/src/forward_declarations.hpp
TonyXiang8787: done

PR #1269, opened 2026-01-19
OpenSSF silver badge: minor refactor C API implementation

mgovers:
Cherry-picked changes from #1264

Re-opened a new PR as #1264 was superceded by #1266.

This PR is also rebased off `main` rather than #1241 to allow easier development.

### DIFF
--- a/power_grid_model_c/power_grid_model_c/src/model.cpp
+++ b/power_grid_model_c/power_grid_model_c/src/model.cpp
@@ -256,8 +256,8 @@ Idx get_stride_size(ConstDataset const* batch_dataset) {
 }
 
 // run calculation
-void calculate_multi_dimensional_impl(MainModel& model, MainModel::Options const& options, MutableDataset const& output_dataset,
-                                      ConstDataset const* batch_dataset) {
+void calculate_multi_dimensional_impl(MainModel& model, MainModel::Options const& options,
+                                      MutableDataset const& output_dataset, ConstDataset const* batch_dataset) {
     // for dimension < 2 (one-time or 1D batch), call implementation directly
     if (auto const batch_dimension = get_batch_dimension(batch_dataset); batch_dimension < 2) {
         calculate_single_batch_dimension_impl(model, options, output_dataset, batch_dataset);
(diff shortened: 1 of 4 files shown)


power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -241,29 +234,33 @@ class MDBatchExceptionHandler : public power_grid_model_c::DefaultExceptionHandl
 
 Idx get_batch_dimension(ConstDataset const* batch_dataset) {
     Idx dimension = 0;
-    while (batch_dataset != nullptr) {
+    ConstDataset const* safe_batch_dataset = safe_ptr_maybe_nullptr(batch_dataset);
```
TonyXiang8787, line 237: What is the purpose here? Batch dataset can be nullptr, so what we are doing here?

power_grid_model_c/power_grid_model_c/src/model.cpp
```diff
@@ -241,29 +234,33 @@ class MDBatchExceptionHandler : public power_grid_model_c::DefaultExceptionHandl
...
+    while (safe_batch_dataset != nullptr) {
         ++dimension;
-        batch_dataset = batch_dataset->get_next_cartesian_product_dimension();
+        safe_batch_dataset =
+            safe_ptr_maybe_nullptr(safe_ptr_get(safe_batch_dataset).get_next_cartesian_product_dimension());
```
TonyXiang8787, line 241: Same here?

power_grid_model_c/power_grid_model_c/src/model.cpp
mgovers: `safe_ptr_maybe_nullptr` is a no-op function (as of now) to indicate that we in fact did think about the contents of the data but that it may still be `nullptr`. However, it may in the future be extended to e.g. check that it is a value in the range of valid pointers (e.g. not exceeding system size, e.g. as a result of `size_t{-1}`).

We currently do not add those additional check, but this helper function serves the purpose of indicating that we indeed can safely load the pointer

mgovers:
> Since we are here, please fix the additional problems in the dataset pointer check. Namely for the attribute pointers.
> 
> https://github.com/PowerGridModel/power-grid-model/blob/b951eff724a39a509bab0e78c957430c89568468/power_grid_model_c/power_grid_model_c/src/dataset.cpp#L131-L137
> 
> https://github.com/PowerGridModel/power-grid-model/blob/b951eff724a39a509bab0e78c957430c89568468/power_grid_model_c/power_grid_model_c/src/dataset.cpp#L165-L171
> 
> https://github.com/PowerGridModel/power-grid-model/blob/b951eff724a39a509bab0e78c957430c89568468/power_grid_model_c/power_grid_model_c/src/dataset.cpp#L197-L203
> 
> These `data` pointers should be allowed to be `nullptr`, if the total number of the elements in the component is zero. Users may still create such an attribute buffer (e.g. empyt vector or empty span, or empty numpy array) in an automatic way. But the `data` pointer maybe `nullptr` which is well defined.

i am implementing that as we speak but it is something we really want to allow? ~currently, it will introduce segmentation violations~ EDIT: my test was incorrect. it works as intended

PR #1286, opened 2026-01-28
Main model refactoring: move calculation logic out of main model

mgovers:
This finishes up the cleanup of main model

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -311,8 +311,8 @@ class MainModelImpl {
             assert(options.optimizer_type == OptimizerType::no_optimization ||
                    (std::derived_from<calculation_type, power_flow_t>));
 
-            return [this, &mutable_comp_coup = state_.comp_coup, &options, &logger](MainModelState const& state,
-                                                                            CalculationMethod calculation_method) {
+            return [this, &mutable_comp_coup = state_.comp_coup, &options,
+                    &logger](MainModelState const& state, CalculationMethod calculation_method) {
                 (void)state; // to avoid unused-lambda-capture when in Release build
                 assert(&state == &state_);
 


CMakeLists.txt
```diff
@@ -13,11 +13,18 @@ project(power_grid_model VERSION ${PGM_VERSION})
...
+if(NOT CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS ${PGM_CXX_STANDARD})
+    set(CMAKE_CXX_STANDARD ${PGM_CXX_STANDARD})
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
-set(CMAKE_C_STANDARD 11)
+if(NOT CMAKE_C_STANDARD OR CMAKE_C_STANDARD LESS ${PGM_C_STANDARD})
+    set(CMAKE_C_STANDARD ${PGM_C_STANDARD})
+endif()
```
mgovers, line 27: this is considered best practice for libraries including shared libraries. E.g. if a user wants to enable C23 or C++26, they should be able to. Our original code prevents them from doing so

TonyXiang8787: What do you mean here exactly? In old way, we force user to compile the library in C++23. But once the library is installed by the user at a given location. To consume the library, user can use whatever they want right?

mgovers: if the user wants to use e.g. `FetchContent` to include PGM into their C++26 project (hypothetically), then we force C++23, even if they set it to C++23. Making it from an exact match into a minimum requirement resolves that.

TonyXiang8787: So you are talking about that user somehow make PGM as part of their source tree. Instead of build and install it separately.

mgovers: yes

power_grid_model_c/power_grid_model/include/power_grid_model/common/counting_iterator.hpp
```diff
@@ -21,4 +21,12 @@ struct IdxRange : public std::ranges::iota_view<Idx, Idx> {
 };
 using IdxCount = typename IdxRange::iterator;
 
+#if __cpp_lib_ranges_enumerate >= 20202L
+using std::views::enumerate;
+#else
+constexpr auto enumerate(std::ranges::viewable_range auto&& rng) {
+    return std::views::zip(IdxRange{std::ssize(rng)}, rng);
```
figueroa1395: `rng` can be easily miss-interpreted. Can you just use `range` instead?
```suggestion
constexpr auto enumerate(std::ranges::viewable_range auto&& range) {
    return std::views::zip(IdxRange{std::ssize(range)}, range);
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -323,14 +264,18 @@ class MainModelImpl {
...
-    template <solver_output_type SolverOutputType, typename MathSolverType, typename YBus, typename InputType,
-              typename PrepareInputFn, typename SolveFn>
+    template <typename MathSolverType, typename YBus, typename PrepareInputFn, typename SolveFn>
         requires std::invocable<std::remove_cvref_t<PrepareInputFn>, Idx /*n_math_solvers*/> &&
-                 std::invocable<std::remove_cvref_t<SolveFn>, MathSolverType&, YBus const&, InputType const&> &&
-                 std::same_as<std::invoke_result_t<PrepareInputFn, Idx /*n_math_solvers*/>, std::vector<InputType>> &&
-                 std::same_as<std::invoke_result_t<SolveFn, MathSolverType&, YBus const&, InputType const&>,
-                              SolverOutputType>
-    std::vector<SolverOutputType> calculate_(PrepareInputFn&& prepare_input, SolveFn&& solve, Logger& logger) {
+                 std::ranges::range<std::invoke_result_t<PrepareInputFn, Idx /*n_math_solvers*/>> &&
```
figueroa1395, line 269: Were the changes from `std::same_as<a, std::vector<b>>` to `std::ranges::range<a, b>` just so it's more general and up to date code? Do you expect to have potential `std::span`s for instance at some point?  Wouldn't it be best to keep it restricted to `std::vector` only if we don't foresee this?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -357,85 +302,32 @@ class MainModelImpl {
...
+            return [this, &comp_coup = state_.comp_coup, &options, &logger](MainModelState const& state,
+                                                                            CalculationMethod calculation_method) {
+                (void)state; // to avoid unused-lambda-capture when in Release build
+                assert(&state == &state_);
+
+                return calculate_<MathSolverProxy<sym>, YBus<sym>>(Calc::preparer(state, comp_coup, options),
+                                                                   Calc::solver(calculation_method, options, logger),
+                                                                   logger);
+            };
+        };
```
figueroa1395: Why not try to get rid of the `state` and `state_` "shenanigans" and do something like (beware of typos)
```suggestion
            return [this, &options, &logger](MainModelState const& state,
                                                                            CalculationMethod calculation_method) {

                return calculate_<MathSolverProxy<sym>, YBus<sym>>(Calc::preparer(state, state.comp_coup, options),
                                                                   Calc::solver(calculation_method, options, logger),
                                                                   logger);
            };
        };
```

power_grid_model_c/power_grid_model/include/power_grid_model/calculation.hpp
```diff
@@ -0,0 +1,121 @@
...
+                                                                               SubArgs&&... sub_args) {
+                    std::forward<SubFunctor>(sub_f).template operator()<calculation_type, sym>(
+                        std::forward<SubArgs>(sub_args)...);
+                },
+                std::forward<Functor_>(f_), std::forward<Args_>(args_)...);
+        },
+        calculation_symmetry, std::forward<Functor>(f), std::forward<Args>(args)...);
+}
+
+template <typename T, typename sym> struct Calculator;
```
figueroa1395, line 76: Nice, I really like these!

power_grid_model_c/power_grid_model/include/power_grid_model/calculation.hpp
```diff
@@ -0,0 +1,121 @@
...
+template <symmetry_tag sym> struct Calculator<power_flow_t, sym> {
```
nitbharambe, line 78:
```suggestion
template <typename T, typename sym> struct Calculator;

template <symmetry_tag sym> struct Calculator<power_flow_t, sym> {
```
Minor

power_grid_model_c/power_grid_model/include/power_grid_model/common/counting_iterator.hpp
mgovers: name conflicts 😢

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: i needed access to the more generic type, otherwise it was not able to infer the type

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: mutable vs immutable. the input state is presumed const, but the comp coup needs to be mutable

power_grid_model_c/power_grid_model/include/power_grid_model/common/counting_iterator.hpp
figueroa1395: @nitbharambe Nitpick: Can you still replace with something else instead of `rng`? It confuses me as I immediately link it to random number generator.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
figueroa1395: @nitbharambe Nitpick: Can you rename to `state_mutable` and `state_immutable` or something along those lines? What do you think?

nitbharambe: > Can you rename to state_mutable
 
Good idea

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -357,85 +302,32 @@ class MainModelImpl {
...
-                return calculate_short_circuit_<sym>(options.short_circuit_voltage_scaling, logger);
-            } else {
-                throw UnreachableHit{"MainModelImpl::calculate", "Unknown calculation type"};
-            }
-        }();
+
+            return [this, &mutable_comp_coup = state_.comp_coup, &options,
+                    &logger](MainModelState const& state, CalculationMethod calculation_method) {
+                (void)state; // to avoid unused-lambda-capture when in Release build
+                assert(&state == &state_);
```
figueroa1395, line 317: Nitpick: Along with the other respective renaming
```suggestion
                assert(&state_const == &state_mutable);
```
Or something like that. Just to directly see the difference (which is then clear related to `mutable_comp_coup`).

nitbharambe: I read from above that you meant this statement. 
This assert statement should simply check if the lambda of get_calculator gets called with same state as the one it is created from. LHS/RHS here is not about constness.

figueroa1395: But I think it is. Because `state_` is mutated via `mutable_comp_coup`, so shouldn't we explicitly state this then?

mgovers: tbf, do we even need this assertion in the first place? it's really just a check given the complexity of the original main model. with the new implementation, if the user specifies the wrong state, that's on them

figueroa1395: Let's just then leave as is and don't make a big deal out of it. It's a nitpick anyway. Resolving.

Issue #1278, opened 2026-01-26
[BUG] readthedocs equations are not visible


nitbharambe:
**Describe the bug**

<img width="2352" height="417" alt="Image" src="https://github.com/user-attachments/assets/09c0449e-adea-4f9c-9692-255ce32dd8db" />

something must have broken the equations because of upgrade of sphinx mentioned in #1267 .

PR #1288, opened 2026-01-28
Fix docs/components

furqan463:
**Fixes issue**: #1278 

multi-line equations inside `$$` delimiter do not render properly. I've replaced them with multiple `$$` math blocks.

### Checks

- [x] review other pages.

### DIFF
--- a/docs/user_manual/components.md
+++ b/docs/user_manual/components.md
@@ -1035,7 +1035,11 @@ $$
 
 Current sensors with `angle_measurement_type` equal to `AngleMeasurementType.local_angle` measure the phase shift
 between the voltage and the current phasor, i.e.,
-$\begin{equation}i\_angle\_measured = voltage\_phase - current\_phase\end{equation}$.
+
+$$
+\text{i\_angle\_measured} = \text{voltage\_phase} - \text{current\_phase}\text{ .}
+$$
+
 As a result, the global current phasor depends on the local voltage phase offset and is obtained using the following
 formula.
 


furqan463:
I could fix equation rendering in `components.md`, but I've also observed the same issue in `calculations.md` with multi-line arrays inside `$$` delimiter. However, I'm unable to find a solution for that.

nitbharambe:
There are errors in other files too like LU solver and in calculations. 

Initial idea about the ticket was to find out a way which does not require modifying each equation with such issue. Maybe there is a way we can resolve this at sphinx or myst or mathjax side directly.

furqan463:
> There are errors in other files too like LU solver and in calculations.
> 
> Initial idea about the ticket was to find out a way which does not require modifying each equation with such issue. Maybe there is a way we can resolve this at sphinx or myst or mathjax side directly.

I had the same intention but couldn't find a way.

docs/user_manual/components.md
```diff
@@ -1320,8 +1387,16 @@ node becomes a PQ node:
...
+$$
+   \begin{eqnarray}
+      & Q_{\text{gen}} = Q_{\text{min}} \text{ or } Q_{\text{max}}
+   \end{eqnarray}
+$$
+$$
+   \begin{eqnarray}
       & |U_{\text{node}}| = \text{calculated from power flow}
    \end{eqnarray}
 $$
```
nitbharambe, line 1330: Can you try this instead?

```
$$\begin{eqnarray}
      & P_{\text{gen}} = P_{\text{specified}} \\
      & Q_{\text{gen}} = Q_{\text{min}} \text{ or } Q_{\text{max}} \\
      & |U_{\text{node}}| = \text{calculated from power flow}
  \end{eqnarray}$$
  ```
  
See https://stackoverflow.com/questions/18860693/mathjax-multi-line-equation-rendering-issue

(Edit: It also suggests changing to align instead of eqnarray though)

Comment (furqan463) [reply docs/user_manual/components.md:1330]: > Can you try this instead?
> 
> ```
> $$\begin{eqnarray}
>       & P_{\text{gen}} = P_{\text{specified}} \\
>       & Q_{\text{gen}} = Q_{\text{min}} \text{ or } Q_{\text{max}} \\
>       & |U_{\text{node}}| = \text{calculated from power flow}
>   \end{eqnarray}$$
> ```
> 
> See https://stackoverflow.com/questions/18860693/mathjax-multi-line-equation-rendering-issue
> 
> (Edit: It also suggests changing to align instead of eqnarray though)

This makes no difference. align* works after building the document, however live preview while editing is not available.

**Comment by furqan463** [comment]
> I allowed the workflow run and `align*` seems to render all the equations correctly. @furqan463 Perhaps you can give it a try at the other instances where this is an issue? There is some [documentation](https://power-grid-model.readthedocs.io/en/stable/advanced_documentation/build-guide.html#documentation) (don't forget to install [Doxygen](https://www.doxygen.nl/manual/install.html)) to build the docs locally, so you can see whether this works or not and not depend on the CI.
> 
> The only remaining thing on the current edited file is to fix formatting, but otherwise it looks good (please double check that I didn't miss any potentially badly rendered equation).
> 
> Side note for @nitbharambe: Because `align` aligns differently, some consecutive equations "look" differently compared to before this broke. I think that's fine, but perhaps we want to align in a specific fashion?

@figueroa1395 thanks for the review. I already built the docs locally, and docs are built in PR without CI. I'll try in other instances, my only concern in the first place was that we loose live preview without `$$` delimiter. So we've 2 options:

1. Keep `$$` delimiter that supports live preview in editor. Trade off: Place all equations separately in `$$` blocks without grouping.
2. Use `align*` delimiter and keep equations grouped together. Trade off: Loos live preview in editor.

**Comment by figueroa1395** [comment]
Hello @furqan463,

The correct approach should be to use both `$$` and `align*`, that way doc generation renders correctly and the preview still works in VS Code.

Taking the following as an example (this is from the current state in this PR):
```
\begin{align*}
    & |Z_{\text{series}}| = u_k*z_{\text{base,transformer}} \\
    & \mathrm{Re}(Z_{\text{series}}) = \frac{p_k}{s_n}*z_{\text{base,transformer}} \\
    & \mathrm{Im}(Z_{\text{series}}) = \sqrt{|Z_{\text{series}}|^2-\mathrm{Re}(Z_{\text{series}})^2}
\end{align*}
```

can be put in between `$$`s, like:
```
$$
\begin{align*}
    & |Z_{\text{series}}| = u_k*z_{\text{base,transformer}} \\
    & \mathrm{Re}(Z_{\text{series}}) = \frac{p_k}{s_n}*z_{\text{base,transformer}} \\
    & \mathrm{Im}(Z_{\text{series}}) = \sqrt{|Z_{\text{series}}|^2-\mathrm{Re}(Z_{\text{series}})^2}
\end{align*}
$$
```

I quickly tried locally and it seems to work without issue.

So basically replacing `eqnarray` with `align*`, while keeping all the `$$` blocks should work as intended.

**Comment by furqan463** [comment]
> Hello @furqan463,
> 
> The correct approach should be to use both `$$` and `align*`, that way doc generation renders correctly and the preview still works in VS Code.
> 
> Taking the following as an example (this is from the current state in this PR):
> 
> ```
> \begin{align*}
>     & |Z_{\text{series}}| = u_k*z_{\text{base,transformer}} \\
>     & \mathrm{Re}(Z_{\text{series}}) = \frac{p_k}{s_n}*z_{\text{base,transformer}} \\
>     & \mathrm{Im}(Z_{\text{series}}) = \sqrt{|Z_{\text{series}}|^2-\mathrm{Re}(Z_{\text{series}})^2}
> \end{align*}
> ```
> 
> can be put in between `$$`s, like:
> 
> ```
> $$
> \begin{align*}
>     & |Z_{\text{series}}| = u_k*z_{\text{base,transformer}} \\
>     & \mathrm{Re}(Z_{\text{series}}) = \frac{p_k}{s_n}*z_{\text{base,transformer}} \\
>     & \mathrm{Im}(Z_{\text{series}}) = \sqrt{|Z_{\text{series}}|^2-\mathrm{Re}(Z_{\text{series}})^2}
> \end{align*}
> $$
> ```
> 
> I quickly tried locally and it seems to work without issue.
> 
> So basically replacing `eqnarray` with `align*`, while keeping all the `$$` blocks should work as intended.

@figueroa1395  I tried locally and also updated 1 equation, line electrical model, here.
It did work with the equation but messed with following text.
Preview works fine but when built, `$$` delimiter before `/begin{align*}` is not treated as math-block delimiter, while `$$` delimiter after `/end{align*}` is treated as start of math block which messes with following text up to next `$$` delimiter in the file.

**Comment by mgovers** [comment]
Since this is a high-prio issue, maybe it's a good idea to first just use `\begin{align*}` everywhere first to fix the rendering on readthedocs and then do a follow-up PR to try to fix the preview

**Comment by figueroa1395** [comment]
@furqan463 You are right, my approach was incorrect. My apologies.

After doing some experimentation locally, I didn't see a different quick and easy way to fix this, providing good rendering and VS code preview at the same time, that is not using `$$` only for each individual equation, as you suggested initially:

> Keep $$ delimiter that supports live preview in editor. Trade off: Place all equations separately in $$ blocks without grouping.

Furthermore, this is a high priority issue, as mentioned by @mgovers, so I suggest to go with that option in all files. Then we have to make sure all equations are rendered correctly. If we later want/need a fancier fix, we can do it as a follow up PR.

In conclusion, let's just use `$$` for each individual equation.

C.C. @nitbharambe

**Comment by furqan463** [comment]
> @furqan463 You are right, my approach was incorrect. My apologies.
> 
> After doing some experimentation locally, I didn't see a different quick and easy way to fix this, providing good rendering and VS code preview at the same time, that is not using `$$` only for each individual equation, as you suggested initially:
> 
> > Keep $$ delimiter that supports live preview in editor. Trade off: Place all equations separately in $$ blocks without grouping.
> 
> Furthermore, this is a high priority issue, as mentioned by @mgovers, so I suggest to go with that option in all files. Then we have to make sure all equations are rendered correctly. If we later want/need a fancier fix, we can do it as a follow up PR.
> 
> In conclusion, let's just use `$$` for each individual equation.
> 
> C.C. @nitbharambe

@figueroa1395 multiple `$$` doesn't work everywhere, e.g., vertical matrices in `calculations.md`. Therefore, I used `align*`.

**Comment by figueroa1395** [comment]
Hello @furqan463,

I decided to dig a little deeper and found the actual solution. I tested both on VS Code preview and a local build, and it works as expected.

The main issue is that `eqnarray` is [deprecated](https://tug.org/pracjourn/2006-4/madsen/madsen.pdf) and that `$$` is intended to be used only for single line equations. Now, `sphynx` just ignores `eqnarray` when it's in a `$$` block, so single line equations render well (both on VS Code and the docs) as intended; however, `$$` doesn't support multi line equations and `eqnarray` being ignored doesn't fix it.

On the other hand, even though `align*` works well, it doesn't support live preview in VSCode because it already generates its own math environment, which becomes redundant with enclosing it in a `$$` block. For this reason, `aligned` is there.`aligned` is meant to be used inside a math environment (see section 3 of https://www.ams.org/arc/tex/amsmath/amsldoc.pdf), hence providing support for multi-line equations while inside a `$$` block which provides preview.

In summary, the solution is to leave every single-line equation in a `$$` block alone (no `aligned`, `align*`, `eqnarray`, etc.), every multi-line equation should be placed in a `$$` block with `\begin{aligned} ... \end{aligned}` (beware to indent the `\begin{aligned}` block at the same level of the `$$` block). In the final result, no `eqnarray` or `align*` should be present, in favor of `$$` blocks for single line equations or `$$ \begin{aligned} ... \end{aligned} $$` blocks for multi line equations. Finally, I didn't see any issue of rendering or with vertical matrices.

Let me know if this helps or if you need a hand.

**Comment by furqan463** [comment]
> Hello @furqan463,
> 
> I decided to dig a little deeper and found the actual solution. I tested both on VS Code preview and a local build, and it works as expected.
> 
> The main issue is that `eqnarray` is [deprecated](https://tug.org/pracjourn/2006-4/madsen/madsen.pdf) and that `$$` is intended to be used only for single line equations. Now, `sphynx` just ignores `eqnarray` when it's in a `$$` block, so single line equations render well (both on VS Code and the docs) as intended; however, `$$` doesn't support multi line equations and `eqnarray` being ignored doesn't fix it.
> 
> On the other hand, even though `align*` works well, it doesn't support live preview in VSCode because it already generates its own math environment, which becomes redundant with enclosing it in a `$$` block. For this reason, `aligned` is there.`aligned` is meant to be used inside a math environment (see section 3 of https://www.ams.org/arc/tex/amsmath/amsldoc.pdf), hence providing support for multi-line equations while inside a `$$` block which provides preview.
> 
> In summary, the solution is to leave every single-line equation in a `$$` block alone (no `aligned`, `align*`, `eqnarray`, etc.), every multi-line equation should be placed in a `$$` block with `\begin{aligned} ... \end{aligned}` (beware to indent the `\begin{aligned}` block at the same level of the `$$` block). In the final result, no `eqnarray` or `align*` should be present, in favor of `$$` blocks for single line equations or `$$ \begin{aligned} ... \end{aligned} $$` blocks for multi line equations. Finally, I didn't see any issue of rendering or with vertical matrices.
> 
> Let me know if this helps or if you need a hand.

@figueroa1395 the issue with vertical matrices is that you can't start bmatrix in one `$$` block and end it in subsequent `$$` block, with matrix rows in between in `$$` blocks.

**Comment by figueroa1395** [comment]
>@figueroa1395 the issue with vertical matrices is that you can't start bmatrix in one $$ block and end it in subsequent $$ block, with matrix rows in between in $$ blocks.

Hello @furqan463,

Vertical matrices, such as the ones found in `calculations.md` are rendered fine if placed in a single `aligned` block within a single `$$` block. This works because even thought the `bmatrix` itself isn't considered a multi-line equation, when doing things such as `x = some_matrix`, the whole expression becomes a multi-line-equation.

Here is how you can fix some of the instances where such matrices appear in the `calculations.md` file:

Starting at line 319 at the at current head of this branch:
```
$$
\begin{aligned}
    x    =  \begin{bmatrix}
            \delta \\
            U
            \end{bmatrix} =
            \begin{bmatrix}
            \delta_2 \\
            \vdots \\
            \delta_N \\
            U_2 \\
            \vdots \\
            U_N
            \end{bmatrix}
    \quad\text{and}\quad
    y    =  \begin{bmatrix}
            P \\
            Q
            \end{bmatrix} =
            \begin{bmatrix}
            P_2 \\
            \vdots \\
            P_N \\
            Q_2 \\
            \vdots \\
            Q_N
            \end{bmatrix}
    \quad\text{and}\quad
    f(x) =  \begin{bmatrix}
            P(x) \\
            Q(x)
            \end{bmatrix} =
            \begin{bmatrix}
            P_{2}(x) \\
            \vdots \\
            P_{N}(x) \\
            Q_{2}(x) \\
            \vdots \\
            Q_{N}(x)
            \end{bmatrix}
\end{aligned}
$$
```

Line 470:
```
$$
\begin{aligned}
        \underline{U}     =     \begin{bmatrix}
                        \underline{U}_1 \\
                        \underline{U}_2 \\
                        \vdots \\
                        \underline{U}_{N_{b}}
                    \end{bmatrix}
\end{aligned}
$$
```

To give a few examples.

Let me know if this is useful or if I'm misinterpreting you in any way. On the other hand, if you wish I can also commit a few lines to give you inspiration if that would help.

Hopefully this helps. Let me know in any case.

**Comment by furqan463** [comment]
> > @figueroa1395 the issue with vertical matrices is that you can't start bmatrix in one $$ block and end it in subsequent $$ block, with matrix rows in between in $$ blocks.
> 
> Hello @furqan463,
> 
> Vertical matrices, such as the ones found in `calculations.md` are rendered fine if placed in a single `aligned` block within a single `$$` block. This works because even thought the `bmatrix` itself isn't considered a multi-line equation, when doing things such as `x = some_matrix`, the whole expression becomes a multi-line-equation.
> 
> Here is how you can fix some of the instances where such matrices appear in the `calculations.md` file:
> 
> Starting at line 319 at the at current head of this branch:
> 
> ```
> $$
> \begin{aligned}
>     x    =  \begin{bmatrix}
>             \delta \\
>             U
>             \end{bmatrix} =
>             \begin{bmatrix}
>             \delta_2 \\
>             \vdots \\
>             \delta_N \\
>             U_2 \\
>             \vdots \\
>             U_N
>             \end{bmatrix}
>     \quad\text{and}\quad
>     y    =  \begin{bmatrix}
>             P \\
>             Q
>             \end{bmatrix} =
>             \begin{bmatrix}
>             P_2 \\
>             \vdots \\
>             P_N \\
>             Q_2 \\
>             \vdots \\
>             Q_N
>             \end{bmatrix}
>     \quad\text{and}\quad
>     f(x) =  \begin{bmatrix}
>             P(x) \\
>             Q(x)
>             \end{bmatrix} =
>             \begin{bmatrix}
>             P_{2}(x) \\
>             \vdots \\
>             P_{N}(x) \\
>             Q_{2}(x) \\
>             \vdots \\
>             Q_{N}(x)
>             \end{bmatrix}
> \end{aligned}
> $$
> ```
> 
> Line 470:
> 
> ```
> $$
> \begin{aligned}
>         \underline{U}     =     \begin{bmatrix}
>                         \underline{U}_1 \\
>                         \underline{U}_2 \\
>                         \vdots \\
>                         \underline{U}_{N_{b}}
>                     \end{bmatrix}
> \end{aligned}
> $$
> ```
> 
> To give a few examples.
> 
> Let me know if this is useful or if I'm misinterpreting you in any way. On the other hand, if you wish I can also commit a few lines to give you inspiration if that would help.
> 
> Hopefully this helps. Let me know in any case.

Got it, I'll try it out and make corrections accordingly.

**Comment by furqan463** [comment]
> the solution is to leave every single-line equation in a `$$` block alone (no `aligned`, `align*`, `eqnarray`, etc.)

@figueroa1395 for single-line equation within `$$` block `&` alignment does not work. I've experimented with `align*`, `aligned` where `&` works but equations are not aligned properly, so I did not touch those equations only fixed those where rendering problem was being experienced.

**Review by nitbharambe — APPROVED** [review]

**Review by figueroa1395 — DISMISSED** [review]
@furqan463 You have solved the rendering problems. Thank you once again for your contribution. We look forward to working with you again. 

 Edit: Fixed `markdown lint` in [d63b768](https://github.com/PowerGridModel/power-grid-model/pull/1288/commits/d63b768b1a1666936187822b16ae8bb441f7e46f). Re-approving.

**Review by figueroa1395 — APPROVED** [review]

File: docs/user_manual/components.md [hunk 2]
```diff
@@ -1036,7 +1035,7 @@ $$
 
 Current sensors with `angle_measurement_type` equal to `AngleMeasurementType.local_angle` measure the phase shift
 between the voltage and the current phasor, i.e.,
-$\text{i_angle_measured} = \text{voltage_phase} - \text{current_phase}$.
+$\begin{equation}i\_angle\_measured = voltage\_phase - current\_phase\end{equation}$.
```
mgovers: this changed the layout

```suggestion
$\begin{equation}\text{i\_angle\_measured} = \text{voltage\_phase} - \text{current\_phase}\end{equation}$.
```

also, is there a specific reason why you needed `\begin{equation}` here?

docs/user_manual/calculations.md
```diff
@@ -616,10 +616,22 @@ See also [the full mathematical workout](https://github.com/PowerGridModel/power
 
 $$
    \begin{eqnarray}
```
mgovers, line 618: any specific reason to keep `eqnarray` here instead of `aligned`?

docs/user_manual/components.md
mgovers: Equation makes sense. After experimentation locally, the correct solution is
```suggestion

$$
\text{i\_angle\_measured} = \text{voltage\_phase} - \text{current\_phase}\text{ .}
$$

```

docs/user_manual/calculations.md
mgovers: After reading the rest of the comments, I'm OK with the solution

mgovers:
Automerge enabled. Please don't edit the PR anymore unless the build fails; any additional changes can go to a separate PR.

PR #1296, opened 2026-02-02
Clarify documentation on buffer get/set value

TonyXiang8787:


power_grid_model_c/power_grid_model_c/include/power_grid_model_c/buffer.h
```diff
@@ -95,8 +96,9 @@ PGM_API void PGM_buffer_set_value(PGM_Handle* handle, PGM_MetaAttribute const* a
  * @param size The size of the buffer in terms of number of elements.
  * @param dest_stride The stride of the destination array in bytes.
  * You can set it to -1, the default stride of the size of the attribute type (like sizeof(double)).
- * If you set it to a positive number, the i-th get-value will retrieve the source data at
+ * If you set it to a positive number, the i-th get-value will write to the destination data at
```
nitbharambe, line 99: This docstring is for `PGM_buffer_get_value` function so the upper one needs changing.

TonyXiang8787: This is actually correct. In doxygen, the dostring is above the function declaration.

PR #1317, opened 2026-03-05
BugFix: Eliminate false positive for meshed observability check with voltage phasor

nitbharambe:
Our meshed observability check can not handle Voltage Phasor. In practice, this can still rise from placing pseudo measurement, in which case the added voltage phasor becomes critical for observability. This PR added a small test case and the correction to this handling. Now it will pass though the network to SE calculation directly if a Meshed network has more than 1 voltage phasor.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
@@ -662,24 +662,24 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
 
     Idx n_voltage_phasor_sensors{};
 
-    // check necessary condition for observability
+    // Check necessary condition for observability
     is_necessary_condition_met = detail::necessary_condition(observability_sensors, n_bus, n_voltage_phasor_sensors,
                                                              measured_values.has_global_angle_current());
     // Early return if necessary condition is not met
-    if (!is_necessary_condition_met) {
+    // Meshed voltage phasor sensor early out: current meshed sufficient-condition implementation cannot handle voltage
+    // phasor sensors
+    if (!is_necessary_condition_met || (n_voltage_phasor_sensors > 1 && !topo.is_radial)) {
         return ObservabilityResult{.is_observable = false,
                                    .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
     }
 
-    //  Sufficient early out, enough nodal measurement equals observable
-    //  Meshed voltage phasor sensor early out: current meshed sufficient-condition implementation cannot handle voltage
-    //  phasor sensors
-    if (observability_sensors.total_injections > n_bus - 2 || (n_voltage_phasor_sensors > 1 && !topo.is_radial)) {
+    // Sufficient early out, enough nodal measurement equals observable
+    if (observability_sensors.total_injections > n_bus - 2) {
         return ObservabilityResult{.is_observable = true,
                                    .is_possibly_ill_conditioned = observability_sensors.is_possibly_ill_conditioned};
     }
 
-    // check the sufficient condition for observability
+    // Check the sufficient condition for observability
     if (topo.is_radial) {
         is_sufficient_condition_met = detail::sufficient_condition_radial_with_voltage_phasor(
             y_bus_structure, observability_sensors, n_voltage_phasor_sensors);


power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -672,13 +672,13 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     }
 
     //  Sufficient early out, enough nodal measurement equals observable
-    if (observability_sensors.total_injections > n_bus - 2) {
+    //  Meshed voltage phasor sensor early out: currect meshed check can not handle voltage phasor
+    if (observability_sensors.total_injections > n_bus - 2 || (n_voltage_phasor_sensors > 1 && !topo.is_radial)) {
         return ObservabilityResult{.is_observable = true,
```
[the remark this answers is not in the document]
Jerry-Jinfeng-Guo: This passthrough is the solution given the lack of voltage phasor handling.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
```diff
@@ -672,13 +672,14 @@ inline ObservabilityResult observability_check(MeasuredValues<sym> const& measur
     }
 
     //  Sufficient early out, enough nodal measurement equals observable
-    if (observability_sensors.total_injections > n_bus - 2) {
+    //  Meshed voltage phasor sensor early out: current meshed sufficient-condition implementation cannot handle voltage
+    //  phasor sensors
+    if (observability_sensors.total_injections > n_bus - 2 || (n_voltage_phasor_sensors > 1 && !topo.is_radial)) {
         return ObservabilityResult{.is_observable = true,
```
TonyXiang8787, line 678: @nitbharambe @Jerry-Jinfeng-Guo 

An early out for voltage phasor sensor should have `is_observable = false`. 

This means, we are cannot sufficiently prove observability, but the we can also not sufficiently prove not-observable. So it leaves to sparse matrix error to decide. But it should not have `is_observable = true` and should not enable pivot perturbation.

On the other hand, the original early out, namely enough injection measurement, should indeed be `is_observable = true`. Because that indeed sufficiently proves observability.

figueroa1395: Would it be possible to add a test that would normally need a pivot perturbation to converge (it's observable), but because we disallow it as it has a voltage phasor sensor, then it doesn't? And mark it as failed... This test would then have to pass when the voltage phasor related stuff is fixed.

Or would that be too hard/not necessary to come up with?

Jerry-Jinfeng-Guo: I think such a test is necessary. Shall we do it in a follow-up and is more geared towards the perturbation?

TonyXiang8787 (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/observability.hpp
Jerry-Jinfeng-Guo: @figueroa1395 after some thinking, your proposal does not make sense. 
A test should only be testing one thing. In this case, whether the presence of extra voltage phasor lead to the use of perturbation due to the return state of `.is_observable` being `false`. Not anything that follows. I will make the test case on this basis.

figueroa1395: @Jerry-Jinfeng-Guo I haven't checked yet the test PR you created (I'll comment again once I do) but why wouldn't the test I propose not make sense? It's basically TDD. It's a test that should pass once we have fixed the observability including phasor voltage sensors. Why only then, because perturbation can only be triggered if `.is_observable = true`, which we don't support yet. I can see such a case appear when more than 2 voltage phasor sensors are present in a meshed grid that includes links in a configuration that would need a perturbation in order for it not to trigger `SparseMatrixError`.

Edit: Merging links into nodes probably makes my test unnecessary, as the whole idea of needed perturbation may need to be revisited.

figueroa1395: https://github.com/PowerGridModel/power-grid-model/pull/1322/changes#r2904360777

PR #1360, opened 2026-04-08
Text logger

figueroa1395:
This PR implements a text logger in the core with multithreaded capability. Unit tests are added.


figueroa1395:
Tests and probably implementation of the multithreaded text logger with flush handler are not yet in place, hence I put this as a draft. The rest can already be reviewed.

Any input is already appreciated, specially regarding the flush handler (which we don't have to expose yet, but I thought I'd be nice to add).

figueroa1395:
I have several `TODO`s that are meant for me - and the reviewers - to consider before merging. Most of the implementation should now work.

What's missing:
- Fix the benchmark and run it with the new interface
- Add an "end-to-end" test to proof the text logger actually logs everything we currently have in the main model.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -116,7 +128,11 @@ class MainModel {
     }
 
     std::unique_ptr<Impl> impl_;
-    MultiThreadedCalculationInfo info_;
+    // TODO(figueroa1395): static here should be okay because NoMultiThreadedLogger doesn't do anything, does it?
+    inline static NoMultiThreadedLogger no_logger_{};
```
mgovers, line 125: can even be constexpr

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -94,12 +103,15 @@ class MainModel {
...
+        // TODO(figueroa1395): This now becomes the caller's responsibility.
+        // logger_->clear();
         JobAdapter<Impl> adapter{std::ref(impl()), std::ref(options)};
-        return JobDispatch::batch_calculation(adapter, result_data, update_data, options.threading, info_);
+        return JobDispatch::batch_calculation(adapter, result_data, update_data, options.threading, *logger_);
     }
 
-    CalculationInfo calculation_info() const { return info_.get(); }
+    // TODO(figueroa1395): I don't think this is needed now because it is the caller's responsibility
+    // CalculationInfo calculation_info() const { return logger_.get(); }
```
mgovers: exactly

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -116,7 +128,11 @@ class MainModel {
     }
 
     std::unique_ptr<Impl> impl_;
-    MultiThreadedCalculationInfo info_;
+    // TODO(figueroa1395): static here should be okay because NoMultiThreadedLogger doesn't do anything, does it?
+    inline static NoMultiThreadedLogger no_logger_{};
+    // TODO(figueroa1395): should we have a shared ptr instead? a unique ptr? can the same logger be shared across
+    // multiple MainModel instances?
+    MultiThreadedLogger* logger_;
```
mgovers: There only needs to be a reference to the parent logger, right? and that one can be `reference_wrapper`

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
figueroa1395: It actually can't because of the underlying `std::mutex`.

figueroa1395:
Benchmark works fine locally. @mgovers could you run it in windows and WLS from this branch locally to confirm please.

Only missing thing is to add the "end-to-end" test.

Btw, since the text logger adds a timestamp already, given the fact the we use an internal timer already (in current implementation), we would have a "double" timestamp. I went with this decision because I'm unsure if everything will go via the timer in the future, and the timer itself doesn't give us a timestamp like we would like it for the text logger. This is definitely something to look at but I consider it out of scope for when we want to polish the loggers. What do you think?

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
```diff
@@ -94,13 +103,10 @@ class MainModel {
     */
     BatchParameter calculate(Options const& options, MutableDataset const& result_data,
                              ConstDataset const& update_data) {
-        info_.clear();
```
mgovers, line 97: please double-check whether the calculation info still works correctly

mgovers:
> Benchmark works fine locally. @mgovers could you run it in windows and WLS from this branch locally to confirm please.

Done. Both Windows and WSL pass

power_grid_model_c/power_grid_model/include/power_grid_model/main_model.hpp
figueroa1395: This works, but now the user is responsible for cleaning between different calculations (if desired) if the same logger is used. Hence the benchmark has been adjusted to this.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -38,6 +39,17 @@ class CalculationInfo : public Logger {
...
+        // no logging
+    }
+    template <LazyLoggingFn Fn>
+    void log(LogEvent /*tag*/, Fn&& /*fn*/) const { // NOLINT(cppcoreguidelines-missing-std-forward) // intentional
+        // no logging
+    }
+    template <LazyLoggingFn Fn>
+    void log(Fn&& /*fn*/) const { // NOLINT(cppcoreguidelines-missing-std-forward) // intentional
+        // no logging
+    }
```
mgovers:
```suggestion
    template <LazyLoggingFn Fn>
    void log(LogEvent /*tag*/, Fn&& fn) const { capturing::into_the_void(std::forward<Fn>(fn)); }
    template <LazyLoggingFn Fn>
    void log(Fn&& fn) const { capturing::into_the_void(std::forward<Fn>(fn)); }
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -38,6 +39,17 @@ class CalculationInfo : public Logger {
     }
     void log(LogEvent tag, double value) override { log_impl(tag, value); }
     void log(LogEvent tag, Idx value) override { log_impl(tag, static_cast<double>(value)); }
+    void log(std::string_view /*message*/) const {
+        // no logging
+    }
```
mgovers: nitpick format
```suggestion
    void log(std::string_view /*message*/) const { /* no logging */ }
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
```diff
@@ -44,13 +44,21 @@ enum class LogEvent : int16_t {
...
 class Logger {
   public:
     virtual void log(LogEvent tag) = 0;
     virtual void log(LogEvent tag, std::string_view message) = 0;
     virtual void log(LogEvent tag, double value) = 0;
     virtual void log(LogEvent tag, Idx value) = 0;
 
+    void log(std::string_view message) { log(LogEvent::unknown, message); }
+    template <LazyLoggingFn Fn> void log(LogEvent tag, Fn&& fn) { log(tag, std::invoke(std::forward<Fn>(fn))); }
+    template <LazyLoggingFn Fn> void log(Fn&& fn) { log(LogEvent::unknown, std::invoke(std::forward<Fn>(fn))); }
```
mgovers, line 60: can you explain why you went for `std::invoke`? not against it, just interested

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -56,6 +56,14 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
         MultiThreadedLoggerImpl* parent_;
     };
 
+    MultiThreadedLoggerImpl() = default;
+    template <typename... Args>
+        requires(sizeof...(Args) > 0 &&
+                 !std::derived_from<std::remove_cvref_t<std::tuple_element_t<0, std::tuple<Args...>>>,
+                                    MultiThreadedLoggerImpl> &&
```
mgovers, line 63: so what happens when it's of the form

```cpp
MultiThreadedLoggerImpl other;
MultiThreadedLoggerImpl current{other, some_other_arg};
```
?
does it need to be explicitly deleted?

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
```diff
@@ -0,0 +1,123 @@
...
+            data_ = std::exchange(other.data_, {});
+            flush_handler_ = std::exchange(other.flush_handler_, {});
+        }
+        return *this;
+    }
+    ~TextLogger() override {
+        // exception swallowing: we try to flush if possible. If not possible, clear data and ignore log
+        try {
+            // only flush if there is a handler and there are contents to flush
+            if (flush_handler_ && data_.rdbuf()->in_avail() > 0) { // only flush if there are contents and a handler
```
mgovers: why not use `!data_.view().empty()`? (please add a comment if there's a good reason)

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
```diff
@@ -0,0 +1,123 @@
...
+        const auto now = system_clock::now();
+        const auto sec = floor<seconds>(now);
+        const auto ms = duration_cast<milliseconds>(now - sec).count();
+
+        // Z stands for UTC time zone
+        // format example: 2026-04-25 14:00:00.000Z
+        return std::format("{:%F %T}.{:03}Z", sec, ms);
+    }
+
+    void log_impl(LogEvent tag, std::string_view message) {
```
mgovers: if you make this a templated member function, you don't need to do explicit `std::to_string`

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
```diff
@@ -0,0 +1,123 @@
...
+  private:
+    static auto timestamp() {
+        using namespace std::chrono;
+        const auto now = system_clock::now();
+        const auto sec = floor<seconds>(now);
+        const auto ms = duration_cast<milliseconds>(now - sec).count();
+
+        // Z stands for UTC time zone
+        // format example: 2026-04-25 14:00:00.000Z
+        return std::format("{:%F %T}.{:03}Z", sec, ms);
```
mgovers, line 75: isn't there a direct formatted std::chrono? or is that C++26?

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+        clear(); // if no handler, discard log
+    }
+
+    TextLogger& merge_into(TextLogger& destination) const {
+        if (&destination == this) {
+            return destination; // nothing to do
+        }
+        destination.data_ << data_.str();
```
mgovers, line 106: prevents unneeded copy
```suggestion
        destination.data_ << data_.view();
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
figueroa1395: I realize why now, because I missed that `.view()` was part of `std::stringstream`. It has a lot of member functions! Good catch!

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
figueroa1395: The copy is needed here. `destination.data_ << data_.view();` would provide a `string_view` which can dangle.

For example, consider the user provided logger and a batch calculation with multithreading enabled. If we use a view here, once the child threads are done with the calculation, they `merge_into` the main logger and then they are deleted. That would be a problem.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
figueroa1395: Good catch!

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
figueroa1395: It doesn't need to be explicitly deleted because in `MultiThreadedLoggerImpl current{other, some_other_arg}` the first argument is also of type `MultiThreadedLoggerImpl`, which the concept forbids. This way, a `MultiThreadedTextLogger` recognizes a constructor with a `FlushHandler` present, but it prevents this constructor from triggering instead of the copy constructor during overload resolution. For instance, it allows:
```c++
MultiThreadedLoggerImpl logger{flush_handler};
MultiThreadedLoggerImpl copy{logger};
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
figueroa1395: No particular reason. Doing `fn()` should suffice here as I require it to be `std::invocable`, but I guess I went with the more "generic" option.

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
figueroa1395: This is the current way as far as I know (cfr. https://en.cppreference.com/w/cpp/chrono/system_clock/formatter.html).

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
figueroa1395, line 60: @mgovers @nitbharambe Should we already make this something like
```suggestion
    template <LazyLoggingFn Fn> void log(LogEvent tag, Fn&& fn, bool invoke = false ) { 
    if (!invoke) { return; }
    log(tag, std::invoke(std::forward<Fn>(fn))); }
    template <LazyLoggingFn Fn> void log(Fn&& fn, , bool invoke = false ) { 
    if (!invoke) { return; }
    log(LogEvent::unknown, std::invoke(std::forward<Fn>(fn))); }
```

to make it truly lazy already?

nitbharambe: Whats the use case compared to not providing text logging based logger while compiling?
Is it for more fine grained control for debugging situations? Then we can name it to enable / disable.

figueroa1395: It's only truly useful if the logging action is expensive. So for the text logger is probably a good idea, and for other loggers it really depends. For the `CalculationInfo` it doesn't really matter. We can also just do this later, it doesn't need to happen now.

mgovers: i think it's fine to leave it as-is.

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
mgovers: can we `move` the data instead?

figueroa1395: I did some reading and it's not so simple unless we have a different data structure or make our own.

In short, we always need at least one copy (see for instance https://stackoverflow.com/questions/37229670/is-there-a-way-to-stdmove-stdstring-into-stdstringstream). However, we currently make two copies, one at `.str()` and another one when writing the data via `<<`.

After some digging, one way to just have one copy (the one that writes the data or `<<`) is to use the low level `rdbuf()`, but `view()` as suggested initially, actually works because `<<` makes the one copy right away, so it can not be dangling.

This means, your initial suggestion was optimal: see [3cc1d5d](https://github.com/PowerGridModel/power-grid-model/pull/1360/commits/3cc1d5dbd7d0cab8757f3f80033b8c2bf350d6bc)

mgovers (approved):

Issue #1249, opened 2026-01-05
[FEATURE] Cancel long-running calculations


mgovers:
**Describe the feature request**
If the user expects a calculation to take milliseconds but then it takes a couple minutes, then there's probably something wrong, and maybe they want to cancel the calculation.

This is especially useful for people running code in (jupyter) notebooks. A simple typo may cause the duration of a calculation to blow up significantly, especially since the introduction of multi-dimensional datasets in #1201. If a user realizes that, they may want to cancel the process as soon as they can without having to kill and restart the entire kernel.

This type of process cancellation typically is handled using `KeyboardInterrupt` (Ctrl+C events), signal handling (e.g. `SIGINT`), by setting a flag or by using stop tokens.

## Considerations

* Cfr. Python documentation (https://docs.python.org/3/library/signal.html#execution-of-python-signal-handlers):
  > A long-running calculation implemented purely in C (such as regular expression matching on a large body of text) may run uninterrupted for an arbitrary amount of time, regardless of any signals received. The Python signal handlers will be called when the calculation finishes.
  
  This has implications:
  * We need to double-check that sending an interrupt to the PGM C API as called from the same thread is possible. Note that this may be platform-dependent.
    * If not, maybe we need to run long operations (PowerGridModel constructors, calculations and (de)serializations) in a separate worker thread from Python.
  * We do not need to consider cancelability on the deepest level. It is fine to only check after every scenario whether a stop token was requested
* Also cfr. Python documentation (https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt and https://docs.python.org/3/library/signal.html#note-on-signal-handlers-and-exceptions):
  > [...] applications that are complex or require high reliability should avoid raising exceptions from signal handlers. They should also avoid catching [KeyboardInterrupt](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) as a means of gracefully shutting down. Instead, they should install their own [SIGINT](https://docs.python.org/3/library/signal.html#signal.SIGINT) handler.
  * We should follow their recommendations.
* XCode has supported `std::jthread` starting with version 26 (cfr. https://developer.apple.com/documentation/xcode-release-notes/xcode-26-release-notes). Other compilers already have supported it for much longer. This means that we can finally use it. It comes with a built-in stop token feature, which enables simplified thread cancellation handling.

## Design Proposal (requires experimentation before finalization)

* [ ] C++ core: support canceling multi-threaded calculations
[...]

PR #1361, opened 2026-04-09
Calculation termination: cancel job dispatch (not yet exposed via C API)

mgovers:
Part of #1249 
Uses the `jthread` introduced in #1358 

* Use `stop_token` to cancel calculations. Stop requests are acknowledged in several stages:
  * before prepare calculate
  * before cache calculations
  * before job dispatch
  * before scenario setup
  * before scenario calculation
  * before scenario winddown
* After every scenario setup is started, a winddown is guaranteed to be called.
  * If the scenario winddown is successful, the main model is guaranteed to be in a good state; ready to be picked up again.
  * This is even true in single scenario calculations (without copy from base model)
  * This implies that (in the future), a keyboard interrupt followed by a rerun will result in successful calculations without having to recreate the main model


tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -431,5 +474,135 @@ TEST_CASE("Test job dispatch logic") {
                 BatchCalculationError);
         }
     }
+    SUBCASE("Test cancel thread") {
+        using namespace std::literals::chrono_literals;
+
+        constexpr auto n_phases_per_scenario = 3; // setup, calculate, winddown
+
+        constexpr auto cancel_delay = 50ms; // arbitrary delay to ensure the cancel signal is sent
```
mgovers:
```suggestion
        constexpr auto n_phases_per_scenario = 3; // setup, calculate, winddown

        constexpr auto cancel_delay = std::chrono::milliseconds{50}; // arbitrary delay to ensure the cancel signal is sent
```

tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -431,5 +474,135 @@ TEST_CASE("Test job dispatch logic") {
...
+            CAPTURE(threading);
+
+            auto const n_threads = JobDispatch::n_threads(n_scenarios, threading);
+            CAPTURE(n_threads);
+
+            SUBCASE("Cancel before/during prepare") {
+                auto adapter = JobAdapterMock{counter};
+                adapter.reset_counters();
+
+                std::stop_source stop_source;
```
mgovers:
```suggestion
                std::stop_source stop_source; // NOLINT(misc-const-correctness) // false positive
```

tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -109,6 +141,16 @@ MultiThreadedLogger& no_logger() {
...
     return instance;
 }
+
+constexpr auto all_scenarios_and_phases(Idx n_scenarios) {
+    using enum CalculationPhase;
+
+    auto result = std::views::cartesian_product(IdxRange{n_scenarios}, std::vector{setup, calculate, winddown}) |
+                  std::ranges::to<std::vector<std::pair<Idx, CalculationPhase>>>();
+    result.emplace_back(-1,
+                        cache_calculate); // cache_calculate is not scenario-specific, so we use -1 as a placeholder
```
mgovers:
```suggestion
    std::vector<std::pair<Idx, CalculationPhase>> result;
    result.emplace_back(-1,
                        cache_calculate); // cache_calculate is not scenario-specific, so we use -1 as a placeholder

    for (Idx const scenario_idx : IdxRange{n_scenarios}) {
        for (CalculationPhase const phase : {setup, calculate, winddown}) {
            result.emplace_back(scenario_idx, phase);
        }
    }

    return result;
```

tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -431,5 +474,135 @@ TEST_CASE("Test job dispatch logic") {
...
+                for (auto const& [cancel_during_scenario, cancel_during_phase] :
+                     all_scenarios_and_phases(n_scenarios)) {
+                    CAPTURE(cancel_during_scenario);
+                    CAPTURE(cancel_during_phase);
+                    std::atomic<bool> cancelling_phase_started = false;
+                    std::atomic<bool> stop_requested = false;
+
+                    auto const check_in_expected_range =
+                        [n_scenarios, n_threads, cancel_scenario = cancel_during_scenario,
+                         cancel_phase = cancel_during_phase](Idx count, CalculationPhase phase) {
```
mgovers:
```suggestion
                    auto const check_in_expected_range =
                        [n_threads, cancel_scenario = cancel_during_scenario,
                         cancel_phase = cancel_during_phase](Idx count, CalculationPhase phase) {
```

tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -109,6 +141,23 @@ MultiThreadedLogger& no_logger() {
...
+                        cache_calculate); // cache_calculate is not scenario-specific, so we use -1 as a placeholder
+
+    for (Idx const scenario_idx : IdxRange{n_scenarios}) {
+        for (CalculationPhase const phase : {setup, calculate, winddown}) {
+            result.emplace_back(scenario_idx, phase);
+        }
+    }
+
+    return result;
+    return result;
```
mgovers:
```suggestion
```

tests/cpp_unit_tests/test_job_dispatch.cpp
```diff
@@ -86,16 +95,39 @@ class JobAdapterMock : public JobInterface {
...
+        ++(counter_->setup_calls);
+    }
+    void winddown_impl() {
+        maybe_await(CalculationPhase::winddown);
+        scenario_idx_ = no_scenario_idx;
+        ++(counter_->winddown_calls);
+    }
+    void maybe_await(CalculationPhase calculation_phase) const {
+        await_call_.and_then(
+            [scenario_idx = scenario_idx_, calculation_phase](AwaitCall const& func) -> std::optional<Idx> {
```
mgovers:
```suggestion
            [scenario_idx = scenario_idx_, calculation_phase](auto const& func) -> std::optional<Idx> {
```

TonyXiang8787:
@mgovers can you elaborate how this works (preferably in some in-code documentation in proper place)?

I see you pass `std::stop_token` around, but I don't see how this object is created.

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -138,20 +152,22 @@ class JobDispatch {
...
+            single_thread_job(stop_token, 0, 1, n_scenarios);
         } else {
             // create parallel threads
             std::vector<std::jthread> threads;
             threads.reserve(n_thread);
-            for (Idx thread_number = 0; thread_number < n_thread; ++thread_number) {
+            for (Idx const thread_number : IdxRange{n_thread}) {
                 // compute each single thread job with stride
-                threads.emplace_back(single_thread_job, thread_number, n_thread, n_scenarios);
+                threads.emplace_back(single_thread_job, stop_token, thread_number, n_thread, n_scenarios);
```
TonyXiang8787, line 164: @mgovers So after reading the cppreference on `std::stop_token`, I understand that the `stop_token` should be the first argument of your functor? Then when creating a `jthread`, it will automatically generate a `stop_token` to the functor.

But here you seem to explicitely pass a `stop_token`, why deviating from standard usage?

mgovers: The reason is that I am using only the getter functionality of `stop_token`. Instead, I am using a `stop_source` (from which you can obtain a stop token). That way, one stop source can trigger multiple stops in multiple threads.

In #1363, I show how this can work: Each handle has a stop source. Expensive operations (like calculations) can be offloaded to a separate thread that is awaited by the main thread. A SIGTERM (e.g.: Ctrl+C command) triggers a `KeyboardInterrupt` or related exceptions. Those are caught in the main thread that awaits the calculation thread. A stop is then requested (atomic, not necessarily lock-free but that is fine in this case, see below) and the exception is reraised. Then, the thread is awaited again as per usual (TBD)

Cfr. the Python docs on signal handlers (specifically https://docs.python.org/3/library/signal.html#note-on-signal-handlers-and-exceptions ) this is fine because we only need to exit on termination command (if we're in a Jupyter notebook, the notebook will take care of the complicated logic handling).

TonyXiang8787: @mgovers does it mean even in single thread calculation, a new thread will also be created for calculation, instead of calculating in main thread?

mgovers: Unfortunately, yes. Hence the experiment in #1363 so that we can discuss the implications.

Alternatively, we can use `async` instead of threads.

The main issue here is that Python handles everything regarding signal handling in the main thread at pseudo-random intervals. If there is no Python code, then there will also never be a call to the signal handler. Cfr. their own documentation in https://docs.python.org/3/library/signal.html#execution-of-python-signal-handlers :

> A long-running calculation implemented purely in C (such as regular expression matching on a large body of text) may run uninterrupted for an arbitrary amount of time, regardless of any signals received. The Python signal handlers will be called when the calculation finishes.

TonyXiang8787: I think creating separate thread for single-thread calculation is red flag for us.

Then we need to think about if this is a desired feature at all from business value perspective. Why do we do this in the first place. For production environment, it is never controlled/interrupted at this level, rather the whole container just gets killed from above (k8s). So I am not sure about the business value of maintaining such a complicated interruption logic in PGM.

mgovers: This is indeed mostly useful for research and experiment environments. In those settings, a typo is very easily made (especially with the cartesian product batch scenarios). I myself have encountered cases where I had to kill calculations after a simple typo that could've left my PC running for hours if I would've let it continue. If I have encountered this more than once, then I'm sure other people will as well. It is not reasonable to ask people to kill and restart their Jupyter notebook session as a consequence of a simple typo. That's what made me investigate this feature.

Maybe we can come up with a way that we can make it opt-in?

TonyXiang8787: In research and experiment environments, asking people to terminate the whole jupyter notebook for typos is very reasonable. Research environments are meant to be unstable, and can be killed at any time.

Note PGM core is mainly a production low-level HPC library. When in doubt if some feature (like supporting interuptions) belongs to the responsibility of PGM or not, we can always try to refer to `openblas` and `lapack`. Do they support such a feature/check?

mgovers: openblas has an issue from 2014 regarding this: `https://github.com/OpenMathLib/OpenBLAS/issues/378`

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -124,12 +129,15 @@ class JobDispatch {
...
-                                                                  std::move(recover_from_bad));
+            auto calculate_scenario = JobDispatch::call_with<Idx>(
+                stop_token, std::move(run), std::move(setup), std::move(winddown),
+                JobDispatch::scenario_exception_handler(exceptions), std::move(recover_from_bad));
 
             for (Idx scenario_idx = start; scenario_idx < n_scenarios; scenario_idx += stride) {
                 Timer const t_total_single{thread_log, LogEvent::total_single_calculation_in_thread};
+                if (stop_token.stop_requested()) {
+                    break;
+                }
```
figueroa1395, line 140: Why not `stop_if_requested(stop_token);` here instead?

power_grid_model_c/power_grid_model/include/power_grid_model/job_dispatch.hpp
```diff
@@ -138,20 +146,22 @@ class JobDispatch {
...
+        requires std::invocable<std::remove_cvref_t<RunSingleJobFn>, std::stop_token /* stop_token */, Idx /*start*/,
+                                Idx /*stride*/, Idx /*n_scenarios*/>
+    static void job_dispatch(std::stop_token const& stop_token, RunSingleJobFn single_thread_job, Idx n_scenarios,
+                             Idx threading) {
         // run batches sequential or parallel
         auto const n_thread = n_threads(n_scenarios, threading);
         if (n_thread == 1) {
             // run all in sequential
-            single_thread_job(0, 1, n_scenarios);
+            single_thread_job(stop_token, 0, 1, n_scenarios);
```
figueroa1395, line 157: Can you elaborate on how this would create an additional thread when running in single thread mode?

figueroa1395 (commented):
Some input/questions for later discussions.

mgovers:
Cfr. off-line discussion: Let's close this PR until we get more clarity on the user value

PR #1367, opened 2026-04-14
Docs: unsupported floating grids in asymmetric calculations

zhen0427:
Documented that 

- asymmetric power flow requires a grounded network (via Yn/Dn transformer or shunt) and
- current short circuit calculations are incorrectly implemented in the abc domain and should be handled in the sequence (0-1-2) domain, with a note added for a future feature request.


docs/user_manual/calculations.md
```diff
@@ -192,12 +196,19 @@ Output:
 - Node voltage magnitude and angle
 - Current flowing through branches and fault.
 
+```{note}
+Short-circuit calculations are currently implemented in the phase (abc) domain and therefore require a grounded network, similar to asymmetric power flow calculations.  
+A future implementation using the sequence (0-1-2) domain is expected to remove this limitation.
```
figueroa1395, line 201: Should we add in this note a "suggestion" like: grounding can happen either through a Yn/Dn transformer or by a shunt. Or is this common EE knowledge? Edit: I see this is already mentioned below. Since this block comes first, perhaps either quote with hyperlink here or just mention here and quote there?

On the other hand, should we also make a note of this in the short circuit algorithm details (i.e. https://power-grid-model.readthedocs.io/en/stable/algorithms/sc-algorithms.html), perhaps with more detail (or not)?

docs/user_manual/calculations.md
```diff
@@ -235,6 +246,17 @@ The option affects which attributes are required and how results are exposed.
...
 
+```{note}
+In power-grid-model, a network is considered grounded if at least one of the following is present:
+- A transformer with grounded neutral (e.g. `winding_from`/`winding_to` of a transformer is set to `wye_n`, `zigzag_n`)
+- A shunt connected to ground
+- Any other component that provides a conductive path to ground
+
+Asymmetric calculations require the network to have a reference to ground.  
+If no such connection exists, the calculation cannot be solved and will fail with a sparse matrix error.
+Our current data validator doesn't cover this check.
```
figueroa1395, line 257: Do we have an issue for this?

docs/user_manual/calculations.md
```diff
@@ -235,6 +246,17 @@ The option affects which attributes are required and how results are exposed.
   For asymmetric calculations voltages are given as line-to-neutral and output contains arrays with values per phase
   for all output variables.
 
+```{note}
+In power-grid-model, a network is considered grounded if at least one of the following is present:
```
figueroa1395, line 250: Maybe good to also add a note about this to the algorithm details? Not sure, but mentioning for consideration.

src/power_grid_model/_core/enum.py
```diff
@@ -106,10 +106,29 @@ class WindingType(IntEnum):
     """Transformer Winding Types"""
 
     wye = 0
+    """
+    Star connection without an explicit neutral reference.
```
figueroa1395, line 110: Can you also check and add a similar message like this to shunt, and other components that add grounding, if missing?

zhen0427: Just checked, and there's no other components have the attribute that defines grounding.

mgovers: at the very least, shunt does, right?

docs/user_manual/calculations.md
zhen0427: I'll rewrite this part based on test results.

src/power_grid_model/_core/enum.py
zhen0427: Added extra note in shunt: In PGM, shunts may also be used to introduce a ground reference in an otherwise floating grid.

docs/algorithms/pf-algorithms.md
```diff
@@ -35,6 +35,10 @@ The following bus types can be present in the system:
 - Voltage controlled bus: a bus with known $P$ and $U$.
   Note: this bus is not supported by power-grid-model yet.
 
+```{note}
+Asymmetric power flow calculations require the network to have a reference to ground. For details and internal solution of asymmetric floating grids calculations in power-grid-model, please refer to[Symmetric vs asymmetric calculations](../user_manual/calculations.md#short-circuit-calculations).
```
mgovers:
```suggestion
Asymmetric power flow calculations require the network to have a reference to ground. For details and internal solution of asymmetric floating grids calculations in power-grid-model, please refer to [Symmetric vs asymmetric calculations](../user_manual/calculations.md#short-circuit-calculations). 
```

docs/user_manual/calculations.md
```diff
@@ -192,6 +192,17 @@ Output:
 - Node voltage magnitude and angle
 - Current flowing through branches and fault.
 
+```{note}
+Short-circuit calculations are currently implemented in the phase (abc) domain and therefore require a grounded network, similar to asymmetric power flow calculations.
```
mgovers: this is true for all except for symmetric faults (ABC). For those, we short-cut and do symmetric calculations, but i do not know whether our current official support and data validator follow that assumption as well. Needs to be double-checked

docs/user_manual/calculations.md
```diff
@@ -192,6 +192,17 @@ Output:
...
+
+In power-grid-model, a network is considered grounded if at least one of the following is present:
+- A transformer with grounded neutral (e.g. `winding_from`/`winding_to` of a transformer is set to `wye_n`, `zigzag_n`).
+- A shunt connected to ground.
+- Any other component that provides a conductive path to ground.
+
+A future implementation using the sequence (0-1-2) domain is expected to remove this limitation.
```
figueroa1395: This is only for short circuit calculations, right?
```suggestion
A future implementation using the sequence (0-1-2) domain is expected to remove this limitation for short circuit calculations.
```

docs/user_manual/calculations.md
```diff
@@ -235,6 +246,12 @@ The option affects which attributes are required and how results are exposed.
   For asymmetric calculations voltages are given as line-to-neutral and output contains arrays with values per phase
   for all output variables.
 
+```{note}
+Asymmetric calculations require the network to have a reference to ground. For the definition of a grounded grid, please refer to [Short circuit calculations](calulations.md#short-circuit-calculations).
+If no such connection exists, the calculation cannot be solved and will fail with a sparse matrix error.
+Currently, power-grid-model addresses this scenario internally by adding a small admittance (implemented as a shunt) to one end of transformers to ensure the network has a reference to ground.
```
figueroa1395: This reads as if the user creates by mistake a floating grid, we can handle it anyway because we internally would add a shunt with small admittance. Can you specify when this is done?

mgovers: we only add a shunt with small admittance when a transformer is added, right? a grid with only a source and a load is still floating

docs/user_manual/calculations.md
nitbharambe: Although approaching via sequence domain can solve this problem, there would not be a future implementation. This is a design choice of PGM. Hence best to remove it.

zhen0427: Removed

docs/user_manual/calculations.md
zhen0427: Both of the support and data validator are also tested before revising the documentation.

docs/algorithms/sc-algorithms.md
```diff
@@ -21,6 +21,11 @@ $$
 This gives the initial symmetrical short circuit current ($I_k^{\prime\prime}$) for a fault.
 This quantity is then used to derive almost all further calculations of short circuit studies applications.
 
+```{note}
+Short-circuit calculations are currently implemented in the phase (abc) domain and therefore require a grounded network, similar to asymmetric power flow calculations.
+A future implementation using the sequence (0-1-2) domain is expected to remove this limitation.
```
mgovers: maybe?
```suggestion
A potential future implementation using the sequence (0-1-2) domain is expected to remove this limitation.
```

docs/user_manual/calculations.md
```diff
@@ -192,6 +192,17 @@ Output:
...
 
+```{note}
+Short-circuit calculations are currently implemented in the phase (abc) domain and therefore require a grounded network, similar to asymmetric power flow calculations.
+
+In power-grid-model, a network is considered grounded if at least one of the following is present:
+- A transformer with grounded neutral (e.g. `winding_from`/`winding_to` of a transformer is set to `wye_n`, `zigzag_n`).
+- A shunt connected to ground.
+- Any other component that provides a conductive path to ground.
+
+A future implementation using the sequence (0-1-2) domain is expected to remove this limitation for short circuit calculations.
```
mgovers: idem, optional

```suggestion
A potential future implementation using the sequence (0-1-2) domain is expected to remove this limitation for short circuit calculations.
```

mgovers (commented):
2 nitpicks, feel free to ignore them

docs/algorithms/sc-algorithms.md
zhen0427: @nitbharambe how do you think?

docs/algorithms/sc-algorithms.md
```diff
@@ -21,6 +21,11 @@ $$
 This gives the initial symmetrical short circuit current ($I_k^{\prime\prime}$) for a fault.
 This quantity is then used to derive almost all further calculations of short circuit studies applications.
 
+```{note}
+Short-circuit calculations are currently implemented in the phase (abc) domain and therefore require a grounded network, similar to asymmetric power flow calculations.
+A potential implementation using the sequence (0-1-2) domain is expected to remove this limitation.
```
figueroa1395: Would it be a good idea to create an issue (to implement using the sequence domain) and reference it here? That way if it gets any traction we can think about implementing it in the future? This because, as far as I know, it isn't in our roadmap at all.

docs/user_manual/calculations.md
```diff
@@ -235,12 +239,34 @@ The option affects which attributes are required and how results are exposed.
   For asymmetric calculations voltages are given as line-to-neutral and output contains arrays with values per phase
   for all output variables.
 
+```{note}
+In power-grid model, asymmetric calculations with certain configurations require the network to have a reference to ground. For details on how floating grids are treated in power-grid-model, please refer to[Floating grid handling](calulations.md#floating-grid-handling).
```
figueroa1395: Small typo
```suggestion
In power-grid model, asymmetric calculations with certain configurations require the network to have a reference to ground. For details on how floating grids are treated in power-grid-model, please refer to [Floating grid handling](calulations.md#floating-grid-handling).
```

docs/user_manual/calculations.md
```diff
@@ -235,12 +239,34 @@ The option affects which attributes are required and how results are exposed.
...
+
+A floating grid issue only arises in specific configurations where:
+
+- A transformer is present, and
+- All involved windings are ungrounded (no star-point grounding, no delta grounding reference, etc.), and
+- No other grounding path (shunt, source grounding, etc.) exists in the network.
+
+In this case, the system may lack a reference for certain sequence components (typically zero-sequence), leading to an ill-posed or singular system.
+
+In current power-grid-model, a shunt with small admittance is added only in transformer-related configurations where a grounding reference is missing. This shunt is connected to one side of the transformer. This is intended to ensure numerical stability in cases where the transformer topology introduces an ungrounded subsystem.
```
figueroa1395: Minor for clarity
```suggestion
Currently in power-grid-model, a shunt with small admittance is added only in transformer-related configurations where a grounding reference is missing. This shunt is connected to one side of the transformer. This is intended to ensure numerical stability in cases where the transformer topology introduces an ungrounded subsystem.
```

figueroa1395 (commented):
Very clear documentation. Just some minor things left.

docs/user_manual/calculations.md
```diff
@@ -235,12 +239,34 @@ The option affects which attributes are required and how results are exposed.
...
+- Numerical solvability in PGM
+  Whether the formulation provides enough reference to compute a unique solution, even if the physical system is not explicitly grounded.
+
+A floating grid issue only arises in specific configurations where:
+
+- A transformer is present, and
+- All involved windings are ungrounded (no star-point grounding, no delta grounding reference, etc.), and
+- No other grounding path (shunt, source grounding, etc.) exists in the network.
+
+In this case, the system may lack a reference for certain sequence components (typically zero-sequence), leading to an ill-posed or singular system.
```
nitbharambe:
```suggestion
In this case, the system may lack a reference for calculating line to ground voltages, leading to an ill-posed or singular system.
```

docs/user_manual/calculations.md
nitbharambe:
```suggestion
In current power-grid-model, a shunt with small admittance is added only in transformer-related configurations where a grounding reference is missing. This shunt is connected to one side of the transformer. This is intended to ensure numerical solvability in cases where the transformer topology introduces an ungrounded subsystem.
```

docs/algorithms/sc-algorithms.md
nitbharambe: Its not in PGM's design to have a sequence domain implementation, hence removing this statement in is the way to go. Just the first statement is alright.

docs/algorithms/sc-algorithms.md
nitbharambe:
```suggestion
Note that this limitation does not exist in the sequence (0-1-2) domain but is present in the phase domain calculation.
```

nitbharambe (commented):
One comment. Rest LGTM

TonyXiang8787 (approved):

PR #1369, opened 2026-04-14
Allow source short circuit power to be updated for batch calculation

### LINKED ISSUE #1156: [FEATURE] Batch for source mode parameters (sk, rx_ration, z01_ratio)
Is it possible to modify the library in such a way that it would be possible to change the parameters of the source mode (sk, rx_ratio, z01_ratio). This would make it more convenient to calculate the model in different network operating modes (max, min and set source modes) without creating a copy of the model with different source parameters.

TonyXiang8787:
Fixes #1156 

# Check list

* [x] Adjust component
* [x] Adjust math model
* [x] Validation test
* [x] Component documentation
* [x] Data validator (no need to adjust, logic is already there)


power_grid_model_c/power_grid_model/include/power_grid_model/component/source.hpp
```diff
@@ -117,9 +114,27 @@ class Source : public Appliance {
...
+            changed = true;
+        }
+        if (!is_nan(new_rx_ratio)) {
+            rx_ratio_ = new_rx_ratio;
+            changed = true;
+        }
+        if (!is_nan(new_z01_ratio)) {
+            z01_ratio_ = new_z01_ratio;
+            changed = true;
+        }
```
mgovers, line 135: wasn't there something like the `set_if_not_nan` function that is used in other components that is shorthand for this?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
```diff
@@ -394,6 +386,10 @@ template <symmetry_tag sym> class YBus {
              std::views::zip(math_model_param_incrmt.shunt_param_to_change, math_model_param_incrmt.shunt_param)) {
             math_model_param_.shunt_param[idx_to_change] = params;
         }
+        for (auto const& [idx_to_change, params] :
+             std::views::zip(math_model_param_incrmt.source_param_to_change, math_model_param_incrmt.source_param)) {
+            math_model_param_.source_param[idx_to_change] = params;
+        }
 
         // process and update affected entries
         update_admittance_entries(get_affected_admittance_entries(math_model_param_incrmt));
```
mgovers, line 395: this function also needs to be adjusted

mgovers (commented):
2 minor comments

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp
TonyXiang8787: No it does not. Source admittance is not part of node admittance matrix. It is only part of power flow equation (outside admittance matrix).

power_grid_model_c/power_grid_model/include/power_grid_model/component/source.hpp
TonyXiang8787: `set_if_not_nan` is meant to be used to set values if the **target** is not `nan`. Here, we set value when the **source** is not `nan`

mgovers: right... we do something like this in `shunt.hpp`:

```cpp
    template <typename T>
        requires std::same_as<T, ShuntInput> || std::same_as<T, ShuntUpdate>
    bool update_params(T shunt_params) {
        bool changed = update_param(shunt_params.g1, g1_);
        changed = update_param(shunt_params.b1, b1_) || changed;
        changed = update_param(shunt_params.g0, g0_) || changed;
        changed = update_param(shunt_params.b0, b0_) || changed;

        if (changed) {
            y1_ = (g1_ + 1.0i * b1_) / base_y_;
            y0_ = (g0_ + 1.0i * b0_) / base_y_;
        }

        return changed;
    }

    static bool update_param(double const& value, double& target) {
        if (is_nan(value) || value == target) {
            return false;
        }

        target = value;
        return true;
    }
```

and we also have `update_real_value<sym>`

but we don't have that consistent across all components. maybe a future improvement but i guess out of scope for now

mgovers (approved):

mgovers:
sorry, my bad, i thought about

> Data validator (no need to adjust, logic is already there)

and second-guessed that statement but you're right: if it is part of the update data created by `initialize_array`, then the data validator will handle the rest of the logic correctly

Issue #867, opened 2025-01-07
Revert disabling mypy call-overload


mgovers:
Revert #863 after the patches to the typing regression in NumPy 2.2.0 are fixed.

> After investigating, seems to be a NumPy typing regression and not a problem on our end. Basically, there are a lot of typing regressions in numpy 2.2.0.
>
> We will probably want to track this overview issue: https://github.com/numpy/numpy/issues/28076
> Note that mypy is also taking action in https://github.com/python/mypy/issues/18343
>
> Minimal repro case for our use case below. When that passes, we should be able to revert this PR.
> 
> ```py
> import numpy as np
> 
> # our use case
> STRUCTURED_ARR = np.array([(1, 2), (3, 4)], dtype=np.dtype([('a', int), ('b', int)]))
> STRUCTURED_ARR['a'] = 10  # mypy error
> 
> # even smaller taken from https://github.com/numpy/numpy/issues/28076
> ARR = np.arange(2)
> ARR = x + 1  # mypy error
> ```

_Originally posted by @mgovers in https://github.com/PowerGridModel/power-grid-model/issues/863#issuecomment-2574841102_

figueroa1395:
@zhen0427 @petersalemink95 this should be closed after #1374 was merged, right?

mgovers:
i'd say yes

> [@zhen0427](https://github.com/zhen0427) [@petersalemink95](https://github.com/petersalemink95) this should be closed after [#1374](https://github.com/PowerGridModel/power-grid-model/pull/1374) was merged, right?

PR #1374, opened 2026-04-16
Revert disabling mypy call-overload

zhen0427:
Revert #863 
Issue #867  (The mentioned test is done, no issue found with current mypy version)

### DIFF
--- a/src/power_grid_model/_core/power_grid_model.py
+++ b/src/power_grid_model/_core/power_grid_model.py
@@ -9,7 +9,7 @@ Main power grid model class
 import itertools
 from enum import IntEnum
 from math import prod
-from typing import Any, overload
+from typing import Any, Literal, overload
 
 import numpy as np
 
@@ -258,9 +258,9 @@ class PowerGridModel:
         output_component_types: ComponentAttributeMapping,
         calculation_type: CalculationType,
         symmetric: bool,
-        is_batch: bool,
+        is_batch: Literal[True, False],
         batch_size: int,
-    ):
+    ) -> Dataset:
         all_component_count = self._get_output_component_count(calculation_type=calculation_type)
         return create_output_data(
             output_component_types=output_component_types,
@@ -877,7 +877,7 @@ class PowerGridModel:
         output_component_types: None | set[ComponentTypeVar] | list[ComponentTypeVar] = ...,
         continue_on_batch_error: bool = ...,
         decode_error: bool = ...,
-        short_circuit_voltage_scaling: ShortCircuitVoltageScaling | str,
+        short_circuit_voltage_scaling: ShortCircuitVoltageScaling | str = ...,
     ) -> SingleRowBasedDataset: ...
     @overload
     def calculate_short_circuit(
@@ -889,7 +889,7 @@ class PowerGridModel:
         output_component_types: ComponentAttributeFilterOptions = ...,
         continue_on_batch_error: bool = ...,
         decode_error: bool = ...,
-        short_circuit_voltage_scaling: ShortCircuitVoltageScaling | str,
+        short_circuit_voltage_scaling: ShortCircuitVoltageScaling | str = ...,
     ) -> SingleColumnarOutputDataset: ...
     @overload
     def calculate_short_circuit(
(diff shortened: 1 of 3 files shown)


figueroa1395 (commented):
@zhen0427 after a bit of reading (cfr. https://github.com/python/mypy/issues/9003#issuecomment-653367674), the CI seems to be failing because of a known bug. My understanding is that `mypy` has trouble understanding the types when you have signatures with `Union`s of `Literal` types, specially if they are complex. This is my understanding of this complaint:
```
src/power_grid_model/_core/power_grid_model.py:265:16: error: No overload variant of "create_output_data" matches argument types "set[Any] | list[Any] | ComponentAttributeFilterOptions | dict[ComponentType, set[Any] | list[Any] | ComponentAttributeFilterOptions | None] | None", "DatasetType", "Any", "bool", "int"  [call-overload]
```
So for this one my recommendation would be to just silence it via (line 265)
```python
all_component_count = self._get_output_component_count(calculation_type=calculation_type)
    return create_output_data(  # type: ignore[call-overload]
...
```

For the other CI complaint:
```
tests/unit/test_power_grid_model.py:254:13: error: No overload variant of "calculate_short_circuit" of "PowerGridModel" matches argument type "str"  [call-overload]
```
I think the issue is that the `_calculate_short_circuit` definition has a default value for its argument `short_circuit_voltage_scaling`, which isn't reflected on the `@overload`s. To fix this you need to make sure the overloads have `short_circuit_voltage_scaling: some_type = ...`. Note that `= ...` is what's missing after the type hint.

These should make CI happy.

mgovers:
> @zhen0427 after a bit of reading (cfr. [python/mypy#9003 (comment)](https://github.com/python/mypy/issues/9003#issuecomment-653367674)), the CI seems to be failing because of a known bug. My understanding is that `mypy` has trouble understanding the types when you have signatures with `Union`s of `Literal` types, specially if they are complex. This is my understanding of this complaint:
> 
> ```
> src/power_grid_model/_core/power_grid_model.py:265:16: error: No overload variant of "create_output_data" matches argument types "set[Any] | list[Any] | ComponentAttributeFilterOptions | dict[ComponentType, set[Any] | list[Any] | ComponentAttributeFilterOptions | None] | None", "DatasetType", "Any", "bool", "int"  [call-overload]
> ```
> 
> So for this one my recommendation would be to just silence it via (line 265)
> 
> ```python
> all_component_count = self._get_output_component_count(calculation_type=calculation_type)
>     return create_output_data(  # type: ignore[call-overload]
> ...
> ```
> 
> For the other CI complaint:
> 
> ```
> tests/unit/test_power_grid_model.py:254:13: error: No overload variant of "calculate_short_circuit" of "PowerGridModel" matches argument type "str"  [call-overload]
> ```
> 
> I think the issue is that the `_calculate_short_circuit` definition has a default value for its argument `short_circuit_voltage_scaling`, which isn't reflected on the `@overload`s. To fix this you need to make sure the overloads have `short_circuit_voltage_scaling: some_type = ...`. Note that `= ...` is what's missing after the type hint.
> 
> These should make CI happy.

the real problem with the first one is a little bit more subtle. i've already found a fix and i'm running tests as we speak.

the second one is an actual bug in our overload statements

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -258,9 +258,9 @@ def _construct_output(
         output_component_types: ComponentAttributeMapping,
         calculation_type: CalculationType,
         symmetric: bool,
-        is_batch: bool,
+        is_batch: Literal[True, False],
```
mgovers, line 261: this is what the issue was... apparently, mypy one one hand assumes that `Literal[True] | Literal[False]` is `bool`, but on the other hand, it cannot understand that if both the overload for `Literal[True]` and the overload for `Literal[False]` are specified, that all edge cases are covered.

The underlying issue is actually very logical if we're talking templates in compiled code (e.g. C++). It's similar to this piece of code

```cpp
template <bool Value> void foo() {
    std::cout << Value ? "true" : "false" << std::endl;
}
void bar(bool val) {
    if (val) { // we need to explicitly statically make the distinction here
        foo<true>();
    } else {
        foo<false>();
    }
}
```

figueroa1395: I understood this was the issue but I didn't get the actual solution then. Nice find!

figueroa1395 (approved):

PR #1384, opened 2026-04-24
Mark node injection measured terminal type in as deprecated in Python

mgovers:
Part of #1334 .

This marks usage of the type `MeasuredTerminalType.node`, `MeasuredTerminalType["node"]`, `MeasuredTerminalType(9)` and `MeasuredTerminalType[9]` as deprecated. For example, the following are now deprecated

```py
from power_grid_model import AttributeType, DatasetType, ComponentType, initialize_array, MeasuredTerminalType

sym_power_sensor_input = initialize_array(DatasetType.input, ComponentType.sym_power_sensor, 1)
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType.node
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType["node"]
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType(9)
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType[9]

asym_power_sensor_input = initialize_array(DatasetType.input, ComponentType.asym_power_sensor, 1)
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType.node
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType["node"]
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType(9)
sym_power_sensor_input[AttributeType.measured_terminal_type] = MeasuredTerminalType[9]
```

NOTE: manually setting the power sensor input hardcodedly to `9` is still allowed. That is, this is still allowed

```py
from power_grid_model import AttributeType, DatasetType, ComponentType, initialize_array

sym_power_sensor_input = initialize_array(DatasetType.input, ComponentType.sym_power_sensor, 1)
sym_power_sensor_input[AttributeType.measured_terminal_type] = 9

asym_power_sensor_input = initialize_array(DatasetType.input, ComponentType.asym_power_sensor, 1)
asym_power_sensor_input[AttributeType.measured_terminal_type] = 9
```

This deprecation warning should be handled by a check in the C++ calculation core.

NOTE: Since this is potentially disruptive to users, the do-not-merge label was added


tests/unit/deprecated/test_deprecated_enum.py
```diff
@@ -0,0 +1,40 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+import warnings
+
+import pytest
+
+from power_grid_model.enum import MeasuredTerminalType
+
+
+def lts_measured_terminal_types():
```
figueroa1395, line 13: What does lts stands for here?

tests/unit/test_enum.py
figueroa1395, line 1: Can you test that the exception to the rule, as in assigning `1` directly and literally would still not be deprecated?

src/power_grid_model/_core/enum.py
```diff
@@ -9,11 +9,51 @@
...
+from enum import EnumMeta, IntEnum, nonmember
 
 # Value names are defined in lower case instead of upper case
 
 
+class _DeprecationAwareEnumMeta(EnumMeta):
+    """
+    Metaclass that supports deprecated enum members.
+
+    In the enum class body, define _deprecated_members as a nonmember dict
```
figueroa1395: Small suggestion because it can be confusing, or it was for me when reading it first.
```suggestion
    In the enum class body, define _deprecated_members as a non-enum-member dict
```

src/power_grid_model/_core/enum.py
```diff
@@ -9,11 +9,51 @@
...
+    mapping deprecated member names to deprecation warning messages. All access
+    patterns (Foo.bar, Foo['bar'], Foo(bar_value), Foo[bar_value]) will emit a DeprecationWarning.
+    """
+
+    def _warn_if_deprecated(cls, member):
+        msg = cls.__dict__.get("_deprecated_members", {}).get(member.name)
+        if msg:
+            warnings.warn(msg, DeprecationWarning, stacklevel=3)
```
figueroa1395, line 30: Wouldn't `FutureWarning` be better here? (cfr. https://docs.python.org/3/library/exceptions.html#FutureWarning).

figueroa1395 (commented):
>NOTE: manually setting the power sensor input hardcodedly to 9 is still allowed. That is, this is still allowed

Can this be disallowed or deprecated from the `AttributeType` point of view?

tests/unit/deprecated/test_deprecated_enum.py
mgovers: long-time support. if you have a better name, i'm happy to hear it 😬

tests/unit/test_enum.py
mgovers: i wonder, but the problem is that direct access does not go via the enum.

> Can this be disallowed or deprecated from the AttributeType point of view?

Nope, since it just directly sets raw int data into the numpy array. If we want to deprecate this, we basically have 2 options:

* [ ] Only in the C++ core (this would also enable deprecations for other users but requires the text logger to function first); or
* [ ] we would have to do a check in Python along the lines of
      ```
      for component_arr in input_data:
          if "measured_terminal_type" in component_arr:
              if np.any(component_arr["measured_terminal_type"] == MeasuredTerminalType.node:
                  mark_deprecated()
      ```
      and i'm not sure whether i like that (feels like a waste of computation effort)

src/power_grid_model/_core/enum.py
mgovers: Does that make sense to do? `nonmember` is the name that the Python built-in `enum` module uses for this specific behavior.

I can also mark it as
```suggestion
    In the enum class body, define _deprecated_members as a 'nonmember' dict
```

or
```suggestion
    In the enum class body, define _deprecated_members as a dict that is explicitly marked 'nonmember'
```

src/power_grid_model/_core/enum.py
mgovers: cfr docs:



>  exception FutureWarning
> 
>     Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.
> 

This is explicitly intended for end users of entire applications, i.e., the people **using** things that under the hood use PGM as a backend (we're not intending to address those users directly), not for the people **writing** those things that under the hood use PGM as a backend (which is our intended purpose)

> exception DeprecationWarning
> 
>     Base class for warnings about deprecated features when those warnings are intended for other Python developers.
> 
>     Ignored by the default warning filters, except in the __main__ module ([PEP 565](https://peps.python.org/pep-0565/)). Enabling the [Python Development Mode](https://docs.python.org/3/library/devmode.html#devmode) shows this warning.
> 
>     The deprecation policy is described in [PEP 387](https://peps.python.org/pep-0387/).

Ergo, this feature is intended for `other Python developers` ==> our use case

Also adding `PendingDeprecationWarning` for completeness:

> exception PendingDeprecationWarning
> 
>     Base class for warnings about features which are obsolete and expected to be deprecated in the future, but are not deprecated at the moment.
> 
>     This class is rarely used as emitting a warning about a possible upcoming deprecation is unusual, and [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) is preferred for already active deprecations.
> 
>     Ignored by the default warning filters. Enabling the [Python Development Mode](https://docs.python.org/3/library/devmode.html#devmode) shows this warning.
> 
>     The deprecation policy is described in [PEP 387](https://peps.python.org/pep-0387/).

`This class is rarely used as emitting a warning about a possible upcoming deprecation is unusual` ==> let's not go down this path

src/power_grid_model/_core/enum.py
figueroa1395: Let's go with your second option then. Stays in line with python naming and it's clearer :)

tests/unit/deprecated/test_deprecated_enum.py
figueroa1395: It's a test so it's fine. Can you add a small comment above it then just so it's clear right away?
```suggestion
# lts = long-term-support
def lts_measured_terminal_types():
```

tests/unit/test_enum.py
figueroa1395: Let's keep it out of scope for now then and consider it when introducing the deprecation in the C++ core (if valuable then)

figueroa1395 (dismissed):
Approving now, but not removing `do-not-merge` label until CI's current issues are fixed.

figueroa1395 (approved):

Issue #1238, opened 2025-12-18
[FEATURE] Drop support for macOS x64


TonyXiang8787:
Apple will drop support for macOS x64 from macOS 27 in 2026. We can then drop support of PGM for x64-based macOS.

PR #1388, opened 2026-04-28
GitHub Actions: bump macos runner images to macos-26 (arm-based)

mgovers:
`macos-26` has been in GA for a couple months now. Let's bump.

~NOTE: `macos-xx` points to ARM by default. If we want to remain on intel cores, we should explicitly specify `macos-26-intel`, but I do not see any reason to do so.~ EDIT: i apparently missed the actual change.

This also comes with a more recent Apple Clang (21, to be specific).

### DIFF
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -138,7 +138,7 @@ jobs:
           install\${{ env.PRESET }}\bin\power_grid_model_package_test; if(!$?) { Exit $LASTEXITCODE }
 
   build-cpp-test-macos:
-    runs-on: macos-15
+    runs-on: macos-26
     strategy:
       matrix:
         build-option: [debug, release]
@@ -177,7 +177,7 @@ jobs:
             os: ubuntu-24.04-arm
             cibw_build: "*_aarch64"
           - platform: macos
-            os: macos-15
+            os: macos-26
             cibw_build: "*"
           - platform: windows
             os: windows-latest
@@ -198,7 +198,7 @@ jobs:
         if: runner.os == 'Windows'
 
       - name: Set up XCode
-        if: matrix.os == 'macos-15'
+        if: matrix.os == 'macos-26'
         uses: maxim-lobanov/setup-xcode@ed7a3b1fda3918c0306d1b724322adc0b8cc0a90 # v1.7.0
         with:
           xcode-version: latest-stable
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -179,12 +179,12 @@ jobs:
           - platform: linux-aarch64
             os: ubuntu-24.04-arm
             cibw_build: "*_aarch64"
-          - platform: macos-aarch64
+          - platform: macos-arm64
             os: macos-26
-            cibw_build: "*"
+            cibw_build: "*_arm64"
           - platform: macos-x64
             os: macos-26-intel
-            cibw_build: "*"
+            cibw_build: "*_x86_64"
           - platform: windows
             os: windows-latest
             cibw_build: "*"
--- a/.github/workflows/build-test-release.yml
+++ b/.github/workflows/build-test-release.yml
@@ -138,10 +138,13 @@ jobs:
           install\${{ env.PRESET }}\bin\power_grid_model_package_test; if(!$?) { Exit $LASTEXITCODE }
 
   build-cpp-test-macos:
-    runs-on: macos-26
     strategy:
       matrix:
         build-option: [debug, release]
+        os: [macos-26, macos-26-intel]
+
+    runs-on: ${{ matrix.os }}
+  
     env:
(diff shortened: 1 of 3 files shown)


figueroa1395:
>NOTE: macos-xx points to ARM by default. If we want to remain on intel cores, we should explicitly specify macos-26-intel, but I do not see any reason to do so.

If we don't, it means we aren't testing on intel based MacOS anymore, aren't we? Is that something we want to do already? Shouldn't we, for now, test both in `macos-26` and `macOS-26-intel`? This is assuming the old `macOS-15`would test on both, but this I'm very doubtful about.

If we weren't testing on Intel already, then it's fine, can you confirm?

Finally, this PR is probably related to https://github.com/PowerGridModel/power-grid-model/issues/1238

mgovers:
> > NOTE: macos-xx points to ARM by default. If we want to remain on intel cores, we should explicitly specify macos-26-intel, but I do not see any reason to do so.
> 
> If we don't, it means we aren't testing on intel based MacOS anymore, aren't we? Is that something we want to do already? Shouldn't we, for now, test both in `macos-26` and `macOS-26-intel`? This is assuming the old `macOS-15`would test on both, but this I'm very doubtful about.
> 
> If we weren't testing on Intel already, then it's fine, can you confirm?

We were **only** testing on intel before, not on arm64. We can decide to matrix test this. Thoughts?

> Finally, this PR is probably related to #1238

:+1:

.github/workflows/check-build-reproducibility.yml
```diff
@@ -126,7 +126,7 @@ jobs:
           Write-Host "Builds are identical"
 
   check-reproducibility-c-macos:
-    runs-on: macos-latest
+    runs-on: macos-26
```
mgovers: we also do not matrix test linux-arm for build reproducibility, so i think this is fine

mgovers: changed anyways

mgovers:
reproducibility build incl. matrix macos-26 + macos-26-intel successful in https://github.com/PowerGridModel/power-grid-model/actions/runs/25099020178

.github/workflows/build-test-release.yml
```diff
@@ -176,8 +179,11 @@ jobs:
           - platform: linux-aarch64
             os: ubuntu-24.04-arm
             cibw_build: "*_aarch64"
```
figueroa1395, line 181: I see we actually build-test arm based ubuntu on python. Should we homogenize for the cpp part and include arm based for the linux cpp part as well?

figueroa1395 (dismissed):
Besides the remark about linux, LGTM.

.github/workflows/build-test-release.yml
mgovers: we can't, it's emulated for arm by CIBW. The actual github actions runner images do not support linux on arm

mgovers:
python build is not working correctly

.github/workflows/build-test-release.yml
mgovers: I now do see that we do use `ubuntu-24.04-arm` as OS using partner runner images (https://github.com/actions/partner-runner-images/blob/main/README.md). Maybe we should add it to the nightly for reproducible build as well

figueroa1395: @mgovers separate PR?

mgovers: yes, test running in https://github.com/PowerGridModel/power-grid-model/actions/runs/25106151623

mgovers: if it fails, let's not dig into it now 😬

figueroa1395 (dismissed):

figueroa1395 (approved):

.github/workflows/build-test-release.yml
```diff
@@ -168,17 +171,20 @@ jobs:
...
-          - platform: macos
-            os: macos-15
-            cibw_build: "*"
+          - platform: macos-arm64
+            os: macos-26
+            cibw_build: "*_arm64"
+          - platform: macos-x64
+            os: macos-26-intel
+            cibw_build: "*_x86_64"
```
TonyXiang8787, line 187: @mgovers why do we split this? The arm instance can run intel-based python perfectly? Spinning one additional intel mac instance seems to be not needed?

mgovers: Santiago and I had a chat about this. Basically, this more explicitly indicates that we actually do run on intel. It will also help us going forward once we drop x86 support for macos (#1238) because it is more clear to just remove the intel once.

figueroa1395: In addition, if I recall correctly, before we were only "simulating" the intel run (not testing natively) and some CI run times were a bit suspicious. This made it transparent.

mgovers: welp apparently i missed this. will create a follow-up PR to revert

PR #1394, opened 2026-05-04
PGM versioning improvement

TonyXiang8787:
We are about to bump major version. This PR prepares the code base to be able to host version inside shared object, and fix a bug on `so` versioning.

### DIFF
--- a/power_grid_model_c/power_grid_model_c/CMakeLists.txt
+++ b/power_grid_model_c/power_grid_model_c/CMakeLists.txt
@@ -31,6 +31,8 @@ file(
 
 target_link_libraries(power_grid_model_c PRIVATE power_grid_model)
 
+target_compile_definitions(power_grid_model_c PRIVATE PGM_VERSION="${PGM_VERSION}")
+
 target_sources(
     power_grid_model_c
     PUBLIC
@@ -44,7 +46,7 @@ set_target_properties(
     power_grid_model_c
     PROPERTIES
         VERSION ${PGM_VERSION}
-        SOVERSION ${PGM_VERSION}
+        SOVERSION ${PGM_VERSION_MAJOR}
         INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE
         INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO TRUE
 )
--- a/power_grid_model_c/power_grid_model_c/src/handle.cpp
+++ b/power_grid_model_c/power_grid_model_c/src/handle.cpp
@@ -18,6 +18,8 @@ using namespace power_grid_model;
 
 using power_grid_model_c::clear_error;
 using power_grid_model_c::compile_time_safe_cast;
+
+constexpr char const* version = PGM_VERSION;
 } // namespace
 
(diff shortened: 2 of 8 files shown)


power_grid_model_c/power_grid_model_c/CMakeLists.txt
```diff
@@ -44,7 +46,7 @@ set_target_properties(
     power_grid_model_c
     PROPERTIES
         VERSION ${PGM_VERSION}
-        SOVERSION ${PGM_VERSION}
```
nitbharambe, line 47: So this earlier use of SOVERSION use was incorrect / redundant right?

nitbharambe: I mean to ask: I see this is an optional property creating same library version, why did we set it in the first place?

TonyXiang8787: It was set without actual good understanding of what it means.

So in Linux/Mac, the generated shared object can have a suffix on version name, e.g., `libpower_grid_model_c.so.VERSION`. The `.VERSION` is to differentiate different versions with breaking changes. If you don't have breaking changes, you should NOT put a new version for the shared object name.

Currently in `main`, it has a unique version for all releases, like `libpower_grid_model_c.so.1.13.69`, which is not good. User will assume every new release has a breaking change in C-API.

You should only put major version into `.VERSION`. In our case, it should be `libpower_grid_model_c.so.1`, as in this PR.

nitbharambe (approved):

power_grid_model_c/power_grid_model_c/src/handle.cpp
```diff
@@ -18,6 +18,8 @@ using namespace power_grid_model;
 
 using power_grid_model_c::clear_error;
 using power_grid_model_c::compile_time_safe_cast;
+
+constexpr char const* version = PGM_VERSION;
```
figueroa1395, line 22: How does the compiler know what is `PGM_VERSION`? Does `cmake` "passes" the variable to it?

TonyXiang8787: Yes

power_grid_model_c/power_grid_model_c/src/handle.cpp
```diff
@@ -52,3 +54,4 @@ char const** PGM_batch_errors(PGM_Handle const* handle) {
     return handle_ref.batch_errs_c_str.data();
 }
 void PGM_clear_error(PGM_Handle* handle) { clear_error(handle); }
+char const* PGM_version(void) { return version; }
```
figueroa1395, line 57: What's the use case for this? If the `*.so.mayor_version` already contains the information for the C API users and you can easily see it on the python side via the wheel/uv/package manager, why do we want this explicit way of exposing it?

TonyXiang8787: This is the packaging convention. That user have some API to retrieve the full version. In C-API it is usually a function. In Python API, it is a `__version__` attribute at root package level.

PR #1397, opened 2026-05-11
use new pgm-build-deps

TonyXiang8787:
Use the new build dependencies in `pgm-build-deps`.


mgovers (approved):
the failing CI was caused by github issues (checkout action failed). not caused by this CI

PR #1370, opened 2026-04-14
Merge links on nodes: Entry point

figueroa1395:
This is the entry point for the _merging links into nodes_ algorithm plus end-to-end testing.

Some additional cleanup is done to also round this up.

PR #1398, opened 2026-05-13
Merge links on nodes: create super nodes

mgovers:
Part of #35 

This actually creates the topological nodes from the regular nodes


power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -425,7 +427,25 @@ struct ComponentTopology {
...
+    IdxVector load_gen_node_idx;
+    std::span<LoadGenType const> load_gen_type;
+    IdxVector voltage_sensor_node_idx;
+    std::span<Idx const> power_sensor_object_idx; // the index is relative to branch, source, shunt or load_gen
+    std::span<MeasuredTerminalType const> power_sensor_terminal_type;
+    std::span<Idx const> current_sensor_object_idx; // the index is relative to branch
+    std::span<MeasuredTerminalType const> current_sensor_terminal_type;
+    std::span<ComponentType const> regulator_type;
+    std::span<Idx const> regulated_object_idx; // the index is relative to branch or branch3
+    std::span<ComponentType const> regulated_object_type;
```
mgovers, line 448: the `span` ones essentially do not change when we go from user nodes to topological nodes, so we can do with a pure view. The branch2, branch3 and appliance ones may change so they need a pass-by-copy.

We can also opt for a `maybe_owning_view`, which can either take by reference, or by value, depending on the situation. I'm investigating that on the side

power_grid_model_c/power_grid_model/include/power_grid_model/common/counting_iterator.hpp
```diff
@@ -29,4 +29,17 @@ constexpr auto enumerate(std::ranges::viewable_range auto&& range_to_enumerate)
...
+#else
+constexpr auto pairwise(std::ranges::viewable_range auto&& range_to_pair) {
+    auto const begin = std::ranges::begin(range_to_pair);
+    auto const end = std::ranges::end(range_to_pair);
+    if (begin == end) {
+        return std::views::zip(std::ranges::subrange(begin, begin), std::ranges::subrange(begin, begin));
+    }
+    return std::views::zip(std::ranges::subrange(begin, end - 1), std::ranges::subrange(begin + 1, end));
+}
+#endif
```
mgovers: in the end, I did not use it in this PR but initially i thought i needed it. I've had the same thing in the past already (multiple times) but removed it every single time, so I was like "ok, might as well keep it this time"

nitbharambe: I think its a handy function. I see we can already use it in `sparse_decode` in grouped_idx_vector maybe. Uses similar  logic.

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+#include <span>
+#include <utility>
+#include <vector>
+
+namespace power_grid_model::supernodes {
+class TopologicalNodeMapping {
+  public:
+    TopologicalNodeMapping() = default;
+    explicit TopologicalNodeMapping(Idx n_topo_nodes_, IdxVector mapping)
+        : mapping_{std::move(mapping)}, n_topo_nodes_{n_topo_nodes_} {
```
nitbharambe:
```suggestion
    explicit TopologicalNodeMapping(Idx n_topo_nodes, IdxVector mapping)
        : mapping_{std::move(mapping)}, n_topo_nodes_{n_topo_nodes} {
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/counting_iterator.hpp
figueroa1395: It's `constexpr` and we shall fully have it when c++23 is fully supported + it's handy. Let's keep it.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -425,7 +427,25 @@ struct ComponentTopology {
     IdxVector regulated_object_idx; // the index is relative to branch or branch3
     std::vector<ComponentType> regulated_object_type;
 
-    Idx n_node_total() const { return n_node + static_cast<Idx>(branch3_node_idx.size()); }
+    constexpr Idx n_node_total() const { return n_node + std::ssize(branch3_node_idx); }
+};
+
+struct ReducedComponentTopology {
+    Idx n_node{}; // num of topological nodes (with reduced links), plus internal nodes for 3-way branches
```
figueroa1395: The "(with reduced links)" part makes this confusing. It should be all topological nodes, which includes user nodes not connected by links and all supernodes, right?

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -436,6 +456,7 @@ using Branch3Connected = std::array<IntS, 3>;
 struct ComponentConnections {
     std::vector<BranchConnected> branch_connected;
     std::vector<Branch3Connected> branch3_connected;
+    std::vector<BranchConnected> link_connected;
```
figueroa1395, line 459: For this `struct` and for `ComponentTopology`, let's add a short comment on why we have split links from branches - for future reference.

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+        check_sanity();
+    }
+
+    auto mapping() const& -> IdxVector const& { return mapping_; }
+    auto mapping() && -> IdxVector { return std::move(mapping_); }
```
figueroa1395, line 37: Maybe I'll find out later, but why not return a `span`? Why explicitly define the move away option?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+#include <iterator>
+#include <ranges>
+#include <span>
+#include <utility>
+#include <vector>
+
+namespace power_grid_model::supernodes {
+class TopologicalNodeMapping {
+  public:
+    TopologicalNodeMapping() = default;
```
figueroa1395: Why the default constructor when below you have an `explicit` one and there is no way to later "populate" the class? I'd argue having a default one makes it easier for us to introduce a bug by mistake. Thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+    IdxVector user_nodes;
+    std::vector<BranchIdx> user_links;
+
+    constexpr auto is_supernode() const noexcept -> bool { return user_nodes.size() > 1 && !user_links.empty(); }
+};
+
+struct ComponentToTopoNodeCoupling {
+    Idx n_topo_nodes{};
+    std::vector<Idx2D> user_nodes_to_topo_nodes;
+    std::vector<Idx2D> user_links_to_topo_nodes;
```
figueroa1395, line 64: Can you ellaborate on this one? It may be good to have some in-code comments for this logic since this was very tricky to design.

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+struct TopologicalNodesAndCoupling {
+    std::vector<TopologicalNode> topo_nodes;
+    ComponentToTopoNodeCoupling coupling;
+};
+
+struct ReducedTopology {
+    ReducedComponentTopology reduced_comp_topo;
+    TopologicalNodesAndCoupling topo_node_coup;
+};
```
figueroa1395, line 75: So at least for now, "the_map" won't live in the `ReducedTopology`? Will we be able to use it later in the output processing? Or if needed there we can add here?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+                          });
+
+    auto const global_graph_ =
+        GlobalGraph{boost::edges_are_unsorted_multi_pass, internal_edges.cbegin(), internal_edges.cend(), n_nodes};
+
+    Idx const num_connected_components = boost::connected_components(global_graph_, vertices.data());
+    return TopologicalNodeMapping{num_connected_components, std::move(vertices)};
+}
+
+// Union-find algorithm:
```
figueroa1395: Rename to the correct algorithm, it's no longer a union-find algorithm (although equivalent). The in-comment haves the naive approaches (which are fine to stay), but maybe it is good to also summarize the chosen solution.

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+// NAIVE (WRONG): [0, 1, 1, 1, 4, 2]  (2 should be super node)
+//   Fixing this retroactively is O(N_links^2)
+//   Unless we use counting sort, which is O(N_nodes + N_links)
+// Result:
+//   [0, 1, 1, 1, 4, 1]
+// Mapping (explanation):
+//   [N0, SN1, SN1, SN1, N4, SN1]
+//   [TN0, TN1, TN1, TN1, TN2, TN1]
+inline TopologicalNodeMapping create_map(ComponentTopology const& comp_topo, ComponentConnections const& comp_conn) {
+    return find_link_connected_components(comp_topo.n_node, comp_topo.link_node_idx, comp_conn.link_connected);
```
figueroa1395, line 136: I don't think we need this intermediate function (`find_link_connected_components`), do we?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+    // sparse directed graph
+    // edge i -> j
+    using GlobalGraph = boost::compressed_sparse_row_graph<boost::bidirectionalS, GlobalVertex, GlobalEdge,
+                                                           boost::no_property, Idx, Idx>;
+
+    IdxVector vertices = IdxRange{n_nodes} | std::ranges::to<IdxVector>();
+
+    std::vector<std::pair<Idx, Idx>> internal_edges;
+    internal_edges.reserve(2 * edges.size());
```
figueroa1395, line 92: Why pre allocating if it's a vector and supernode structure is expected to be small?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+    constexpr Idx n_topo_nodes() const { return n_topo_nodes_; }
+    constexpr Idx n_user_nodes() const { return std::ssize(mapping_); }
+
+  private:
+    IdxVector mapping_;
+    Idx n_topo_nodes_{};
+
+    constexpr void check_sanity() const noexcept {
+        // the mapping should be a valid partition of user nodes into topological nodes
+        assert(n_topo_nodes_ == 0 || (n_topo_nodes_ == std::ranges::max(mapping_) + 1));
```
figueroa1395: Is `std::ranges::max(mapping_)` if `mapping_` is empty well defined?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+namespace detail {
+
+inline TopologicalNodeMapping find_link_connected_components(Idx n_nodes, std::vector<BranchIdx> const& edges,
+                                                             std::vector<BranchConnected> const& edge_connections) {
+    struct GlobalEdge {};
+    struct GlobalVertex {};
+
+    // sparse directed graph
+    // edge i -> j
+    using GlobalGraph = boost::compressed_sparse_row_graph<boost::bidirectionalS, GlobalVertex, GlobalEdge,
```
figueroa1395, line 86: I think that using `boost::undirectedS` instead of `boost::bidirectionalS` may be cleaner. Then you can change it below from
```c++
internal_edges.emplace_back(from, to);
internal_edges.emplace_back(to, from);
```
to
```c++
internal_edges.emplace_back(from, to);
```
and allocate less space for `internal_edges`.

Or is there something I'm missing?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+    internal_edges.reserve(2 * edges.size());
+    std::ranges::for_each(std::views::zip(edges, edge_connections),
+                          [&internal_edges](auto const& vertices_and_connectivity) {
+                              if (BranchConnected const& connectivity = std::get<1>(vertices_and_connectivity);
+                                  connectivity[0] == 0 || connectivity[1] == 0) {
+                                  return; // only add edge if both sides are connected
+                              }
+                              BranchIdx const& vertices = std::get<0>(vertices_and_connectivity);
+                              auto const from = vertices[0];
+                              auto const to = vertices[1];
```
figueroa1395, line 101: Maybe a sanity check like 
```c++
assert(from < n_nodes && to < n_nodes);
```
for when we split links from branches in `main_core/topology.hpp`?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+                                       .coupling = {.n_topo_nodes = topo_node_mapping.n_topo_nodes(),
+                                                    .user_nodes_to_topo_nodes = std::move(user_node_topo_node_coup),
+                                                    .user_links_to_topo_nodes = std::move(user_link_topo_node_coup)}};
+};
+
+// Use the topo_node_mapping to remap every part of comp_topo such that the output nodes are the topological nodes;
+// not the user nodes. This effectively removes all link-related stuff from the math topology.
+inline ReducedComponentTopology construct_reduced_topology(ComponentTopology const& comp_topo,
+                                                           TopologicalNodeMapping const& topo_node_mapping) {
+    auto remap_nodes = [&topo_node_mapping]<typename IdxType>(std::vector<IdxType> const& user_nodes) {
```
figueroa1395, line 197: Clever stuff, nicely done!

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+                                                  std::ranges::to<std::vector>();
+
+    std::vector<Idx2D> user_link_topo_node_coup =
+        enumerate(std::views::zip(comp_topo.link_node_idx, comp_conn.link_connected)) |
+        std::views::transform([&node_mapping, &topo_nodes](auto const& idx_link_and_connectivity) {
+            auto const& [link_idx, link_nodes_and_connectivity] = idx_link_and_connectivity;
+            auto const& [link_nodes, link_connected] = link_nodes_and_connectivity;
+            if (link_connected[0] == 0 || link_connected[1] == 0) {
+                return Idx2D{.group = disconnected, .pos = disconnected};
+            }
```
figueroa1395: I'm thinking about this edge case:
If we have a link that's disconnected on one side, but it's not part of any other super node, and its connected side is then connected to a node to an active line, what will happen to that link? It clearly isn't part of a supernode so it will be excluded from the link solver algorithm, but it will also be excluded from the regular topology. What happens then? Is this expected? We still have to have output for it on the connected side.

tests/cpp_unit_tests/test_supernodes.cpp
```diff
@@ -0,0 +1,162 @@
...
+                .link_connected = {{1, 1}},
+            };
+            auto const topo_node_mapping = detail::create_map(comp_topo, comp_conn);
+            CHECK(topo_node_mapping.n_topo_nodes() == 2);
+            REQUIRE(std::ranges::max(topo_node_mapping.mapping()) + 1 == topo_node_mapping.n_topo_nodes());
+            REQUIRE(topo_node_mapping.n_user_nodes() == topo_node_mapping.mapping().size());
+            REQUIRE(topo_node_mapping.n_user_nodes() == comp_topo.n_node);
+            CHECK(topo_node_mapping.mapping() == IdxVector{0, 0, 1});
+        }
+        SUBCASE("Disconnected link => not remapped") {
```
figueroa1395, line 53: The mentioned edge case in my previous comment is similar to this, however, can you add it here as well too? Just to make sure the mapping handles it well.

tests/cpp_unit_tests/test_supernodes.cpp
```diff
@@ -0,0 +1,162 @@
...
+                    std::views::transform([](TopologicalNode const& node) -> auto& { return node.user_links; }),
+                std::vector<std::vector<BranchIdx>>{{}, {}, {{2, 4}}, {}, {}}));
+        }
+    }
+    SUBCASE("construct_reduced_topology") {
+        // TODO: Add test implementation
+    }
+    SUBCASE("reduce_topology") {
+        // TODO: Add test implementation
+    }
```
figueroa1395, line 297: Reminder to add these :)

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -0,0 +1,216 @@
...
+                                                            ComponentConnections const& comp_conn,
+                                                            TopologicalNodeMapping const& topo_node_mapping) {
+    auto const& node_mapping = topo_node_mapping.mapping();
+
+    std::vector<TopologicalNode> topo_nodes(topo_node_mapping.n_topo_nodes());
+    std::vector<Idx2D> user_node_topo_node_coup = enumerate(node_mapping) |
+                                                  std::views::transform([&topo_nodes](auto const& idx_and_topo) {
+                                                      auto const& [user_node, topo_node] = idx_and_topo;
+                                                      topo_nodes[topo_node].user_nodes.push_back(user_node);
+                                                      return Idx2D{.group = topo_node, .pos = user_node};
```
nitbharambe, line 149: Maybe we can already start with separate alias for same type. 

I already might be getting confused with the terminology of user / topo / supernodes (And then maybe we need to rename or define the terms in comments.)

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -412,6 +413,7 @@ struct ComponentTopology {
     Idx n_node{};
     std::vector<BranchIdx> branch_node_idx;
     std::vector<Branch3Idx> branch3_node_idx;
+    std::vector<BranchIdx> link_node_idx;
```
nitbharambe, line 416: This comment is intended for future refactor:

cfr. https://github.com/PowerGridModel/power-grid-model/pull/1398/changes#r3272590347

Lets define what index means. 
PGM maintainers can understand that these links are ordered on the basis of their order in container. But maybe not new contributors.

I say this because now there might be overlap in content between branch and link.

nitbharambe:
no new comments from me @mgovers

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers:
```suggestion
    Idx n_node{}; // num of topological nodes, including internal nodes for 3-way branches
```

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: lifetime: the mapping object is a temporary object in the design proposal. I tried to follow the design

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: generally, it's good to have default-constructible objects (and it's also what i used during the original stub implementation) but you're right, this is no longer necessary. will remove

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: `"the map"` is a sparse mapping and does not make the info needed to get the index accessible. instead, i create 2 separate mappings in the topo_node_coup, which have that `"the map"` decoded in a way that can be used directly by not only nodes but also links

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: indeed but i don't really mind it either IMO (happy to remove it if you disagree tho)

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: why **not** preallocate?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: nope, undefined. good catch!

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: i tried that initially but it seems like the `connected_components` algorithm expects a directed graph when using the compressed sparse graph representation...

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: i don't want to provide `n_nodes` into the lambda captures by default, but i'll add a separate line that does this assertion

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: it should not be loaded. partially connected branches do contribute to the diagonal elements of the Ybus, but after this whole merging-links-on-nodes ordeal is finished, one of the key assumptions is that links have vanishing impedance and reactance relative to the rest of the grid, so they should get null output.

However, you're right in that that differs from the existing output, in which we assume fixed finite impedance.

@nitbharambe and/or @TonyXiang8787 can you confirm that this is indeed correct?

tests/cpp_unit_tests/test_supernodes.cpp
mgovers: that one is tested and handled correctly, right?

and later on in `SUBCASE("Disconnected link => not included in user links")` it is explicitly tested that they are disconnected in `create_topological_nodes`

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: do you have any specific suggestions?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
TonyXiang8787: Single-side connected links should not participate on calculations. For the output, if the node at the connected side is energized, the link should also have the output `energized=1`. Otherwise `energized=0`. All other quantities (power, current, etc) should be zero.

This is exactly the same behavior as current links.

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: as discussed: let's scope it out of this whole merging links epic.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: scoped out until the merging links is actually fully released, as in the meantime, branches actually still include links (only after GA will this no longer be the case)

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: will rename `create_map` to `find_link_connected_components`

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: comments added

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: cfr. discussion: it's probably not the right location to make such a comment: either we do make the distinction or we don't (feature flag). adding an explicit comment on **why** we split up would defeat the purpose of a feature flag

AnkurArohi:
Hey @mgovers I would like to understand the concept of super node what is it mathematically are you simply merging the switches or links with low impedance or changing something in jacobian

mgovers:
> Hey @mgovers I would like to understand the concept of super node what is it mathematically are you simply merging the switches or links with low impedance or changing something in jacobian

the intention is to merge the nodes that are connected by switches and links into one super node, by assuming that links and switches have infinitesimal impedance, and therefore vanishing when compared against normal impedances.

this effectively means that we're splitting up the calculations into a calculation at macro scale (basically the net grid) and separately one at micro scale (within each such supernode). the jacobian then effectively changes structure as a result, because we have less nodes in the net grid. in particular, the unstable sections of the jacobian disappear from the equations.

for the micro-grid within a supernode, we make the same assumption that we currently make in `mean` that all links have the same impedance.

AnkurArohi:
> > Hey @mgovers I would like to understand the concept of super node what is it mathematically are you simply merging the switches or links with low impedance or changing something in jacobian
> 
> the intention is to merge the nodes that are connected by switches and links into one super node, by assuming that links and switches have infinitesimal impedance, and therefore vanishing when compared against normal impedances.
> 
> this effectively means that we're splitting up the calculations into a calculation at macro scale (basically the net grid) and separately one at micro scale (within each such supernode). the jacobian then effectively changes structure as a result, because we have less nodes in the net grid. in particular, the unstable sections of the jacobian disappear from the equations.
> 
> for the micro-grid within a supernode, we make the same assumption that we currently make in `mean` that all links have the same impedance.

are you trying to say that you are still using the impedance of the switch and link rather than completely ignoring it and in order to do so you are creating another micro jacobian and using mean of all impedances in that switch or line?

To this end I would like to point out that this mean value will still be quite small and can very well lead to instabilities, I have some suggestions

- Model it as ignored impedance and later provide virtual injections to cater the voltage drop, for state estimation it is important to respect the measurements which may be associated with these units on both sides so modelling it as virtual injection which then takes care of voltage drop is elegant way to handle this 
- Alternatively instead of taking mean you can take the max value (it shouldnt be very different)

mgovers:
> are you trying to say that you are still using the impedance of the switch and link rather than completely ignoring it and in order to do so you are creating another micro jacobian and using mean of all impedances in that switch or line?
> 
> To this end I would like to point out that this mean value will still be quite small and can very well lead to instabilities

1. In the large (macro-level) jacobian, any contribution of links will no longer exist
2. On the micro-level jacobian, there are only links, and they are therefore assigned the same constant impedance, or equivalently, only `1`'s (which is what we do in reality), since the system of equations and its solution are equivalent up to that same constant scale factor. This is well-defined (and the algorithm is already implemented in #1370 and related)
3. We then use the macro-level PF/SE solution as known injections to the PF equation on the micro-level grid to distribute the active and reactive power flows through the switches/links

NOTE: we do not model lines with low impedance in this way. lines with low impedance are still treated the same as the current PGM does.

> I have some suggestions
> 
>     * Model it as ignored impedance and later provide virtual injections to cater the voltage drop, for state estimation it is important to respect the measurements which may be associated with these units on both sides so modelling it as virtual injection which then takes care of voltage drop is elegant way to handle this

Can you elaborate? (or maybe @nitbharambe you can jump in here?) Our current intention is to completely neglect the voltage drop over links while aggregating any voltage measurements on the affected nodes (Kalman filter). We also have not supported power and current sensors on links from the start.

>     * Alternatively instead of taking mean you can take the max value (it shouldnt be very different)

This is i guess similar to what we do

AnkurArohi:
@mgovers I understand now , so using the micro logic is not solving the problem of low impedance on links, switches, lines, but thats ok because this is primarily an issue for the transmission state estimation.
In this case, instead of complexing the logic with micro part, it would be fine to assume the fixed 1 value by another realistic one directly.

mgovers:
Exactly.

Maybe out of personal interest, can you explain in what way transmission grids differ from distribution grids here?

nitbharambe:
> to cater the voltage drop,

Voltage drop between?

Maybe the term "impedance" is being misrepresented. Summarizing whole feature:

We basically dissolve links into a single "supernode". Then perform usual PF or SE only on the higher level grid without any link components. This leaves us with voltages and injections from/to normal nodes and the supernodes.
Then within a supernode, we assume power flows equally through all links. Based on this micro-level topology, we redistribute the injection flows to these links. The voltage result of all nodes contained within this supernode is given out equal.

Hence we do away with the old way of assuming link with low impedance and resolve the stability issue.

AnkurArohi:
> Exactly.
> 
> Maybe out of personal interest, can you explain in what way transmission grids differ from distribution grids here?

Absolutely, so in distribution grids the impedance has a lower X/R ratio in comparision to the transmission grid and is usually also less metered at switches or bus couplers.
Hence this usecase has less relevance. BUT its an industry problem to solve the cases when there is a measurement at bus coupler which state estimation should use, this improves the observability (2*no._of_meas -1) but since the switches have low or very very low impedance the solvers model it as a diect conneciton to eradicate inversion and Jacobian issues (this is what you are also solving but , using micro node you are simply merging it and provising assumptive impedance)

AnkurArohi:
> > to cater the voltage drop,
> 
> Voltage drop between?
> 
> Maybe the term "impedance" is being misrepresented. Summarizing whole feature:
> 
> We basically dissolve links into a single "supernode". Then perform usual PF or SE only on the higher level grid without any link components. This leaves us with voltages and injections from/to normal nodes and the supernodes. Then within a supernode, we assume power flows equally through all links. Based on this micro-level topology, we redistribute the injection flows to these links. The voltage result of all nodes contained within this supernode is given out equal.
> 
> Hence we do away with the old way of assuming link with low impedance and resolve the stability issue.

voltage drop on the ends of switches is always there in reality, switch and links should have drops even if very very very low.
whats the definition of link in here? Why cant we simply merge the nodes and run one power flow or SE

mgovers:
@TonyXiang8787 i think your input is also valuable here.

TonyXiang8787:
Hi @AnkurArohi,

Thanks for your question and fruitful discussion. While many points have been addressed by @nitbharambe  and @mgovers, I try to elaborate a bit at conceptual level.

# Purpose of link in PGM

Link in PGM is a conceptual component. It allows user to create a node-to-node interconnection where the user wish ignore the impedance and voltage drop. Realistically, this basically means switches, inter-bus couplings, or any short conductors. If user wish to model these interconnection and ignore the impedance/voltage drop, they should use`link`.

On the other hand, if user does want to model the (real measured) impedance of a switch (or bus coupling, or similar), they should model it as a `line`. Line itself, is also a conceptual component. It means anything with a defined impedance.

# Current implementation of link in PGM

In current internal implementation, the link is just modeled as a super small impedance and participate PF, SE, and SC calculations as a regular line. This is less ideal, but works in many cases. Sometimes it results in numerical instability due to the super small impedance, thus super large admittance.

However, if you try to put power/current sensor in links for SE, the calculation will not converge due to ill-conditioned system. So we explicitly forbid sensors on links. This is a limitation in PGM. And yes, for your TSO related use cases, where measurements are on bus-coupling, or switches, those measurements cannot be properly represented in PGM due to this limitation.

# Proposed new feature: merging nodes into super node

So the new feature #35 is about to merge nodes which are connected by links, into super node. And links disappear entirely in the actual PF, SE, and SC calculations. The mathematical core does not see links at all. After the calculation, we still need to put some results into links, that's the micro-grid part @mgovers and @nitbharambe are referring. It is merely a post processing to distribute power/current into links.

NOTE: since links are removed entirely in mathematical core, there can still be no sensors on links. This means your TSO use-case (measurement on bus-coupling) will still not be calculated by PGM.

# Further future feature: sensors on links

After #35 is done, we will further think about if and how it is possible to model sensors on links for SE. My gut feeling is that this should still be possible. But that require a long-term thinking and more mathematical consideration. This is a long-term ambition which will only be considered after #35.

AnkurArohi:
@TonyXiang8787 that makes my guess clear this here is a post processing result handling part and also the concept of using super node does not solve the inherent problem of impedance less or extremely low impedance based lines and switches with measurements.
I am not demeaning the work here but just want to see and understand what the concept is ,gap is and what the usability is

The industry needs those kind of ideas and mathematical solution which would handle these scenarios with precision and in case that we want to say PGM is also the tool for TSO or mixed DSO TSO networks. (we talked about this shortly at the summit)

Great work
I have ideas and i think also a solution PR for the cases of measurements on switches with low impedance and its handling for SE and PF without illconditioning 

happy to share if needed

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
figueroa1395: My intention was more geared towards: if a new contributor is reading the code and stumbles upon this, the person would find it initially strange that we split links from branches. So my idea of a comment was to clarify that it's split because we don't have links into the mathematical core, but we rather handle them in post-processing with the link solver.

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
figueroa1395: @mgovers In short, this should be handled then only at the output post-processing stage, right? If so, can you add a todo so we don't forget?

Edit: I see you handled it in 4b697c61b3a0e7659735928341b58422524f3f87

tests/cpp_unit_tests/test_supernodes.cpp
```diff
@@ -153,10 +153,113 @@ TEST_CASE("Test Supernodes") {
         }
     }
     SUBCASE("construct_reduced_topology") {
```
figueroa1395, line 157: Can you add some additional test cases?
- A minimal case without any branches nor links, basically a single node with a source.
- A simple case without branches and only links.
- A complex case resembling a circle with a node in between where the circle is made up of links and branches and the center node is connected crazily to the nodes making up the circle with branches and links. This to stress test the algorithm in a similar way we did in https://github.com/PowerGridModel/power-grid-model/blob/9f117a0b6aef1dd3d1b90a2481ee422165e18501/tests/cpp_unit_tests/test_link_solver.cpp#L941
- A simple case where a link is disconnected on one side.

figueroa1395 (commented):
Looks good! I would like a few additional test cases, but it's ready to be merged.

tests/cpp_unit_tests/test_supernodes.cpp
mgovers: > A simple case where a link is disconnected on one side.

this is not relevant for this function (there's no `comp_conn` and everything is handled in the `TopologicalNodeMapping`)

I've added test cases for the rest

figueroa1395 (dismissed):
Nothing more from my side, looks good!

TonyXiang8787:
> @TonyXiang8787 that makes my guess clear this here is a post processing result handling part and also the concept of using super node does not solve the inherent problem of impedance less or extremely low impedance based lines and switches with measurements. I am not demeaning the work here but just want to see and understand what the concept is ,gap is and what the usability is
> 
> The industry needs those kind of ideas and mathematical solution which would handle these scenarios with precision and in case that we want to say PGM is also the tool for TSO or mixed DSO TSO networks. (we talked about this shortly at the summit)
> 
> Great work I have ideas and i think also a solution PR for the cases of measurements on switches with low impedance and its handling for SE and PF without illconditioning
> 
> happy to share if needed

Hi @AnkurArohi,

Thanks for your reply. Can you also make a feature request issue for allow sensors on links? This would be follow-up to the #35. Please also elaborate the use-cases.

mgovers:
do-not-merge label added so that we wait until Monday to merge

figueroa1395 (approved):
No additional changes have been made. It's ready for merging.

PR #1401, opened 2026-05-18
Improve and reduce usage of universal reference and forward

TonyXiang8787:
# Background

This PR fixes a code quality problem. In the code base, universal/forwarding references and the corresponding `std::forward` are excessively used. Most of the usage is not justified. Especially, the usage of `std::forward` are merely to satisfy the `clang-tidy` and `sonar` rules. This PR removes those usage and adjusts the rules.

# Functor object as function argument should be passed by value

One misuse of universal reference is wrongly declaring functor object in function argument as universal reference. See below:

```c++
template <class Functor, class... Args>
decltype(auto) call_functor(Functor&& f, Args&&... args) {
    return std::forward<Functor>(f)(std::forward<Args>(args)...);
}
```

The [CPP guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#clambdas-function-objects-and-lambdas) is clear: Function objects should be cheap to copy (and therefore [passed by value](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#rf-in)).

Defining functor argument as reference is not correct. The PR changes this back to pass by value. See below. Note the declaring universal reference for `args` and use `std::forward` on `args` is correct, which is arguably the only correct usage of `std::forward` in PGM code base.

```c++
template <class... Args>
decltype(auto) call_functor(functor_c auto f, Args&&... args) {
    return f(std::forward<Args>(args)...);
}
```

## Requirements on PGM functors

Oppositions of this change may argue that in general C++ programming, we have indeed sometimes legitimate usage of universal references for functors. Thus this PR proposes to strengthen the requirements of custom defined functors in PGM code base as follows.

* The functor class should be cheap to copy.
* The functor class should not have rvalue overload for `operator()`. See [this comment](https://github.com/PowerGridModel/power-grid-model/pull/1401#issuecomment-4668871115) for detail.

We define our own functor concept:

https://github.com/PowerGridModel/power-grid-model/blob/1babbae70bc7f2ec76f44a4f206d88169f634bf4/power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp#L143-L146

# Over-generalization of interface using range concept based template

In the current code base, range concept based templates are extensively used, where in many places the usage is not justified. For example, in the following code segment, the `IndPtr` is defined as range concept based template.

https://github.com/PowerGridModel/power-grid-model/blob/d1874c87a33fc704ca9166ae69a8af2156dc5401/power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp#L51-L58

However, in the entire PGM code base, including tests, there is no single caller of this function using a non-vector argument. Over-generalization of this argument to a range concept based template, and thus the usage of univeral reference and `std::forward`, is not justified. We should just change it back to simple vector.

https://github.com/PowerGridModel/power-grid-model/blob/fc3800ea0fd57faf2f62057b3fdaa6611e5e96c2/power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp#L51-L57

## Order of consideration for argument

We should strongly prefer simple argument type if possible, instead of over-generalization. The order for range like argument should be as follows:

* `std::vector<T> const&`, or `std::vector<T>&`
* `std::span<const T>`, or `std::span<T>`. Note this is pass by value.
* Only if the above two cannot satisfy the need of code base, looking for more advanced range concept based templates. 

## Decision for specific cases

For `sparse_encode` and `sparse_decode`, we can just use the view option as described in next section. But define separate vector overload for convenience. 

https://github.com/PowerGridModel/power-grid-model/blob/f190471001ee34be4a0de2b1a756f1b8d8fd821a/power_grid_model_c/power_grid_model/include/power_grid_model/common/grouped_index_vector.hpp#L54-L61

# Ambigious ownership declaration when using universal reference of range concept based template

Now, there are many places in PGM code base where we have legitimate usage of advanced range concept based templates as argument, because the callers have very diversified input types. Currently, we geneneralize the following arguments:

```c++
func(std::vector<T> const& r);
func(std::vector<T>& r);
```

to a `std::ranges::viewable_range` with universal reference, and using `std::foward` to some view.

```c++
template <std::ranges::viewable_range R>  auto func(R&& r) {
    auto view = std::views::all(std::forward<R>(r));
    // proceed using view
}
```
This changes the semantic meaning of the original intention, which is the root cause of so many `clang-tidy` warnings and using `std::forward` to satisfy `clang-tidy`. Specifically, note the difference below.

* In original intention, passing `std::vector` as const or mutable reference, communicate clearly that:
  * The function **does not** take the ownership of the argument (vector), in **any** circumstances.
  * The function merely read (for const) or mutate (for mutable) the values inside the argument.
  * Because of that, there is not `std::forward` stuff at all.
* The implementation using universal reference of ranges has now an ambiguity:
  * If caller passes a `lvalue` reference, the function does not take over the ownership.
  * If caller passes a `rvalue` reference, the function **might potentially** take over the ownership.
  * Now, in all the practical callers of the entire PGM code base, the caller **never** expects the function to take over the ownership.

The main problem is that we choose an abstraction (universal reference) to allow the function to potentially take over the ownership, while this is never the case in the entire PGM code base.

## Proposal: pass view by value

To solve this ambiguity problem, this PR proposes a new way to define function argument interface, pass a custom `non_owning_view_c` by value. See the definition below:

https://github.com/PowerGridModel/power-grid-model/blob/1bac75b5cc654608dcf8d150e95250bfc0cccbf0/power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp#L118-L124

Then the function argument is passed by value:

```c++
auto func(non_owning_view_c auto r) {
    // proceed using r, no forward stuff needed
}
```

In this way, it is clearly communicated that this function **never** takes over the ownership, which is the case for the entire PGM code base.

Now, if you want to put a real owning container into the function, it will not compile. So, we define a convenient `by_ref` adaptor to translate the owning container to a view.

```c++
std::vector<T> some_vec;
func(by_ref(some_vec)); // converting some_vec to a view
```

See the definition of `by_ref` below. We choose to define our own `by_ref` instead of using `ref_view` from STL, because we want to allow temporary owning ranged to be captured by const reference.

https://github.com/PowerGridModel/power-grid-model/blob/1bac75b5cc654608dcf8d150e95250bfc0cccbf0/power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp#L126-L141

If the caller passes owning container frequently, it might be also good to define a separate overload of `func` to improve convenience:

```c++
auto func(std::vector<T> const& vec) {
    return func(by_ref(vec));
}
```

In this way, the caller can just call as usual:

```c++
std::vector<T> some_vec;
func(some_vec); // will use the vector overload
```

# Other unjustified usage of universal references

There are other places where usage of universal reference is not justified. A simple const reference, or mutable lvalue reference is sufficient for the situation.

## Iterator operators

In `operator+` (and all operators) of iterator, the argument never needs to be universal reference, as below:

https://github.com/PowerGridModel/power-grid-model/blob/d1874c87a33fc704ca9166ae69a8af2156dc5401/power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp#L85-L93

We can easily to change it back to a const reference.

https://github.com/PowerGridModel/power-grid-model/blob/fc3800ea0fd57faf2f62057b3fdaa6611e5e96c2/power_grid_model_c/power_grid_model/include/power_grid_model/common/iterator_facade.hpp#L83-L90

## Deducing this functions

In some places, the deducing this functions are defined with universal reference of `self`. This allows a temperory object being able to be call this method.

https://github.com/PowerGridModel/power-grid-model/blob/d1874c87a33fc704ca9166ae69a8af2156dc5401/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp#L19-L26

However, in the entire PGM code base, there is no use-case of temperory object calling this method. This is again an over-generalization. We should just use a simple `lvalue` reference.

https://github.com/PowerGridModel/power-grid-model/blob/fc3800ea0fd57faf2f62057b3fdaa6611e5e96c2/power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp#L19-L26

# clang-tidy and sonar cloud warning

Originally, this PR proposed to silence `clang-tidy` and `SonarCloud` warning globally. However, after deep investigation, it was found that the root cause is not over-use of `std::forward`, but over-use of universal references in the first place. After removing the universal reference (and thus `std::forward`), there is no `clang-tidy` and `SonarCloud` warning.


power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
```diff
@@ -556,11 +556,11 @@ template <dataset_type_tag dataset_type_> class Dataset {
     template <class type_getter, class ComponentType, typename Func,
               class StructType = DataStruct<typename type_getter::template type<ComponentType>>>
         requires std::invocable<Func, SpanRange<StructType>> && std::invocable<Func, RangeObject<StructType>>
-    auto for_each_component(Func&& func, Idx scenario = invalid_index) const {
+    auto for_each_component(Func func, Idx scenario = invalid_index) const {
```
mgovers, line 559: this will cause any function to be copied unless they are in-place constructed. for small lambdas, that is alright, but if a lambda captures a larger object by value (or rvalue), then this can result in significant performance regressions. why live on the edge if we can just either capture by value or by reference, whichever is more efficient (which is the use case of perfect forwarding)?

mgovers (commented):
i would argue against globally disabling the check of both clang-tidy and sonar cloud. i get that there are multiple separate reasons why one would use it, and i get that it clutters the codebase, but i generally am of the opinion that the people working for sonar and clang are probably more aware of the details etc. than we are. also, even within the maintainer team, not everyone is as aware as you are as to when which use case is used. IMO, we should better be safe than sorry.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: See my first header: The [CPP guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#clambdas-function-objects-and-lambdas) is clear: Function objects should be cheap to copy (and therefore [passed by value](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#rf-in)).

If a functor object is big and expensive to copy, we should fix the functor definition. The functor should be cheap to copy and passed by value.

TonyXiang8787:
> i would argue against globally disabling the check of both clang-tidy and sonar cloud. i get that there are multiple separate reasons why one would use it, and i get that it clutters the codebase, but i generally am of the opinion that the people working for sonar and clang are probably more aware of the details etc. than we are. also, even within the maintainer team, not everyone is as aware as you are as to when which use case is used. IMO, we should better be safe than sorry.

Replying to your general comments @mgovers. The actual legitimate use of `std::forward` in the entire PGM code base is only this:

```c++
template <class... Args>
decltype(auto) some_func(Args&&... args) {
    return some_other_func(std::forward<Args>(args)...);
}
```

Even for this case, missing `std::forward` will only cause real performance problem if `some_other_func` actually has a rvalue overload which consumes (takes the ownership) the (possibly big) input parameter.

So the proposal to disable rules globally is based on three major facts:

1. The places where use of `std::forward` is actually making sense are overwhelmingly obvious for the context of PGM code base. We are not creating arbitrary places where you need a `std::forward`. The argument that normal developers will regularly miss that obvious use of `std::forward` is weak.
2. Even for those cases, the consequence of missing a `std::forward` is near to zero. It only matters if the subsequent function has a rvalue overload taking over the resource ownership.
3. Based on the two points above, the benefit of avoiding the hypothetical problems as stated in two points above by enabling clang-tidy and sonar warning, simply does not outweigh the drawback of decrease of code readability and decrease (external) contributor motivation when they see full of `//NOLINT` everywhere.


And no, for this particular case, sonar and clang-tidy is wrong in big way. They simply cannot distinguish difference use-cases of universal reference in a deterministic way.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
```diff
@@ -51,18 +51,18 @@ template <class T> constexpr CType ctype_v = ctype_t<T>::value;
...
+        return f.template operator()<double>(std::forward<Args>(args)...);
     case c_double3:
-        return std::forward<Functor>(f).template operator()<RealValue<asymmetric_t>>(std::forward<Args>(args)...);
+        return f.template operator()<RealValue<asymmetric_t>>(std::forward<Args>(args)...);
     case c_int8:
-        return std::forward<Functor>(f).template operator()<int8_t>(std::forward<Args>(args)...);
+        return f.template operator()<int8_t>(std::forward<Args>(args)...);
     case c_int32:
-        return std::forward<Functor>(f).template operator()<int32_t>(std::forward<Args>(args)...);
+        return f.template operator()<int32_t>(std::forward<Args>(args)...);
```
figueroa1395: As a learning exercise for me I have the following question. I clearly understand your point as your quotes from the C++ core guidelines are clear, however, if I look at [this](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f19-for-forward-parameters-pass-by-tp-and-only-stdforward-the-parameter) guideline and specifically the way they construct their example, this feels like a case in which to perfect forward `f`. My confusion comes from the wording:
>If the object is to be passed onward to other code and not directly used by this function, we want to make this function agnostic to the argument const-ness and rvalue-ness.
In that case, and only that case, make the parameter TP&& where TP is a template type parameter...

But then in the example they also forward the functor:
```c++
template<class F, class... Args>
inline decltype(auto) invoke(F&& f, Args&&... args)
{
    return forward<F>(f)(forward<Args>(args)...);
}
```
so which one is it? This example is confusing me a bit.

figueroa1395:
@mgovers @TonyXiang8787 Can we do a knowledge sharing for everyone before we merge this? I can prepare it if so. It's good to be aligned first and make sure we all understand the subtleties well.

TonyXiang8787:
> @mgovers @TonyXiang8787 Can we do a knowledge sharing for everyone before we merge this? I can prepare it if so. It's good to be aligned first and make sure we all understand the subtleties well. 

The intention is indeed to discuss and agree with the maintainers, then merge (possibly with adjustments).

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/meta_data.hpp
TonyXiang8787: Thanks for pointing out. This is indeed contradictory of guidelines self. They are community maintained anyway.

By looking at the actual case, there is nothing to forward here for functor object self (not the argument to the functor call). I do not see any reason to preserve ravlue correctness.

figueroa1395: I have a follow up question on this. Is there any performance impact if we were to keep it? I don't think so, right? So the argument is mostly let's follow the guidelines and make it more readable, is that correct?

TonyXiang8787: You are correct. Performance is not the issue here. Mainly about readability.

And also which guideline to follow: functor should be passed by value, or the one you mentioned.

figueroa1395:
A few additional points to add to this discussion:
- `std::ranges::viewable_range` requires perfect forwarding (together with `std::views::all`) to work correctly with `std::vector` and `std::span` (with all their `&`, `const`, etc. versions).
- According to the CPP guidelines, i.e.
>The [CPP guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#clambdas-function-objects-and-lambdas) is clear: Function objects should be cheap to copy (and therefore [passed by value](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#rf-in)).

but in practice that's not really true, as I believe with capture `[*this]` in `lambda`s, so before copying blindly, we may want to re-evaluate whether we are using correct patterns as that is not cheap to copy.
- The above mentioned rule doesn't consider that in general function objects, the `operator()` can be overloaded based on value category and cv-qualification of the callable object itself, and here perfect forwarding it matters. Consider:
```c++
struct CustomFunctor {
    void operator()() & {
        std::cout << "called on lvalue\n";
    }

    void operator()() && {
        std::cout << "called on rvalue\n";
    }
};

template <typename F>
void plain_invoke(F&& f) {
    f();  // f is a named variable => always an lvalue expression
}

int main(){
CustomFunctor f{};

plain_invoke(f); // correct behavior, output is "called on lvalue\n"
plain_invoke(CustomFunctor{}); // incorrect behavior, output "called on lvalue\n" and expected "called on rvalue\n"

return 0;
}
```

I just wanted to highlight a few points I came across lately for this discussion.

TonyXiang8787:
Hi @figueroa1395 

> * `std::ranges::viewable_range` requires perfect forwarding (together with `std::views::all`) to work correctly with `std::vector` and `std::span` (with all their `&`, `const`, etc. versions).

It does not requires perfect forwarding to work correctly. It needs to be a universal reference, if the caller wants to pass an owning-mutable container, still no forward is needed. I will come back to this point later in the PR description.

NOTE: this is under the assumption that your function (and follow-up deeper calls) does not intend to take over the ownership of the resource.

We should not generalize to assume all function calls on ranges will potentially take over some ownership.


> > The [CPP guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#clambdas-function-objects-and-lambdas) is clear: Function objects should be cheap to copy (and therefore [passed by value](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines?lang=en&utm_source=openai#rf-in)).
> 
> but in practice that's not really true, as I believe with capture `[*this]` in `lambda`s, so before copying blindly, we may want to re-evaluate whether we are using correct patterns as that is not cheap to copy.
> 
> * The above mentioned rule doesn't consider that in general function objects, the `operator()` can be overloaded based on value category and cv-qualification of the callable object itself, and here perfect forwarding it matters. Consider:

You are correct here. Generically, there can be a case where `operator()` has a rvalue reference overload. The question is: do we have this in our code base? As far as I know, not. I will address this too in the decription later. 

The same goes with the cheap to copy functor. It's not that functors are always cheap to copy. But I think **we can define cheap to copy, pass by value, as the coding rules of PGM core.**

TonyXiang8787:
@mgovers @figueroa1395 

This PR is ready to be discussed and reviewed again. It had many changes since the first idea.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -36,10 +36,8 @@ class CalculationInfo : public Logger {
     void log(LogEvent tag, double value) override { log_impl(tag, value); }
     void log(LogEvent tag, Idx value) override { log_impl(tag, static_cast<double>(value)); }
     void log(std::string_view /*message*/) const { /* ignore all such events for now */ }
-    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn&& fn) const {
-        capturing::into_the_void(std::forward<Fn>(fn));
-    }
-    template <LazyLoggingFn Fn> void log(Fn&& fn) const { capturing::into_the_void(std::forward<Fn>(fn)); }
+    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn fn) const { capturing::into_the_void(fn); }
+    template <LazyLoggingFn Fn> void log(Fn fn) const { capturing::into_the_void(fn); }
```
mgovers: shouldn't this be `std::move`
```suggestion
    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn fn) const { capturing::into_the_void(std::move(fn)); }
    template <LazyLoggingFn Fn> void log(Fn fn) const { capturing::into_the_void(std::move(fn)); }
```

if not, then it should be simply

```suggestion
    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn /*fn*/) const {}
    template <LazyLoggingFn Fn> void log(Fn /*fn*/) const {}
```
which unfortunately is still an unnecessary copy but maybe will be optimized away in most cases

power_grid_model_c/power_grid_model/include/power_grid_model/common/dummy_logging.hpp
```diff
@@ -20,10 +20,8 @@ class NoLogger : public Logger {
     void log(LogEvent /*tag*/, double /*value*/) override { /* no logging */ }
     void log(LogEvent /*tag*/, Idx /*value*/) override { /* no logging */ }
     void log(std::string_view /*message*/) const { /* no logging */ }
-    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn&& fn) const {
-        capturing::into_the_void(std::forward<Fn>(fn));
-    }
-    template <LazyLoggingFn Fn> void log(Fn&& fn) const { capturing::into_the_void(std::forward<Fn>(fn)); }
+    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn fn) const { capturing::into_the_void(fn); }
+    template <LazyLoggingFn Fn> void log(Fn fn) const { capturing::into_the_void(fn); }
```
mgovers: see my other comment

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -345,8 +345,9 @@ inline void add_diag(Eigen::ArrayBase<DerivedA>& x, Eigen::ArrayBase<DerivedB> c
     x.matrix().diagonal() += y.matrix();
 }
 template <rk2_tensor DerivedA, column_vector DerivedB>
+// NOLINTNEXTLINE(cppcoreguidelines-rvalue-reference-param-not-moved)
 inline void add_diag(Eigen::ArrayBase<DerivedA>&& x, Eigen::ArrayBase<DerivedB> const& y) {
```
mgovers:
```suggestion
inline void add_diag(Eigen::ArrayBase<DerivedA>&& x, Eigen::ArrayBase<DerivedB> const& y) { // NOSONAR
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -15,26 +15,25 @@ namespace power_grid_model::main_core::utils {
...
+template <typename Tuple, std::size_t... Indices>
+constexpr void run_functor_with_tuple_index_return_void(functor_c auto functor,
+                                                        std::index_sequence<Indices...> /*unused*/) {
     if constexpr (sizeof...(Indices) == 1) {
-        (std::forward<Functor>(functor).template operator()<std::tuple_element_t<Indices, Tuple>>(), ...);
+        (functor.template operator()<std::tuple_element_t<Indices, Tuple>>(), ...);
     } else {
         (functor.template operator()<std::tuple_element_t<Indices, Tuple>>(), ...);
-        capturing::into_the_void(std::forward<Functor>(functor));
+        capturing::into_the_void(functor);
```
mgovers:
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/core_utils.hpp
```diff
@@ -15,26 +15,25 @@ namespace power_grid_model::main_core::utils {
...
                                                          std::index_sequence<Indices...> /*unused*/) {
     if constexpr (sizeof...(Indices) == 1) {
-        return std::array {
-            std::forward<Functor>(functor).template operator()<std::tuple_element_t<Indices, Tuple>>()...
-        };
+        return std::array { functor.template operator()<std::tuple_element_t<Indices, Tuple>>()... };
     } else {
         auto result = std::array { functor.template operator()<std::tuple_element_t<Indices, Tuple>>()... };
-        capturing::into_the_void(std::forward<Functor>(functor));
+        capturing::into_the_void(functor);
```
mgovers:
```suggestion
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -300,7 +299,7 @@ class MainModelImpl {
             return prepare_input_(get_n_math_solvers<ModelType>(state_));
         }();
         // calculate
-        return [this, &logger, &input, solve_ = std::forward<SolveFn>(solve)] {
+        return [this, &logger, &input, solve_ = solve] {
```
mgovers: sonar warning

mgovers (commented):
couple minor remarks remaining, rest looks good

TonyXiang8787:
@mgovers comments are addressed.

power_grid_model_c/power_grid_model/include/power_grid_model/common/dummy_logging.hpp
```diff
@@ -20,10 +20,8 @@ class NoLogger : public Logger {
     void log(LogEvent /*tag*/, double /*value*/) override { /* no logging */ }
     void log(LogEvent /*tag*/, Idx /*value*/) override { /* no logging */ }
     void log(std::string_view /*message*/) const { /* no logging */ }
-    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn&& fn) const {
-        capturing::into_the_void(std::forward<Fn>(fn));
-    }
-    template <LazyLoggingFn Fn> void log(Fn&& fn) const { capturing::into_the_void(std::forward<Fn>(fn)); }
+    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn /* fn */) const {}
+    template <LazyLoggingFn Fn> void log(Fn /* fn */) const {}
```
mgovers: sonar
```suggestion
    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn /* fn */) const { /*do nothing*/ }
    template <LazyLoggingFn Fn> void log(Fn /* fn */) const { /*do nothing*/ }
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -36,10 +36,8 @@ class CalculationInfo : public Logger {
     void log(LogEvent tag, double value) override { log_impl(tag, value); }
     void log(LogEvent tag, Idx value) override { log_impl(tag, static_cast<double>(value)); }
     void log(std::string_view /*message*/) const { /* ignore all such events for now */ }
-    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn&& fn) const {
-        capturing::into_the_void(std::forward<Fn>(fn));
-    }
-    template <LazyLoggingFn Fn> void log(Fn&& fn) const { capturing::into_the_void(std::forward<Fn>(fn)); }
+    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn /*fn*/) const {}
+    template <LazyLoggingFn Fn> void log(Fn /*fn*/) const {}
```
mgovers: sonar

```suggestion
    template <LazyLoggingFn Fn> void log(LogEvent /*tag*/, Fn /*fn*/) const { /*do nothing*/ }
    template <LazyLoggingFn Fn> void log(Fn /*fn*/) const { /*do nothing*/ }
```

mgovers (dismissed):
also ok with merging without resolving the sonar warnings.

figueroa1395:
If no rush, I'll take a look later today or tomorrow morning.

If you want to get this out, I'll just read it after it's merged and feel free to ignore this comment.

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
figueroa1395, line 559: Why not add `functor_c` here as well? In case this is used with a `lambda`.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
figueroa1395: I think we can and should also add `functor_c` to this concept.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -101,22 +103,48 @@ template <class T, class... Ts>
...
+namespace detail {
+template <class> struct is_owning_view : std::false_type {};
+template <class R> struct is_owning_view<std::ranges::owning_view<R>> : std::true_type {};
+} // namespace detail
+template <class R>
+concept non_owning_view_c = std::ranges::view<R> && !detail::is_owning_view<std::remove_cvref_t<R>>::value;
+
+// by ref adaptor to pass to functions which accepts std::ranges::view
+template <class R>
+    requires(std::ranges::viewable_range<R> && !std::ranges::view<R> && !std::ranges::borrowed_range<R>)
```
figueroa1395: Just to make sure I got it right, this makes sure the range is viewable, but that it isn't a view itself (or else why are we making another view), and what's the last about?

In [cppreference](https://en.cppreference.com/cpp/ranges/borrowed_range) I see: _The concept `borrowed_range` defines the requirements of a range such that a function can take it by value and return iterators obtained from it without danger of dangling._

Is it because if we can take the range by value and trivially get iterators from this, then why do we need a view? Can you give me such an example?

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
```diff
@@ -101,22 +103,48 @@ template <class T, class... Ts>
...
+constexpr auto by_ref(R& r) noexcept {
+    return std::ranges::ref_view(r);
+}
+template <class R>
+    requires(std::ranges::viewable_range<R> && !std::ranges::view<R> && !std::ranges::borrowed_range<R>)
```
figueroa1395: I see this clause appears three times, can you make it a named `concept`?

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -345,8 +345,9 @@ inline void add_diag(Eigen::ArrayBase<DerivedA>& x, Eigen::ArrayBase<DerivedB> c
     x.matrix().diagonal() += y.matrix();
 }
 template <rk2_tensor DerivedA, column_vector DerivedB>
-inline void add_diag(Eigen::ArrayBase<DerivedA>&& x, Eigen::ArrayBase<DerivedB> const& y) {
-    std::move(x).matrix().diagonal() += y.matrix();
+// NOLINTNEXTLINE(cppcoreguidelines-rvalue-reference-param-not-moved)
+inline void add_diag(Eigen::ArrayBase<DerivedA>&& x, Eigen::ArrayBase<DerivedB> const& y) { // NOSONAR
```
figueroa1395, line 349: Is the intention of this overload then to capture r-values only? So when the user purposely moves it in the function call?

Wouldn't we then have to move this back out? Maybe I'm missing something because of some `Eigen` thing.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -415,20 +416,18 @@ inline auto all_zero(RealValue<asymmetric_t> const& value) { return (value == Re
 //
 // The function assumes that the current value is normalized and new value should be normalized with scalar
 template <symmetry_tag sym, class Proxy>
+// NOLINTNEXTLINE(cppcoreguidelines-missing-std-forward)
```
figueroa1395, line 419: Similar question to above. Wouldn't we need to then move out the result if we expect r-value only here?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
```diff
@@ -280,28 +278,28 @@ inline UpdateChange update_component(ComponentContainer& components, Updates&& c
                 *changed_it++ = sequence_single;
             }
         },
-        std::views::all(std::forward<Updates>(component_updates)), sequence_idx);
+        std::views::all(component_updates), sequence_idx);
```
figueroa1395: Is `std::views::all` still needed given `component_updates` should satisfy `non_owning_view_c`?

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
```diff
@@ -17,53 +17,53 @@ class JobInterface {
   public:
     // the multiple  NOSONARs are used to avoid the complaints about the unnamed concepts
     template <typename Self, typename ResultDataset>
-    void calculate(this Self&& self, ResultDataset const& result_data, Idx pos, Logger& logger)
+    void calculate(this Self& self, ResultDataset const& result_data, Idx pos, Logger& logger)
```
figueroa1395, line 20: Can you also add to the description or later the documentation the deliberate choice of not having `this Self&&` when using `deducing this`? As in, this now forbids a temporary from calling `calculate`. This isn't something relevant currently, but it may be later and a new developer may run into a strange error which is easy to pinpoint with some docs.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -244,7 +243,7 @@ class MainModelImpl {
         auto const& component_sequence = std::get<component_index>(sequence_idx);
 
         if (!cached_inverse_update.empty()) {
-            update_component<CompType, permanent_update_t>(cached_inverse_update, component_sequence);
+            update_component<CompType, permanent_update_t>(by_ref(cached_inverse_update), component_sequence);
```
figueroa1395, line 246: So the choice is that it's the caller's responsibility to use `by_ref` here or else it won't compile. But if a new function is added, we also must be disciplined then, right?

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp
```diff
@@ -92,12 +92,12 @@ class Handle {
         }
     }
 
-    template <typename Func, typename... Args> auto call_with(Func&& func, Args&&... args) const {
-        if constexpr (std::is_void_v<decltype(std::forward<Func>(func)(get(), std::forward<Args>(args)...))>) {
-            std::forward<Func>(func)(get(), std::forward<Args>(args)...);
+    template <typename Func, typename... Args> auto call_with(Func func, Args&&... args) const {
```
figueroa1395, line 95: Shouldn't we also add the `func_c` here?

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
```diff
@@ -55,25 +55,24 @@ class UnsupportedPGM_CType : public PowerGridError {
 };
 
 template <class Functor, class... Args>
```
figueroa1395, line 57: Same here `functor_c` missing

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
```diff
@@ -55,25 +55,24 @@ class UnsupportedPGM_CType : public PowerGridError {
...
     default:
         throw UnsupportedPGM_CType();
     }
 }
 
 template <class Functor, class... Args>
-decltype(auto) pgm_type_func_selector(MetaAttribute const* attribute, Functor&& f, Args&&... args) {
-    return pgm_type_func_selector(MetaData::attribute_ctype(attribute), std::forward<Functor>(f),
-                                  std::forward<Args>(args)...);
+decltype(auto) pgm_type_func_selector(MetaAttribute const* attribute, Functor f, Args&&... args) {
```
figueroa1395, line 74: And here

tests/native_api_tests/test_api_model.cpp
```diff
@@ -96,10 +96,10 @@ void check_exception(PowerGridError const& e, PGM_ErrorCode const& reference_err
 }
 
 template <typename Func, class... Args>
-void check_throws_with(Func&& func, PGM_ErrorCode const& reference_error, std::string_view reference_err_msg,
+void check_throws_with(Func func, PGM_ErrorCode const& reference_error, std::string_view reference_err_msg,
```
figueroa1395, line 101: Do we make the deliberate choice to not enforce `functor_c` for tests to make our life easier? If yes, can you add a small comment? This shouldn't affect releases at all, right?

figueroa1395 (commented):
Should we add documenting these new internal guidelines in the advanced documentation as part of the scope of this change?

power_grid_model_c/power_grid_model/include/power_grid_model/auxiliary/dataset.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/update.hpp
TonyXiang8787: You are right. Removed

power_grid_model_c/power_grid_model/include/power_grid_model/job_interface.hpp
TonyXiang8787: I will have a separate PR regarding advanced coding guidelines.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
TonyXiang8787: Yes, we have to be disciplined for that pattern. In a follow-up PR I will add this to the guideline.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp
TonyXiang8787: No, you can't. This is `cpp` wrapper, which does not have access to the header-only core.

And the `cpp` wrapper is now c++17 compliant.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
TonyXiang8787: Same here.

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/utils.hpp
TonyXiang8787: Same here.

tests/native_api_tests/test_api_model.cpp
TonyXiang8787: Same here.

When you access stuff via C-API, you don't have access to internal header only core. But I can add a comment here.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
TonyXiang8787: The last is about the range is owning some resources. So we can make a view of that, to be passed by value of certain function which accept `non_owning_view`.

power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp
TonyXiang8787: done

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
TonyXiang8787: Yes. It's meant to capture a temperary proxy. There is nothing to move there. This is a false warning from `clang-tidy`. It cannot detect the 3rd use case of rvalue, which is capture temperary proxy.

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
TonyXiang8787: Same here. This one is meant to capture proxy object. `clang-tidy` cannot detect this and trigger a false warning.

figueroa1395 (approved):

PR #1407, opened 2026-05-26
Reproducible build: fix MSVC build reproducibility

mgovers:
The MSVC build reproducibility failed recently (see e.g. https://github.com/PowerGridModel/power-grid-model/actions/runs/26438097171/job/77825539883)

The underlying reason seems to be related to a couple things related to the debug directories: (tested using `dumpbin /HEADERS <first> <second>` and also with a full check using `dumpbin /ALL <first> <second>`)

## Changes in this PR

* For the reproducibile build check, migrated from `RelWithDebInfo` to `Release` to get rid of more debug symbols
  * NOTE: THIS MAY CAUSE `CMakeUserPresets.json` TO BREAK FOR SOME DEVELOPERS
* set `/debug:None` to force `Debug Directories` to be reproducible (creates a `None` string instead)
* set `/NOCOFFGRPINFO` to remove the `COFFGRP` directory from the `Debug Directories`
* set `/LTCG` to disable the default fast linking code generation (which is not fully disabled by disabling incremental build)

### First
```
  Debug Directories

        Time Type        Size      RVA  Pointer
    -------- ------- -------- -------- --------
    E39168F9 cv            79 006D2934   6D1934    Format: RSDS, {9310574C-CE7A-FD1A-032B-721FC6B9EABA}, 1, D:\a\power-grid-model\power-grid-model\cpp_build\ci-msvc-reproducible\bin\power_grid_model_c.pdb
    E39168F9 feat          14 006D29B0   6D19B0    Counts: Pre-VC++ 11.00=0, C/C++=46, /GS=46, /sdl=0, guardN=37
    E39168F9 coffgrp      33C 006D29C4   6D19C4    4C544347 (LTCG)
    E39168F9 repro         24 006D2D00   6D1D00    4C 57 10 93 7A CE 1A FD 03 2B 72 1F C6 B9 EA BA 4A 7A A3 30 4A 32 40 08 69 87 BA 73 F9 68 91 E3
```

### Second
```
  Debug Directories

        Time Type        Size      RVA  Pointer
    -------- ------- -------- -------- --------
    63B9CD42 cv            79 006D2934   6D1934    Format: RSDS, {EFB674F5-E861-8C49-023F-F88F6283BE43}, 1, D:\a\power-grid-model\power-grid-model\cpp_build\ci-msvc-reproducible\bin\power_grid_model_c.pdb
[...]

PR #1416, opened 2026-06-01
Reproducible build: Address fail at action 665

nitbharambe:
Manual run: 
https://github.com/PowerGridModel/power-grid-model/actions/runs/26764227321
https://github.com/PowerGridModel/power-grid-model/actions/runs/26802385103

We added /LTCG in https://github.com/PowerGridModel/power-grid-model/pull/1407. This option replaces the default fast linking code generation. However LTCG itself is not deterministic. Searched around and maybe this option works to disable this optimization. Although i am still not sure if this makes the optimization deterministic enough or not. Hence we might bump into similar issues later. 

image

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -267,6 +267,7 @@
       },
       "cacheVariables": {
         "PGM_ENABLE_DEV_BUILD": "OFF",
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "OFF",
         "CMAKE_SHARED_LINKER_FLAGS": "/debug:None /Brepro /INCREMENTAL:NO /NOCOFFGRPINFO"
       }
     },
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -267,7 +267,7 @@
       },
       "cacheVariables": {
         "PGM_ENABLE_DEV_BUILD": "OFF",
-        "CMAKE_SHARED_LINKER_FLAGS": "/debug:None /Brepro /INCREMENTAL:NO /LTCG /NOCOFFGRPINFO"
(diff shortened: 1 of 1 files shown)


figueroa1395:
Your previous run was successful. Running manually again at https://github.com/PowerGridModel/power-grid-model/actions/runs/26802385103 for an additional try. If it passes the second time, let's merge it and address again later if needed.

CMakePresets.json
```diff
@@ -267,6 +267,7 @@
       },
       "cacheVariables": {
         "PGM_ENABLE_DEV_BUILD": "OFF",
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "OFF",
```
figueroa1395, line 270: Assuming this works, why not disable it for all builds? My understanding from a quick read is that in can increase runtime performance at the cost of compile time, but we are already heavily templated so that's mostly taken care of. Also, shouldn't we homogenize how we ship across platforms and ideally the same for a reproducible build and the release one (as in the actual release build should be reproducible)?

mgovers: (removed my older comments as they are not valid)

We want IPO (especially in Release builds) because performance is important. Only if reproducibility is extremely important can we do without.

That said, it may be good to disable it if a reproducible build is requested

TonyXiang8787: We are not required to put reproducible build as default build option.

TonyXiang8787 (approved):

CMakePresets.json
nitbharambe: Also note that the difference is only in the debug directories section. Rest of the content is same before and after the PR.

Issue #1418, opened 2026-06-02
[FEATURE] Allow branch into itself


mgovers:
### Describe the feature request

Occasionally, there may be branches that have both (for branch3: multiple) ends connecting to the same node. While it may be caused by a data error, it is actually physically OK (the branch will act similarly to a branch that's disconnected on one side). Therefore, there is no reason to actually reject this as a data error.

This will be relevant for implementation of #35, because it means that multiple user nodes will be combined into one super node. Since it is OK to connect any two separate user nodes with branches, merging nodes on links without supporting branches into itself would actually result in a regression.

### Community Meeting

- [ ] I wish to discuss this feature in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

PR #1419, opened 2026-06-02
Branch into itself: core implementation

mgovers:
Part of #1418 

Contributes towards #35 

NOTE: this only affects the core. No validation cases are added in this PR. While it is of course possible that there are bugs introduced that will be caught by such validation cases, since both the docs and the data validator at this point in time still reject branches into itself, the behavior is still unspecified (subcategory of undefined behavior) and, therefore, such a bug should not be considered to affect users, because it is the user's responsibility to not invoke undefined behavior.


power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
```diff
@@ -202,12 +203,12 @@ class Topology {
             auto const [i, j] = branch_node_idx;
             auto const [i_status, j_status] = branch_connected;
             // node_i - node_j
-            if (i_status != 0 && j_status != 0) {
+            if (i_status != 0 && j_status != 0 && i != j) {
```
figueroa1395, line 206: This was the only issue? That we were explicitly creating the edge for the self connecting branch?

tests/cpp_unit_tests/test_link.cpp
```diff
@@ -52,7 +52,7 @@ TEST_CASE("Test link") {
 
     SUBCASE("Invalid branch") {
```
figueroa1395: Change subcase name?

tests/cpp_unit_tests/test_y_bus.cpp
```diff
@@ -41,28 +41,31 @@ TEST_CASE("Test y bus") {
...
+    //   --- 0 ---            |   \--- 6 ----/
+    //                        X
     MathModelTopology topo{};
     MathModelParam<symmetric_t> param_sym;
     topo.phase_shift.resize(4, 0.0);
     topo.branch_bus_idx = {
         {1, 0}, // branch 0 from node 1 to 0
         {1, 2}, // branch 1 from node 1 to 2
         {2, 3}, // branch 2 from node 2 to 3
+        {2, 2}, // branch 6 from node 2 to 2 (loop into itself)
```
figueroa1395: The ASCII drawing shows 6 as a different branch. You have branches 3 and 6 mixed up.

tests/cpp_unit_tests/test_y_bus.cpp
```diff
@@ -75,29 +78,31 @@ TEST_CASE("Test y bus") {
...
+        19.0 + 2.0i,          // 1, 0 -> {0, 1}tf + {1, 0}ft
+        20.0 + 1.0i + 5.0,    // 1, 1 -> {0, 1}tt + {1, 0}ff + {1,2}ff
+        6.0,                  // 1, 2 -> {1,2}ft
+        7.0,                  // 2, 1 -> {1,2}tf
+        8.0 + (21.0i + 22.0i + 22.0i + 21.0i) + 9.0i + 16.0 + 1000.0i,
+        // 2, 2 -> {1,2}tt + ({3,3}ff + {3,3}ft + {3,3}{tf} + {3,3}tt)
+        //            + {2,3}ff + {3, 2}tt + {2,-1}ff
+        10.0i + 15.0,         // 2, 3 -> {2,3}ft + {3,2}tf
+        11.0i + 14.0,         // 3, 2 -> {2,3}tf + {3,2}ft
+        12.0i + 13.0 + 200.0i // 3, 3 -> {2,3}tt + {3,2}ff + shunt(1)
```
figueroa1395: I'm very confused at this. Was it wrong before? It seems like you made the comments the actual values and that the comments before didn't align with the values.

tests/cpp_unit_tests/test_y_bus.cpp
```diff
@@ -220,26 +226,47 @@ TEST_CASE("Test y bus") {
...
+
+        CHECK(ybus.size() == 1);
+        CHECK(ybus.nnz() == nnz);
+        CHECK(indptr == ybus.row_indptr());
+        CHECK(col_indices == ybus.col_indices());
+        CHECK(bus_entry == ybus.bus_entry());
+        CHECK(lu_transpose_entry == ybus.lu_transpose_entry());
+        CHECK(y_bus_entry_indptr == ybus.y_bus_entry_indptr());
+    }
+    SUBCASE("Branch into itself") {
```
figueroa1395, line 262: Can you add a 3-bus test that is cyclic and that has one additional branch into itself on one of the buses, or even two? This to make sure that we didn't miss anything in the y-bus logic since it's very complex.

tests/cpp_unit_tests/test_topology.cpp
```diff
@@ -32,30 +32,43 @@
...
+ *   /                  \        v          /    $0                           (b2)                        (b1)
+ *  [0:s0,lg0]         (b0)-$2- [2+v0+v2]  /      \                            |                           |
+ *   +p0+p11            /    +p15         X       (b3)                         $0                          $0
+ *    \                $1                /         \  \                        X                           |
+ *     \          +p16/                 /           $1 $2                      [10]                        [11:lg1+p6]
+ *      1 -->+p2+p10 [3+v3:s3X,h0] -- 2              \  \
+ *                    |         |                     [12]
+ *                    |         |                                                                 /--$0--\
+ *                    \----8----/                                   [9:s2X+p3,h2]          [13:s3]---$1---(b4)
+ *                                                                                                \--$2--/
```
figueroa1395, line 50: The drawing is a bit hard to understand, but why the addition of the disconnected grid [9]?

tests/cpp_unit_tests/test_topology.cpp
```diff
@@ -192,25 +209,32 @@ TEST_CASE("Test topology") {
         {.group = 0, .pos = 0},
         {.group = 0, .pos = 4},
         // 4 5 6
-        {.group = 1, .pos = 2}, // Topological node 4 has become node 2 in mathematical model (group) 1
-        {.group = 1, .pos = 3},
+        {.group = 1, .pos = 4}, // Topological node 4 has become node 2 in mathematical model (group) 1
```
figueroa1395: Adjust the comment as it is no longer valid. Perhaps the same happens in other below.

tests/cpp_unit_tests/test_topology.cpp
```diff
@@ -32,30 +32,43 @@
...
+ *   /                  v         v            \                                        v   v
+ *  [4:s0,lg1]          [3] <-7- [0+v0+v1]      5                                        [1]
+ *   +pf0+pf1           ^     +pf4               v                                        ^
+ *    \                6                         [3]                                      2
+ *     \          +pf3/                           ^ ^                                     X
+ *      1 ->+pt1+pt2 [1+v2:h0] -- 2 --X            6 7
+ *                    |     |                       \ \
+ *                    |     |                        [2]
+ *                    \--4--/
+ *
```
figueroa1395, line 67: I find it interesting how after adding (b3) and [12] in the topology, the math model is labeled like this for group 1. Is this expected? I would have expected [5] to remain [2] not [12] to become 2, or is this normal because of the special handling for transformer ranking?

figueroa1395 (commented):
I'm almost done, but checking out the topology tests is not an easy task. I will finish tomorrow with the last part, but so far, so good.

In addition, I will also dig into the SE to double check again if nothing breaks there.

tests/cpp_unit_tests/test_y_bus.cpp
mgovers: the order changed... but indeed there are some comments that did not align

tests/cpp_unit_tests/test_y_bus.cpp
mgovers: whoops... so many potential points of error. good catch

power_grid_model_c/power_grid_model/include/power_grid_model/topology.hpp
mgovers: yes exactly

tests/cpp_unit_tests/test_topology.cpp
mgovers: because this topology case catches all possibilities, including a disconnected grid with disconnected source (which should be excluded from all calculations; no math solver should be created). Note that it was also there in the original, but the location was a bit awkward so i had to move it

tests/cpp_unit_tests/test_topology.cpp
mgovers: transformer ranking has nothing to do with this - it's restricted to transformer powerflow. The reason why the ranking changes is just "coincidental" (in the sense that it's an implementation detail of the boost DFS algorithm)

That said: good catch, we should check how the transformer ranking algorithm handles this.

mgovers: i suspect that stack popping may have to do with it

tests/cpp_unit_tests/test_topology.cpp
figueroa1395: I didn't notice it in the original, my bad. Thanks!

tests/cpp_unit_tests/test_topology.cpp
figueroa1395: If it's boost's DFS internals, that's fine, and it makes sense. Let's not spend too much time on it, but just wanted to point it out.

tests/cpp_unit_tests/test_topology.cpp
```diff
@@ -32,30 +32,43 @@
...
+ *                                      Math model #2:
+ *                                                       /--0-->
+ *                                                 [1:s0]---1-->[0]
+ *                                                       \--2-->
  *
  * Extra fill-in:
```
figueroa1395, line 73: This is very cumbersome, but we should also check if the fill ins need to be updated. Or are these unrelated to the above?

tests/cpp_unit_tests/test_y_bus.cpp
```diff
@@ -41,28 +41,31 @@ TEST_CASE("Test y bus") {
...
         {1, 0}, // branch 0 from node 1 to 0
         {1, 2}, // branch 1 from node 1 to 2
         {2, 3}, // branch 2 from node 2 to 3
-        {3, 2}, // branch 3 from node 3 to 2
-        {0, 1}, // branch 4 from node 0 to 1
-        {2, -1} // branch 5 from node 2 to "not connected"
+        {2, 2}, // branch 3 from node 2 to 2 (loop into itself)
+        {3, 2}, // branch 4 from node 3 to 2
+        {0, 1}, // branch 5 from node 0 to 1
+        {2, -1} // branch 6 from node 2 to "not connected"
```
figueroa1395: According to the drawing:
```suggestion
        {3, 2}, // branch 6 from node 3 to 2
        {0, 1}, // branch 4 from node 0 to 1
        {2, -1} // branch 5 from node 2 to "not connected"
```

tests/cpp_unit_tests/test_y_bus.cpp
figueroa1395: I see now what happened, some numbers changed for example from
```
17.0 + 104.0i
```
to
```
4.0i + 17.0 + 100.0i
```
which is the same. Now, the second was in the comments and now it's self documented in code. Understood!

tests/unit/test_error_handling.py
```diff
@@ -118,8 +117,7 @@ def test_handle_invalid_branch_error():
     line_input[AT.from_node] = [0]
     line_input[AT.to_node] = [0]
 
-    with pytest.raises(InvalidBranch):
-        PowerGridModel(input_data={CT.node: node_input, CT.line: line_input})
+    PowerGridModel(input_data={CT.node: node_input, CT.line: line_input})
```
figueroa1395, line 121: Can you also modify the test names? As they are no longer "invalid". Same below.

figueroa1395 (commented):
> since both the docs and the data validator at this point in time still reject branches into itself, the behavior is still unspecified (subcategory of undefined behavior) and, therefore, such a bug should not be considered to affect users, because it is the user's responsibility to not invoke undefined behavior.

I agree, but lets make sure we all agree on this so we don't have to circle back later. We also may want to send a deprecation notice via the mailing list/community meeting or similar.

On the other hand, I scanned again NRSE and I didn't see any problem thus far. There isn't any explicit check against having self connecting branches + the math seems to hold as long as the ybus entries are fine, which seems to be the case as well.

Finally, I noticed that we may have an issue with observability, as I don't think it will recognize "flow" sensors on a self connecting branch since this will only appear in the y-bus diagonal, whereas the check looks for the upper diagonal parts. This needs to be further checked, but something to keep and potentially tackle in a follow up.

src/power_grid_model/errors.py
```diff
@@ -16,8 +16,6 @@
     IDNotFound,
     IDWrongType,
     InvalidArguments,
-    InvalidBranch,
-    InvalidBranch3,
```
figueroa1395, line 20: I've tried making this comment several times from the VS code extension and it's been rejected thus far for whatever reason. So let me make it here as a standalone one:

I don't think we should remove these from here as we might break users if they are using these errors in a custom way in their pipeline. I think the deprecation warning should be enough for now. What do you think?

tests/cpp_unit_tests/test_y_bus.cpp
mgovers: done

tests/cpp_unit_tests/test_topology.cpp
mgovers: fill-ins will never be updated by this feature because they are introduces by off-diagonal entries, which are non-existent for branches into itself

src/power_grid_model/errors.py
mgovers: good one, i originally added the deprecation warning for exactly that reason but forgot in the chaos this morning. good catch

power_grid_model_c/power_grid_model/include/power_grid_model/common/three_phase_tensor.hpp
```diff
@@ -187,6 +187,16 @@ inline ComplexValue<asymmetric_t> piecewise_complex_value(T<Derived> const& x) {
...
+// basic operations
+using Eigen::arg;
+using Eigen::conj;
+using Eigen::imag;
+using Eigen::real;
+using std::arg;
+using std::conj;
+using std::imag;
+using std::real;
+
```
figueroa1395, line 199: Why is this needed here now?

mgovers: recent version of clangd/clang-tidy was complaining. felt too simple to explicitly scope out into a separate PR

tests/cpp_unit_tests/test_observability.cpp
```diff
@@ -57,23 +57,23 @@ TEST_CASE("Observable voltage sensor - basic integration test") {
...
-    topo.power_sensors_per_shunt = {from_sparse, {0}};
-    topo.power_sensors_per_branch_from = {from_sparse, {0, 1, 2}};
-    topo.power_sensors_per_branch_to = {from_sparse, {0, 0, 0}};
-    topo.current_sensors_per_branch_from = {from_sparse, {0, 0, 0}};
-    topo.current_sensors_per_branch_to = {from_sparse, {0, 0, 0}};
-    topo.voltage_sensors_per_bus = {from_sparse, {0, 1, 1, 1}};
+    topo.branch_bus_idx = {{0, 1}, {1, 2}, {1, 1}};
+    topo.sources_per_bus = {from_dense, {0}, 3};
+    topo.shunts_per_bus = {from_dense, {}, 3};
+    topo.load_gens_per_bus = {from_dense, {}, 3};
```
figueroa1395, line 63: This example is "trivial" in the sense that we have no `loadgens` so we get "free" zero-injections sensors at each node. Does it even make sense to try to come up with an example in which we have loads on all nodes and the sensor on the self-loop branch MUST be used to achieve observability? I haven't been able to and I'm wondering if we should even try.

mgovers: added a separate case

tests/cpp_unit_tests/test_topology.cpp
```diff
@@ -32,30 +32,43 @@
...
  *      0 ----->+p13 [1+v1:lg3+p7]             /     /                     \        /                  \        /
- *     /          +p14\            5 ---X--- [4] <- 6                       $2     $1                   $1     $2
- *    /                $0         /          ^                               \    /                      \    /
- *   /                  \        v          /                                 (b2)                        (b1)
- *  [0:s0,lg0]         (b0)-$2- [2+v0+v2]  /                                   |                           |
- *   +p0+p11            /    +p15         X                                    $0                          $0
- *    \                $1                /         [9:s2X+p3,h2]               X                           |
- *     \          +p16/                 /                                     [10]                        [11:lg1+p6]
- *      1 -->+p2+p10 [3+v3:s3X,h0] -- 2
+ *     /          +p14\            5 ---X--- [4] <- 6                      $2     $1                   $1     $2
```
nitbharambe, line 39: Lets do away with ASCII art soon. This has reached the point of being unreadable and unmaintainable.
Ideas welcome.

Option 1:
PGM-DS can save visualization topology but not individual data. If the data of arrays are stored as well then maybe we can save all.
This way would need a small script to create PGM array and save its outcome.

Option 2:
We formalize https://github.com/PowerGridModel/power-grid-model-io/pull/393 with some changes. Store that somewhere within the project. 
Then a small script to create grid and save outcome.

Option 3:
Small independent and minimal script to generate pdf or ascii art again

figueroa1395: I highly support these. Let's scope it out though and brainstorm about it!

mgovers: unfortunately, this is a little bit more tricky on the level of unit tests than it is for validation cases

tests/cpp_unit_tests/test_observability.cpp
mgovers: extended with multiple other tests

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -744,6 +744,47 @@ TEST_CASE("Test Transformer ranking") {
                                "source side:\n  Transformer IDs: 10");
         }
     }
+
+    SUBCASE("Transformer into itself") {
+        // Test that a transformer with an edge to itself is handled correctly
+        // ========Test Grid========
+        //      [source 0]
+        //          |
+        //     [trafo 10] (WRONG: has an edge to itself)
```
figueroa1395: What do you mean by this?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
```diff
@@ -744,6 +744,47 @@ TEST_CASE("Test Transformer ranking") {
...
+        //          |
+        //      [node 2]
+
+        TestState state;
+        std::vector<NodeInput> const nodes{
+            {.id = 0, .u_rated = 150e3}, {.id = 1, .u_rated = 10e3}, {.id = 2, .u_rated = 10e3}};
+        main_core::add_component<Node>(state.components, nodes, 50.0);
+
+        std::vector<TransformerInput> const transformers{get_transformer(10, 0, 1, BranchSide::from)
+                                                         // Transformer with an edge to itself (wrong)
```
figueroa1395: I also don't get it here, this isn't a trafo into itself, but 11 is. what are you trying to say here?

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: ugh remnant of copy/pasting. good catch
```suggestion
        //     [trafo 10]
```

tests/cpp_unit_tests/test_tap_position_optimizer.cpp
mgovers: also fixed

figueroa1395 (approved):

PR #1420, opened 2026-06-03
Node injection sensor deprecation: suppress deprecation warnings in data validator + tests

mgovers:
Relates to #1334 

In #1384, node injection sensor types were marked as deprecated. However, this caused users to get deprecation warnings if they run the data validator, even if they themselves did not use node injection sensors, because the data validator internally checks for `MeasuredTerminalType.node`. Similarly, running `pytest` also resulted in a multitude of deprecation warnings.

This PR fixes both.


figueroa1395 (approved):
Tested locally and didn't run into any issue.

PR #1423, opened 2026-06-05
Reproducible build: Failing again at action 667 and 669 nightly

nitbharambe:
The reproducibility build has still not been fixed. I suspect these options would reenable the interprocedural optimisation. Hence disabling it specifically for reproducible builds

Failed at nightly-s 667, 669, 673, 674

Manual runs:
https://github.com/PowerGridModel/power-grid-model/actions/runs/26874783996
https://github.com/PowerGridModel/power-grid-model/actions/runs/26873653208


figueroa1395 (approved):
LGTM. Let's see if this helps nightly.

PR #1428, opened 2026-06-08
Compilation times: reorder cpp unit-tests

mgovers:
Reorder cpp unit test files into separate subdirectories, following the same order as for the header files and the CMake targets

Rationale: Recently, our compilation times have increased. This PR does not solve the individual compilation time for each file but it will make it more performant if built in parallel as the files with long compilation times are not longer all at the back. It also homogenizes the codebase.

Contributes to #1429


tests/cpp_unit_tests/CMakeLists.txt
```diff
@@ -139,18 +28,3 @@ target_link_libraries(
 )
 
 doctest_discover_tests(power_grid_model_unit_tests)
```
figueroa1395, line 31: Was it intentional to leave these behind?

figueroa1395:
Additionally, don't forget to modify the documentation (build-guide) to reflect the new targets for `cmake`.

tests/cpp_unit_tests/CMakeLists.txt
mgovers: yes because these unit test files were not yet moved to a subdir. should i do that now?

mgovers:
> Additionally, don't forget to modify the documentation (build-guide) to reflect the new targets for `cmake`.

~no updates required here AFAICT~ nvm, looked at the wrong place; edited

tests/cpp_unit_tests/CMakeLists.txt
figueroa1395: I guess we might do it for all the files and homogenize at once.

figueroa1395: Offline discussion: This already aligns with the header file hierarchy. No need to modify anything else.

Resolving.

figueroa1395 (approved):

PR #1433, opened 2026-06-11
Update NRPF for better memory optimization

### LINKED ISSUE #1422: [IMPROVEMENT] Memory improvement in Newton Raphson Power Flow: Reuse Jacobian memory
### Problem faced

@TonyXiang8787 suggested a neat memory efficiency improvement for NRPF. We initialize the nodal voltages to a results of run of a `linear` method calculation:

https://github.com/PowerGridModel/power-grid-model/blob/aa0fc3ec2c4c30186b227ff67ba0845082e8e022/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp#L236-L243

Hence we allocate `linear_mat_data`, calculate then deallocate it. This extra memory overhead is avoidable and might be significant for very large grids.  

### Improvement proposal

Instead of solving the initial linear calculation on the complex domain, we can solve it in real domain instead. 
1. Hence we split the linear matrix into real and imaginary domain. 
2. Solve equations.
3. Reuse the matrix for Jacobian by casting to `std::vector<PFJacBlock<sym>>`. 
4. Convert u_initial_real, u_initial_imag to u_magnitude and u_theta.  

This would involve implementing the entire linear algorithm into real / imaginary form. This is simply rewriting existing code in different form.

Note: This issue is marked as `good-first-issue` because its re-implementation of existing logic. However the output from this issue would be a bit larger.

### Community Meeting

- [ ] I wish to discuss this improvement in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

akshatowl:
removing the linear allocations and splitting computation into real and imaginary components, attempting to address #[1422](https://github.com/PowerGridModel/power-grid-model/issues/1422)


nitbharambe:
Thanks for the contribution @akshatowl ! We shall conduct a review soon. Meanwhile there are some build failures which you can address. I see the DCO action can be resolved as well.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +228,79 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
           bus_types_(y_bus.size(), BusType::pq),
           voltage_regulators_per_load_gen_{std::ref(topo.voltage_regulators_per_load_gen)} {}
 
-    // Initilize the unknown variable in polar form
+    // Initialize the unknown variable in polar form using a linear voltage guess solved in the real domain.
+    // This implementation reuses the class-level Jacobian matrix (data_jac_) and RHS vector (del_x_pq_)
+    // to eliminate temporary memory allocations.
+    // The complex system (G + jB)(Ur + jUi) = Ir + jIi is mapped to the real system:
+    // [[-B, G], [-G, -B]] [Ui, Ur]^T = [Ir, -Ii]^T
```
figueroa1395: The way this expression is written is a little bit bizarre, why not something like
```suggestion
    // [[G, -B], [B, G]] [Ur, Ui]^T = [Ir, Ii]^T
```
instead?

I see the logic below corresponds to this notation and it looks correct, but why this starting point?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +228,79 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+    // This mapping ensures that H/N/M/L sub-blocks align with the NR Jacobian at flat start.
     void initialize_derived_solver(YBus<sym> const& y_bus, PowerFlowInput<sym> const& input,
                                    SolverOutput<sym>& output) {
-        using LinearSparseSolverType = SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>>;
+        // Reset reused buffers to zero
+        std::fill(data_jac_.begin(), data_jac_.end(), PFJacBlock<sym>{});
+        std::fill(del_x_pq_.begin(), del_x_pq_.end(), ComplexPower<sym>{});
```
figueroa1395: This is more expressive:
```suggestion
        std::ranges::fill(data_jac_, PFJacBlock<sym>{});
        std::ranges::fill(del_x_pq_, ComplexPower<sym>{});
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +228,79 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
 
-        ComplexTensorVector<sym> linear_mat_data(y_bus.nnz_lu());
-        LinearSparseSolverType linear_sparse_solver{y_bus.row_indptr_lu(), y_bus.col_indices_lu(), y_bus.lu_diag()};
-        typename LinearSparseSolverType::BlockPermArray linear_perm(y_bus.size());
+        // Map network admittance to real-domain system
+        IdxVector const& map_lu_y_bus = y_bus.map_lu_y_bus();
+        ComplexTensorVector<sym> const& ydata = y_bus.admittance();
+        for (Idx k = 0; k < data_jac_.size(); ++k) {
```
figueroa1395: The CI is complaining about implicit casting happening here between `Idx` and `data_jac_.size()`, you can easily solve it with `ssize()` or `narrow_cast<Idx>(size())`, it's up to you. It should look like
```suggestion
        for (Idx k = 0; k < data_jac_.ssize(); ++k) {
```
This also happens in other places, so please fix them too.

In addition, please use a more explicit naming for `k`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +228,79 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+    // This implementation reuses the class-level Jacobian matrix (data_jac_) and RHS vector (del_x_pq_)
+    // to eliminate temporary memory allocations.
+    // The complex system (G + jB)(Ur + jUi) = Ir + jIi is mapped to the real system:
+    // [[-B, G], [-G, -B]] [Ui, Ur]^T = [Ir, -Ii]^T
+    // This mapping ensures that H/N/M/L sub-blocks align with the NR Jacobian at flat start.
     void initialize_derived_solver(YBus<sym> const& y_bus, PowerFlowInput<sym> const& input,
                                    SolverOutput<sym>& output) {
-        using LinearSparseSolverType = SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>>;
+        // Reset reused buffers to zero
+        std::fill(data_jac_.begin(), data_jac_.end(), PFJacBlock<sym>{});
```
nitbharambe: It would be nice if you create a `RealLinearCalcBlock` with new getters. Then cast this vector later to one of `PFJacBlock`s. Same logic, but more readable.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +228,79 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+            Idx const k_y_bus = map_lu_y_bus[k];
+            if (k_y_bus != -1) {
+                set_linear_block(data_jac_[k], ydata[k_y_bus]);
+            }
+        }
+
+        IdxVector const& bus_entry = y_bus.lu_diag();
+        auto const& load_gens_per_bus = this->load_gens_per_bus_.get();
+        auto const& sources_per_bus = this->sources_per_bus_.get();
+        for (auto const& [bus_number, load_gens, sources] :
```
figueroa1395, line 270: I see you "replicated" the logic in `detail::prepare_linear_matrix_and_rhs` but for the real domain. The logic looks correct, but can you split this function similarly as the other one? It is rather long and complex. Splitting in something like
```
 for (auto const& [bus_number, load_gens, sources] : enumerated_zip_sequence(load_gens_per_bus, sources_per_bus)) {
Idx const diagonal_position = bus_entry[bus_number];
add_loads(...);
add_sources(...);
...
}
```
or similar would be a good idea.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
nitbharambe: Or maybe just add new getters: real_real(), real_imag(), imag_real(), imag_imag() and rename PFJacBlock -> NRPFBlock

In the same spirit, you can also rename PolarPhasor -> NRPFRhs.
And add new getters i_real(), i_imag()

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +228,79 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+
+        for (Idx i = 0; i != this->n_bus_; ++i) {
+            //single phase
+            if constexpr (is_symmetric_v<sym>) {
+                output.u[i] = {del_x_pq_[i].q(), del_x_pq_[i].p()};
+            } else {
+                output.u[i] = ComplexValue<asymmetric_t>{RealValue<asymmetric_t>{del_x_pq_[i].q()},
+                                                         RealValue<asymmetric_t>{del_x_pq_[i].p()}};
+            }
+        }
```
figueroa1395: I believe you can simply let the templates solve this, and a `range` based `for` loop is more desirable here. Something along the lines of
```suggestion
        for (Idx i : std::views::iota(Idx{}, this->n_bus_) ) {
                output.u[i] = ComplexValue<sym>{RealValue<asymmetric_t>{del_x_pq_[i].q()},
                                                         RealValue<sym>{del_x_pq_[i].p()}};
        }
```

figueroa1395 (commented):
Additional comments and remarks:
- Please take a look at the CI and resolve those complains. I believe I mentioned them in the in-code comments , but double check in case I missed anything.
- CI is complaining about formatting. In order to resolve easily, you can run `uv run pre-commit run` and this will re-format for you, among other things. Moreover, if you build from source, `uv sync` and then `uv run pre-commit run` you will also locally check that the tests pass at the python side; this way you will make sure that your changes in the core make sense throughout the whole pipeline locally.
- DCO check is failing and I see you attempted to fix it by looking at the commit history. Unfortunately, in order to fix it, you must amend all the previous commits and add a sign-off to those as well; having one commit signed off is not enough. This means you need to force push (danger mode). I can't do this for you but if you need a hand just ping me.

Finally, this looks very good! And the (unit and validation) tests all passed on the first try. Great job and thank you for the contribution!

Don't hesitate to ping us if you need or want a hand. We'll take another look once you have resolved our initial comments.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -517,6 +573,25 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+    }
+
+    // Helper to map complex injection I = Y * U to real-domain RHS [Ir, -Ii]^T.
+    static void add_linear_rhs(ComplexPower<sym>& rhs, ComplexTensor<sym> const& y, ComplexValue<sym> const& u) {
+        auto const g = real(y);
+        auto const b = imag(y);
+        auto const ur = real(u);
+        auto const ui = imag(u);
+        rhs.p() += (dot(g, ur) - dot(b, ui)); // Ir
+        rhs.q() -= (dot(b, ur) + dot(g, ui)); // -Ii
```
nitbharambe:
```suggestion
        auto const i_rhs = dot(y, u);
        rhs.i_real() += real(i_rhs);
        rhs.i_imag() += imag(i_rhs);
```
Along with suggestion from https://github.com/PowerGridModel/power-grid-model/pull/1433/changes#r3395596490

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: Thank you for this! data_jac_.ssize() does not compile for me so for the time being I've used std::ssize() on the vector instead

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: Thank you for the suggestion, I have considered this in the latest commit

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: Thank you for the feedback, I've accounted for this in the latest commit

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: Thank you for the suggestion, I agree. This should reflect in my latest commit

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: I think I was confused and initially wanted to copy the NR jacobian structure but I have now reverted the logic to the expression you suggested because I think its better to understand and it might be better for the solver because it is standard in terms of the algebra

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +235,48 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
           bus_types_(y_bus.size(), BusType::pq),
           voltage_regulators_per_load_gen_{std::ref(topo.voltage_regulators_per_load_gen)} {}
 
-    // Initilize the unknown variable in polar form
+    // Initialize the unknown variable in polar form using a linear voltage guess solved in the real domain.
+    // This implementation reuses the class-level Jacobian matrix (data_jac_) and RHS vector (del_x_pq_)
+    // to eliminate temporary memory allocations.
+    // The complex system (G + jB)(Ur + jUi) = Ir + jIi is mapped to the real system:
+    // [[-B, G], [G, B]] [Ui, Ur]^T = [Ir, Ii]^T
```
TonyXiang8787: What is the reason you order the voltage as first imaginary part, then real part?

Why [Ui, Ur], not [Ur, Ui]?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -160,18 +160,45 @@ J.L -= -dQ_cal_m/dV
...
+    using Block<double, sym, false, 2>::operator=;
+
+    // Mapping: Eq0 = Real(I), Eq1 = Imag(I)
+    GetterType<0, 0> i_real() { return this->template get_val<0, 0>(); }
+    GetterType<1, 0> i_imag() { return this->template get_val<1, 0>(); }
+
+    // Mapping: Var0 = Ui, Var1 = Ur
+    GetterType<0, 0> u_imag() { return this->template get_val<0, 0>(); }
+    GetterType<1, 0> u_real() { return this->template get_val<1, 0>(); }
+};
```
TonyXiang8787: I don't think you need a separate class for this.

Instead, just add following getters in the original class ` PolarPhasor`

```
i_real
i_imag
u_real
u_imag
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -160,18 +160,45 @@ J.L -= -dQ_cal_m/dV
...
+    template <int r, int c> using GetterType = typename Block<double, sym, true, 2>::template GetterType<r, c>;
+    using Block<double, sym, true, 2>::Block;
+    using Block<double, sym, true, 2>::operator=;
+
+    // System: [[-B, G], [G, B]] * [Ui, Ur]^T = [Ir, Ii]^T
+    GetterType<0, 0> real_imag() { return this->template get_val<0, 0>(); }
+    GetterType<0, 1> real_real() { return this->template get_val<0, 1>(); }
+    GetterType<1, 0> imag_imag() { return this->template get_val<1, 0>(); }
+    GetterType<1, 1> imag_real() { return this->template get_val<1, 1>(); }
+};
```
TonyXiang8787: Same goes here, you do not need a new class, just add getters to original class `PFJacBlock `

```
real_imag
real_real
imag_imag
imag_real
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -182,21 +209,11 @@ template <symmetry_tag sym> struct PolarPhasor : public Block<double, sym, false
...
-// class for complex power
 template <symmetry_tag sym> using ComplexPower = PolarPhasor<sym>;
 
-// class of pf block
-// block of incomplete power flow jacobian
-// non-diagonal H, N, M, L
-// [ [H = dP/dTheta, N = V * dP/dV],
-// [M = dQ/dTheta = -N, L = V * dQ/dV = H] ]
-// Hij = Gij .* sij - Bij .* cij = L
-// Nij = Gij .* cij + Bij .* sij = -M
```
TonyXiang8787, line 194: Why remove this comments? They are pretty helpful for readers

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +244,49 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
                                    SolverOutput<sym>& output) {
-        using LinearSparseSolverType = SparseLUSolver<ComplexTensor<sym>, ComplexValue<sym>, ComplexValue<sym>>;
+        // Reset reused buffers to zero
+        std::ranges::fill(data_jac_, PFJacBlock<sym>{});
+        std::ranges::fill(del_x_pq_, PolarPhasor<sym>{});
+
+        // Use span views to treat buffers as linear solve calculation blocks
+        auto const jac_view =
+            std::span{reinterpret_cast<RealLinearCalcBlock<sym>*>(data_jac_.data()), data_jac_.size()};
+        auto const rhs_view = std::span{reinterpret_cast<RealLinearCalcRhs<sym>*>(del_x_pq_.data()), del_x_pq_.size()};
```
TonyXiang8787: If you follow my advice to not to define new classes, but add more getters in existing class, you do not need to do ugly `reintepret_cast` here.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: Thank you for the comment, My latest commit should reflect this for both the new wrapper class I had initially created

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: I'm sorry about this, This was completely by accident and my latest commit recovers this

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: My main motivation to go for this for because I formulated my admittance matrix as -B G G B instead of G -B B G which I presume is usually what is done. For an initial guess, the jacobian matrix H N M L. G would appear in N and M so with Ui, Ur for linear guesses I want to always map G to the off-diagonal. This was based on my initial understanding of the codebase and the NR solver, would be happy to revert the logic to the standard Ur, Ui pattern if that makes the code more readable and is preferred

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
figueroa1395: My mistake, `std::ssize()` is what I meant!

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
TonyXiang8787: I propose we make the order as regular complex number order, so Ur and then Ui

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
figueroa1395: Resolving in favor of https://github.com/PowerGridModel/power-grid-model/pull/1433#discussion_r3401471612

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -432,6 +470,45 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+            //                   coeff of Ur in Imag Eq = B = -imag(Y_load) = -Q
+            auto const y_load = -conj(s);
+            add_diag(block.real_imag(), -imag(y_load));
+            add_diag(block.real_real(), real(y_load));
+            add_diag(block.imag_imag(), real(y_load));
+            add_diag(block.imag_real(), imag(y_load));
+        }
+    }
+
+    void add_linear_initial_guess_sources(IdxRange const& sources, Idx bus_number, PFJacBlock<sym>& block,
```
figueroa1395: `bus_number` is no longer used as pointed by the CI.
```suggestion
    void add_linear_initial_guess_sources(IdxRange const& sources,  PFJacBlock<sym>& block,
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: Sure, I can work on getting that done

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -180,9 +181,14 @@ template <symmetry_tag sym> struct PolarPhasor : public Block<double, sym, false
...
     GetterType<1, 0> q() { return this->template get_val<1, 0>(); }
+
+    // linear solve path getters: x0 = Ur, x1 = Ui and Eq0 = Real, Eq1 = Imag
+    GetterType<0, 0> i_real() { return this->template get_val<0, 0>(); }
+    GetterType<1, 0> i_imag() { return this->template get_val<1, 0>(); }
+    GetterType<0, 0> u_real() { return this->template get_val<0, 0>(); }
+    GetterType<1, 0> u_imag() { return this->template get_val<1, 0>(); }
 };
 
-// class for complex power
```
TonyXiang8787, line 185: Please keep this comment, this is just a type alias to help readability

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -212,9 +225,8 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
   public:
     using sym = sym_type;
 
-    using SparseSolverType = SparseLUSolver<PFJacBlock<sym>, ComplexPower<sym>, PolarPhasor<sym>>;
-    using BlockPermArray =
-        typename SparseLUSolver<PFJacBlock<sym>, ComplexPower<sym>, PolarPhasor<sym>>::BlockPermArray;
+    using SparseSolverType = SparseLUSolver<PFJacBlock<sym>, PolarPhasor<sym>, PolarPhasor<sym>>;
+    using BlockPermArray = typename SparseLUSolver<PFJacBlock<sym>, PolarPhasor<sym>, PolarPhasor<sym>>::BlockPermArray;
```
TonyXiang8787: Please keep the original definition. Although `ComplexPower` and `PolarPhasor` is the same thing, use two names expresses the intent.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,23 +240,49 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
-                                              this->sources_per_bus_.get(), output, linear_mat_data);
-        linear_sparse_solver.prefactorize_and_solve(linear_mat_data, linear_perm, output.u, output.u);
+        for (Idx const i : std::views::iota(Idx{}, this->n_bus_)) {
+            output.u[i] =
+                ComplexValue<sym>{RealValue<sym>{del_x_pq_[i].u_real()}, RealValue<sym>{del_x_pq_[i].u_imag()}};
+        }
 
         set_u_ref_and_bus_types(input, output.u);
 
-        // get magnitude and angle of start voltage
```
TonyXiang8787, line 247: Restore this comment

TonyXiang8787:
Hi @akshatowl,

The PR is in good shape. Please fix the last CI build errors. Then we can consider to merge.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
figueroa1395: @akshatowl I believe this is the only thing left for you to address (to fix CI) and we shall be good to merge the PR :)

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -432,6 +473,45 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+                                        PowerFlowInput<sym> const& input) {
+        for (Idx const load_number : load_gens) {
+            auto const& s = input.s_injection[load_number];
+            // Y_load = P - jQ -> G=P, B=-Q
+            // System: [[G, -B], [B, G]] [Ur, Ui]^T = [Ir, Ii]^T
+            // Diagonal mapping: coeff of Ur in Real Eq = G = real(Y_load) = P
+            //                   coeff of Ui in Real Eq = -B = imag(Y_load) = Q
+            //                   coeff of Ur in Imag Eq = B = -imag(Y_load) = -Q
+            //                   coeff of Ui in Imag Eq = G = real(Y_load) = P
+            auto const y_load = -conj(s);
```
TonyXiang8787: `auto const` with `eigen` expression template can be tricky. You get a proxy object instead of real copied value.

You need to define it as the real type:

```suggestion
            ComplexValue<sym> const y_load = -conj(s);
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -517,6 +597,24 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
             data_jac_[diagonal_position].l() += block_mm.l();
         }
     }
+
+    // Helper to map complex admittance to real-domain block matrix using PFJacBlock getters.
+    // This mapping ensures G aligns with the N/M sub-blocks as in the NR Jacobian.
+    static void set_linear_block(PFJacBlock<sym>& block, ComplexTensor<sym> const& y) {
+        auto const g = real(y);
+        auto const b = imag(y);
```
TonyXiang8787: Same here

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -517,6 +597,24 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+        block.real_imag() = -b;
+        block.real_real() = g;
+        block.imag_imag() = g;
+        block.imag_real() = b;
+    }
+
+    // Helper to map complex injection I = Y * U to real-domain RHS [Ir, Ii]^T using PolarPhasor getters.
+    static void add_linear_rhs(PolarPhasor<sym>& rhs, ComplexTensor<sym> const& y, ComplexValue<sym> const& u) {
+        auto const i_rhs = dot(y, u);
```
TonyXiang8787: same here

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: That makes sense, I can work on this. I just wanted to clarify something. For this particular instance, shouldn't the type to use be `ComplexValue<sym>` ? Since we actually extract the real and imaginary parts in the `add_diag `method below this

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: @figueroa1395 , Sorry about this, for some reason the diff for this change didn't land before. I have removed it from the function signature and it should reflect in the commits now.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
figueroa1395: I believe you are right for this case. I've triggered the CI run and it should show if there's any issue. In addition, don't forget to address this also in the places mentioned at the comments below (https://github.com/PowerGridModel/power-grid-model/pull/1433#discussion_r3411768533 and https://github.com/PowerGridModel/power-grid-model/pull/1433#discussion_r3411769278)

TonyXiang8787: Sorry, it should be indeed `ComplexValue`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
```diff
@@ -228,19 +242,46 @@ class NewtonRaphsonPFSolver : public IterativePFSolver<sym_type, NewtonRaphsonPF
...
+            Idx const diagonal_position = bus_entry[bus_number];
+            add_linear_initial_guess_loads(load_gens, data_jac_[diagonal_position], input);
+            add_linear_initial_guess_sources(sources, data_jac_[diagonal_position], del_x_pq_[bus_number], y_bus,
+                                             input);
+        }
 
-        ComplexTensorVector<sym> linear_mat_data(y_bus.nnz_lu());
-        LinearSparseSolverType linear_sparse_solver{y_bus.row_indptr_lu(), y_bus.col_indices_lu(), y_bus.lu_diag()};
-        typename LinearSparseSolverType::BlockPermArray linear_perm(y_bus.size());
+        // Solve: [Ui, Ur] result in [p, q]
```
TonyXiang8787: This comment needs to adjust the order

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/newton_raphson_pf_solver.hpp
akshatowl: Got it, thank you! The latest commit should address all of the comments that were open

TonyXiang8787 (approved):
Hi @akshatowl,

Thanks again for your great contribution. The PR is ready to go into the merge queue.

PR #1434, opened 2026-06-12
Node injection sensor deprecation: future behavior if experimental

mgovers:
Part of #1334 

Add future behavior for deprecation of node injection sensor as an experimental feature.
This is a step towards introducing a feature toggle for merging nodes on links, but it **does not yet introduce** the feature toggle.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -409,6 +409,18 @@ class MainModelImpl {
         }
     }
 
+    void check_no_future_deprecations(Options const& /*options*/, ConstDataset const* /*batch_dataset*/) const {
+        ModelType::run_functor_with_all_component_types_return_void([this]<typename CT>() {
+            if constexpr (requires(CT c) { c.get_terminal_type(); }) {
+                if (std::ranges::any_of(state_.components.template citer<CT>(), [](auto const& sensor) {
+                        return sensor.get_terminal_type() == MeasuredTerminalType::node;
+                    })) {
+                    throw InvalidMeasuredTerminalType{MeasuredTerminalType::node, CT::name};
+                }
+            }
+        });
+    }
+
   private:
     template <solver_output_type SolverOutputType>
     void output_result(MathOutput<std::vector<SolverOutputType>> const& math_output, MutableDataset const& result_data,
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
@@ -413,8 +413,8 @@ class MainModelImpl {
         ModelType::run_functor_with_all_component_types_return_void([this]<typename CT>() {
             // only flow sensors have get_terminal_type() so we can filter on it
             if constexpr (requires(CT c) {
-                { c.get_terminal_type() } -> std::same_as<MeasuredTerminalType>;
-            }) {
+                              { c.get_terminal_type() } -> std::same_as<MeasuredTerminalType>;
+                          }) {
                 if (std::ranges::any_of(state_.components.template citer<CT>(), [](auto const& sensor) {
                         return sensor.get_terminal_type() == MeasuredTerminalType::node;
                     })) {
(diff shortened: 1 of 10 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -409,6 +409,18 @@ class MainModelImpl {
         }
     }
 
+    void check_no_future_deprecations(Options const& /*options*/, ConstDataset const* /*batch_dataset*/) const {
+        ModelType::run_functor_with_all_component_types_return_void([this]<typename CT>() {
+            if constexpr (requires(CT c) { c.get_terminal_type(); }) {
+                if (std::ranges::any_of(state_.components.template citer<CT>(), [](auto const& sensor) {
+                        return sensor.get_terminal_type() == MeasuredTerminalType::node;
+                    })) {
+                    throw InvalidMeasuredTerminalType{MeasuredTerminalType::node, CT::name};
```
figueroa1395, line 421: It's abusing the notation, but would it make sense to do something like
```suggestion
                    throw InvalidMeasuredTerminalType{std::format("Deprecation error: {}", MeasuredTerminalType::node), CT::name};
```
Just to further distinguish it a bit more without introducing anything else? Like a small convention for these rare deprecation errors.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -409,6 +409,18 @@ class MainModelImpl {
         }
     }
 
+    void check_no_future_deprecations(Options const& /*options*/, ConstDataset const* /*batch_dataset*/) const {
+        ModelType::run_functor_with_all_component_types_return_void([this]<typename CT>() {
+            if constexpr (requires(CT c) { c.get_terminal_type(); }) {
```
figueroa1395: At first I didn't realize how would this check distinguish flow sensors from voltage sensors and realized that `get_terminal_type` is only part of the former. Can you add a short comment for easy pinpoint?

figueroa1395 (commented):
Ready to approve after small remarks are resolved.

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers: i personally think it would just convolute things even more. people enable `experimental_features`, they might as well deal with it

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
mgovers:
```suggestion
            // only flow sensors have get_terminal_type() so we can filter on it
            if constexpr (requires(CT c) { { c.get_terminal_type() } -> std::same_as<MeasuredTerminalType>; }) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -409,6 +409,19 @@ class MainModelImpl {
         }
     }
 
+    void check_no_future_deprecations(Options const& /*options*/, ConstDataset const* /*batch_dataset*/) const {
+        ModelType::run_functor_with_all_component_types_return_void([this]<typename CT>() {
+            // only flow sensors have get_terminal_type() so we can filter on it
+            if constexpr (requires(CT c) { { c.get_terminal_type() } -> std::same_as<MeasuredTerminalType>; }) {
```
mgovers:
```suggestion
            if constexpr (requires(CT c) {
                { c.get_terminal_type() } -> std::same_as<MeasuredTerminalType>;
            }) {
```

power_grid_model_c/power_grid_model/include/power_grid_model/main_model_impl.hpp
```diff
@@ -409,6 +409,21 @@ class MainModelImpl {
         }
     }
 
+    void check_no_future_deprecations(Options const& /*options*/, ConstDataset const* /*batch_dataset*/) const {
+        ModelType::run_functor_with_all_component_types_return_void([this]<typename CT>() {
+            // only flow sensors have get_terminal_type() so we can filter on it
+            if constexpr (requires(CT c) {
+                { c.get_terminal_type() } -> std::same_as<MeasuredTerminalType>;
+            }) {
```
mgovers:
```suggestion
            if constexpr (requires(CT c) {
                              { c.get_terminal_type() } -> std::same_as<MeasuredTerminalType>;
                          }) {
```

figueroa1395 (approved):

PR #1439, opened 2026-06-15
Remove Redundant Overload

furqan463:
The `calculate_power_flow()` overloads at line 473 and 487 both are similar except that the second one has argument `update_data: None = ...,`, while the first one omits it. However, the `update_data: None = ...,` covers the case of not passing the `update_data` at all.

Reference to the comment https://github.com/PowerGridModel/power-grid-model/pull/1178#discussion_r3411489945
- [x] IDE Autocomplete working
- [x] mypy run passed

### DIFF
--- a/src/power_grid_model/_core/power_grid_model.py
+++ b/src/power_grid_model/_core/power_grid_model.py
@@ -478,20 +478,6 @@ class PowerGridModel:
         error_tolerance: float = ...,
         max_iterations: int = ...,
         calculation_method: CalculationMethod | str = ...,
-        threading: int = ...,
-        output_component_types: None | set[ComponentTypeVar] | list[ComponentTypeVar] = ...,
-        continue_on_batch_error: bool = ...,
-        decode_error: bool = ...,
-        tap_changing_strategy: TapChangingStrategy | str = ...,
-    ) -> SingleRowBasedDataset: ...
-    @overload
-    def calculate_power_flow(
-        self,
-        *,
-        symmetric: bool = ...,
-        error_tolerance: float = ...,
-        max_iterations: int = ...,
-        calculation_method: CalculationMethod | str = ...,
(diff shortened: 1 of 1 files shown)


figueroa1395 (commented):
I see that there are test cases in place that already cover this. If CI doesn't complain, this is good to go then. I'll approve once the checks are green.

src/power_grid_model/_core/power_grid_model.py
```diff
@@ -470,20 +470,6 @@ def _calculate_short_circuit(  # noqa: PLR0913
...
-        symmetric: bool = ...,
-        error_tolerance: float = ...,
-        max_iterations: int = ...,
-        calculation_method: CalculationMethod | str = ...,
-        threading: int = ...,
-        output_component_types: None | set[ComponentTypeVar] | list[ComponentTypeVar] = ...,
-        continue_on_batch_error: bool = ...,
-        decode_error: bool = ...,
-        tap_changing_strategy: TapChangingStrategy | str = ...,
-    ) -> SingleRowBasedDataset: ...
```
mgovers, line 486: can you please double-check for any similar redundant overloads in `calculate_state_estimation` and `calculate_short_circuit`?

furqan463: no redundant overloads in `calculate_state_estimation` and `calculate_short_circuit`. All of their overloads have distinct output types.

mgovers:
> I see that there are test cases in place that already cover this. If CI doesn't complain, this is good to go then. I'll approve once the checks are green.

`@overload` is semantics only. we do not have tests that check for the existence of overloads explicitly. The code runs through the implementation only. Only `mypy` may catch missing implementations if and when hit.

TonyXiang8787 (approved):

PR #1445, opened 2026-06-23
Use release bot and environment for release in main

TonyXiang8787:


.github/workflows/build-test-release.yml
```diff
@@ -299,18 +299,11 @@ jobs:
         build-and-test-python,
         build-and-test-conda,
       ]
-    # always run publish job but fail at the first step if previous jobs have been failed
-    if: always()
+    if: (inputs.create_release)
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
+    environment: release
```
mgovers, line 304: shouldn't this obtain its environment from the current environment (e.g. `environment: ${{ vars.ENVIRONMENT_STAGE }}`)

TonyXiang8787: It's the other way around. You specify you have to use `release` environment. If it is not allowed due to branch or other rules, the job will fail.

TonyXiang8787: See here
https://docs.pypi.org/trusted-publishers/using-a-publisher/

.github/workflows/ci.yml
```diff
@@ -77,8 +76,8 @@ jobs:
     name: Publish to PyPI
     runs-on: ubuntu-latest
     permissions:
-      contents: write
       id-token: write # Required for Trusted Publishing
+    environment: release
     needs: build-test-release
     if: (github.event_name == 'workflow_dispatch') || github.event_name == 'push'
```
vincentkoppen, line 82: Do we also want to limit it here to only on the `main` branch for workflow_dispatch? 
Or is the workflow_dispatch even still needed, since the release bot only works for `main` right? 

Currently I'm allowed to trigger this action on a feature branch and maybe get access to any of the secrets in the release environment if there is any compromise in the code?

.github/workflows/build-test-release.yml
mgovers: ok

.github/workflows/ci.yml
TonyXiang8787: Please look at context. For PGM repo, you can run the whole CI with `create_release` set to `false` on any branch with workflow_dispatch.

But only when you run `workflow_dispatch` on `main`, you are allowed to set `create_release` set to `true`

mgovers: nowhere does it say this in the checkboxes and there are also no checks inbetween that raise an error if `create_release: true` on a branch that is not `main`. "allowed" can be interpreted in 2 ways: "should not" and "can't"

nitbharambe:
The changes look alright given the secrets are set up.

.github/workflows/ci.yml
TonyXiang8787: When you run the CI in non-main with create release to true, when the release job tries to fetch the environment `release`, it will fail the job.

That's what GA says. But let's see.

vincentkoppen: Ah check, didn't know that `release` env was limited to a branch. 

So the options to create a pre-release is removed entirely (which is fine from my point of view): 
- feature branches can't access the `release` environment and/or release bot
- a workflow dispatch with create_release is true on the main branch, will become a regular release: 
https://github.com/PowerGridModel/power-grid-model/blob/9c942bd887fe9714e7301b5e9519f49b7a345960/.github/workflows/build-test-release.yml#L343

So my preference would be to make this clear from the options instead of failing because an environment isn't allowed to be used. To do so, we could remove the boolean from the workflow call and change the call below to only be true if it is a push event
https://github.com/PowerGridModel/power-grid-model/blob/9c942bd887fe9714e7301b5e9519f49b7a345960/.github/workflows/ci.yml#L36-L44


If we really want to keep to option to manually trigger the action on `main` for a new release, we could of course keep the boolean. Then I would say that we: 
- rename the variable to not include (pre-) anymore and maybe even mention it only works on `main`
- Change the above workflow call to include a check that we are on main to be more explicit in the skip (instead of fail/error)

TonyXiang8787: Yes, all that should be improved and clarified. For now, this PR is about get thing working without blockage.

TonyXiang8787:
Manual trigger run in https://github.com/PowerGridModel/power-grid-model/actions/runs/28027781585

mgovers (approved):
experiment in https://github.com/PowerGridModel/power-grid-model/actions/runs/28027781585/job/82965801464 was successful

TonyXiang8787:
The release does not work:
https://github.com/PowerGridModel/power-grid-model/actions/runs/28044882907/job/83026915088

Environment secrets cannot be easily used in reusable workflow.
https://github.com/actions/runner/issues/1490

 We need to factor the github release step out.

Issue #739, opened 2024-09-26
[FEATURE] asymmetric implementation of generic_branch


petersalemink95:
For the discussions about `generic_branch` see #729.

The asymmetric implementation is more difficult and needs more thinking. Some options could be:

- Provide r1, x1, b1, g1, r2, x2, b2, g2 for positive and negative sequence and y_ff, y_ft, y_tf, y_tt for zero sequence
- Provide r_series / x_series and g/b_shunt_from/to
- Provide r1, x1, b1, g1, r2, x2, b2, g2, r0, x0, b0, g0, so provide the PI model for all sequences: this way only YnYn transformers can be calculated in asym calculation

mgovers:
maybe `GenericLine` and `GenericTransformerYNyn`, ...?

TonyXiang8787:
Hi @sudo-ac, thanks for the suggestion. 

If we define so many `GenericBranch*`, I wonder what generic we do have here. We really need to think hard about how to implement asymmetric generic branch. @petersalemink95, please also add this option in the list of choices to make.

For now, is it sufficient for you @sudo-ac to first implement the symmetric `GenericBranch` for your own use-cases? If yes, we are looking forward to seeing a PR about that.

petersalemink95:
@TonyXiang8787 added it as an option to this issue.

@sudo-ac Thanks for the suggestion! As @TonyXiang8787  mentioned we might need to think a bit longer about the best solution. Would just implementing symmetric calculations for now suit your need? Your proposal would still be valid if we decide to implement that in a later stage if we implement only the symmetric calculations now.

petersalemink95:
Thanks for all the input @sudo-ac!
We're looking forward to reviewing the symmetric PR! (But enjoy your holiday first ;) )

We will include your latest suggestions in further discussions about the asymmetric generic branch.

petersalemink95:
Hi @sudo-ac,

Thanks for working out your idea!
I would like to propose to wait with further investigation of asymmetric generic branch until a user presents itself. 
That way we can design the correct implementation including the perspective of the actual user.

Let's keep this issue open, so further discussions can take place here.

Issue #733, opened 2024-09-24
[FEATURE] Ideal source handling


mgovers:
## Describe the feature request

An ideal source is represented by having infinite short circuit power (`sk`). However, the current implementation of the power-grid-model will raise a validation error when `sk == inf`:
![image](https://github.com/user-attachments/assets/202ef196-f96e-4b77-8f7a-999088f76ce7)

This ticket is about supporting ideal sources for power flow and state estimation calculation types. (And undefined behavior for short circuit calculations.)

## Background

When `sk` is extremely large but not infinite (e.g. `sk == 10e50`), the power flow and state estimation calculation types are already numerically stable, cfr. #458 . Unfortunately, this value is arbitrarily set. We currently require our users to provide such an arbitrarily large value when they desire an ideal source. Even more so, within the PGM, intermediate calculations may cause an arbitrarily large value to become infinite, e.g.:

* `10e200 * 10e200 = 10e400 > std::numeric_limits<double>::max() (~10e306)` => `10e200 * 10e200` will result in `inf`

A single such `inf` value in the internal calculations will result in `NaN` values in the output, i.e. nonsense output.

## Desired solution

* [ ] In the C++ implementation of the Source component type in https://github.com/PowerGridModel/power-grid-model/blob/main/power_grid_model_c/power_grid_model/include/power_grid_model/component/source.hpp :
  * [ ] Create a `static constexpr double ideal_source_sk = 10e50;` constant.
    * [ ] Note that this is arbitrarily chosen (mention this in a comment)
    * [ ] This value is so large that it does not affect numerical stability for practical use cases, while still being small enough to never cause infinities and NaN values in the calculations. (please verify with simple examples)
  * [ ] In the constructor of the Source component type, set `sk` to the minimum of the user-specified `sk` and `ideal_source_sk`.
    * [ ] Note that this is only correct for power flow and state estimation
    * [ ] For short circuit calculations, the `sk` heavily influences the calculation result, and we consider this to be undefined behavior (UB; see below)
* [ ] Update the validation cases
  * [ ] Support `sk == inf` for power flow and state estimation
  * [ ] Do not support `sk == inf` and `sk > 10e50` for short circuit calculations due to deliberately UB
* [ ] In the documentation on the Source component type : 
  * [ ] On the input attributes: https://github.com/PowerGridModel/power-grid-model/blob/main/docs/user_manual/components.md?plain=1#L388
    * [ ] Mention that `sk == inf` represents an ideal source
[...]

Issue #489, opened 2024-02-08
[FEATURE] single phase load/generation


TonyXiang8787:
# Background

We now have `asym_load` and `asym_gen` to represent asymmetric load and generation. If user want to connect a single phase load, he/she still needs to define `p_specified` for all three phases, two of which are zeros. This is waste of memory and calculation.

# Proposal

Create a `single_phase_load` and `single_phase_gen` component. Allow user to specify a phase code and single value of active/reactive power.

petersalemink95:
Besides saving memory, as mentioned in the issue by @TonyXiang8787, there is another use case: state estimation with single phase loads. This will mainly happen on the LV grid. Currently, Power Grid Model cannot handle single phase loads in state estimation.

If we want to be able to perform state estimation on grids with single phase loads we would need the following:
- single phase load/generation (this issue)
- single phase voltage sensor #933 and/or single phase power sensor #490

Issue #188, opened 2023-01-09
Phase shifter transformer


TonyXiang8787:

TonyXiang8787:
This is a request by `pandapower`. They have so-called ideal shifter transformer, where the phase shift will change by changing the tap position.

https://pandapower.readthedocs.io/en/latest/elements/trafo.html

This will be very rare in distribution grids. We highly doubt if we are ever going to implement this.

scud-soptim:
I believe we can conclude that PSTs can be mapped using the Generic Branch.

petersalemink95:
@scud-soptim awesome! Do you maybe have a PST equivalent of a GenericBranch that we could implement as a validation test? We would leave this issue open, in case anyone wants a dedicated PST component, but would be great if we can already use GenericBranch :)

scud-soptim:
@petersalemink95 PSTs are documented here: https://power-grid-model.readthedocs.io/en/stable/examples/Generic%20Branch%20Example.html

mgovers:
@scud-soptim I am looking at the example in https://power-grid-model.readthedocs.io/en/stable/examples/Generic%20Branch%20Example.html#generic-branch-representation-for-line-transformer-and-pst and I noticed the following:

The PST seems to not be an ideal phase shifter, because it has a non-zero reactance. I also do not think that PGM currently actually supports an ideal phase shifter, for the same reasons that an ideal transformer needs to be modeled as a link (see also #1456). Is this something that should be added to the example as a sidenote?

scud-soptim:
Good point.

I agree that the PST in the example should not be interpreted as an ideal phase shifter. It is modeled as a Generic Branch with non-zero impedance, so it represents a practical PST model rather than an ideal phase-shifting element.

I would avoid saying that an ideal phase shifter is currently supported through the Generic Branch, because setting the series impedance to zero is likely not a valid or numerically well-defined branch representation. This is similar to the discussion around ideal transformers in #1456.

So yes, adding a short sidenote to the example would be helpful: the example represents a practical PST with impedance, not an ideal phase shifter.

Welthulk:
I would like to clarify the idea of extending `GenericBranch`.

I do **not** mean modelling an ideal phase shifter or ideal transformer by setting the branch impedance to zero. That would not be a well-defined admittance-branch representation, because the normal branch stamp relies on

```text
y = 1 / z
```

and for `z = 0` this becomes singular.

[...]

PR #1456, opened 2026-06-30
Docs: explain how to model grid component types not provided by PGM

mgovers:
This documents how to model some of the component types not explicitly provided by PGM:

* [x] Ideal transformer (modeled as link)
* [x] Grounding transformer/earthing transformer (modeled as shunt)
* [x] Choke coil/reactance coil (modeled as line, explicitly referring to #739 regarding as to why the recommendation is to model it as line and not as generic branch)
* [x] Cable (trivially modeled as line)

## Potential extensions

If you want me to, I can also add other components to the list (many of which can be directly taken from PGM-IO: https://github.com/PowerGridModel/power-grid-model-io/blob/main/src/power_grid_model_io/config/excel/vision_en.yaml):

* [ ] Phase shifter transformer (#188)
* [ ] Transformer loads (modeled using a transformer, a node and a load/gen)
* [ ] Synchronous generators (identical to `sym_gen`)
* [ ] Wind turbines (trivially modeled as `sym_gen`)
* [ ] PVs (trivially modeled as `sym_gen`)
* [ ] Capacitor (modeled as shunt)
* [ ] Reactor (modeled as shunt)
* [ ] Single-phase loads (#489 , but notice https://github.com/PowerGridModel/power-grid-model/issues/489#issuecomment-2751247228 which states that this modeling will not work for state estimation, so that would require an explicit `{warning}` statement)
* [ ] Ideal source (this would put the handling on the user-side, rather than internally in PGM as per #733)
* [ ] Compound load (modeled as 3 separate loads representing the individual ZIP components, rather than to implement it internally in PGM as per #168)

or i can create a follow-up ticket with this

### DIFF
--- a/docs/user_manual/non-pgm-components.md
+++ b/docs/user_manual/non-pgm-components.md
@@ -1,0 +1,64 @@
+<!--
+SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+
+SPDX-License-Identifier: MPL-2.0
+-->
+
+# Grid Component Modeling
+
+```{note}
+This document is about components not explicitly supported by PGM that can still be modeled using PGM components.
+For documentation on the components supported by PGM, please refer to [Components](./components.md).
+```
+
+The power grid consists of many different component types and the list is ever-increasing.
+As a result, it would be an impossible task to make the list of components supported by PGM exhaustive.
+Moreover, many grid component types share electrical attributes, so that one such grid component may be modeled as
+another.
+This document contains a non-exhaustive list of such modeling examples.
+New contributions and ideas for modeling grid components are very welcome!
+
+## Ideal transformer
+
+An ideal transformer can be modeled as a [link](./components.md#link). <!-- markdownlint-disable-line descriptive-link-text line-length -->
+To this end, it is explicitly allowed to place a link between two different voltage levels.
+Trivially, no additional electrical parameters need to be specified.
+
+```{note}
+The [transformer](./components.md#transformer) component as provided by the PGM models a real transformer in the grid,
+i.e., with finite impedance.
+This directly conflicts with the properties of an ideal transformer, so the PGM
+[transformer](./components.md#transformer) cannot be used to model ideal transformers.
+```
+
+## Grounding transformer
+
+A grounding transformer (a.k.a. earthing transformer) can be modeled as a [shunt](./components.md#shunt) with only
+zero-sequence parameters (and positive-sequence parameters set to $0$).
+In particular, a grounding transformer with resistance $r$ and reactance $x$ can be modeled as follows.
+
+$$
+\begin{aligned}
+g_1 &= 0 \\
+b_1 &= 0 \\
+g_0 &= \Re\left\{\frac{1}{r + j x}\right\} \\
+b_0 &= \Im\left\{\frac{1}{r + j x}\right\}
+\end{aligned}
+$$
+
+## Choke coil
+
+A choke coil (a.k.a. reactance coil) can be modeled as a [line](./components.md#line) with its shunt parameters
+(`c1`, `tan1`, `c0`, `tan0`) set to $0$.
+
+```{note}
+A choke coil can also be modeled using a [generic branch](./components.md#generic-branch).
+However, at the time of writing (June 2026; see
+[PGM issue #739](https://github.com/PowerGridModel/power-grid-model/issues/739)), this component does not yet support
+zero-sequence parameters. As a result, asymmetric calculations are not supported, and this model currently only works
+for symmetric calculations.
+```
+
+## Cable
+
+A cable is equivalent to a line.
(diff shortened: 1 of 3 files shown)


docs/user_manual/grid-component-modeling.md
```diff
@@ -0,0 +1,69 @@
+# Grid Component Modeling
+
+```{note}
+This document is about components not explicitly/directly supported by PGM that can still be modeled using PGM
```
figueroa1395: Using "explicitly" is enough. Adding "directly" made me wonder if we do support them or not in the end. It creates confusion.

```suggestion
This document is about components not explicitly supported by PGM that can still be modeled using PGM
```

docs/user_manual/non-pgm-components.md
figueroa1395, line 1: I'd just call this file `other-components` and the same for the title in the docs themselves.

docs/user_manual/grid-component-modeling.md
```diff
@@ -0,0 +1,69 @@
...
+g_1 &= 0 \\
+b_1 &= 0 \\
+g_0 &= \Re\left\{\frac{1}{r + j x}\right\} \\
+b_0 &= \Im\left\{\frac{1}{r + j x}\right\}
+\end{aligned}
+$$
+
+## Choke coil
+
+A choke coil (a.k.a. reactance coil) can be modeled as a line without shunt parameters:
```
figueroa1395: You are missing `line` and `shunt` hyperlinks.

docs/user_manual/grid-component-modeling.md
```diff
@@ -0,0 +1,69 @@
...
+
+```{note}
+A choke coil can also be modeled using a [generic branch](./components.md#generic-branch).
+However, at the time of writing (June 2026), [generic branch](./components.md#generic-branch) does not yet support
+zero-sequence parameters, so this model would only work for symmetric calculations.
+```
+
+## Cable
+
+A cable can be trivially modeled as a line.
```
figueroa1395: This can be read as you should know how to do this. Maybe a better phrasing is: a cable is equivalent to a line in PGM. Or something like that.

docs/user_manual/grid-component-modeling.md
figueroa1395: Aren't `g` and `b` the shunt parameters according to above? If not, the phrasing is a bit confusing.

figueroa1395 (commented):
Math looks fine, but perhaps an eye from @nitbharambe  is needed to double check the EE stuff.

About the other components:
> Phase shifter transformer (https://github.com/PowerGridModel/power-grid-model/issues/188)

There is already an example which documents how to model it via a generic branch, so let's add it.

For the rest of the mentioned components, since there isn't yet a mention in PGM nor a request to add them (correct me if I'm wrong), or there are still implementation steps missing, let's scope them out.

docs/user_manual/grid-component-modeling.md
mgovers: changed to `with its shunt parameters set to $0$`. this is about `c1`, `tan1`, `c0` and `tan0`

mgovers: added `line` hyperlink. `shunt` should not have a hyperlink here because it is not about the parameters of a `shunt`, but about the parameters of the line that are `shunt` (`c1`, `tan1`, `c0`, `tan0`)

docs/user_manual/non-pgm-components.md
mgovers: i wonder... "other components" suggests that this is a continuation of the `components.md` but it's not: it's a document that lives in parallel.

maybe: `non-pgm-components`?

mgovers:
> Math looks fine, but perhaps an eye from @nitbharambe is needed to double check the EE stuff.
> 
> About the other components:
> 
> > Phase shifter transformer (#188)
> 
> There is already an example which documents how to model it via a generic branch, so let's add it.
> 
> For the rest of the mentioned components, since there isn't yet a mention in PGM nor a request to add them (correct me if I'm wrong), or there are still implementation steps missing, let's scope them out.

regarding the PST: cfr. https://github.com/PowerGridModel/power-grid-model/issues/188#issuecomment-4851056384 , the PST described by the Generic Branch example is not an ideal PST (non-zero reactance). Do you still think it's a good idea to add it?

docs/user_manual/non-pgm-components.md
figueroa1395: What about `alien-components`? it sounds too esoteric though. `non-pgm-components` is fine.

figueroa1395:
>regarding the PST: cfr. https://github.com/PowerGridModel/power-grid-model/issues/188#issuecomment-4851056384 , the PST described by the Generic Branch example is not an ideal PST (non-zero reactance). Do you still think it's a good idea to add it?

Let's then add it later if needed once we get a response. For now, scope it out.

figueroa1395 (approved):

PR #1461, opened 2026-07-06
fix: allow zero mutual asym-line values

### LINKED ISSUE #1460: [BUG] Allow mutual impedances to be 0  in asym line data validator
### Describe the bug

Our asym lines lines via validator do not allow mutual impedances to be 0.  Practically, they can be equal to 0

- [ ] Change greater_than_zero to all_greater_than_or_equal_to_zero in validate_asym_line for mutual r / x / c params.  Change corresponding tests.
- [ ] Change documentation in components.md

Note that our calculation core does allow this.
A way to bypass it while still using validator would be to use `validate_input_data` or `validate_batch_data`, then filter all `NotGreaterThanError` errors with field in `AT.r_ba,AT.r_ca,AT.r_cb,AT.x_ba,AT.x_ca,AT.x_cb,AT.r_na,AT.r_nb,AT.r_nc,<s>AT.r_nn</s>,AT.x_na,AT.x_nb,AT.x_nc,<s>AT.x_nn</s>,AT.c_ba,AT.c_ca,AT.c_cb`

looooown2006:
## Summary

- allow zero-valued mutual asym-line resistance/reactance entries while keeping self terms strictly positive
- allow zero-valued mutual asym-line capacitance matrix entries while keeping diagonal capacitance and `c0`/`c1` strictly positive
- update the asym-line component docs and validation tests for the new `>= 0` mutual-field contract

Closes #1460

## Validation

- `python -m py_compile src/power_grid_model/validation/_validation.py tests/unit/validation/test_input_validation.py`
- `git diff --check`
- AST parse of the changed Python files

I also attempted the focused pytest path locally, but this checkout needs the compiled `power_grid_model_c.dll`. A direct source run failed at import with the missing shared library, and an editable install did not get past build dependency setup on this machine after several minutes without output.

### DIFF
--- a/tests/unit/validation/test_input_validation.py
+++ b/tests/unit/validation/test_input_validation.py
@@ -1111,30 +1111,30 @@ def test_asym_line_input_data(input_data):
     validation_errors = validate_input_data(input_data, symmetric=True)
     assert validation_errors is not None
     assert NotGreaterThanError(CT.asym_line, AT.r_aa, [55], 0) in validation_errors
-    assert NotGreaterThanError(CT.asym_line, AT.r_ba, [55], 0) in validation_errors
+    assert NotGreaterOrEqualError(CT.asym_line, AT.r_ba, [55], 0) in validation_errors
     assert NotGreaterThanError(CT.asym_line, AT.r_bb, [55], 0) in validation_errors
-    assert NotGreaterThanError(CT.asym_line, AT.r_ca, [55], 0) in validation_errors
-    assert NotGreaterThanError(CT.asym_line, AT.r_cb, [55], 0) in validation_errors
+    assert NotGreaterOrEqualError(CT.asym_line, AT.r_ca, [55], 0) in validation_errors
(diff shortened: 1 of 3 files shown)


tests/unit/validation/test_input_validation.py
```diff
@@ -1189,6 +1189,33 @@ def test_asym_line_input_data(input_data):
     )
 
 
+def test_asym_line_input_data__zero_mutual_impedances_are_valid():
```
nitbharambe, line 1192: I guess this test can be removed since already incorporated in the above test. Testing is unitized at the component level.

nitbharambe:
@looooown2006 , This PR and #1462 were created almost simultaneously hence there was no assigned person for it. This PR has a comment left to address. 
Hence for this reason, we decided to go ahead #1462 . Thanks for contributing, and apologies for having to close it. Please feel free to contribute for any other current / upcoming issues.

PR #1462, opened 2026-07-06
Allow mutual impedances/capacitances to be zero in asym line validator

### LINKED ISSUE #1460: [BUG] Allow mutual impedances to be 0  in asym line data validator

zain-asif-dev:
Fixes #1460.

### Motivation

The `validate_asym_line` validator required every entry of the r/x/c matrices (including the mutual/off-diagonal terms) to be strictly greater than zero. Physically, mutual impedances/capacitances between phases (and between a phase and neutral) can legitimately be zero, e.g. when the conductors are sufficiently decoupled. The calculation core already allows this; only the validator was too strict, causing valid asym line data to be rejected.

### Modifications

- Split the asym line r/x/c matrix fields in `validate_asym_line` into:
  - "self" (diagonal) fields (`r_aa`, `r_bb`, `r_cc`, `x_aa`, `x_bb`, `x_cc`, `c_aa`, `c_bb`, `c_cc`), which keep using `all_greater_than_zero` (still required to be `> 0`, since a conductor's self-impedance can't be zero).
  - "mutual" (off-diagonal) fields (`r_ba`/`r_ca`/`r_cb`, `r_na`/`r_nb`/`r_nc`/`r_nn`, and the equivalent `x_*`/`c_*` fields), which now use `all_greater_than_or_equal_to_zero` (allowed to be `>= 0`), matching the field list called out in the issue.
- Updated the valid-values (`> 0` -> `>= 0`) for the mutual fields in `docs/user_manual/components.md`.
- Updated the existing `test_asym_line_input_data` unit test to expect `NotGreaterOrEqualError` (instead of `NotGreaterThanError`) for the mutual fields, since the test data uses `-1` (still invalid under the relaxed `>= 0` rule) for all of these fields.

### Result

Asym line input data with zero-valued mutual impedances/capacitances is now accepted by the validator, while negative values and zero self-impedances are still correctly rejected.

### Testing

- Ran the full `tests/unit/validation/` suite against the modified `_validation.py` (installed the `power-grid-model` 1.13.108 wheel from PyPI, which has an identical pure-Python validation layer for this version, and overlaid the modified file to avoid a full C++ core rebuild in this environment): 300 passed, 1 skipped (unrelated).
- Checked `ruff check`/`ruff format --check` on the modified files.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/user_manual/components.md
+++ b/docs/user_manual/components.md
@@ -347,30 +347,30 @@ If the neutral values are not provided, the last row and column from the above m
 | name   | data type | unit       | description                       | required                     |  update  | valid values |
 |--------|-----------|------------|-----------------------------------|------------------------------|:--------:|:------------:|
 | `r_aa` | `double`  | ohm (Ω)    | Series serial resistance aa       | &#10004;                     | &#10060; |    `> 0`     |
-| `r_ba` | `double`  | ohm (Ω)    | Series serial resistance ba       | &#10004;                     | &#10060; |    `> 0`     |
+| `r_ba` | `double`  | ohm (Ω)    | Series serial resistance ba       | &#10004;                     | &#10060; |    `>= 0`    |
 | `r_bb` | `double`  | ohm (Ω)    | Series serial resistance bb       | &#10004;                     | &#10060; |    `> 0`     |
-| `r_ca` | `double`  | ohm (Ω)    | Series serial resistance ca       | &#10004;                     | &#10060; |    `> 0`     |
-| `r_cb` | `double`  | ohm (Ω)    | Series serial resistance cb       | &#10004;                     | &#10060; |    `> 0`     |
+| `r_ca` | `double`  | ohm (Ω)    | Series serial resistance ca       | &#10004;                     | &#10060; |    `>= 0`    |
+| `r_cb` | `double`  | ohm (Ω)    | Series serial resistance cb       | &#10004;                     | &#10060; |    `>= 0`    |
 | `r_cc` | `double`  | ohm (Ω)    | Series serial resistance cc       | &#10004;                     | &#10060; |    `> 0`     |
-| `r_na` | `double`  | ohm (Ω)    | Series serial resistance na       | &#10024; for a neutral phase | &#10060; |    `> 0`     |
-| `r_nb` | `double`  | ohm (Ω)    | Series serial resistance nb       | &#10024; for a neutral phase | &#10060; |    `> 0`     |
-| `r_nc` | `double`  | ohm (Ω)    | Series serial resistance nc       | &#10024; for a neutral phase | &#10060; |    `> 0`     |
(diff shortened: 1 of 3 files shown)


docs/user_manual/components.md
```diff
@@ -347,30 +347,30 @@ If the neutral values are not provided, the last row and column from the above m
...
 | `r_cc` | `double`  | ohm (Ω)    | Series serial resistance cc       | &#10004;                     | &#10060; |    `> 0`     |
-| `r_na` | `double`  | ohm (Ω)    | Series serial resistance na       | &#10024; for a neutral phase | &#10060; |    `> 0`     |
-| `r_nb` | `double`  | ohm (Ω)    | Series serial resistance nb       | &#10024; for a neutral phase | &#10060; |    `> 0`     |
-| `r_nc` | `double`  | ohm (Ω)    | Series serial resistance nc       | &#10024; for a neutral phase | &#10060; |    `> 0`     |
-| `r_nn` | `double`  | ohm (Ω)    | Series serial resistance nn       | &#10024; for a neutral phase | &#10060; |    `> 0`     |
+| `r_na` | `double`  | ohm (Ω)    | Series serial resistance na       | &#10024; for a neutral phase | &#10060; |    `>= 0`    |
+| `r_nb` | `double`  | ohm (Ω)    | Series serial resistance nb       | &#10024; for a neutral phase | &#10060; |    `>= 0`    |
+| `r_nc` | `double`  | ohm (Ω)    | Series serial resistance nc       | &#10024; for a neutral phase | &#10060; |    `>= 0`    |
+| `r_nn` | `double`  | ohm (Ω)    | Series serial resistance nn       | &#10024; for a neutral phase | &#10060; |    `>= 0`    |
```
nitbharambe: Lets not enable this for nn values. So r_nn, x_nn needs to be the way it was. Reason being that it would need special handling at C++ core.

zain-asif-dev:
Makes sense, reverted r_nn and x_nn back to strictly greater than zero and kept the rest of the mutual terms allowing zero.

nitbharambe (approved):

PR #1463, opened 2026-07-07
Docs: improve documentation on slack bus + ideal source

mgovers:
Document answers in https://github.com/orgs/PowerGridModel/discussions/107


docs/user_manual/non-pgm-components.md
```diff
@@ -62,3 +62,14 @@ for symmetric calculations.
...
+
+## Ideal source
+
+An ideal source  (a.k.a. slack bus) can be modeled using a regular PGM [source](./components.md#source) with a high
+`sk`, e.g. $s_k = 10^{50}\,\text{VA}$.
+
+```{warning}
+At the time of writing `sk` cannot be too high, as it may result in unresolved infinities (see
+[PGM issue #733](https://github.com/PowerGridModel/power-grid-model/issues/733), which is a `good-first-issue` and
+can be picked up by anyone who would like to).
```
figueroa1395: I think the "too high" wording is very ambiguous. Let's explicitly asy `sk` can't be infinite and just point to the issue.

docs/user_manual/calculations.md
```diff
@@ -523,3 +523,25 @@ computing.
...
+
+In principle, all calculation methods treat all nodes in the grid the same.
+That also means that even slack buses don't get "special treatment".
+The reason behind this is clear: if there are multiple slack buses in the grid, it is not uniquely determined which one
+should get priority.
+However, it is still useful to select a slack bus for multiple reasons.
+Firstly, because the grid state is only determined up to a constant complex phase, the slack bus can resolve the
+ambiguity by choosing the gauge.
+Secondly, because isolated islands in the grid (i.e., connected components in the topological meaning of the word) can
+most easily be determined using depth-first or breadth-first search from some "starting node".
```
figueroa1395: I think these don't add much to the above and might just confuse people. With what's written below it's very clear how we select slack buses in PGM.

docs/user_manual/calculations.md
```diff
@@ -523,3 +523,25 @@ computing.
...
+
+In PGM, slack buses are selected as follows.
+
+1. [Source](./components.md#source) nodes are slack buses.
+2. If there are multiple isolated islands, each island's source is selected as the slack bus.
+3. If there are multiple sources within an isolated island, it is
+   [implementation-defined](../advanced_documentation/terminology.md#implementation-defined) which source node is
+   selected as the slack bus.
+   The simplest choice is to select the first-occuring source in the input data, but note that this behavior may change
+   over time.
```
figueroa1395: Maybe add 4. Slack bus as an ideal source. or something like that and point to the `## Ideal source` written below.

docs/user_manual/non-pgm-components.md
mgovers: but `10^200` is also too high because it is squared and then becomes infinity in the calculations, hence `10^50`

docs/user_manual/calculations.md
mgovers: i'll just add a note underneath it.

docs/user_manual/calculations.md
mgovers: removed

docs/user_manual/non-pgm-components.md
mgovers: rephrased

TonyXiang8787:
Hi @mgovers @figueroa1395,

I wonder the usefulness of such a documentation for slack bus. In PGM implementation, we do not use slack bus at all. The name "slack bus" is just some convenient programming variable for the first found source in an island. It does not really have physical meaning.

Explaining this to the user does not seem to be useful.

We can indeed, only explain how to set source parameter so that it will behaves like a real power system slack bus (ideal source). That has actualy meaning for the user.

figueroa1395:
@TonyXiang8787 This was asked in a recent [post](https://github.com/orgs/PowerGridModel/discussions/107#discussioncomment-17558542), so we thought that since we answered it there, perhaps other users would find that information useful as well, so why not add it to the documentation, since the effort is minimal.

figueroa1395 (dismissed):

nitbharambe (dismissed):

nitbharambe:
Its actually been asked multiple times in past, including the discussion thread. And the response has indeed been what you mention. This text itself in this PR does not seem to have incorrect statements if we are only comparing slack bus and source bus. It would indeed be misleading though.

> Hi @mgovers @figueroa1395,
> 
> I wonder the usefulness of such a documentation for slack bus. In PGM implementation, we do not use slack bus at all. The name "slack bus" is just some convenient programming variable for the first found source in an island. It does not really have physical meaning.
> 
> Explaining this to the user does not seem to be useful.
> 
> We can indeed, only explain how to set source parameter so that it will behaves like a real power system slack bus (ideal source). That has actualy meaning for the user.

figueroa1395 (approved):

nitbharambe:
As per offline discussions, we decided to go ahead with it.

llru-fhv:
> Hi @mgovers @figueroa1395,
> 
> I wonder the usefulness of such a documentation for slack bus. In PGM implementation, we do not use slack bus at all. The name "slack bus" is just some convenient programming variable for the first found source in an island. It does not really have physical meaning.
> 
> Explaining this to the user does not seem to be useful.
> 
> We can indeed, only explain how to set source parameter so that it will behaves like a real power system slack bus (ideal source). That has actually meaning for the user.

Sorry, but I got surprised to read "In PGM implementation, we do not use slack bus at all". Then, how do you solve the power flow if no slack bus is used? From a mathematical point of view, one has to define the slack node (to set the reference angle) to match the number of variables with the number of power flow equations (then to solve these nonlinear equations, e.g., with Newton Raphson algorithm). From a physical point of view, the selection of the slack bus is also relevant, and normally assigned to the largest generation unit, if there are multiples. Or to the low voltage side of a transformer for low voltage networks.

mgovers:
> > Hi @mgovers @figueroa1395,
> > I wonder the usefulness of such a documentation for slack bus. In PGM implementation, we do not use slack bus at all. The name "slack bus" is just some convenient programming variable for the first found source in an island. It does not really have physical meaning.
> > Explaining this to the user does not seem to be useful.
> > We can indeed, only explain how to set source parameter so that it will behaves like a real power system slack bus (ideal source). That has actually meaning for the user.
> 
> Sorry, but I got surprised to read "In PGM implementation, we do not use slack bus at all". Then, how do you solve the power flow if no slack bus is used? From a mathematical point of view, one has to define the slack node (to set the reference angle) to match the number of variables with the number of power flow equations (then to solve these nonlinear equations, e.g., with Newton Raphson algorithm). From a physical point of view, the selection of the slack bus is also relevant, and normally assigned to the largest generation unit, if there are multiples. Or to the low voltage side of a transformer for low voltage networks.

Hi @llru-fhv,

Yes exactly, that is where this whole PR came from. We are gradually trying to split the documentation for users from the internal documentation. Quoting the form of the internal documentation we settled on:

> PGM does not have the common notion of a slack bus in the sense of an ideal source, both for computational and
> ambiguity reasons when there are multiple sources in the grid.
> However, it is still useful to select a slack bus to fix the otherwise ambiguous complex phase.
> 
> In PGM, slack buses are selected as follows.
> 
> 1. [Source](./components.md#source) nodes are slack buses.
> 2. If there are multiple isolated islands, each island's source is selected as the slack bus.
> 3. If there are multiple sources within an isolated island, it is
>    [implementation-defined](../advanced_documentation/terminology.md#implementation-defined) which source node is
>    selected as the slack bus.
>    The simplest choice is to select the first-occuring source in the input data, but note that this behavior may change
>    over time.
> 
> ```{note}
> The slack bus used by PGM internally differs from the slack bus in the sense of an ideal source.
> For documentation on ideal source modeling, please refer to [its documentation](./non-pgm-components.md#ideal-source).
> ```
> 

So, yes, we do in fact assign a slack bus internally (the Mathematical interpretation you mentioned), but it is not a slack bus in the classical sense (the Physical interpretation you mentioned).

The reason we make this distinction is that - in principle - there is no "order" in the power flow equations - all nodes are equal. Choosing the gauge is required but arbitrary (by definition of gauge symmetry), so WLOG we can do what we are doing.

I hope that somewhat explains the comment.

PR #1448, opened 2026-06-24
Branch into itself: validation cases

mgovers:
Validation cases for lines in to itself

Part of #1418 

## generated using the following script

```py
from pathlib import Path
from re import match
from power_grid_model import CalculationType, CalculationMethod, PowerGridModel, ComponentType as CT, AttributeType as AT
from power_grid_model._core.dataset_definitions import ComponentType
from power_grid_model._core.enum import TapChangingStrategy
from power_grid_model.errors import PowerGridBatchError
from power_grid_model.utils import json_deserialize_from_file, json_serialize_to_file, get_dataset_scenario
from power_grid_model.validation.assertions import assert_valid_batch_data, assert_valid_input_data


CALCULATION_FUNCS = {
    CalculationType.power_flow: lambda x: x.calculate_power_flow,
    CalculationType.state_estimation: lambda x: x.calculate_state_estimation,
    CalculationType.short_circuit: lambda x: x.calculate_short_circuit
}


class TestCase:
    def __init__(self, name, self_branch_type, n_branches_into_itself, n_regular_branches):
        self.name = name
        self.self_branch_type = self_branch_type
        self.n_branches_into_itself = n_branches_into_itself
        self.n_regular_branches = n_regular_branches


def main(test_case: TestCase, calculation_type: CalculationType, symmetric: bool | None = None, tap_changing_strategy: TapChangingStrategy | None = None):
    print(f"Running {calculation_type.name} calculation with symmetric={symmetric}")

    if (symmetric is None) != (calculation_type == CalculationType.short_circuit):
        raise ValueError("symmetric must be specified for power flow and state estimation calculations, and must not be specified for short circuit calculations")

    calculation_func = CALCULATION_FUNCS[calculation_type]
    base_path = Path(f"tests/data/{calculation_type.name}/{test_case.name}")
    prefix = "sc" if calculation_type == CalculationType.short_circuit else ("sym" if symmetric else "asym")

    print("load base path:", base_path)
    input_data = json_deserialize_from_file(base_path / "input.json")
    update_data = json_deserialize_from_file(base_path / "update_batch.json")
    model = PowerGridModel(input_data=input_data)

    assert_valid_input_data(input_data=input_data, calculation_type=calculation_type, symmetric=symmetric)
    assert_valid_batch_data(input_data=input_data, update_data=update_data, calculation_type=calculation_type, symmetric=symmetric)

    print("calculate")
    kwargs = {"update_data": update_data}
    match(calculation_type):
[...]

PR #1466, opened 2026-07-09
Branch into itself: make validator + docs accept branches into itself

### LINKED ISSUE #1418: [FEATURE] Allow branch into itself

mgovers:
Closes #1418 

* [x] Depends on #1448 , `do-not-merge` label added to reflect this
  * (yes, it is technically independent from it but it if we first review the validation cases, we have evidence that it also works as intended before we officially make it GA)

### DIFF
--- a/src/power_grid_model/validation/_validation.py
+++ b/src/power_grid_model/validation/_validation.py
@@ -626,7 +626,6 @@ def validate_branch(data: SingleDataset, component: CT) -> list[ValidationError]
     errors = validate_base(data, component)
     errors += _all_valid_ids(data, component, AT.from_node, CT.node)
     errors += _all_valid_ids(data, component, AT.to_node, CT.node)
-    errors += _all_not_two_values_equal(data, component, AT.to_node, AT.from_node)
     errors += _all_boolean(data, component, AT.from_status)
     errors += _all_boolean(data, component, AT.to_status)
     return errors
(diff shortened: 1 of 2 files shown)


src/power_grid_model/validation/_validation.py
```diff
@@ -626,7 +626,6 @@ def validate_branch(data: SingleDataset, component: CT) -> list[ValidationError]
     errors = validate_base(data, component)
     errors += _all_valid_ids(data, component, AT.from_node, CT.node)
     errors += _all_valid_ids(data, component, AT.to_node, CT.node)
-    errors += _all_not_two_values_equal(data, component, AT.to_node, AT.from_node)
```
mgovers, line 629: this is untested in `main` and this is not the moment to start testing this particular chunk of code

nitbharambe (approved):

Issue #78, opened 2022-06-01
[FEATURE] Uncertainty margin for state estimation


TonyXiang8787:

jaccoheres:
@TonyXiang8787 I guess this is the a posteriori uncertainty of the estimated state?

TonyXiang8787:
@jaccoheres yes you are correct.

jaccoheres:
The use cases for this within the Alliander 'Inzicht in Basislast' team is the following:

- The state-estimator in PGM is used to kalibrate the Bottom-Up loads of PULSE per secondary substation/MV busbar to the EMS measurements at the start of every MV feeder and maybe other measurements in the feature. These kalibrated/estimated values are then used in load flow calculations, sometimes the whole loadprofile is used but more often the peak loads (sometimes within a certain time window) are used. However, due to the fact that we usually only have one measurement per MV feeder, the peak loads calculated via the state-estimation are underestimating the real peak loads, because the individual peaks are smeared out over the whole feeder. We would like to a an a posteriori uncertainty of the calculated load per timesteps, so that we could take into account the probablity that the real load is higher then the estimated load. In this way, we could estimate better peak loads.
- The estimated loads are input for forecasts, which are currently deterministic. There is however much urgence to move to probabilistic forecasts. Then the current loads should also be probabilistic in order to achieve this, so it would be helpfull to have a uncertainty of the load at all points in time.

An estimation of the uncertainty of the states could also be helpfull in deciding what the best location for new measurements would be, in order to decrease the uncertainty.

It would be of course very nice to see the influence of each parameter on the uncertainty of the estimated values, but it would be already quite something to have an idea of the uncertainty of each estimated value (voltage, current, power) on it's own, so just one positive scalar per quantity per timestep, based on the chosen sigma's.

PR #1472, opened 2026-07-13
Uncertainty quantification: add selective inverse and its tests

Plusero:
Relates to https://github.com/PowerGridModel/power-grid-model/issues/78

Selective inverse for block sparse matrices is a component for the Uncertainty Quantification of State Estimation. 

Changes proposed in this PR:
1. Implement `inplace_selective_inverse_with_prefactorized_matrix` in sparse LU solver.
2. Add tests for selective inverse for block sparse matrices.
3. Add documentation for both dense and selective inverse.


docs/algorithms/dense-and-selective-inverse.md
```diff
@@ -0,0 +1,245 @@
+<!--
+SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+
+SPDX-License-Identifier: MPL-2.0
+-->
+
+# Dense and selective matrix inverses
```
mgovers, line 7: please also add this to the respective `{toctree}` section in the `index.md` so that it shows up in the sidebar

docs/algorithms/dense-and-selective-inverse.md
mgovers, line 1: nitpick but we have a convention that after every sentence in markdown, we have a newline. feel free to close this if you think it's too much work, or one of us can pick it up

docs/algorithms/dense-and-selective-inverse.md
Plusero: Sure, will do soon.

docs/algorithms/dense-and-selective-inverse.md
```diff
@@ -0,0 +1,245 @@
...
+1. Buffer the packed pivot block, the original $\mathbf{U}_{pm}$ blocks, and the original $\mathbf{L}_{mp}$
+   blocks.
+2. Compute the lower blocks $\mathbf{Z}_{kp}$.
+3. Compute the upper blocks $\mathbf{Z}_{pj}$.
+4. Compute $\mathbf{Z}_{pp}$ last, using the newly computed lower blocks.
+
+This order permits the LU storage to be overwritten in place without losing values needed by the current step.
+
+#### Why row/column buffers are needed
+
```
mgovers, line 206: might be an idea to also refer to the LU solver docs as some of these things are very similar

docs/algorithms/dense-and-selective-inverse.md
Plusero: I will follow this convention.

docs/algorithms/dense-and-selective-inverse.md
Plusero: Done

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -303,18 +303,37 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+            }
+
+            // Restore A^-1_ij per sparse entry: Z_ij = Q_i * Z_ij * P_j.
+            for (Idx row = 0; row < size_; ++row) {
+                for (Idx idx = row_indptr_[row]; idx < row_indptr_[row + 1]; ++idx) {
+                    data[idx] =
+                        (block_perm_array[row].q * data[idx].matrix() * block_perm_array[col_indices_[idx]].p).array();
+                }
+            }
+        }
```
TonyXiang8787: Please factor this out to a separate private function. Then here you can just call the function before the closing `}` above.

docs/algorithms/dense-and-selective-inverse.md
Plusero: It will take some work indeeded. I also encountered the length limit of each line by markdownlint. I will fix this tomorrow.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Plusero: Done. Note that I did not put the extracted function call outside the `if constexpr (!is_block)` statement (line313). Doing that would cause compilation to fail because the call would also be compiled for the scalar specialization, while the extracted function and its calls are constrained with `requires is_block`.

docs/algorithms/dense-and-selective-inverse.md
Plusero: Done.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -645,6 +651,105 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+    // are left solves (Ax=b), so they cannot be reused here.
+    static Tensor multiply_inverse_unit_lower_right(Tensor const& pivot, Tensor block)
+        requires is_block
+    {
+        for (Idx block_col = block_size - 1; block_col > -1; --block_col) {
+            for (Idx other_col = block_size - 1; other_col > block_col; --other_col) {
+                block.col(block_col) -= pivot(other_col, block_col) * block.col(other_col);
+            }
+        }
+        return block;
```
TonyXiang8787, line 766: Why you cannot use forward substitution function as you did for backwards as above?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -645,6 +651,105 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+
+        // Buffer LU values from row/column p before overwriting them with Z blocks.
+        Tensor const pivot = data[pivot_idx];
+        std::vector<Tensor> const u_row(data.begin() + u_start, data.begin() + u_end); // u_row is contagious.
+        std::vector<Tensor> l_col(n_off_diagonal); // l column is scattered we need to extract them.
+        IdxVector l_indices(n_off_diagonal);
+        for (Idx offset = 0; offset < n_off_diagonal; ++offset) {
+            l_indices[offset] = find_entry(col_indices_[u_start + offset], pivot_row_col);
+            l_col[offset] = data[l_indices[offset]];
+        }
```
TonyXiang8787: Use a lambda to make `l_col` also const

```c++
        std::vector<Tensor> const l_col = [&]() { 
            std::vector<Tensor> res(n_off_diagonal); // l column is scattered we need to extract them.
            IdxVector l_indices(n_off_diagonal);
            for (Idx offset = 0; offset < n_off_diagonal; ++offset) {
                l_indices[offset] = find_entry(col_indices_[u_start + offset], pivot_row_col);
                res[offset] = data[l_indices[offset]];
            }
            return res;
       }();
```

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -645,6 +651,105 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+    }
+
+    // Find the data index of entry (row, col), which must exist in the filled LU pattern.
+    Idx find_entry(Idx row, Idx col) const {
+        auto const found =
+            std::lower_bound(col_indices_.begin() + row_indptr_[row], col_indices_.begin() + row_indptr_[row + 1], col);
+        assert(found != col_indices_.begin() + row_indptr_[row + 1]);
+        assert(*found == col);
+        return narrow_cast<Idx>(std::distance(col_indices_.begin(), found));
+    }
```
TonyXiang8787: In all the call to `find_entry`, you always try to find a tranpose position of `(j,i)` when you already know the original position `(i,j)` and its data index `k`.

We have actually pre-stored all transpose mapping to convenient track in the `YBusStruct`

https://github.com/Plusero/power-grid-model/blob/2e38c3ac2f9b63668c4582af08762bd6b7d34643/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/y_bus.hpp#L115-L119

You might want to use that. But that's not passed into LU Sparse solver yet. You can change that.

TonyXiang8787: After some digging, it seems that you are looking up more than just transpose. This is not valid.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -645,6 +651,105 @@ template <class Tensor, class RHSVector, class XVector> class SparseLUSolver {
...
+        Tensor sum = Tensor::Zero();
+        for (Idx m_offset = 0; m_offset < n_off_diagonal; ++m_offset) {
+            sum += dot(u_row[m_offset], data[l_indices[m_offset]]);
+        }
+        data[pivot_idx] =
+            LUFactor::inverse_factorized_block(pivot.matrix()).array() - multiply_inverse_upper_left(pivot, sum);
+    }
+
+    // Find the data index of entry (row, col), which must exist in the filled LU pattern.
+    Idx find_entry(Idx row, Idx col) const {
```
TonyXiang8787: In the factorization part, there is also a `lower_bound` to find entry. Can you refactor that one to call this too?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: I see you need `l_indices` also below. Then it becomes more difficult. You need to use a structured binding and return a `std::pair` in the lambda. Feel free to consider if it is worth it or not.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: Ah I see your explanation. Can you just add this function into the `LUFactor` class?

Plusero: Done. I extracted the implementation into `DenseLUFactor`.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Plusero: Let me try.

Plusero: Done.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Plusero: Done. Used a lambda and `std::pair` as suggested.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: Now, let's try some more optimization points:

1. You are using `lower_bound` to search the whole row, namely `indptr[row]:indptr[row+1]`. However, in a given context, we might have narrower search space than the whole row. You can think about adding some optional arguments to limit the search space which can be narrower than `indptr[row]:indptr[row+1]`. Think about what is logical based on where it is called under which context. 
2. The `lower_bound` is using binary search, which is less efficient than a simple linear search if the search space is very small. You can have a branch to first decide the search size, if it is smaller than `linear_search_threshold`, you use linear search instead of binary search.

TonyXiang8787: Usually the rule of thumb for the threshold is something like `linear_search_threshold=16`

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: For this one, we can also think about further making the code more concise. You have defined some functions in `LUFactor` for right or left solve. We also have still some more routines doing similar things, see below:

https://github.com/Plusero/power-grid-model/blob/9469be2a1da2e51b14b8e17e05fbde661d06160f/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp#L421-L437

https://github.com/Plusero/power-grid-model/blob/9469be2a1da2e51b14b8e17e05fbde661d06160f/power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp#L450-L474

Think about, can we define a single function in `LUFactor` to solve all of them? Using some template argument to do compile time `if constexpr` like branching. So including, right or left solve, L or U solve, all combinations.

Plusero: Done. I have put the 4 different solves (Left L, Left U, Right L, Right U) in one template.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -152,36 +163,36 @@ template <rk2_tensor Matrix> class DenseLUFactor {
...
+
+        for (int step = 0; step < size; ++step) {
+            int const index = forward_traversal ? step : size - 1 - step;
+            for (int previous_step = 0; previous_step < step; ++previous_step) {
+                int const previous_index = forward_traversal ? previous_step : size - 1 - previous_step;
+                if constexpr (side == TriangularSolveSide::left) {
+                    rhs.row(index) -= lu_matrix(index, previous_index) * rhs.row(previous_index);
+                } else {
+                    rhs.col(index) -= lu_matrix(previous_index, index) * rhs.col(previous_index);
+                }
```
TonyXiang8787: use `int8_t` for small integer indexing.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -152,36 +163,36 @@ template <rk2_tensor Matrix> class DenseLUFactor {
...
                  std::same_as<typename RHSDerived::Scalar, Scalar> && rk2_tensor<LUDerived> &&
                  (LUDerived::RowsAtCompileTime == size) && (LUDerived::ColsAtCompileTime == size) &&
-                 (RHSDerived::RowsAtCompileTime == size))
+                 ((side == TriangularSolveSide::left && RHSDerived::RowsAtCompileTime == size) ||
+                  (side == TriangularSolveSide::right && RHSDerived::ColsAtCompileTime == size)))
     {
-        for (int8_t row = 0; row < size; ++row) {
-            for (int8_t col = 0; col < row; ++col) {
-                rhs.row(row) -= lu_matrix(row, col) * rhs.row(col);
+        constexpr bool forward_traversal = (side == TriangularSolveSide::left) == (factor == TriangularFactor::lower);
```
TonyXiang8787: static constexpr

Plusero: Why? I don't get the idea.

TonyXiang8787: only static constexpr garantees that  it will be evaluated in compile time.

Plusero: Done.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Plusero: Done. Now `find_entry` accepts narrower search space. 

Narrower search space is used when we do Schur-complement update in LU factorization, i.e., `prefactorize`, where we update unfactorized rows from left to right and use `find_entry`. When updating a row, we start at the first unfactorized element/block of the row and end at the end of the row.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -46,6 +46,17 @@ inline void perturb_pivot_if_needed(double perturb_threshold, Scalar& value, dou
...
+// Triangular solve variants:
+// | Side  | Factor | Traversal | Updated RHS view | Operation  | Divide by diagonal?      |
+// |-------|--------|-----------|------------------|------------|--------------------------|
+// | Left  | L      | Forward   | Row              | L^-1 * rhs | No; L has unit diagonal  |
+// | Left  | U      | Backward  | Row              | U^-1 * rhs | Yes                      |
+// | Right | L      | Backward  | Column           | rhs * L^-1 | No; L has unit diagonal  |
+// | Right | U      | Forward   | Column           | rhs * U^-1 | Yes                      |
+// Left-lower and left-upper solves are commonly called forward and backward substitution, respectively.
+enum class TriangularSolveSide : bool { left, right };
+enum class TriangularFactor : bool { lower, upper }; // L or U
```
TonyXiang8787: @mgovers 

Do you still have comments on this construction? If not, feel free the resolve the conversation and approve the PR.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -152,36 +163,38 @@ template <rk2_tensor Matrix> class DenseLUFactor {
...
+                 ((side == TriangularSolveSide::left && RHSDerived::RowsAtCompileTime == size) ||
+                  (side == TriangularSolveSide::right && RHSDerived::ColsAtCompileTime == size)))
     {
-        for (int8_t row = 0; row < size; ++row) {
-            for (int8_t col = 0; col < row; ++col) {
-                rhs.row(row) -= lu_matrix(row, col) * rhs.row(col);
+        static constexpr bool forward_traversal =
+            (side == TriangularSolveSide::left) == (factor == TriangularFactor::lower);
+
+        for (int8_t step = 0; step < size; ++step) {
```
mgovers: you can also uses ranges:

```suggestion
        for (auto step : std::views::iota(int8_t{0}, size)) {
```
or
```suggestion
        for (auto step : IdxRanges{size}) {
```

you also don't need to restrict to `int8_t`/`uint8_t`; internally, the type will be promoted to 64-bit because it's running on a 64-bit architecture. by restricting to `int8_t` you actually introduce the risk that if `size >= 128`, then it will wrap-around, which causes potentially hard-to-find bugs

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
mgovers: i have been thinking about this.

personally, i'm more in favor of using tags/types (see e.g. how we handle symmetric vs asymmetric calculations using `symmetry_tag`, `symmetric_t` and `asymmetric_t`)

if you really want to go down the enum route, then please make it an `int8_t`, not a `bool`. the layout of `bool` types are unspecified in the standard (in particular, it is not specified whether the underlying representation is signed or unsigned and whether it is 32-bit or 8-bit) and it can lead to unexpected changes if we ever refactor parts of this code.

it may also be worth thinking about combining the two into a single enum as clearly there is a distinct difference between each of the 4 operations. in-place, you can then do a `switch`/`case` statement with `[[fallthrough]]`s to select the desired traversal, or (if everything is compile-time), then you can use `requires` clauses and/or `constexpr if` statements to select the correct behavior

mgovers: see https://github.com/PowerGridModel/power-grid-model/pull/1472#discussion_r3601058312 . will close this comment in favor of the other one

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
```diff
@@ -152,36 +163,38 @@ template <rk2_tensor Matrix> class DenseLUFactor {
...
-        requires(std::same_as<typename LUDerived::Scalar, Scalar> &&
-                 std::same_as<typename RHSDerived::Scalar, Scalar> && rk2_tensor<LUDerived> &&
-                 (LUDerived::RowsAtCompileTime == size) && (LUDerived::ColsAtCompileTime == size) &&
-                 (RHSDerived::RowsAtCompileTime == size))
-    {
-        for (int8_t row = size - 1; row > -1; --row) {
-            for (int8_t col = size - 1; col > row; --col) {
-                rhs.row(row) -= lu_matrix(row, col) * rhs.row(col);
+            if constexpr (factor == TriangularFactor::upper) {
+                if constexpr (side == TriangularSolveSide::left) {
```
mgovers, line 193: i really like this change btw

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: The `size` of the static tensor is already limited to `int8_t`. Do you think it still makes sense to use `Idx` here in the loop?

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: I do not agree with using one enum/tag. These are two orthogonal options.

For choice of enum, tag type, I am more in favor of enum. @Plusero  it's up to you to decide.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
mgovers: in that case, you can make it
```suggestion
        for (auto step : std::views::iota(int8_t{0}, size)) {
```

or we can create an `Int8Range` similar to how we have an `IdxRange`, or an `IntegerRange` which deduces the type of the integer based on the size type (`IdxRange` may be a specialization of it using `IntegerRange<Idx>`)

TonyXiang8787: Maybe just use `Idx` in the loop? It's a local variable without any significant memory footprint.

Plusero: I would like to go with `for (auto step : IdxRanges{size}) {...}`, since we already use `IdxRange` in many other places of the repo. It also looks the most readable to me.

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
Plusero: I have updated `bool` to `int8_t`.
For enum or tag, I prefer to use enum since tags like
```
struct left_t {};
struct right_t {};
struct lower_t {};
struct upper_t {};
```
does not express the orthogonal nature of these options.

In: https://github.com/PowerGridModel/power-grid-model/pull/1472/changes/8530e358d0d900004f24dac37307a0e5fd6b193b

power_grid_model_c/power_grid_model/include/power_grid_model/math_solver/sparse_lu_solver.hpp
TonyXiang8787: Let's go for `IdxRange`

Plusero: I used `IdxRange`. Implemented in https://github.com/PowerGridModel/power-grid-model/pull/1472/changes/d62152fac4ad63ea8c71c82311d71b1f1cc0e478 .

docs/algorithms/dense-and-selective-inverse.md
TonyXiang8787: Please address this last comment then we are ready.

Plusero: I added one subsection to show the similarities and differences between LU solver and inverse, in https://github.com/PowerGridModel/power-grid-model/pull/1472/changes/6d105137b7836317df70b2b31e0e4c06e4c91ac3.

TonyXiang8787 (approved):

PR #1474, opened 2026-07-14
Clang-tidy: clang-tidy-22 compatibility

mgovers:
### `bugprone-throwing-static-initialization`

This is an actual bug in our code. We're initializing something in the user space without knowing the actual storage of users.

```cpp
error: initialization of 'complete_state_json' with static storage duration may throw an exception that cannot be caught [bugprone-throwing-static-initialization,-warnings-as-errors]
[build]   109 | auto const complete_state_json = R"json({
[build]       |            ^
```

### `cppcoreguidelines-use-enum-class`

Check enabled, but in-place `NOLINT` for the C API and wherever we're testing the C API.

```cpp
error: enum 'PGM_CalculationMethod' is unscoped, use 'enum class' instead [cppcoreguidelines-use-enum-class,-warnings-as-errors]
[build]   155 | enum PGM_CalculationMethod {
[build]       |      ^
```

### `modernize-use-scoped-lock`
```cpp
: error: use 'std::scoped_lock' instead of 'std::lock_guard' [modernize-use-scoped-lock,-warnings-as-errors]
[build]   220 |             std::lock_guard<std::mutex> const lock(calls_mutex);
[build]       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[build]       |             std::scoped_lock
```

### `-readability-container-contains`

In-place scoped out for the C++ API
```cpp
error: use 'contains' to check for membership [readability-container-contains,-warnings-as-errors]
[build]   358 |                 output_component_attribute_filters.find(component_meta) == output_component_attribute_filters.end()) {
[build]       |                                                    ^~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[build]       |                 !                                  contains component_meta)
```

### `readability-use-concise-preprocessor-directives`
```cpp
error: preprocessor condition can be written more concisely using '#ifdef' [readability-use-concise-preprocessor-directives,-warnings-as-errors]
[build]    16 | #if defined _WIN32
[build]       |  ^~~~~~~~~~~~~~~~~
[build]       |  ifdef _WIN32
```

### `readability-redundant-parentheses`

```cpp
error: redundant parentheses around expression [readability-redundant-parentheses,-warnings-as-errors]
[build]    40 |     (void)(argc);
[build]       |           ^    ~
```

### `readability-redundant-typename`

```cpp
error: redundant 'typename' [readability-redundant-typename,-warnings-as-errors]
[build]   106 |     constexpr decltype(auto) operator[](this Self const& self, typename Self::difference_type idx) {
[build]       |                                                                ^~~~~~~~
error: redundant 'typename' [readability-redundant-typename,-warnings-as-errors]
[build]   372 | template <class... T> using Container = typename container_impl::container_trait<T...>::type;
[build]       |                                         ^~~~~~~~
```

## Extended checks

previously agreed, now extended

### `bugprone-derived-method-shadowing-base-method`

in-place `NOLINT`: we want to **know** when we're deliberately doing this.

```cpp
error: 'power_grid_model::meta_data::json_converter::JsonConverter::visit_float32' shadows method with the same name in class 'null_visitor' [bugprone-derived-method-shadowing-base-method,-warnings-as-errors]
[build]   146 |     bool visit_float32(float v) { return visit_float64(v); }
[build]       |     ^
```

### `modernize-avoid-c-style-cast`

```cpp
error: C-style casts are discouraged; use static_cast/const_cast/reinterpret_cast [modernize-avoid-c-style-cast,-warnings-as-errors]
[build]   291 |         solve_with_prefactorized_matrix((std::vector<Tensor> const&)data, block_perm_array, rhs, x);
[build]       |     
```

### `modernize-use-designated-initializers`

```cpp
error: use designated initializer list to initialize 'VoltageSensorCalcParam<asymmetric_t>' (aka 'UniformComplexRandVar<asymmetric_t>') [modernize-use-designated-initializers,-warnings-as-errors]
[build]   147 |         return {u, u_variance};
[build]       |                ^~~~~~~~~~~~~~~
[build]       |                 .value= .variance=
```

### `bugprone-unhandled-exception-at-new`

C APIs are not allowed to throw. However, `new` can throw. In the current `main`, we have uncaught `new` (specifically `new PGM_Handle{}` and `new PGM_Options{}`). These will result in hard crashes.

The solution is to `#define PGM_NOEXCEPT noexcept` for C++ and `#define PGM_NOEXCEPT` for C users. The bindings work in both cases, because the extern signature disables `noexcept` keywords. For internal linkage, however, this will catch errors like

```cpp
error: missing exception handler for allocation failure at 'new' [bugprone-unhandled-exception-at-new,-warnings-as-errors]
[build]    27 |     return new PGM_Handle{}; // NOSONAR(S5025)
[build]       |            ^
```

### `cert-dcl03-c,hicpp-static-assert,misc-static-assert`

It depends on library version whether this is `constexpr` or not. We can't enable this yet.

```cpp
error: found assert() that could be replaced by static_assert() [cert-dcl03-c,hicpp-static-assert,misc-static-assert,-warnings-as-errors]
[build]   382 |             assert(std::ssize(branch_connected) == 2); // NOSONAR(R354)
[build]       |             ^~~~~~
[build]       |             static_assert
```

### `cppcoreguidelines-pro-bounds-avoid-unchecked-container-access`

We use this _everywhere_

```cpp
error: possibly unsafe 'operator[]', consider bounds-safe alternatives [cppcoreguidelines-pro-bounds-avoid-unchecked-container-access,-warnings-as-errors]
[build]    55 |         std::fill(std::begin(result) + indptr_view[group], std::begin(result) + indptr_view[group + 1], group);
[build]       |                                                                                 ^~~~~~~~~~~~~~~~~~~~~~
```

### `google-runtime-float`

This one is a big one. Since we use complex numbers **everywhere**, and very often combine it with asymmetric values, this would require changes everywhere.

```cpp
error: 'long double' type from literal suffix 'L' is not portable and should not be used [google-runtime-float,-warnings-as-errors]
[build]   235 |     return RealValue<asymmetric_t>{arg(ComplexValue<asymmetric_t>{exp(1.0i * phase)})};
[build]       |                                   
```

It discourages `1.0i` in favor of `DoubleComplex{0.0, 1.0}`.

However, cases like `exp(1.0i * phase)` **can** be rewritten in the faster form `polar(1.0, phase)` (to be created for asymmetric), so the check is not entirely nonsensical. We need to discuss, but out of scope of this PR.

### `llvm-prefer-static-over-anonymous-namespace`

Using anonymous namespace "nukes" the symbol out of the implementation. This clang-tidy check discourages doing that in favor of `static`, so that, if you ever need it, you can use it from another translation unit without changing the behavior of the original translation unit in which it was defined. However, in our codebase, the whole point is that you can, e.g., have multiple test files that declare the exact same function name and interface, potentially with even different implementations, without them affecting eachother. Using anonymous namespaces is the superior way here. Since the core is header-only as well, there's no point in enabling this check for our purposes.

```cpp
error: function 'execute_test' is declared in an anonymous namespace; prefer using 'static' for restricting visibility [llvm-prefer-static-over-anonymous-namespace,-warnings-as-errors]
[build]   595 | void execute_test(CaseParam const& param, T&& func) noexcept {
[build]       |      ^
```

### `portability-avoid-pragma-once`

as previously agreed, internally in the C++ core, we can use `#pragma once`
```cpp
error: avoid 'pragma once' directive; use include guards instead [portability-avoid-pragma-once,-warnings-as-errors]
[build]     5 | #pragma once
[build]       | ^
```


power_grid_model_c/power_grid_model_c/src/handle.cpp
```diff
@@ -23,27 +23,35 @@ constexpr char const* version = PGM_VERSION;
 } // namespace
 
 // create and destroy handle
-PGM_Handle* PGM_create_handle() {
-    return new PGM_Handle{}; // NOSONAR(S5025)
```
mgovers, line 27: this was an actual bug: `new` can throw `std::bad_alloc` and it is uncaught, resulting in a potential hard crash

power_grid_model_c/power_grid_model_c/src/options.cpp
```diff
@@ -16,44 +17,48 @@ using namespace power_grid_model;
 
 using power_grid_model_c::call_with_catch;
 using power_grid_model_c::safe_ptr_get;
+using power_grid_model_c::create;
+using power_grid_model_c::destroy;
 } // namespace
 
 // options
-PGM_Options* PGM_create_options(PGM_Handle* /* handle */) {
-    return new PGM_Options{}; // NOSONAR(S5025)
```
mgovers, line 23: this was an actual bug: `new` can throw `std::bad_alloc` and it is uncaught, resulting in a potential hard crash

.clang-tidy
```diff
@@ -66,6 +69,8 @@ readability-*,
 -readability-redundant-member-init,
 -readability-uppercase-literal-suffix,
 -bugprone-unchecked-optional-access,
+
+-google-runtime-float
```
mgovers: we probably need a proper discussion on this

.clang-tidy
```diff
@@ -58,6 +60,7 @@ modernize-*,
 performance-*,
 portability-*,
 -portability-template-virtual-member-function,
+-portability-avoid-pragma-once,
```
mgovers, line 63: we deliberately use `#pragma once`

.clang-tidy
```diff
@@ -38,6 +39,7 @@ hicpp-*,
 llvm-*,
 -llvm-else-after-return,
 -llvm-header-guard,
+-llvm-prefer-static-over-anonymous-namespace,
```
mgovers, line 42: the intention is precisely to make it module-private and therefore, it should **not** be `static`

.clang-tidy
```diff
@@ -25,6 +25,7 @@ cppcoreguidelines-*,
 -cppcoreguidelines-no-malloc,
 -cppcoreguidelines-non-private-member-variables-in-classes,
 -cppcoreguidelines-owning-memory,
+-cppcoreguidelines-pro-bounds-avoid-unchecked-container-access,
```
mgovers, line 28: we do this everywhere

TonyXiang8787:
@mgovers can we wait until #1401 is properly reviewed and merged? I can foresee some conflicts

mgovers:
> @mgovers can we wait until #1401 is properly reviewed and merged? I can foresee some conflicts

definitely. will add `do-not-merge` label

power_grid_model_c/power_grid_model_c/src/handle.cpp
```diff
@@ -23,27 +23,34 @@ constexpr char const* version = PGM_VERSION;
...
+        return nullptr;
+    }
 }
-void PGM_destroy_handle(PGM_Handle* handle) {
-    delete handle; // NOSONAR(S5025)
+void PGM_destroy_handle(PGM_Handle* handle) noexcept {
+    try {
+        delete handle; // NOSONAR(S5025)
+    } catch (...) {    // NOSONAR(S2486) // NOLINT(bugprone-empty-catch)
+    }
```
mgovers:
```suggestion
void PGM_destroy_handle(PGM_Handle* handle) noexcept {
    delete handle; // NOSONAR(S5025)
```

.clang-tidy
mgovers: as discussed: explicitly scoped out, but we can in the near future add a `polar` function replicating the `std::polar` behavior.

power_grid_model_c/power_grid_model_c/src/handle.cpp
```diff
@@ -23,27 +23,31 @@ constexpr char const* version = PGM_VERSION;
 } // namespace
 
 // create and destroy handle
-PGM_Handle* PGM_create_handle() {
-    return new PGM_Handle{}; // NOSONAR(S5025)
+PGM_Handle* PGM_create_handle() noexcept {
+    try {
+        return new PGM_Handle{}; // NOSONAR(S5025)
+    } catch (...) {
+        return nullptr;
```
figueroa1395, line 30: So this becomes the user responsibility to check that their handle isn't `nullptr`? Should we add a small check (if not present yet) in the python side for this? Do we have a check in place for the `cpp` API object?

Also maybe good to explicitly mention this in the documentation somewhere.

power_grid_model_c/power_grid_model_c/src/safe_memory_handling.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+    char const* what() const noexcept final { return msg_.c_str(); }
+
+  private:
+    std::string msg_;
+};
+
+template <class T, typename... Args>
+    requires std::constructible_from<T, Args...>
+inline T* create(Args&&... args) {
```
figueroa1395, line 28: I think I'd be more in favor of a different naming convention. We already use different sort of `create*`s in the codebase and we might want to be more explicit. What about:
- `pgm_new`?
- `custom_new`?
- other?

power_grid_model_c/power_grid_model_c/src/safe_memory_handling.hpp
```diff
@@ -0,0 +1,38 @@
...
+    try {
+        return new T{std::forward<Args>(args)...}; // NOSONAR(S5025)
+    } catch (std::bad_alloc const& ex) {
+        throw FailedConstruction{ex};
+    }
+}
+template <class T> inline void destroy(T* ptr) {
```
figueroa1395, line 35: Same as above:
- `pgm_delete`?
- `custom_delete`?
- other?

tests/cpp_unit_tests/component/test_current_sensor.cpp
```diff
@@ -26,7 +26,7 @@ TEST_SUITE_BEGIN("test_current_sensor");
 
 namespace power_grid_model {
 namespace {
-auto const r_nan = RealValue<asymmetric_t>{nan};
+auto r_nan() { return RealValue<asymmetric_t>{nan}; }
```
figueroa1395, line 29: Maybe out of scope, but can't this (and other similar ones be `constexpr`? We may not have a `constexpr` constructor in place, but should/can we?

tests/cpp_validation_tests/test_validation.cpp
```diff
@@ -359,33 +359,52 @@ void assert_result(OwningDataset const& owning_result, OwningDataset const& owni
 }
 
 // root path
+inline std::filesystem::path data_dir() {
 #ifdef POWER_GRID_MODEL_VALIDATION_TEST_DATA_DIR
-// use marco definition input
-std::filesystem::path const data_dir{POWER_GRID_MODEL_VALIDATION_TEST_DATA_DIR};
+    // use marco definition input
```
figueroa1395: Who's Marco? 😆

tests/package_tests/.clang-tidy
figueroa1395, line 1: So this introduces these `clang-tidy` checks to the tests? I'm fine with the chance.

tests/package_tests/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-cmake_minimum_required(VERSION 3.23)
+cmake_minimum_required(VERSION 3.25)
```
figueroa1395, line 5: What's stopping us from going to 4ish?

tests/package_tests/CMakeLists.txt
```diff
@@ -13,6 +13,13 @@ set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 
 find_package(power_grid_model REQUIRED CONFIG)
 
+# Treat power_grid_model headers as system headers to suppress warnings from them
+set_target_properties(
+    power_grid_model::power_grid_model_c
+    power_grid_model::power_grid_model_cpp
+    PROPERTIES SYSTEM ON
+)
+
```
figueroa1395, line 22: So ignore any warnings coming from the core and focus on tests parts only?

figueroa1395 (commented):
Minor comments, but LGTM.

power_grid_model_c/power_grid_model_c/src/handle.cpp
mgovers: hmm... good question. let me think about it for a minute

tests/cpp_unit_tests/component/test_current_sensor.cpp
mgovers: not possible because Eigen is not (yet) constexpr. They are working hard on it tho. calling a `constexpr` function of something that can explicitly not be `constexpr` is ill-formed, no diagnostic required.

tests/package_tests/CMakeLists.txt
mgovers: nothing really but also there's no hard requirement

tests/package_tests/CMakeLists.txt
mgovers: yes

power_grid_model_c/power_grid_model_c/src/safe_memory_handling.hpp
mgovers: i would like to keep `create` as a proxy for `PGM_create`. this keeps the parallel between the interface and the internal implementation

power_grid_model_c/power_grid_model_c/src/handle.cpp
mgovers: we tacitly already do this. however, due to the nature of failing handle creation, i did add such a check. I also homogenized some of the similar comments in the rest of the codebase

power_grid_model_c/power_grid_model_c/src/safe_memory_handling.hpp
figueroa1395: But we do use it at several places, beyond `PGM_create`, and then it becomes confusing for that same reason in other places. In addition, in the end is just a proxy for `new` that `throw`s a nicer exception.

But I also don't have a very strong opinion. If you think this is fine feel free to resolve then.

mgovers: since it's internal anyway, let's scope it out for now. we can always change later

figueroa1395 (dismissed):

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/dataset.hpp
```diff
@@ -347,15 +354,14 @@ struct OwningDataset {
...
+            return detail::contains(irrelevant_components, component);
         };
 
         for (Idx component_idx{}; component_idx != ref_info.n_components(); ++component_idx) {
             auto const& component_name = ref_info.component_name(component_idx);
             auto const& component_meta = MetaData::get_component_by_name(dataset.get_info().name(), component_name);
             // skip components not in the filter
-            if (enable_filters &&
-                output_component_attribute_filters.find(component_meta) == output_component_attribute_filters.end()) {
+            if (enable_filters && detail::contains(output_component_attribute_filters, component_meta)) {
```
nitbharambe:
```suggestion
            if (enable_filters && !detail::contains(output_component_attribute_filters, component_meta)) {
```

nitbharambe (approved):

PR #1481, opened 2026-07-20
OpenSSF gold badge: UBSan hardening in CMakePresets.json

mgovers:
This replaces the always-on hardening proposed in #1241 as that requires more discussions.

In the meantime, we can already get the benefits of getting hardening in the C/C++ tests.

## Testing

To test that it works, you can create

```cpp
TEST_CASE("Test UBSan") {
    int k = 0x7fffffff;
    k += 1;
    CHECK(k == 0x00000000); // overflow has occurred
}
```

If UBSan is correctly enabled, you should get something along the lines of
```cpp
ctest]     Start 64: Counting Iterator
[ctest] 1/1 Test #64: Counting Iterator ................Subprocess aborted***Exception:   0.07 sec
[ctest] tests/cpp_unit_tests/common/test_counting_iterator.cpp:15:7: runtime error: signed integer overflow: 2147483647 + 1 cannot be represented in type 'int'
[ctest] SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior tests/cpp_unit_tests/common/test_counting_iterator.cpp:15:7 
```

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -148,6 +148,9 @@
     {
       "name": "msvc-debug",
       "displayName": "Debug (MSVC)",
+      "environment": {
+        "SANITIZER_FLAGS": "/RTC1"
+      },
       "inherits": ["msvc-base", "debug"]
     },
     {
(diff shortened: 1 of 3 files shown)


CMakePresets.json
```diff
@@ -120,7 +120,7 @@
     {
       "name": "unix-sanitizer",
       "environment": {
-        "SANITIZER_FLAGS": "-fsanitize=address"
+        "SANITIZER_FLAGS": "-fstack-protector -fsanitize=address,pointer-compare,undefined -fno-sanitize-recover"
```
mgovers, line 123: or now, i only enabled hardening on Unix. Should I also enable it on Windows (using `/RTC1`)?

TonyXiang8787 (dismissed):

CMakePresets.json
```diff
@@ -434,6 +500,14 @@
...
       "configurePreset": "clang-release"
     },
+    {
+      "name": "apple-clang-debug",
+      "configurePreset": "apple-clang-debug"
+    },
+    {
+      "name": "apple-clang-release",
+      "configurePreset": "apple-clang-release"
+    },
```
TonyXiang8787, line 510: Can you elaborate here why we suddenly need a separate apple clang preset?

mgovers: Yes, on xcode, `-lubsan` is linked autonatically (ships with compiler). On LLVM clang, you explicitly need to link it. But if you link it with xcode, it fails to compile. Thus, we need both presets unfortunetaly, or we would have to handle that logic in the CMakeLists

TonyXiang8787 (approved):

PR #1482, opened 2026-07-21
Serialization: C (de)serialization example

figueroa1395:
When investigating the bug in #1339, I put some effort into creating a C script for debugging purposes (see https://github.com/PowerGridModel/power-grid-model/issues/1339#issuecomment-4109518070). Furthermore, I polished that  script and added it to the C API examples


power_grid_model_c_example/serialization.c
```diff
@@ -0,0 +1,160 @@
...
+        "    ]\n"
+        "  }\n"
+        "}";
+
+    // create handle
+    PGM_Handle* handle = PGM_create_handle();
+
+    // create deserializer from JSON string
+    PGM_Deserializer* deserializer = PGM_create_deserializer_from_null_terminated_string(handle, json_data, PGM_json);
+    assert(PGM_error_code(handle) == PGM_no_error);
```
mgovers: here and below: should not be `assert`s as they result in hard crashes in Debug builds but silently continue in Release builds (which may not be valid, and therefore result in undefined behavior). Instead, it should `printf` that there's an unexpected error (e.g. out-of-memory) and exit `1` (optionally do this in a separate function)

figueroa1395: You are right, however, I'm following the convention of the original example (`main.c`). I can change it but then I should probably change both. Or, since it's an example of C-API usage, should we leave those details to the power users? It isn't intended as a "test" for us anyway.

TonyXiang8787: It should be exactly like this actually. The input is pre-setup and there should be no error in this case. So an `assert` is about debug check.

mgovers: so that means that if the handle (or any other object for all that matters) can't be created due to corrupted/insufficient contiguous memory, then this will result in a segfault in release builds. that doesn't sound correct.

mgovers: `new` can throw `bad_alloc` as well, not just throw if the underlying constructor throws

TonyXiang8787: This is not a production-grade application but a demostration example of how C-API works. 

The `assert` tells user that you need to check for errors.

How user is going to actually handle this kind of stuff in their production application/library, is outside the scope of this example.

mgovers: I understand, but dereferencing a nullptr is explicitly undefined behavior of type undefined behavior. this is not an assertion, this is dangerous. You can create a weak `safe_assert` that raises a `SIGTERM` instead if you really want to go down that road.

TonyXiang8787: Yes, so C-API users are supposed to understand all about error code, `nullptr` stuff and more. They are supposed to deal with this properly in their production library/application. It is not our job to teach C-API user that generic knowledge.

The `assert` in the demonstration example is to inform user that this can fail, which is enough information for the user. Note this is not meant to be run in production, but a demostration.

figueroa1395: See [a57a5a0](https://github.com/PowerGridModel/power-grid-model/pull/1482/commits/a57a5a091560d8609100f388b2917cc8faeb7b30) and [c24f554](https://github.com/PowerGridModel/power-grid-model/pull/1482/commits/c24f554e775ca33831e783b526c9f791c7474966). Resolving.

power_grid_model_c_example/main.c
```diff
@@ -36,6 +40,13 @@ We do batch calculation with 3 scenarios, with the following mutation
 #include <stdio.h>
 #include <stdlib.h>
 
+#define CHECK_ERROR(handle)                                                                                            \
```
mgovers: maybe rename to `CHECK_NO_ERROR`, `CHECK` or `ASSERT_NO_ERROR`?

power_grid_model_c_example/serialization.c
```diff
@@ -0,0 +1,171 @@
...
+#include "power_grid_model_c.h"
+#include "power_grid_model_c/dataset_definitions.h"
+
+#include <assert.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#define CHECK_ERROR(handle)                                                                                            \
```
mgovers: maybe rename to `CHECK_NO_ERROR`, `CHECK` or `ASSERT_NO_ERROR`?

power_grid_model_c_example/serialization.c
```diff
@@ -0,0 +1,171 @@
...
+    if (PGM_error_code(handle) != PGM_no_error) {                                                                      \
+        fprintf(stderr, "PGM error: %s\n", PGM_error_message(handle));                                                 \
+        PGM_destroy_handle(handle);                                                                                    \
+        return 1;                                                                                                      \
```
mgovers: optionally use `abort()` (https://en.cppreference.com/c/program/abort), which indicates that you deliberately do not clean up any resources but just kill the process

mgovers (dismissed):
feel free to close the comments. they are minor.

mgovers (approved):

PR #1483, opened 2026-07-21
Add support for musllinux-aarch64 wheels

figueroa1395:


pyproject.toml
```diff
@@ -205,16 +205,16 @@ test-groups = ["test"]
 test-command = "pytest {package}/tests"
 # we do not support 
 #    PyPy
-#    musllinux in aarch64
-#    disable Python 3.14 for now until it is released
-skip = ["pp*", "*-musllinux_aarch64"]
+#    disable Python 3.15 for now until it is released
```
TonyXiang8787, line 208: You do not actually put skip to py315 in the list

mgovers: is there a hard reason to skip 3.15 for now? If I read the CIBW docs, it says that it's still opt-in rather than opt-out at this time (for the right reasons). However, at some point, it will be built by default. I don't think that's an issue.

TonyXiang8787: CIBW will enable 3.15 before numpy ships 3.15 wheels, and at that moment, our build will fail.

figueroa1395: Done in [ec30ae9](https://github.com/PowerGridModel/power-grid-model/pull/1483/commits/ec30ae9abb1593314d53a7be7cf7b7d6d3fd4bfb)

TonyXiang8787 (approved):

PR #1499, opened 2026-07-27
Dependencies: lock refresh + pre-commit from dependabot

mgovers:
* Using Dependabot to update python packages has many benefits for both security and dev experience.
* Dependabot now groups dependency bumps into a single PR where possible
* Rather than depend on external pre-commit hooks, we can also use the already installed dependencies in the environment. This enables configuring everything exactly as intended.

Stil TBD: what to do with the (p)npm dependencies `markdownlint-cli` and `biome`? (see #1502  and #1501 )


.github/dependabot.yml
```diff
@@ -7,4 +7,23 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "cron"
+      cronjob: "0 8 * * 1" # Every Monday at 8:00 UTC
```
mgovers: this prevents us from having repo A running dependabot on Monday, repo B on Tuesday and repo C on Saturday, meaning we have to switch contexts way less often after this PR

pyproject.toml
```diff
@@ -41,9 +41,10 @@ lint = [
     "pre-commit",
```
mgovers, line 41: we probably want to add `pre-commit-uv` here

.pre-commit-config.yaml
```diff
@@ -3,54 +3,58 @@
...
-    # Ruff version.
-    rev: v0.15.22
-    hooks:
-      # Run the linter.
-      - id: ruff-check
-        name: ruff-check
-        args: [ --fix ]
-      # Run the formatter.
+        name: reuse
+        entry: uv run --frozen reuse lint
```
mgovers, line 10: TBD: `uv run --frozen` is considered good practice but does not work with custom index. What to do?

FYI: the underlying reason is that `pre-commit` would stash the `uv.lock`, meaning the custom environment actually uses the original lockfile, not the one specified by the user.

Possible workarounds (in arbitrary order; there is no 'nice' solution):

* [ ] don't use `uv run` at all but only use directly `pytest`
  * This requires users to always pre-load the virtual environment instead of using `uv run`
* [ ] use `uv run` (status quo)
  * This will always upgrade, even if that means the lockfile goes completely out of sync between different environments. Potentially hard to track issues
* [ ] exclude `uv.lock` from index on every laptop
  * Means including when switching branches or weird issues may arise.
  * Requires active editing on each device.
* [ ] developers using a custom index would edit the `pre-commit-config.yaml` to use `uv run` and anyone not bound to that can symply use `uv run --locked`

.github/workflows/check-code-quality.yml
figueroa1395, line 1: Why the very specific differences between some dependencies using `--locked` and `--frozen`? It seems to be like only `dev` dependencies are `--locked` (`lint` and `code-generation`). Why is that? 

Shouldn't we just have them all with `--locked`? for this CI check (and potentially the rest?)?

.github/workflows/sonar.yml
```diff
@@ -48,8 +48,8 @@ jobs:
 
       - name: Python test and coverage
         run: |
-          CC=clang CXX=clang++ uv sync --no-default-groups --group test
-          uv run pytest
+          CC=clang CXX=clang++ uv sync --locked --no-default-groups --group test
+          uv run --frozen pytest
```
figueroa1395, line 52: Similar question to the previous one. Why specifically `--frozen` here? In CI, I'd argue all dependencies should be "super" up to date.

.pre-commit-config.yaml
figueroa1395: I like the first option best I think. My understanding is that now users must manage their environment correctly whereas the `pre-commit` did it before.

It might be "cumbersome", but gives full control. We may want to even do
```
entry: .venv/bin/reuse lint
```

For example, so users don't have to manually activate the environment as one normally doesn't if using `uv` for instance.

If a user manages multiple environments at the same time, they should be able to modify the pre-commit however they want. it should also be possible to add a check and raise a nice error if the user doesn't have `.venv` or just has a different name to it.

package-lock.json
figueroa1395, line 1: Are you just getting ahead of dependabot, or did it not pick these up? Just double checking.

pyproject.toml
```diff
@@ -41,9 +41,10 @@ lint = [
...
     "ruff",
     "mypy",
-    "numpy",                # needed for mypy
-    "gersemi",              # CMake linter
-    "clang-format ~= 18.0", # C++ formatter
+    "numpy",                             # needed for mypy
+    "gersemi",                           # CMake linter
+    "clang-format ~= 18.0",              # C/C++ formatter
+    "reuse[charset-normalizer,chardet]", # license check
```
figueroa1395, line 47: What's `[charset-normalizer,chardet]`?

pyproject.toml
```diff
@@ -80,6 +81,9 @@ Discussion = "https://github.com/orgs/PowerGridModel/discussions"
 [project.entry-points."cmake.root"]
 power_grid_model = "power_grid_model._core.power_grid_model_c"
 
+[tool.uv]
+exclude-newer = "7 days"
```
figueroa1395, line 85: 👍

pyproject.toml
figueroa1395: Isn't this potentially conflicting with some of the options in https://github.com/PowerGridModel/power-grid-model/pull/1499/changes#r3671710405? If I understand correctly this just uses `uv` by default, so again using a custom index would be an issue, right?

.github/workflows/check-code-quality.yml
mgovers: They are different purposes:

1. `sync` using `--locked`, meaning: sync everything you have, use the current lockfile. Raise an ERROR if you encounter any yanked releases of dependencies, but otherwise, use the lockfile exactly as-is. This is ideal for CI.
2. `run` using `--frozen`: assuming you've already run `sync` before, you don't need to check for yanked releases of dependencies anymore. This is also the configuration you need to use in production (because you don't want to "drop" your runtime)

So basically, `--locked` is stricter than `--frozen`, while `--frozen` is faster.

Technically, we could let everything run `--locked`, but there's not really a reason to run it twice. Even more so, it may actually introduce confusion if CI passes a sync but failed on a run because a release of a dependency is yanked while CI is already running.

.github/workflows/sonar.yml
mgovers: sync `--locked`, run `--frozen`

package-lock.json
mgovers: i had to run the sync job multiple times because of the dependencies. because of the custom index i use on my device, i couldn't run `uv lock --upgrade` locally

pyproject.toml
mgovers: optional dependency groups of `reuse` that are also used in the original pre-commit hook. They are used to handle utf-8 encoding. Without it, a utf-8 encoded file can't be loaded unless we specify some other flags, but that just makes `reuse` all the more unreadable.

pyproject.toml
mgovers: indeed it conflicts.

.github/workflows/check-code-quality.yml
figueroa1395: Clear explanation. I didn't realize the first `uv sync --locked` was setting up the following `uv run --frozen` steps. The same then applies for the code generation bit at the end.

Thanks! Resolving.

.github/dependabot.yml
mgovers:
```suggestion
      cronjob: "0 4 * * 2" # Every Tuesday at 4:00 UTC
```

.github/dependabot.yml
```diff
@@ -7,4 +7,23 @@ updates:
...
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 7
+
+  - package-ecosystem: "uv"
+    directory: "/"
+    schedule:
+      interval: "cron"
+      cronjob: "0 8 * * 1" # Every Monday at 8:00 UTC
```
mgovers:
```suggestion
      cronjob: "0 4 * * 2" # Every Tuesday at 4:00 UTC
```

.pre-commit-config.yaml
mgovers, line 10:
```suggestion
        entry: .venv/bin/reuse lint  # enable use of custom indexes
```

mgovers: @TonyXiang8787 @petersalemink95 what do you think about this? is this fine?

mgovers: Due to differences between Windows and Unix-like platforms (`.venv/Scripts` vs `.venv/bin`), this is not an option.

The decision was made to keep it the user's problem. A workaround exists and I will add it to the docs.

mgovers: Documented in https://github.com/PowerGridModel/.github/pull/112

mgovers: Cf. discussion offline: this cannot be done. Instead, documentation added in https://github.com/PowerGridModel/.github/pull/112

TonyXiang8787 (approved):

.github/dependabot.yml
```diff
@@ -7,4 +7,23 @@ updates:
...
+    directory: "/"
+    schedule:
+      interval: "cron"
+      cronjob: "0 4 * * 2" # Every Tuesday at 4:00 UTC
+    groups:
+      all-dependencies:
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 7
```
TonyXiang8787, line 29: Do we have duplicated `uv` update now? One in dependabot and one in our custom actions?

mgovers: yes. i did not find a way to combine the two as they represent updates from different ecosystems

mgovers: there is this pattern:

```yaml
version: 2

multi-ecosystem-groups:
  infrastructure:
    schedule:
      interval: "weekly"
    assignees: ["@platform-team"]
    labels: ["infrastructure", "dependencies"]

updates:
  - package-ecosystem: "docker"
    directory: "/"
    patterns: ["nginx", "redis", "postgres"]
    multi-ecosystem-group: "infrastructure"

  - package-ecosystem: "terraform"
    directory: "/"
    patterns: ["aws", "terraform-*"]
    multi-ecosystem-group: "infrastructure"
```

but i personally am not a fan of it either

mgovers: if you prefer, i can convert it, tho

TonyXiang8787: I do not understand. Once you have this dependabot, `uv.lock` will be updated automatically by that.

Why do we still need our custom actions to update `uv.lock`?

mgovers: i also misunderstood your comment.

After this PR, we have, on a weekly basis, 2 dependabot PRs. The github action no longer needs to update the uv.lock. However, I keep it in the code for now, because we can just remove the entire action in one go after #1502 and #1501 are merged.

PR #1501, opened 2026-07-28
CI: custom json formatter

mgovers:
In light of recent events, we are aiming to reduce our amount of dependencies - in particular the `npm` ones. Unfortunately, that means we need to revert some of the changes in #1256 in favor of small custom scripts.

Relates to #1499


scripts/format_json.py
```diff
@@ -0,0 +1,64 @@
...
+EXCLUDE = [
+    re.compile(rf"{REPO_PATH}/docs/.*"),
+    re.compile(rf"{REPO_PATH}/package-lock\.json"),
+    re.compile(rf"{REPO_PATH}/\.vscode/.*"),
+    re.compile(rf"{REPO_PATH}/\.devcontainer/.*"),
+    re.compile(rf"{REPO_PATH}/tests/data/.*/input\.json"),
+    re.compile(rf"{REPO_PATH}/tests/data/.*/.*_output.*\.json"),
+    re.compile(rf"{REPO_PATH}/tests/data/.*/update_batch\.json"),
+    re.compile(rf"{REPO_PATH}/tests/unit/deprecated/data/."),
+]
```
mgovers, line 23: we need this, or there will be some ugly error-prone repeated code between the code quality workflow and this repo

.pre-commit-config.yaml
```diff
@@ -47,15 +47,9 @@ repos:
         language: system
         pass_filenames: true
         types: [ cmake ]
-  - repo: https://github.com/igorshubovych/markdownlint-cli  # outside pypi 
-    rev: v0.49.1
-    hooks:
-      - id: markdownlint
-        args: ["--fix"]
-  # format .json and .jsonc files
```
mgovers, line 55: re-add if we choose not to merge #1502

.pre-commit-config.yaml.jinja
```diff
@@ -47,15 +47,9 @@ repos:
         language: system
         pass_filenames: true
         types: [ cmake ]
-  - repo: https://github.com/igorshubovych/markdownlint-cli  # outside pypi 
-    rev: v{{ markdownlint_cli }}
-    hooks:
-      - id: markdownlint
-        args: ["--fix"]
-  # format .json and .jsonc files
```
mgovers, line 55: re-add if we choose not to merge #1502

package-lock.json
mgovers, line 1: re-add if we choose not to merge #1502

package-lock.json.license
mgovers, line 1: re-add if we choose not to merge #1502

package.json
mgovers, line 1: re-add if we choose not to merge #1502

package.json.license
mgovers, line 1: re-add if we choose not to merge #1502

kornerc:
@mgovers I like to idea to get rid of npm.

One alternative came into my mind to a custom JSON formatter: Biome has [standalone binaries](https://biomejs.dev/guides/manual-installation) (it’s written in Rust).
However, I think it’s unfortunately rather difficult to integrate this nicely into pre-commit and to manage the versions properly.

Unfortunately, this PR to publish Biome as Python package is currently not progressing https://github.com/biomejs/biome/pull/8818

In summary I guess a custom JSON formatter - like your drafted one - might be the best solution (until there is a Python package for Biome).

mgovers:
> @mgovers I like to idea to get rid of npm.
> 
> One alternative came into my mind to a custom JSON formatter: Biome has [standalone binaries](https://biomejs.dev/guides/manual-installation) (it’s written in Rust). However, I think it’s unfortunately rather difficult to integrate this nicely into pre-commit and to manage the versions properly.
> 
> Unfortunately, this PR to publish Biome as Python package is currently not progressing [biomejs/biome#8818](https://github.com/biomejs/biome/pull/8818)
> 
> In summary I guess a custom JSON formatter - like your drafted one - might be the best solution (until there is a Python package for Biome).

Hi @kornerc,

Thank you for the input. We really appreciate it, especially after the great in-depth investigation you have done so far. 🤞 fingers crossed that that PR for biome will be published soon...

scripts/format_json.py
```diff
@@ -0,0 +1,70 @@
...
+    re.compile(rf"{REPO_PATH}/uv.lock"),
+]
+
+
+def format_file(path: str) -> bool:
+    """Reformat a single JSON file in place. Return True if it changed."""
+    file_path = Path(path)
+    original = file_path.read_text(encoding="utf-8-sig")
+
+    print(file_path)
```
mgovers:
```suggestion
```

scripts/format_json.py
```diff
@@ -0,0 +1,70 @@
...
+        p = Path(path)
+        full_path = str(p.resolve())
+        return p.suffix in (".json", ".jsonc") and not any(rex.match(full_path) for rex in EXCLUDE)
+
+    return [path for path in paths if _include_path(path)]
+
+
+def main(paths: list[str]) -> int:
+    format_paths = to_include(paths)
+    print(format_paths)
```
mgovers:
```suggestion
```

docs/advanced_documentation/build-guide.md
```diff
@@ -504,12 +504,12 @@ For optimal usage, open the folder using `code <project_dir>` from a terminal th
...
 ```
 
-For automatic formatting of JSON(C) files, you will need to have [Node.js](https://nodejs.org/) installed.
-You may also need to update the recommended extension Biome to the latest version using
-`manage > install specific version` for optimal up-to-date support.
+For automatic formatting of JSON(C) files, we use a custom formatter in `scripts/format_json.py`.
 Alternatively, as usual, you can also use
 {{ "[`pre-commit`]({}/CONTRIBUTING.md#pre-commit-hooks)".format(gh_link_head_blob) }} to keep all files
 correctly formatted before committing.
+You can use a formatter of your choice as well, but be aware that our CI may reject the format, so we always recommend
```
figueroa1395, line 511: Can this be something annoying to external users? "Easy" to fix with the custom formatter, but that's assuming it's "perfect", hence the business question.

code_generation/data/attribute_classes/input.json
```diff
@@ -29,12 +29,18 @@
       "attributes": [
         {
           "data_type": "ID",
-          "names": ["from_node", "to_node"],
+          "names": [
+            "from_node",
+            "to_node"
+          ],
```
figueroa1395, line 35: We previously used `biome` and besides is `npm` dependency, we thought it was fine. So my question is, do we want to re-format things again, or if we decide to have a custom formatter, do we want to keep it as close to `biome`, at least in the migration step, as possible?

tests/data/power_flow/automatic-tap-regulator/achievable-voltage-above-band-any/params.json
```diff
@@ -1,9 +1,9 @@
 {
   "calculation_method": "newton_raphson",
   "tap_changing_strategy": "any_valid_tap",
-  "rtol": 1e-5,
+  "rtol": 1e-05,
```
figueroa1395, line 4: I'd argue the old version is better. But this is subjective. This comment is mostly another instance of https://github.com/PowerGridModel/power-grid-model/pull/1501/changes#r3674359730

tests/data/power_flow/pandapower/components/asymmetric/node/params.json
```diff
@@ -1,5 +1,8 @@
 {
-  "calculation_method": ["newton_raphson", "iterative_current"],
-  "rtol": 1e-5,
-  "atol": 1e-4
+  "calculation_method": [
+    "newton_raphson",
+    "iterative_current"
+  ],
+  "rtol": 1e-05,
+  "atol": 0.0001
```
figueroa1395, line 7: Where do we draw the line to use scientific notation?

tests/data/power_flow/vision-validation-network/params.json
```diff
@@ -1,20 +1,20 @@
...
     "i_to": 0.5,
-    "s_to": 0.5e3,
-    "p": 0.5e3,
-    "q": 0.5e3,
+    "s_to": 500.0,
+    "p": 500.0,
+    "q": 500.0,
     "i": 0.5,
-    "s": 0.5e3
+    "s": 500.0
```
figueroa1395, line 18: I personally like this version more, but same comment as before.

scripts/format_json.py
figueroa1395: I'd argue that if we add this custom formatter, let's also include the following three
```
re.compile(rf"{REPO_PATH}/tests/data/.*/input\.json"),
re.compile(rf"{REPO_PATH}/tests/data/.*/.*_output.*\.json"),
re.compile(rf"{REPO_PATH}/tests/data/.*/update_batch\.json"),
```
I don't see what's the problem. We would get a BIG PR but it's just formatting.

scripts/format_json.py
figueroa1395, line 1: Now that I see how simple the custom formatter is, I really don't see why we shouldn't go with this at least as a temporary solution. It isn't perfect but introduces some readability and consistence. Maintenance burden is minimal.

Given the proposed solution, the initial small comments about specific style can probably be omitted.

figueroa1395 (commented):
This is generally fine by me, but let's discuss with the rest of the team before proceeding.

docs/advanced_documentation/build-guide.md
mgovers: in the end, our CI will be leading. Generally, our previous requirements on formatting differed from the norm. The new one is more in line with what's considered "normal"

scripts/format_json.py
mgovers: yeah, that was my thought as well 😬 the only piece of extra configuration we could do is to disallow `NaN` (in line with the JSON standard; we currently allow it as the default of the `json` module) or we would have to create our custom serializer and i really would not like to go down that path.

scripts/format_json.py
mgovers: currently, we do a lot of compact formatting in our input/output/update. that would no longer be possible and also significantly increase test data size with literally only whitespaces due to indentation. this was also discussed when introducing `biome`

scripts/format_json.py
nitbharambe: To make this case strong, it would be better to fill this with full documentation / intent so that AI agent can take over maintenance of this tool. Maybe we define a schema. Something that is understandable to both us + AI.

kornerc: I briefly looked over the source code and its functionality looks pretty similar to the pre-commit-hook [pretty-format-json](https://github.com/pre-commit/pre-commit-hooks/blob/4189d189e039fd49eb235357c9d9977b026c6c8e/pre_commit_hooks/pretty_format_json.py): it uses Python's `json.loads` and afterwards `json.dumps` for formatting (but additionally adds a `\n` in the end).

`pretty_format_json` was also my first attempt in the "[Biome PR](https://github.com/PowerGridModel/power-grid-model/pull/1256)".
Based on the wish (https://github.com/PowerGridModel/power-grid-model/pull/1256#issuecomment-3722536930) to have max indentation I found Biome as suitable alternative.

Thus, it might be a possibility to revisit `pretty_format_json` instead of the custom formatter.
This would also allow to have consistent JSON-formatting across different `power-grid-model` repos by just applying the pre-commit-hook.

pre-commit allows also individual [include and exclude patterns](https://pre-commit.com/index.html#pre-commit-configyaml---hooks) per hook.

mgovers: you're right. i thought it was not available via PyPI but it is (we would like to get rid of git repo hooks as dependencies for governance reasons, so that would have been a dealbreaker)

that said, some people are leaning towards not having a JSON formatter at all, so that the barrier for external contributions (in particular test cases) is lowered, if ever so slightly. Formatting is fairly opinionated, so it will always come with some downsides.

Thanks for the input! We'll keep you updated!

mgovers: I see that pretty_format_json is unmaintained (1 watching, 1 star, last release 2020) and the code is also... questionable in some places (e.g. eval to call node to pretty-print errors, feels like a security risk we are better off not having). I think by this time it is safer to use the custom implementation. Thoughts?

figueroa1395: A custom implementation is fine from my side.

kornerc: > I see that pretty_format_json is unmaintained (1 watching, 1 star, last release 2020) and the code is also... questionable in some places (e.g. eval to call node to pretty-print errors, feels like a security risk we are better off not having). I think by this time it is safer to use the custom implementation. Thoughts?

Apparently there is the [pre-commit-hook](https://github.com/pre-commit/pre-commit-hooks/blob/4189d189e039fd49eb235357c9d9977b026c6c8e/pre_commit_hooks/pretty_format_json.py) (6k stars)  and an own [repo](https://github.com/weaming/pretty-format-json) (1 star)  called _pretty-format-json_ which seem not to be related to each other.
The pre-commit-hook seems to be much more trustworthy.
However, it is to the best of my knowledge not on PyPi.
Since it _must be_ on PyPi, I think a custom formatter is the best idea.
(Sorry for the noise)

mgovers: It's been great input nevertheless! Thanks for the time and effort you've spent on this!

.github/workflows/check-code-quality.yml
```diff
@@ -61,10 +61,10 @@ jobs:
           echo "Running markdownlint"
           npx markdownlint --fix .
 
-      - name: Run Biome
+      - name: Run Json formatter
         run: |
-          echo "Running Biome"
-          npx biome format --write --files-ignore-unknown=true --no-errors-on-unmatched
+          echo "Running Json formatter"
+          git ls-files -z -- '*.json' '*.jsonc' | xargs -0 -r uv run --frozen scripts/format_json.py
```
kornerc: `*.jsonc` with comments or trailing commas cannot be loaded by `json.loads()`.

**Comment:**
```python
import json

print(json.loads(
"""
{
  // comment
  "a": 1
}
"""
))
```

**Trailing comma:**
```python
import json

print(json.loads(
"""
{
  "a": 1,
}
"""
))
```

Even if `json.loads()` would be able to loads .jsonc files with comments, the comments would get lost.

Therefore, it might be a good decision to _not format_ `.*jsonc` files

mgovers: good catch

mgovers:
Awaiting further investigation; closing for now

PR #1502, opened 2026-07-28
CI: use pymarkdownlnt

mgovers:
In light of recent events, we are aiming to reduce our amount of dependencies - in particular the `npm` ones. That means we need to migrate to a markdown linter in the PyPI ecosystem.

Relates to #1499 and #1501 .

Important things to note:

* wide tables and KaTeX need to be explicitly marked `<!-- pyml disable line-length-->` and `<!-- pyml enable line-length-->`.
  * There is NO way to do this inline yet.
  * It needs to be done at the block-level (no indentation), meaning you can only do a full disable and enable for that table/block...


.pre-commit-config.yaml.jinja
```diff
@@ -47,11 +47,23 @@ repos:
...
+        types: [ markdown ]
+        require_serial: true
+      - id: pymarkdown-scan
+        name: pymarkdown-scan
+        alias: markdown-check
+        entry: uv run --frozen pymarkdown scan
+        language: system
+        pass_filenames: true
+        types: [ markdown ]
+        require_serial: true
```
mgovers, line 65: for some weird reason, `require_serial: true` is needed to prevent a new environment from being created and used. I tried digging really deeply into it, including setting environment variables, etc, but pre-commit kept trying to create a new environment but not pulling data from the pypi mirror index i specified but from the pythonhosted one. I still haven't figured out why but it seems to have something to do with different environments and multithreading

docs/advanced_documentation/build-guide.md
```diff
@@ -355,9 +363,11 @@ conda create --yes -p C:\conda_envs\cpp_pkgs -c conda-forge libboost-headers eig
 
 Set `CMAKE_PREFIX_PATH` so CMake can locate the C++ libraries. In PowerShell:
 
+<!-- pyml disable commands-show-output-->
 ```powershell
 $env:CMAKE_PREFIX_PATH = C:\conda_envs\cpp_pkgs\Library
 ```
+<!-- pyml enable commands-show-output-->
```
mgovers, line 366: some kind of weird false positive where a `$` is interpreted as unix `$`, not as Windows `$`

.pre-commit-config.yaml.jinja
mgovers: See https://github.com/PowerGridModel/power-grid-model/pull/1499#discussion_r3671710405

.pre-commit-config.yaml
```diff
@@ -47,11 +47,23 @@ repos:
...
+      - id: pymarkdown-scan
+        name: pymarkdown-scan
+        alias: markdown-check
+        entry: uv run --frozen pymarkdown scan
+        language: system
+        pass_filenames: true
+        types: [ markdown ]
+        require_serial: true
+        # format .json and .jsonc files
   # format .json and .jsonc files
```
figueroa1395: Isn't this for `.md` files? not `json`, right? Also, the line is duplicated.

.github/workflows/check-code-quality.yml
```diff
@@ -52,14 +52,14 @@ jobs:
...
-          npm ci
+          echo "Running pymarkdown"
+          git ls-files -z -- '*.md' | xargs -0 -r uv run --frozen pymarkdown fix
 
-      - name: Run markdownlint
+      - name: Install npm packages
         run: |
-          echo "Running markdownlint"
-          npx markdownlint --fix .
+          npm ci
```
figueroa1395, line 63: This is to be removed altogether is the custom json formatter is accepted, right? It was just a bit weird to see these changes here.

figueroa1395 (commented):
Even though a "newer" tool, it seems to have a good number of [downloads](https://pypistats.org/packages/pymarkdownlnt) and be actively maintained. My only concern is that it seems like it's mainly maintained by one guy (cfr. [github](https://github.com/jackdewinter/pymarkdown)).

Other than that, I have no issue with it. In any case we can always switch or drop such a tool entirely.

.pre-commit-config.yaml
mgovers: Yup, whoops, caused by merge conflict...
```suggestion
  # format .json and .jsonc files
```

.github/workflows/check-code-quality.yml
mgovers: Yes indeed. This is just github diff that's messed up because this line was also at the start of this block but apparently it thinks i "removed" that line but i did not

docs/user_manual/components.md
```diff
@@ -910,6 +948,7 @@ measurement.
 
 2. If neither `p_sigma` nor `q_sigma` are provided, `power_sigma` represents the standard deviation of the apparent
 power. In this case, infinite value of `power_sigma` disables the entire measurement.
+<!-- pyml enable line-length-->
```
mgovers:
```suggestion
```

pyproject.toml
```diff
@@ -201,6 +204,13 @@ show_column_numbers = true
 non_interactive = true
 install_types = true
 
+[tool.pymarkdown]
+plugins.heading-increment.enabled = false
+plugins.no-duplicate-heading.enabled = false
+plugins.line-length.line_length = 120
+plugins.line-length.heading_line_length = 120
+plugins.line-length.code_block_line_length = 120
```
mgovers: TODO: check if we can disable line length on code blocks and tables. if not, then disable entirely

nitbharambe:
Some 3 markdownlint ignores are present in the docs. Would you do a global find on them?

docs/algorithms/lu-solver.md
```diff
@@ -1214,9 +1217,10 @@ $$
 \end{bmatrix}
 $$
 
+<!-- pyml disable line-length -->
 * The regular $L_{\infty}$-norm is $\max\left\{20+20+2+2,30+3,100,3+1\right\} = \max\left\{44,33,100,4\right\} = 100$.
 * The block-wise infinity norm with diagonals would be
-  <!-- markdownlint-disable-next-line line-length -->
```
figueroa1395, line 1219: I assume a similar per-line disabling command doesn't exist in `pymarkdownlint`? And now we have to do it instead over blocks?

That's fine by me, it's the same as before but just a different "style".

docs/user_manual/data-model.md
```diff
@@ -14,6 +14,7 @@ The components types are organized in an inheritance-like hierarchy.
 A sub-type has all the attributes from its parent type.
 The hierarchy tree of the component types is shown below.
 
+<!-- pyml disable line-length -->
```
figueroa1395, line 17: This is the only other type of instance where we have this disable block, right? When we have graphs and these we can't exclude from the check like tables, correct?

figueroa1395 (commented):
I see the potential issues were addressed. The remaining "exception" blocks from the linter are acceptable to me: only long equations and graphs, as far as I can tell.

The table issue has been resolved.

LGTM.

docs/user_manual/data-model.md
mgovers: yes, we want a 120-line code block width, but we can't do that for mermaid, so we just disable it explicitly

docs/algorithms/lu-solver.md
mgovers: > I assume a similar per-line disabling command doesn't exist in pymarkdownlint? And now we have to do it instead over blocks?

exactly. There is a `disable-next-line` pragma but it does not work on indented code blocks like enumerations

mgovers:
Awaiting further investigation; closing for now

PR #1520, opened 2026-08-06
Merge links on nodes: create user node output from topo node output

mgovers:
Part of #35 .

* [ ] Create the topological nodes from the user nodes + links
* [ ] Accumulate the appliances to get the bus injection on each user node


tests/data/power_flow/dummy-test/sym_output.json
```diff
@@ -18,7 +18,7 @@
         "u_pu": 0.84,
         "u": 8400.0,
         "u_angle": 0.0,
-        "p": -0.00961,
+        "p": 0.0,
```
mgovers, line 21: this was actually a rounding issue in the original `main` as a result of how we used Ybus to calculate the bus injection. Although the difference is small, it is significant enough to fail the tests on this branch, so it needed to be updated.

This branch is correct, as the bus injection is the sum of all customers' injections (sources, loads and generators), and there are no customers on this node.

tests/data/state_estimation/distribution-case/sym_output_batch.json
mgovers, line 1: i decided to make the output here exact: node injection is the sum of all customers' injections (sources, loads and generators). If there is exactly 1 customer, then the bus injection can be exactly determined

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
```diff
@@ -83,4 +84,89 @@ template <std::derived_from<Regulator> ComponentType, class ComponentContainer>
 constexpr auto get_math_id(MainModelState<ComponentContainer> const& state, Idx topology_sequence_idx) {
     return state.topo_comp_coup->regulator[topology_sequence_idx];
 }
+
+template <std::same_as<TopologicalNode> Component, class ComponentContainer>
+    requires model_component_state_c<MainModelState, ComponentContainer, Node>
+constexpr auto comp_base_sequence_cbegin(MainModelState<ComponentContainer> const& state) {
+    return state.topo_comp_coup->node.cbegin();
+}
```
figueroa1395: This overload isn't needed. If I remove everything compiles and tests run successfully.

Do you remember why was it added?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/math_output_queries.hpp
```diff
@@ -0,0 +1,62 @@
...
+template <solver_output_type SolverOutputType>
+constexpr auto get_voltage_output(MathOutput<std::vector<SolverOutputType>> const& math_output, Idx2D const& math_id) {
+    if constexpr (steady_state_solver_output_type<SolverOutputType>) {
+        return math_output.solver_output[math_id.group].u[math_id.pos];
+    } else if constexpr (short_circuit_solver_output_type<SolverOutputType>) {
+        return math_output.solver_output[math_id.group].i[math_id.pos];
+    } else {
+        static_assert(false, "Unsupported solver output type for voltage output retrieval");
+    }
+}
```
figueroa1395: - This one is not used anywhere. Would we need it considering the implementation in `topological_node_output.hpp`?
- Returning `i` for the `short_circuit_solver_output_type<SolverOutputType>` branch seems strange due to the function name. Is this intended?
Due to the above, I believe this is from a "legacy" implementation. Can I remove it?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/math_output_queries.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+template <steady_state_solver_output_type SolverOutputType>
+constexpr auto get_bus_injection_output(MathOutput<std::vector<SolverOutputType>> const& math_output,
+                                        Idx2D const& math_id) {
+    return math_output.solver_output[math_id.group].bus_injection[math_id.pos];
+}
```
figueroa1395: I don't think this is no longer needed either per the implementation in `topological_node_output.hpp`. Can I remove it?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -133,29 +55,40 @@ constexpr void produce_output(MainModelState<ComponentContainer> const& state, C
...
+constexpr auto output_result(Component const& node, MainModelState<ComponentContainer> const& state,
+                             MathOutput<std::vector<SolverOutputType>> const& math_output, Idx2D const& topo_id) {
+    auto const& math_id = get_math_id<Component>(state, topo_id.group);
+
     if (math_id.group == disconnected) {
         return node.get_null_sc_output();
     }
-    return node.get_sc_output(solver_output[math_id.group].u_bus[math_id.pos]);
+
+    return node.get_sc_output(math_output.solver_output[math_id.group].u_bus[math_id.pos]);
```
figueroa1395, line 84: So instead of processing the short circuit output via the super node output, we directly do it at the final output step?

This is fine now, but when we do use the link algorithm, we'd need to change this via `supernode_output` as well. Wouldn't we?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topological_node_output.hpp
figueroa1395, line 1: Tied to comment https://github.com/PowerGridModel/power-grid-model/pull/1520/changes#r3759633981. I see that all the logic involving short circuit was removed for the node output via the supernode. Was that intended or was that mistakenly removed while cherry picking?

At the moment everything works because we don't yet use the link algorithm, but we will need it. I assume it was a mistake, but I want to confirm it.

Cfr., [this](https://github.com/PowerGridModel/power-grid-model/pull/1520/changes/df3a3a77edb64d49f9144f88197ce00fa87514a2#diff-81382e6fdbbb534a0ea062b1ecd3f8a8d72290a2bb2b09d0c6e1e19bcbc8f16a) is where most of the removal happened.

tests/cpp_unit_tests/main_core/test_topological_node_output.cpp
figueroa1395, line 1: Tied to https://github.com/PowerGridModel/power-grid-model/pull/1520/changes#r3759686460. The relevant short circuit tests were removed. Was that intended?

Cfr., [this](https://github.com/PowerGridModel/power-grid-model/pull/1520/changes/df3a3a77edb64d49f9144f88197ce00fa87514a2#diff-81382e6fdbbb534a0ea062b1ecd3f8a8d72290a2bb2b09d0c6e1e19bcbc8f16a) is where most of the removal happened.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/math_output_queries.hpp
mgovers: right, this was one of my attempts to make things work. let me actually create a new branch to see if i can get it to work cause it is a good change. after that, will cherry-pick out of this one

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
mgovers: i do remember but indeed no longer relevant 😬

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
mgovers: see https://github.com/PowerGridModel/power-grid-model/pull/1529 , voltage can be obtained directly from the topological node as it's equal across all user nodes in the same supernode

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topological_node_output.hpp
mgovers: it was intended. there's no node output except voltage which can be directly obtained from the supernode. no need to copy everything

tests/cpp_unit_tests/main_core/test_topological_node_output.cpp
mgovers: yes it was intended. there is no short circuit output for nodes. They are re-added for links in https://github.com/PowerGridModel/power-grid-model/pull/1527

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/math_output_queries.hpp
mgovers: also cherry-picked out into #1529

mgovers:
We'd like to squash this PR first given the length of the thread and complexity from cherry-picking changes out into other branches. After that is done, we can remove the `do-not-merge` label.

mgovers:
done + do-not-merge label removed

figueroa1395 (approved):

PR #1529, opened 2026-08-12
Chore: cleanup output.hpp

mgovers:
Cleanup after #1520 .

The conversion from math output to output data can benefit from a serious cleanup, especially because it's become a bit tricky to extend. During work on #1520 (contributing to #35 ), some initial work was added. However, since it did not directly contribute to the output, it was cherry-picked out of that PR. Since it's still beneficial to the overall PGM core, here it is.

NOTE: this is not a final state but a step in the right direction. It has become clear that there's some duplication that can be cleaned up during follow-up work


nitbharambe (approved):

mgovers:
going to be a big merge conflict with #1545 . Trying to rebase the branch. `do-not-merge` label added

nitbharambe (dismissed):

TonyXiang8787 (approved):

PR #1530, opened 2026-08-12
Release artifacts: add C/C++ binaries

mgovers:
* [x] Use CPack to create a platform-compatible binary with a name of the form `libpower-grid-model-<major>.<minor>-<platform>-<architecture>`
* [x] Upload those binaries as release artifacts

This enables using pre-built binaries in `libpower-grid-model-java` codegen.


.github/workflows/build-test-release.yml
```diff
@@ -87,6 +87,18 @@ jobs:
...
+      - name: Package binary
+        if: matrix.compiler == 'gcc' && matrix.build-option == 'release'
+        run: cpack --preset ${{ env.PRESET }}
+
+      - name: Upload binary
+        if: matrix.compiler == 'gcc' && matrix.build-option == 'release'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: libpower-grid-model-linux-x64
+          path: cpp_build/${{ env.PRESET }}/libpower-grid-model-*.tar.gz
```
mgovers, line 105: we also need a separate pipeline for linux aarch64

mgovers:
As discussed: release artifacts will depend on a relatively modern version of glibc, which defeats the purpose, as the Java chain should also be able to work with older glibc versions.

PR #1534, opened 2026-08-14
Revert "Merge links on nodes: create user node output from topo node output"

mgovers:
Reverts PowerGridModel/power-grid-model#1520

We found an issue regarding node injection output. The intention is to have the Kirchoff-calculated output in the node injection output, while the individual injection outputs have the directly-calculated injections. That way, the user is able to calculate the power mismatch themselves from the difference between the Kirchoff node injection output and the sum of all injections on each node.

In #1520, the node injection output was changed to the sum of the individual outputs. This is a bug. Since fixing it may carry over the weekend, let's revert the change and make sure that `main` is in a good state again before the weekend.


nitbharambe:
I am still at loss about what the issue is. 
It would be nice if we can add a validation case reproducing the bug.

By the description, is this what you mean by "power mismatch"?: 

- the del P/Q itself is incorrectly calculated: 
    -  This is not the case. It gets calculated using input.s_injection and calcualted p,q during jacobian creation. equivalent as y_bus.bus_injection but calculated from other way.

- Using output_data["node"]["p"], users can find out what the final deviation of power is after end of iterations. This can be useful since we provide error_tolerance in voltage p.u. as opposed to delta P/Q which is what most users are accustomed to. Hence this bus_injection becomes noticeable at a higher error tolerance.
    - We never documented this intent. We dont guarantee it either. It is a nice way to obtain it. But it can indeed be redirected via logger in future. I am not aware of any user using this actively by checking at each iteration to enforce error_tolerance in power instead of voltage.

mgovers:
> * the del P/Q itself is incorrectly calculated:
>   * This is not the case. It gets calculated using input.s_injection and calcualted p,q during jacobian creation. equivalent as y_bus.bus_injection but calculated from other way.

Indeed, this was not the intention here.

> * Using output_data["node"]["p"], users can find out what the final deviation of power is after end of iterations. This can be useful since we provide error_tolerance in voltage p.u. as opposed to delta P/Q which is what most users are accustomed to. Hence this bus_injection becomes noticeable at a higher error tolerance.
>   * We never documented this intent. We dont guarantee it either. It is a nice way to obtain it. But it can indeed be redirected via logger in future. I am not aware of any user using this actively by checking at each iteration to enforce error_tolerance in power instead of voltage.

This is indeed the main reason and also what I understood was the original intention of @TonyXiang8787 , but indeed it is undocumented. The main decision now is: do we want to officially support it or not?

That said, if we revert this PR, we buy ourselves time to make this decision.

> It would be nice if we can add a validation case reproducing the bug.

https://github.com/PowerGridModel/power-grid-model/pull/1534/changes#diff-f84f9ef1068e20c3b08ccce77f602717a1ddd3d840705cfd903704cc57146a86 is a validation case for this

mgovers:
As discussed offline: we do not consider node injection output via Ybus to be a stable feature (it was undocumented). Using the node injection output to obtain the power mismatch may be considered a "hack".

PR #1540, opened 2026-08-18
Fix sensor output for disconnected measured components

### LINKED ISSUE #1464: [BUG] inconsistent sensor output behavior on disconnected branches vs appliances
### Describe the bug

When running state estimation with a disabled power/current sensor on a disabled branch, the output is empty (`energized = 0, p_residual = 0, q_residual = 0`).

However, when running state estimation with a disabled power/current sensor on a disabled shunt or load/gen, the output is not empty (`energized = 1, p_residual = 6924.0000000000009, q_residual = -69280`).

This behavior is unexpected

### Input Data Validity

- [x] I checked the validity of input data using the helper function ([examples here](https://power-grid-model.readthedocs.io/en/stable/examples/Validation%20Examples.html)).

## Input data

```json
{
  "version": "1.0",
  "type": "input",
  "is_batch": false,
  "attributes": {},
  "data": {
    "node": [
      {
        "id": 1,
        "u_rated": 10e3
      }
    ],
    "line": [
      {
        "id": 5,
        "from_node": 1,
        "to_node": 1,
        "from_status": 0,
        "to_status": 0,
        "r1": 1000000.0,
        "x1": 1000000.0,
        "c1": 2.0e-5,
        "tan1": 0.1,
        "r0": 1000000.0,
        "x0": 1000000.0,
        "c0": 1.0e-6,
        "tan0": 0.2
      }
    ],
    "shunt": [
      {
        "id": 7,
        "node": 1,
        "status": 0,
        "g1": 6.283185307179587e-4,
        "b1": 6.283185307179587e-3,
        "g0": 6.283185307179587e-5,
        "b0": 3.141592653589793e-4
      }
    ],
    "source": [
      {
        "id": 10,
        "node": 1,
        "status": 1,
        "u_ref": 1.05,
        "sk": 1e12
      }
    ],
    "sym_voltage_sensor": [
      {
        "id": 13,
        "measured_object": 1,
        "u_measured": 6062.1931942119563,
        "u_sigma": 0.1,
        "u_angle_measured": -1.4832085023115637e-06,
        "u_angle_sigma": 0.1
      }
    ],
    "asym_power_sensor": [
      {
        "id": 16,
        "measured_object": 5,
        "measured_terminal_type": 0,
        "power_sigma": 5.0,
        "p_measured": [1154.0, 1153.0, 1155.0],
        "q_measured": [-11540.0, -11550.0, -11550.0]
      },
      {
        "id": 17,
        "measured_object": 7,
        "measured_terminal_type": 3,
        "power_sigma": 10.0,
        "p_measured": [2308.0, 2306.0, 2310.0],
        "q_measured": [-23080.0, -23100.0, -23100.0]
      }
    ]
  }
}
```
## Update data

```json
{
  "version": "1.0",
  "type": "update",
  "is_batch": true,
  "attributes": {},
  "data": [
      {
          "line": [
              {
                  "id": 5,
                  "from_status": 0,
                  "to_status": 0
              }
          ],
          "shunt": [
              {
                  "id": 7,
                  "status": 1
              }
          ],
          "asym_power_sensor": [
              {
                  "id": 16,
                  "power_sigma": "inf"
              }
          ]
      },
      {
          "line": [
              {
                  "id": 5,
                  "from_status": 1,
                  "to_status": 1
              }
          ],
          "shunt": [
              {
                  "id": 7,
                  "status": 0
              }
          ],
          "asym_power_sensor": [
              {
                  "id": 17,
                  "power_sigma": "inf"
              }
          ]
      }
  ]
}
```

## To reproduce

Run symmetric state estimation with `newton_raphson`

## Symmetric output data (current)

```json
{
  "version": "1.0",
  "type": "sym_output",
  "is_batch": true,
  "attributes": {},
  "data": [
    {
      "asym_power_sensor": [
        {"id": 16, "energized": 0, "p_residual": 0, "q_residual": 0},
        {"id": 17, "energized": 1, "p_residual": -14400.013604963937, "q_residual": 143960.13604963932}
      ]
    },
    {
      "asym_power_sensor": [
        {"id": 16, "energized": 1, "p_residual": -7200.0068024819693, "q_residual": 71980.068024819673},
        {"id": 17, "energized": 1, "p_residual": 6924.0000000000009, "q_residual": -69280}
      ]
    }
  ]
}
```

### To Reproduce

Run state estimation with e.g. `newton_raphson` calculation method and `symmetric = True`

### Expected behavior

The sensor output on the disabled shunt is empty

## Symmetric output data (expected)

```json
{
  "version": "1.0",
  "type": "sym_output",
  "is_batch": true,
  "attributes": {},
  "data": [
    {
      "asym_power_sensor": [
        {"id": 16, "energized": 0, "p_residual": 0, "q_residual": 0},
        {"id": 17, "energized": 1, "p_residual": -14400.013604963937, "q_residual": 143960.13604963932}
      ]
    },
    {
      "asym_power_sensor": [
        {"id": 16, "energized": 1, "p_residual": -7200.0068024819693, "q_residual": 71980.068024819673},
        {"id": 17, "energized": 0, "p_residual": 0, "q_residual": 0}
      ]
    }
  ]
}
```

### Screenshots

_No response_

### Additional context

This will need to be fixed in the `output.hpp`.

The underlying reason is that, topologically, branches are treated differently from appliances for caching reasons:

* disabling branches (and source appliances) changes the topology and therefore, the entire topology needs to be rebuilt
* disabling appliances (except sources) does not change the topology - only the Ybus parameters (shunts) or even only the actual values (loads). The topology is therefore cached. In particular, appliances will always be added to the topology and the actual output for those components is created based on the internal status - not the topological configuration.

However, the current sensor output only looks at the topological status of the measured object. This results in different behavior depending on whether the topological status is updated or not when the measured object status changes.

Instead of checking the measured object topology, the measured object status should be used to determine whether the sensor should get `null_output`.

### Community Meeting

- [ ] I wish to discuss this bug in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

appleweiping:
## Summary

- Return null sensor output when a measured source, shunt, load, or generator is disabled, or when the measured terminal of a branch or branch3 is disconnected.
- Preserve the topology-group check for grids that are not connected to a source, and apply the same terminal-status semantics to power and current sensors.
- Preserve reduced main-core component-container compatibility with compile-time guards for component types that a container omits.
- Re-enable the existing symmetric/asymmetric batch regression cases, add focused C++ coverage, and document how sensor-output `energized` is derived at the measured terminal.

## Root cause

Sensor output only checked whether the measured object had a connected mathematical topology group. Status-only updates for shunts, loads, and generators do not rebuild that topology, so a disabled appliance could still produce `energized == 1` and non-zero residuals. A partially disconnected branch or branch3 also retains a mathematical group through its connected terminals, so the group alone does not describe whether the terminal measured by a sensor is connected.

The output path now keeps the topology check and additionally reads the actual status of the measured appliance or measured branch terminal. Power and current sensors share the same terminal-status mapping.

Because the measured terminal is selected at runtime, all lookup branches are instantiated at compile time. Compile-time type guards preserve support for reduced component containers that omit unused measured-object types.

## Validation

- `uv run --frozen pytest` — 1778 passed, 4 skipped, 13 xfailed; 97.13% coverage, using the newly built C++ core.
- Release `power_grid_model_c` build — passed, including production `model.cpp` template instantiation and DLL linking.
- Focused C++ main-core build and CTest selection — 9/9 passed, including symmetric/asymmetric power and current sensor cases.
- The two affected dataset fixture families — 9/9 passed.
- `uv run --frozen ruff check .`, `uv run --frozen reuse lint`, clang-format dry-run, and `git diff --check` — passed.
- The old behavior returns energized output for a sensor on a disconnected terminal of a partially connected branch/branch3; the new regression cases require null output with zero residuals.

The temporary Python migration reference was removed after the maintainer resolved the C++ test-migration thread.

No public API, schema, or generated file is changed.

Closes #1464

Assisted-by: OpenAI Codex


docs/user_manual/components.md
```diff
@@ -764,6 +764,8 @@ The state estimator uses the data to evaluate the state of the grid with the hig
 
 A sensor only has output for state estimation.
 For other calculation types, sensor output is undefined.
+Residual values are calculated only when the sensor output is marked as energized (`energized == 1`). When
+`energized == 0`, all residual values are zero.
```
mgovers: `energized` is a calculated value, not a user input. please explicitly mention the underlying meaning instead.

tests/data/state_estimation/dummy-test-line-into-itself-minimal/params.json
```diff
@@ -22,9 +22,5 @@
     "p_residual": 2e-3,
     "q_residual": 1e-2,
     "u_residual": 1e-3
-  },
-  "xfail": {
-    "raises": "AssertionError",
-    "reason": "Energized behaves differently between sensors on disabled shunts and on disabled branches (https://github.com/PowerGridModel/power-grid-model/issues/1464)"
```
mgovers, line 28: :+1:

tests/data/state_estimation/dummy-test-line-into-itself-minimal/sym_output_batch.json
```diff
@@ -40,11 +40,11 @@
       ],
       "asym_power_sensor": [
         {"id": 16, "energized": 1, "p_residual": -7200.0068024819684, "q_residual": 71980.068024819659},
-        {"id": 17, "energized": 0, "p_residual": 6924, "q_residual": -69280}
+        {"id": 17, "energized": 0, "p_residual": 0, "q_residual": 0}
```
mgovers, line 43: :+1:

tests/data/state_estimation/dummy-test-line-into-itself/asym_output_batch.json
```diff
@@ -71,12 +71,12 @@
         {"id": 13, "energized": 1, "p_residual": [-51770.4106748376, -51847.591438489362, -51839.393964931936], "q_residual": [-25635.054141311452, -25643.885638187403, -25585.132207542603]},
         {"id": 15, "energized": 1, "p_residual": [-10156.957454722986, -10157.477366083891, -10155.854711926075], "q_residual": [101562.71307273903, 101562.20801004104, 101562.97424445581]},
         {"id": 16, "energized": 1, "p_residual": [-10110.025940894742, -10194.044413720736, -10080.765281441463], "q_residual": [101307.65023838131, 101349.52127187414, 101201.17907509189]},
-        {"id": 17, "energized": 0, "p_residual": [2308, 2306, 2310], "q_residual": [-23090, -23090, -23090]},
-        {"id": 18, "energized": 0, "p_residual": [2304, 2310, 2296], "q_residual": [-23040, -23040, -23000]}
+        {"id": 17, "energized": 0, "p_residual": [0, 0, 0], "q_residual": [0, 0, 0]},
+        {"id": 18, "energized": 0, "p_residual": [0, 0, 0], "q_residual": [0, 0, 0]}
```
mgovers, line 75: :+1:

tests/data/state_estimation/dummy-test-line-into-itself/params.json
```diff
@@ -23,9 +23,5 @@
     "q_residual": 2e1,
     "u_residual": 1e-2,
     "u_angle_residual": 1e-5
-  },
-  "xfail": {
-    "raises": "AssertionError",
-    "reason": "Energized behaves differently between sensors on disabled shunts and on disabled branches (https://github.com/PowerGridModel/power-grid-model/issues/1464)"
```
mgovers, line 29: :+1:

tests/unit/test_sensor_output_energized.py
mgovers, line 1: please make this as C++ unit tests, not Python unit tests. Feel free to leave this in for now for reference until you've finished with the C++ unit tests, and then remove this file once those have been accepted.

docs/user_manual/components.md
```diff
@@ -764,6 +764,12 @@ The state estimator uses the data to evaluate the state of the grid with the hig
 
 A sensor only has output for state estimation.
 For other calculation types, sensor output is undefined.
+In state estimation output, `energized` is calculated from the measured object rather than supplied as sensor input.
+It is `1` when the measured object is electrically connected to a source. For an appliance (`source`, `shunt`, `load`,
+or `generator`), its connection `status` must also be `1`. The value is independent of whether the measurement is
+included in the estimation. For a `branch` or `branch3`, connectivity is evaluated for the component as a whole: an
+open terminal does not make the sensor output unenergized while the component still contributes to the energized grid.
+If the measured object is not energized, `energized` is `0` and all residual values are zero.
```
mgovers: this should be summarized. Now, it also contains a lot of explanation on the energized state of the measured object. This is duplication. This section should restrict its explicit scope to the sensor output.

docs/user_manual/components.md
```diff
@@ -764,6 +764,8 @@ The state estimator uses the data to evaluate the state of the grid with the hig
 
 A sensor only has output for state estimation.
 For other calculation types, sensor output is undefined.
+The sensor output attribute `energized` is derived from the energized state of the measured object and is independent
+of whether the measurement is included in state estimation. When `energized == 0`, all residual values are zero.
```
mgovers:
```suggestion
The sensor output attribute `energized` is derived from the energized state of the measured object, regardless
of whether the measurement is included in state estimation.
If the measured object is not energized, all residual values are zero.
```

mgovers: or otherwise

```suggestion
The sensor output attribute `energized` is derived from the energized state of the measured object, regardless of
of whether the measurement itself was enabled or disabled.
If the measured object is not energized, all residual values are zero.
```

mgovers:
Hi @appleweiping, Something seems to have gone wrong with the base branch. I had to change the base branch due to some blocking things in the root, but the base branch was not automatically updated. Can you please re-open the PR and point the base branch to `PowerGridModel:main`?

nitbharambe:
I see this PR got closed because underlying branch was deleted. That branch and PR is now merged. 
@appleweiping would you please create a new PR where we can continue with the contribution?

nitbharambe:
Nevermind, resolved! We will continue with review here.

mgovers:
Hi @appleweiping,

It looks like compilation is failing for MSVC and clang-tidy for unreachable code and misc-include-cleaner reasons, respectively. If you need any help, please let us know and we'll have a look.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -246,7 +246,46 @@ constexpr auto output_result(Component const& power_sensor, MainModelState<Compo
...
+        case branch_to:
+            [[fallthrough]];
+        case branch3_1:
+            [[fallthrough]];
+        case branch3_2:
+            [[fallthrough]];
+        case branch3_3:
+            [[fallthrough]];
+        case node:
+            return true;
```
TonyXiang8787: @mgovers @appleweiping Does this PR only partially fixes problem?

With this PR, you indeed solve the problem with sensors on disconnected appliances. But for energized branch or branch3, if one side is disconnected and there is a sensor on this side, even after this PR, the sensor value will still be wrongfully calculated instead of getting a null output.

To fix that part also, you cannot use fall through here for all branch switch cases. Instead, you need to check the actual status of that side of the branch/branch3.

appleweiping: Thanks for raising this, Tony. I intentionally kept partially disconnected branch/branch3 sensors topology-governed based on the scope previously confirmed in #1464 and the earlier review guidance in #1468: when the component still contributes to the mathematical grid, the sensor remains energized.

This proposed terminal-specific behavior would change that agreed contract. @mgovers, could you please confirm which behavior should apply? If terminal-specific behavior is intended, should it apply only to power sensors, or to current sensors as well? I’ll wait for alignment before changing the implementation and tests.

mgovers: for branches, the sensor on the disconnected side of a partially disconnected branch can still get a non-zero reading if the other side of the branch is connected to source, right @TonyXiang8787 ? (e.g. zero-sequence)

or should we also identically put it to `0` because it application of Kirchoff's law may still result in an approximation only?

TonyXiang8787: @mgovers 

Ideally, for energized branch/branch3, if a sensor is placed at the disconnected side, the sensor self should be not energized and have everything output as zero.

mgovers: alright, in that case @appleweiping , please follow this recommendation. The topological check still needs to be in place to handle grids without a source, but instead of only checking the status for appliances, it should also be checked for branches on the side to which the sensor is connected.

mgovers: please also add a test case for this

nitbharambe: >  should it apply only to power sensors, or to current sensors as well?

Same behaviour of disconnected output for both power and current sensor.

tests/cpp_unit_tests/main_core/test_main_core_output.cpp
```diff
@@ -12,18 +12,165 @@
...
+    output[0].u.resize(1);
+    output[0].bus_injection.resize(1);
+    output[0].branch.resize(3);
+    output[0].source.resize(1);
+    output[0].shunt.resize(1);
+    output[0].load_gen.resize(2);
+    return output;
+}
+
+template <symmetry_tag sym> void check_zero(RealValue<sym> const& value) {
```
nitbharambe, line 147: There is a check_close in test_math_solver_common and test_measured_values, check_nan_preserving_equality in test_voltage_sensor. (And maybe something else). 

Feel free to refactor all this to a new util file within test. Or you can scope it out of this PR if you wish.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
appleweiping: Implemented in 9e16db9. Power and current sensor outputs now retain the topology-group check and also gate on the status of the measured branch/branch3 terminal. The C++ regressions cover distinct partial-status patterns for branch (`0/1`) and branch3 (`1/0/1`), topology-disconnected output, exact zero residuals, and both calculation symmetries. The Release C API build, focused CTest (9/9), and full Python suite (1778 passed) all pass.

tests/cpp_unit_tests/main_core/test_main_core_output.cpp
appleweiping: Thanks. I have scoped the broader test-helper refactor out of this PR. This local helper checks the exact-zero contract of null sensor output, while consolidating the other approximate/NaN-aware helpers would widen the change beyond #1464.

TonyXiang8787 (approved):
Hi @appleweiping,

Thank you again for your contribution. The PR is in good shape and will be added to the merge queue if CI succeeds.

TonyXiang8787 (changes requested):
Hi @appleweiping,

The code is now breaking due to another change in #1546 causing the code not compatible anymore. Please have a look and adjust accordingly.

appleweiping:
Thanks, Tony. I adapted the current-sensor output test state to the `Edge` topology base introduced by #1546 in commit 91cc2ff, while retaining `Branch` because the sensor object sequence is still branch-relative.

I verified the change against a fresh snapshot of the merged head with GCC 14 / C++23: the main-core target builds and all 9 test cases (266 assertions) pass. `clang-format --dry-run --Werror` also passes. The full CI has been triggered again.

TonyXiang8787 (approved):

PR #1541, opened 2026-08-18
Temporarily skip clang-tidy

Jerry-Jinfeng-Guo:
Latest clang-tidy is failing. 

At most one of https://github.com/PowerGridModel/power-grid-model/pull/1541 and https://github.com/PowerGridModel/power-grid-model/pull/1542 can be merged.

PR #1542, opened 2026-08-18
Pin to earlier pgm dep

Jerry-Jinfeng-Guo:
Clang-tidy is failing. Pin to earlier pgm dep version to avoid disabling clang-tidy all together.

At most one of https://github.com/PowerGridModel/power-grid-model/pull/1541 and https://github.com/PowerGridModel/power-grid-model/pull/1542 can be merged.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -63,7 +63,7 @@ jobs:
       
       - name: Install pgm-build-deps
         run: |
-          uv tool install pgm-build-deps
+          uv tool install pgm-build-deps@2026.08.10.64
           pgm-build-setup-ga-ci
 
       - name: Set build target in case of workflow dispatch


.github/workflows/clang-tidy.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       
       - name: Install pgm-build-deps
         run: |
-          uv tool install pgm-build-deps
+          uv tool install pgm-build-deps@2026.08.10.64
```
mgovers, line 66: a sustainable way would be to make this a rolling 7-day window. can you please investigate how that works for `uv tool`?

also, we do not have an answer yet as to **why** it is failing. This PR just fixes CI but does not solve the underlying issue: a failure. We need to answer some questions:

1. is clang-tidy-18 part of the problem?
   1. does a newer clang-tidy version fix it?
   2. can/should we bump it now already (as we want to do in the near future once ubuntu-26.04 runners become GA anyways)?
3. is boost 1.92.0 broken?
   1. should we skip this version?
   2. should we report the problem?

mgovers (approved):
it's a hotfix. follow-up investigation can come later

PR #1545, opened 2026-08-21
Merge links into nodes: Create link output from link solver algorithm

figueroa1395:
Part of https://github.com/PowerGridModel/power-grid-model/issues/35

Follow up of https://github.com/PowerGridModel/power-grid-model/pull/1536

This PR should activate the Link output with two paths (old and new) + add the steady state output for `i`

PR #1550, opened 2026-08-25
Merge links into nodes: link registration refactoring (no functional change)

Jerry-Jinfeng-Guo:
Based on the presence of node injection sensors, this PR adds two Link registration path: 
 - Link component will be registered as is and sit next to Branch if there is no node injection sensors; 
 - Else Link component will be registered as Branch and sit withinin branch index.

However, the two paths aren't activated yet, so this PR becomes refactoring only in preparation for the activation.

The activation of the dual code path happens in https://github.com/PowerGridModel/power-grid-model/pull/1584.


Jerry-Jinfeng-Guo:
This PR is blocked by https://github.com/PowerGridModel/power-grid-model/pull/1545  due to the link output being necessary in `produce_output`

figueroa1395:
> This PR is blocked by #1545 due to the link output being necessary in `produce_output`

@Jerry-Jinfeng-Guo https://github.com/PowerGridModel/power-grid-model/pull/1545 is now in merge queue. That should get you unblocked.

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -154,9 +158,14 @@ inline TopologicalNodesAndCoupling create_topological_nodes(ComponentTopology co
                 return Idx2D{.group = disconnected, .pos = disconnected};
             }
 
+            // early out if either is disconnected since the link solver doesn't consume semi-disconnected links
+            if (from_conn == disconnected || to_conn == disconnected) {
+                return Idx2D{.group = topo_node, .pos = disconnected};
+            }
+
```
figueroa1395, line 165: Two bugs are in here. This is consumed by the link solver algorithm, but it doesn't understand disconnected sides of a link, and that breaks its premise. In addition, the "enumeration" that was coming from the user nodes (i.e. links per supernode being numbered [{{0, 1},{2, 3}}, {4, 5}] instead of [{{0, 1},{2, 3}}, {0, 1}]), was not suitable either.

This is a bit of a hacky fix, as now `user_links` doesn't really contain the user links indexed by the user nodes, but rather it's own indexing per super node. That said, I added it here to attempt the fix. Since this works, it's probably a better idea to put it where the link solver algorithm consumes user links or elsewhere; this only works because currently `user_links` isn't consumed elsewhere besides for the link solver algorithm (we can probably delete it in the final clean up for v2).

tests/cpp_unit_tests/test_supernodes.cpp
```diff
@@ -165,7 +165,7 @@ TEST_CASE("Test Supernodes") {
             CHECK(std::ranges::equal(
                 topo_nodes.topo_nodes |
                     std::views::transform([](TopologicalNode const& node) -> auto& { return node.user_links; }),
-                std::vector<std::vector<BranchIdx>>{{{0, disconnected}}, {{disconnected, 1}}, {{2, 4}}, {}, {}}));
+                std::vector<std::vector<BranchIdx>>{{{0, disconnected}}, {{disconnected, 0}}, {{0, 1}}, {}, {}}));
```
figueroa1395: These result from https://github.com/PowerGridModel/power-grid-model/pull/1550/changes#r3921543997, hence you can see here it's hacky.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/calculation_input_preparation.hpp
```diff
@@ -284,8 +284,10 @@ prepare_short_circuit_input(main_model_state_c auto const& state, ComponentToMat
     for (Idx const fault_idx : IdxRange{state.components.template size<Fault>()}) {
         auto const& fault = state.components.template get_item_by_seq<Fault>(fault_idx);
         if (fault.status()) {
-            auto const node_idx = state.components.template get_seq<Node>(fault.get_fault_object());
-            auto const topo_bus_idx = state.topo_comp_coup->node[node_idx];
+            auto const user_node_idx = state.components.template get_seq<Node>(fault.get_fault_object());
+            auto const topo_node_idx =
+                state.reduced_topology->topo_node_coup.coupling.user_nodes_to_topo_nodes[user_node_idx].group;
+            auto const topo_bus_idx = state.topo_comp_coup->node[topo_node_idx];
```
figueroa1395, line 290: This is the SC bug. Wrong mapping was used.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -146,9 +146,9 @@ constexpr auto output_result(Component const& link, MainModelState<ComponentCont
 }
 template <std::same_as<Link> Component, class ComponentContainer, short_circuit_solver_output_type SolverOutputType>
     requires model_component_state_c<MainModelState, ComponentContainer, Component>
-inline auto output_result(Component const& link, MainModelState<ComponentContainer> const& /* state */,
+inline auto output_result(Component const& link, MainModelState<ComponentContainer> const& state,
                           MathOutput<std::vector<SolverOutputType>> const& math_output, Idx2D const& topo_id) {
-    if (topo_id.group == disconnected) {
+    if (topo_id.group == disconnected || state.topo_comp_coup->node[topo_id.group].group == disconnected) {
```
figueroa1395, line 154: This was another bug. The output didn't understand that a link was not energized if it's statuses where on but it wasn't connected to a source.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
```diff
@@ -92,10 +92,24 @@ constexpr auto comp_base_sequence_cbegin(MainModelState<ComponentContainer> cons
...
+constexpr auto comp_base_sequence_cbegin(MainModelState<ComponentContainer> const& state) {
+    auto const& link_topo_ids = state.reduced_topology->topo_node_coup.coupling.user_links_to_topo_nodes;
+
+    if (std::ranges::ssize(link_topo_ids) == get_component_size<Link>(state.components)) {
+        // new path: links are not branches
+        return link_topo_ids.cbegin();
+    } else {
+        // legacy path: links are branches
+        return state.topo_comp_coup->branch.cbegin() + get_component_sequence_offset<Edge, Link>(state.components);
+    }
```
figueroa1395, line 112: The rest of the issues so far, were related to the "two-path dispatch", similar to this. Where in the old path Branches included links, so the offset was different to the new path (or the container to use was different), where branches and links are separated.

I think it's good that we spam TODOs in every such two-code-path place introduced here, so we can easily pinpoint for the cleanup after v2.

tests/data/power_flow/dummy-test-line-into-itself/asym_output_batch.json
Jerry-Jinfeng-Guo, line 1: Test changes are from https://github.com/PowerGridModel/power-grid-model/pull/1569

tests/data/power_flow/dummy-test-line-into-itself/sym_output_batch.json
Jerry-Jinfeng-Guo, line 1: Test changes are from https://github.com/PowerGridModel/power-grid-model/pull/1569

tests/data/power_flow/handling-of-links/merge-nodes-by-links-1-line-with-line-to-self/sym_output.json
Jerry-Jinfeng-Guo, line 1: Test changes are from https://github.com/PowerGridModel/power-grid-model/pull/1569

tests/data/short_circuit/dummy-test-line-into-itself/sc_output_batch.json
Jerry-Jinfeng-Guo, line 1: Test changes are from https://github.com/PowerGridModel/power-grid-model/pull/1569

Jerry-Jinfeng-Guo:
@TonyXiang8787 @nitbharambe please have a look at [123e0e5](https://github.com/PowerGridModel/power-grid-model/pull/1550/commits/123e0e5933c05d197ca47cc4cfe3fdb51c9a9dc1), whether the updated calculation is correct with respect to our new way of calculating with merged links.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -276,10 +276,15 @@ template <std::derived_from<GenericVoltageSensor> Component, class ComponentCont
...
+                             MathOutput<std::vector<SolverOutputType>> const& math_output, Idx const /* node_seq */) {
     using sym = decode_symmetry_v<SolverOutputType>;
 
-    Idx2D const node_math_id = state.topo_comp_coup->node[node_seq];
+    Idx const sensor_seq = get_component_sequence_offset<GenericVoltageSensor, Component>(state.components) +
+                           get_component_sequence_idx<Component>(state.components, voltage_sensor.id());
+    Idx const node_seq = state.comp_topo->voltage_sensor_node_idx[sensor_seq];
+    auto const& node_math_id = node_seq < std::ranges::ssize(state.topo_comp_coup->node)
+                                   ? state.topo_comp_coup->node[node_seq]
+                                   : state.topo_comp_coup->voltage_sensor[sensor_seq];
```
Jerry-Jinfeng-Guo, line 287: A fallback mechanism is needed to handle reduced topology wrt sensor indexing. This is the issue that caused the Pytest segfault.

Jerry-Jinfeng-Guo:
Final steps for this PR: mark for V2 clean ups in the code comments.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -270,10 +276,15 @@ template <std::derived_from<GenericVoltageSensor> Component, class ComponentCont
           steady_state_solver_output_type SolverOutputType>
     requires model_component_state_c<MainModelState, ComponentContainer, Component>
 constexpr auto output_result(Component const& voltage_sensor, MainModelState<ComponentContainer> const& state,
-                             MathOutput<std::vector<SolverOutputType>> const& math_output, Idx const node_seq) {
+                             MathOutput<std::vector<SolverOutputType>> const& math_output, Idx const /* node_seq */) {
     using sym = decode_symmetry_v<SolverOutputType>;
 
-    Idx2D const node_math_id = state.topo_comp_coup->node[node_seq];
+    Idx const sensor_seq = get_component_sequence_offset<GenericVoltageSensor, Component>(state.components) +
+                           get_component_sequence_idx<Component>(state.components, voltage_sensor.id());
```
mgovers: can we use a query short-hand here similar to how we do it for the other component types?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
mgovers: does it make sense to simply directly use the reduced topology instead of the original `comp_topo` to generate the `node_seq` where this function is called?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
mgovers: yes, an explicit detectable pattern like `TODO(github-username): cleanup v2` is much more easily recognizable than `legacy path`

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -229,31 +249,75 @@ constexpr void register_connections_components(ComponentContainer const& compone
...
     using TopologyTypesTuple = ModelType::TopologyTypesTuple;
     main_core::utils::run_functor_with_tuple_return_void<TopologyTypesTuple>(
         [&components, &comp_topo]<typename CompType>() {
-            detail::register_topology_components<CompType>(components, comp_topo);
+            if constexpr (!std::same_as<CompType, Edge>) {
+                detail::register_topology_components<CompType>(components, comp_topo);
+            }
         });
+    if (has_node_injection_sensors) {
+        // old path: all edges (branches + links) in branch_node_idx
```
mgovers: same as before, here and elsewhere: a pattern like `TODO(github-username): cleanup v2: ...` is much clearer and even automatically detectable, while `old/legacy/new path` requires exhaustively listing those options to find all of them.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
```diff
@@ -506,7 +506,8 @@ struct ReducedComponentTopology {
 
 struct TopologicalNode {
     IdxVector user_nodes;
-    std::vector<BranchIdx> user_links;
+    std::vector<BranchIdx>
+        user_links; // TODO(figueroa1395): temporarily doesn't have user nodes, but internal nodes while exploring
```
mgovers: given the `while exploring`, adding a comment here so that we don't accidentally merge before that TODO is resolved

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_preparation.hpp
```diff
@@ -134,7 +134,12 @@ inline void rebuild_topology(typename ModelType::MainModelState& state, SolverPr
 
     // clear old solvers
     reset_solvers(state, solver_context, solvers_cache_status);
-    ComponentConnections const comp_conn = main_core::construct_components_connections<ModelType>(state.components);
+    // Determine the path based on whether link_node_idx is populated (new path) or empty (old path)
```
mgovers, line 138: please also add an explicit `TODO(github-username): cleanup v2: ...` here

tests/data/short_circuit/dummy-test-line-into-itself/sc_output_batch.json
figueroa1395: Since this is a bug anyway (previously already in main), we accept the results and make the actual fix in #1566, right? Just making sure I understand the logic of it

Jerry-Jinfeng-Guo: @nitbharambe

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
figueroa1395: A side note about this: I re-used `user_links` for the fix because it wasn't being used anywhere else (yet?). I foresee a couple of options:
- Rename this variable to `topo_links` or similar, since this is only used as input for the link algorithm. If this is done the `struct` may need renaming.
- Same as above but also keep `user_links` if there may be a use for it.
- Keep as it was, for now at least, and instead re-apply this logic elsewhere. Potentially at the link algorithm by adding a helper function that just "re-numbers" the `user_links` correctly, but this may be extra/unnecessary work.
- Something else?

So the fix is there to stay, this is about where since I just applied it where it was "easier" to see whether that was it or not.

tests/cpp_unit_tests/test_supernodes.cpp
```diff
@@ -165,7 +165,7 @@ TEST_CASE("Test Supernodes") {
             CHECK(std::ranges::equal(
                 topo_nodes.topo_nodes |
                     std::views::transform([](TopologicalNode const& node) -> auto& { return node.user_links; }),
-                std::vector<std::vector<BranchIdx>>{{{0, disconnected}}, {{disconnected, 1}}, {{2, 4}}, {}, {}}));
+                std::vector<std::vector<BranchIdx>>{{}, {}, {{0, 1}}, {}, {}}));
```
figueroa1395, line 168: Commenting again here since https://github.com/PowerGridModel/power-grid-model/pull/1550#discussion_r3921547034 is outdated (resolved it to make it easier), but it may still be relevant because of  https://github.com/PowerGridModel/power-grid-model/pull/1550#discussion_r3948313370 and https://github.com/PowerGridModel/power-grid-model/pull/1550#discussion_r3921543997.

mgovers:
> @TonyXiang8787 @nitbharambe please have a look at [123e0e5](https://github.com/PowerGridModel/power-grid-model/pull/1550/commits/123e0e5933c05d197ca47cc4cfe3fdb51c9a9dc1), whether the updated calculation is correct with respect to our new way of calculating with merged links.

confirmed the issue is fixed in the new code path but still remains in the old code path. Added a separate test case that includes a node injection sensor to show this.

tests/data/state_estimation/link/asym_output.json
figueroa1395, line 1: As a first set of eyes (out of the required several) this looks fine:
- The nodes with `0.0` `p` and `q` outputs are expected, as there is no injection. This aligns perfectly with the new (improved) way of calculating the node injection not using the `y-bus`.
- The "flows" on the `transformer` and `link` are "more-zero" now as well. Both are consistent and this is expected.

As stated, someone else should double check these values too before we make the final call.

figueroa1395 (commented):
Some of my old comments were just informative. Feel free to resolve those when you consider appropriate.

tests/data/state_estimation/link/asym_output.json
mgovers: in the new, improved way of calculating link output, indeed, these are the correct values.

However, this validation case, as-is, does not catch all edge cases of #1567 . I created a separate validation case that captures this (CI failing).

To actually solve the root cause of the problem, fixing of #1567 shall be done off `main` on a separate branch in #1576 combined with a repro case in a to-be-created PR (still in progress, experiments running on `pgm/experiment/issue-1567`, e.g. https://github.com/PowerGridModel/power-grid-model/actions/runs/34205950554 )

figueroa1395:
[202af72](https://github.com/PowerGridModel/power-grid-model/pull/1550/commits/202af7264d7c4feb4d854866353d3e6e7a147796) is the merge commit into the branch of the "old path" bugfix [PR](https://github.com/PowerGridModel/power-grid-model/pull/1576). I've deliberately kept the values of the link SE validation case from this branch and not from the upstream one because here, due to the absence of a node injection sensor, the new path is taken.

The values of the validation test with a node injection sensor, which trigger the old code path, are still to be addressed and at this stage it's expected to fail with `AssertionError`. This is WIP (see [this](https://github.com/PowerGridModel/power-grid-model/pull/1550#discussion_r3956098670) comment).

`do-not-merge` label added until we merge the upstream branch first.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
```diff
@@ -286,21 +309,21 @@ ComponentTopology construct_topology(typename ModelType::ComponentContainer cons
...
             }
         });
+    if (has_node_injection_sensors) {
+        // old path: all edges (branches + links) in branch_node_idx
+        detail::register_topology_components<Edge>(components, comp_topo, has_node_injection_sensors);
+    } else {
+        // new path: branches only in branch_node_idx, links in link_node_idx
+        detail::register_topology_components<Branch>(components, comp_topo);
+        detail::register_topology_components<Link>(components, comp_topo);
+    }
```
mgovers, line 335: why was this reverted? this results in segfaults, because e.g. a current sensor needs to be registered on a branch, which already needs to be registered before the current sensor is registered.

mgovers: reverted

figueroa1395: It was a merge conflict resolution mistake. Fixed in [3d758c8](https://github.com/PowerGridModel/power-grid-model/pull/1550/commits/3d758c8b11f93f982e36092f7bdf8ac307607f98)

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers:
```suggestion
    std::vector<BranchIdx> internal_edges;
```

tests/data/power_flow/handling-of-links/merge-nodes-by-links-1-line-line-to-super-node-itself-II/params.json
```diff
@@ -1,9 +1,5 @@
 {
   "calculation_method": ["newton_raphson"],
   "rtol": 3e-7,
-  "atol": 1e-7,
-  "xfail": {
-    "raises": "AssertionError",
-    "reason": "The handling of links is not fully implemented."
-  }
+  "atol": 1e-7
```
mgovers: 🚀

tests/data/power_flow/handling-of-links/merge-nodes-by-links-1-line-line-to-super-node-itself/params.json
```diff
@@ -1,9 +1,5 @@
 {
   "calculation_method": ["newton_raphson"],
   "rtol": 3e-7,
-  "atol": 1e-7,
-  "xfail": {
-    "raises": "AssertionError",
-    "reason": "The handling of links is not fully implemented."
-  }
+  "atol": 1e-7
```
mgovers: 🚀

tests/data/power_flow/handling-of-links/merge-nodes-by-links-1-line-with-line-to-self/sym_output.json
```diff
@@ -98,13 +98,13 @@
...
+                "q_from": 5541259.927274941,
+                "i_from": 1199.94392461469216,
+                "s_from": 21800322.716069363,
                 "p_to": -21084319.029628307,
-                "q_to": -5195613.111142274,
-                "i_to": 1195.2496789651816,
-                "s_to": 21715038.672399823
+                "q_to": -5541259.927274941,
+                "i_to": 1199.94392461469216,
+                "s_to": 21800322.716069363
```
mgovers: These values are significantly large (differing ~5%). We need to double check these

tests/data/power_flow/handling-of-links/merge-nodes-by-links-1-line/params.json
```diff
@@ -1,9 +1,5 @@
 {
   "calculation_method": ["newton_raphson"],
   "rtol": 3e-7,
-  "atol": 1e-7,
-  "xfail": {
-    "raises": "AssertionError",
-    "reason": "The handling of links is not fully implemented."
-  }
+  "atol": 1e-7
```
mgovers: 🚀

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
figueroa1395: Repurposed in [67522ff](https://github.com/PowerGridModel/power-grid-model/pull/1550/commits/67522ff2b9ccc0774a9485959e808404d306c998)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topology.hpp
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model/pull/1550/changes#diff-2520d6443f9ad4b92ab98381c38c011e8831dfa54b8f59c5d0b23ef085c566f6R287

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/state_queries.hpp
figueroa1395: There is a TODO in place. Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_preparation.hpp
Jerry-Jinfeng-Guo: https://github.com/PowerGridModel/power-grid-model/pull/1550/changes#diff-5542c64914fcdf22001b2278b6310a7cc625a3d51035a7cb22c62ec06c4c8933R137

tests/data/power_flow/handling-of-links/merge-nodes-by-links-1-line-with-line-to-self/sym_output.json
mgovers: https://github.com/PowerGridModel/power-grid-model/pull/1569/changes#r3933485843

tests/cpp_unit_tests/test_supernodes.cpp
mgovers: resolved

tests/data/short_circuit/dummy-test-line-into-itself/sc_output_batch.json
mgovers: still TBD: do we merge this PR first or do we merge #1569 first?

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
```diff
@@ -129,14 +129,18 @@ inline TopologicalNodesAndCoupling create_topological_nodes(ComponentTopology co
 
     std::vector<Idx2D> user_link_topo_node_coup =
         enumerate(std::views::zip(comp_topo.link_node_idx, comp_conn.link_connected)) |
-        std::views::transform([&node_mapping, &topo_nodes](auto const& idx_link_and_connectivity) {
+        std::views::transform([&node_mapping, &topo_nodes,
```
mgovers, line 132: TODO: clean up in separate PR (either immediately do so or add TODO comment)

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
mgovers: also scope out to separate PR

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
mgovers: maybe also scope out to separate PR

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: resolved

tests/data/state_estimation/link/asym_output.json
mgovers: #1580 is in merge queue. resolving

power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
mgovers: resolved

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
figueroa1395: This was addressed/changed in [f770018](https://github.com/PowerGridModel/power-grid-model/pull/1550/commits/f7700189d6fdc9728f5882134911cce0c45c5f5a).  @mgovers was still able to reproduce the segfault on python only, the commit fixes and explains it.

In short, segfaults were not being triggered in C++ because in C++ we don't create output buffers (columnar data usage) for attributes for which we don't have a specific testing output. That was the case for voltage sensor in the failing test (we were able to find the culprit by hard throwing manually near this location after it was found when debug symbols were enabled for python). However, in Python we create "null" output for all output buffers in the validation tests even if we don't specify it, hence triggering the issue that seemed very obscure, but it was just an out of bounds memory read for the voltage sensor since its query was incorrect.

Resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
figueroa1395: Resolving as this is no longer relevant. See https://github.com/PowerGridModel/power-grid-model/pull/1550/changes/3cbd4f68853cd32961fecb80f2feb4b4bd61a224#r3965977001.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -134,25 +142,31 @@ constexpr auto output_result(Component const& node, MainModelState<ComponentCont
...
 template <std::same_as<Link> Component, class ComponentContainer, short_circuit_solver_output_type SolverOutputType>
     requires model_component_state_c<MainModelState, ComponentContainer, Component>
-inline auto output_result(Component const& link, MainModelState<ComponentContainer> const& /* state */,
+inline auto output_result(Component const& link, MainModelState<ComponentContainer> const& state,
                           MathOutput<std::vector<SolverOutputType>> const& math_output, Idx2D const& topo_id) {
-    if (topo_id.group == disconnected) {
+    if (topo_id.group == disconnected || state.topo_comp_coup->node[topo_id.group].group == disconnected) {
         return link.get_null_sc_output();
     }
+    if (topo_id.pos == disconnected) {
```
nitbharambe, line 167: What can we deduce from the special addressal of -1 / disconnected in this PR? 
Can there be a case of group=-1 yet state.topo_comp_coup->node[topo_id.group].group = -1?
or 
both of above as -1 and yet pos =-1? (Edit: possible and documented in calculation_parameters.hpp)
(No need to respond, Give me some time to figure this out :D)

tests/data/short_circuit/dummy-test-line-into-itself/sc_output_batch.json
nitbharambe: Either way works. If this PR goes first, we merge considering the bug in 1566. Then modify this test again in #1570 which considers merged links and solved bug.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
nitbharambe: Cant seem to identify the answer to first question.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
nitbharambe: @figueroa1395 And these dont get registered with group=-1?

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
mgovers: > Can there be a case of group=-1 yet state.topo_comp_coup->node[topo_id.group].group = -1?

if `topo_id.group == -1`, then `state.topo_comp_coup->node` does not contain it, so the former has priority over the latter.

However, what i think you meant is

> Can there be a case of group!=-1 yet state.topo_comp_coup->node[topo_id.group].group = -1?

In that case, the answer is yes: the former tells us to which topological node the link is connected. The latter tells us that there is no source connected to the connected component that the topological node is in.

mgovers: it would probably be more accurate to `assert` the first statement rather than checking the condition, but i guess there's no harm in keeping it as-is. branch prediction may help us out here.

tests/data/short_circuit/dummy-test-line-into-itself/sc_output_batch.json
mgovers: let's focus on this one first then. #1570 can wait.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
mgovers: Indeed, the `group=-1` is about the topo node that the link belongs to, not the math solver group. Unfortunate re-use of types, I am aware. See also https://github.com/PowerGridModel/power-grid-model/pull/1550/changes#r3966051309

power_grid_model_c/power_grid_model/include/power_grid_model/supernodes.hpp
mgovers: i dug into this and it's fairly hard to refactor without making the logic unnecessarily complex. resolving.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
figueroa1395: > @figueroa1395 And these dont get registered with group=-1?

@nitbharambe To elaborate a bit more, there are now two different used for `Idx2d`. The first one (original one) is to track to which math solver a component belongs to or if it's disconnected (`group`) and which index within it's component list is its own within that solver (`pos`); on the other hand `Idx2d` is also used to classify supernodes.

In the sense of supernodes, the `group` tells us to which supernode/topological node does this node belong to, and `pos` its internal enumeration within that supernode structure. Then, if we want to know if a link (which is part of a supernode node) is energized or connected to a source, we need to check that the nodes it is attached to correspond to a topological node (from the macro topology perspective) that is connected to a source. That is what that line is doing.

nitbharambe: Ahh right. I think this also answers the below question

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
figueroa1395: @nitbharambe The answer to this one is actually different to the previous one.

What happens is that I added [this](https://github.com/PowerGridModel/power-grid-model/pull/1550/changes#diff-3e824f0bfb3c501da7853fa421b91ced2b42ecf6822552bf9c290b06e8b8f47eR141-R143) and [this](https://github.com/PowerGridModel/power-grid-model/pull/1550/changes#diff-3e824f0bfb3c501da7853fa421b91ced2b42ecf6822552bf9c290b06e8b8f47eR161-R169) pieces of logic because the `link_solver` can't handle links that are disconnected on either or both sides. And instead, we should handle that case directly at output by producing `get_energized_zero_output`.

This is what happening here, if either side of the link is off (not necessarily non connected to a source - that's what happens above), the we get null output. And that's what `pos` is doing here.

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/output.hpp
```diff
@@ -60,11 +60,19 @@ constexpr bool measured_terminal_active(MeasuredTerminalType const terminal_type
         using enum MeasuredTerminalType;
 
     case branch_from:
-        return measured_component_active<Edge>(state, obj_seq,
-                                               BranchSide::from); // TODO(mgovers): cleanup v2: change back to Branch
+        // TODO(mgovers): cleanup v2: only keep Branch code path
+        if (state.comp_topo != nullptr && state.comp_topo->link_node_idx.empty()) {
```
figueroa1395:
```suggestion
        if (state.comp_topo->link_node_idx.empty()) {
```

Here and below this isn't needed. It's an extra runtime check there only for a test. And if we do it here we should do it everywhere. 

This is an FYI, I have a change locally, but I'll collect all (if I find any more) and push at once.

figueroa1395: Resolved in [2df4d3c](https://github.com/PowerGridModel/power-grid-model/pull/1550/commits/2df4d3c7d7be8ddfd0cbc5abf688819a75fdab9a)

mgovers: :+1:

mgovers:
Benchmark done in #1583 against `v1.13.163` and seems OK.

mgovers (approved):
No changes to validation cases needed anymore :tada: LGTM

nitbharambe (approved):

figueroa1395 (approved):

mgovers:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=1550) **Quality Gate failed**
> Failed conditions  [76.9% Coverage on New Code](https://sonarcloud.io/component_measures?id=PowerGridModel_power-grid-model&pullRequest=1550&metric=new_coverage&view=list) (required ≥ 80%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=PowerGridModel_power-grid-model&pullRequest=1550)

Coverage is expected due to #1584

Issue #1554, opened 2026-08-26
[BUG] Result-production ~10x slower on main since #1536


marhofmann:
### Describe the bug

While I wanted to finish the second part of the job from #1551, I noticed my branch no longer applied cleanly on main, so I rebased onto the merge of #1536. Before writing up the numbers again I re-measured, and the baseline had moved a lot more than I expected.

Same benchmark, same grid, same machine, ten interleaved rounds in one session:

| | before #1536 (`c763f484`) | current main (`e3a8308`) | |
|---|---:|---:|---:|
| `Produce output` | 0.2268 s | 2.4487 s | **10.8× slower** |
| Math Calculation | 1.6983 s | 2.0757 s | 1.22× slower |
| whole batch run | 2.2681 s | 4.9217 s | **2.17× slower** |

`tests/benchmark_cpp`, radial grid, 2605 nodes, 1000-scenario symmetric Newton-Raphson, MSVC `/O2`. The `Produce output` ranges don't overlap (0.220–0.232 against 1.869–2.554), and `tests/benchmark_cpp` is byte-identical between the two commits, so it isn't a changed workload.

Two things that surprised me:

**The benchmark grid has no links at all.** So this is the cost of the link machinery running on a grid that has nothing for it to do. `compute_link_solver` is called for every topological node rather than only merged ones, and each call allocates a few vectors (`get_total_injection_per_node()`, `links | ranges::to<std::vector>()`, the link output) on top of `node_injection` and `node_flow_from_branch` per node, and a full vector copy into `bus_injection`.

**Math Calculation got 22% slower too**, even though #1536 only added comment lines to `math_solver`. Since the runs were interleaved I don't think that's drift. My guess is allocator pressure from the churn above spilling over, but that's a guess, not something I've confirmed.

I haven't tried to fix it. You know this code far better than I do, and I didn't want to guess at intent while the link work is still in flight. Happy to help measure whatever you try.

### Input Data Validity

- [x] I checked the validity of input data using the helper function ([examples here](https://power-grid-model.readthedocs.io/en/stable/examples/Validation%20Examples.html)).

### To Reproduce

No input data needed! This reproduces with the C++ benchmark that's already in the repo.

Build `power_grid_model_benchmark_cpp` at each of the two commits and run it. Nothing else changes; `tests/benchmark_cpp` is byte-identical between them.

```bash
# before
git checkout c763f484
[...]

PR #1555, opened 2026-08-26
Give per-user-node vectors inline storage

marhofmann:
Each topological node holds one value per user node, and only link-merged supernodes have more than one: in a grid without links every topological node has exactly one. Three of these vectors are rebuilt on every scenario (node_injection, node_flow_from_branch and bus_injection), so each node asks the allocator for room to store a single complex number three times per scenario.

Give them inline room for one element, so the ordinary node does not allocate. Link-merged nodes still work, they just allocate as before. The indexing interface is unchanged. As a side effect the copy from node_injection into bus_injection stops being a heap copy, because both sides are now the same inline type.

Expose the container generically rather than at the point of use: a new common/small_vector.hpp holds a single using declaration for SmallVector, alongside the other type aliases in common/. Other call sites can then reach for it without pulling in boost directly, and the underlying container can be swapped in one place. boost::container::small_vector is header-only and Boost::headers is already an INTERFACE dependency of the power_grid_model target, so this adds nothing to the build. The inline capacity stays a per-use-site decision; here it is 1 because that is exactly what an ordinary topological node holds.

Measured on tests/benchmark_cpp (radial grid, 2605 nodes, 1000-scenario symmetric Newton-Raphson batch), MSVC /O2, against the parent commit, 10 interleaved rounds per arm on an otherwise idle machine, medians:

  produce output    2.4487 s -> 2.2470 s   1.09x
  math calculation  2.0757 s -> 2.1361 s   control, unchanged within noise
  whole batch run   4.9217 s -> 4.7514 s   1.04x

The change won 9 of the 10 paired rounds, so the effect is real, but the 0.20 s median gain sits inside main's own 0.69 s run-to-run spread, so treat it as modest rather than decisive.

The gain is smaller than it would have been before #1536, because most of the per-node allocation cost now sits in compute_link_solver, which this does not touch. See the linked issue.

All ten C++ test suites pass, 176999 assertions, built with warnings as errors, including the handling-of-links cases that exercise supernodes holding more than one user node.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/calculation_parameters.hpp
@@ -8,6 +8,7 @@
 #include "common/enum.hpp"
 #include "common/grouped_index_vector.hpp"
 #include "common/maybe_owning_view.hpp"
+#include "common/small_vector.hpp"
 #include "common/statistics.hpp"
 #include "common/three_phase_tensor.hpp"
 
@@ -410,10 +411,14 @@ struct OptimizerOutput {
 
 template <typename T> struct SupernodeOutput;
 
+// One entry per user node of a topological node. Only link-merged supernodes hold more than one, so
+// inline room for a single element keeps the ordinary node from allocating at all.
+template <symmetry_tag sym> using UserNodeValueVector = SmallVector<ComplexValue<sym>, 1>;
+
 template <steady_state_solver_output_type SolverOutputType> struct SupernodeOutput<SolverOutputType> {
     using sym = decode_symmetry_v<SolverOutputType>;
 
-    ComplexValueVector<sym> bus_injection;     // user bus output
+    UserNodeValueVector<sym> bus_injection;    // user bus output
     std::vector<BranchSolverOutput<sym>> link; // user link
 };
 template <short_circuit_solver_output_type SolverOutputType> struct SupernodeOutput<SolverOutputType> {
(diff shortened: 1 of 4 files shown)


marhofmann:
Linked issue #1554

mgovers (dismissed):
Hi @marhofmann,

Thanks you once again for your contribution. Everything looks good :+1:

Apologies for the change in #1536 that has reduced the benefits of this branch. My educated guess is that that's just repetition of the same/a similar performance regression, in which case applying small vectors there would solve it. I will dive into #1554 now to investigate, but since this PR looks good, let's merge it. It might come in useful for the investigation. Would it be alright if you do another benchmark analysis after our investigation?

marhofmann:
There was a format issue. I fixed it

mgovers (approved):

mgovers:
@TonyXiang8787 the bug report #1554 is separate from this and was a new finding during implementation of this ticket. I am working on a fix for #1554 based off this branch.

marhofmann:
@mgovers 
Now that #1556 exists I could finally measure this properly and the numbers look quite different from what I put in the description. My 1.09× was measured against unfixed main, where the link solver was drowning everything else out.

`Produce output`, 2605-node grid, 1000-scenario batch, 8 interleaved rounds:

| build | `Produce output` | vs pre-#1536 |
| :--- | ---: | ---: |
| pre-#1536 | 0.222 s | — |
| `main` | 1.971 s | 8.9× |
| `main` + #1556 | 0.421 s | 1.90× |
| **`main` + #1556 + this PR** | **0.249 s** | **1.12×** |

So on top of #1556 this is worth 1.69×, and together they get ~98.5% of the regression back. Whole run 3.93 s → 2.19 s.
The ~0.03 s left is probably the branch-flow accumulation #1536 added, which is needed work. So I doubt there's much more to win on the input side or in the link solver.

mgovers:
Perfect news and once again thank you for the extensive analysis!

mgovers:
I wonder how many other locations can benefit significally from the small vector. Topology reduction is a very likely candidate.

PR #1556, opened 2026-08-27
Merge links into nodes: Fix performance regression topological node output

### LINKED ISSUE #1554: [BUG] Result-production ~10x slower on main since #1536

mgovers:
Fixes #1554 

Relates to #1555

There are potentially some other locations remaining where SmallVector can make a difference (in particular, topological node construction on the input side and the vectors internally in the link solver implementation), but for that, #1555 needs to be merged first. This fix is standalone. Instead, this PR fixes on the big fish.

Credits to @marhofmann for finding the issue.

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topological_node_output.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topological_node_output.hpp
@@ -162,6 +162,10 @@ template <symmetry_tag sym, typename LinkSolver>
     requires std::invocable<LinkSolver, std::vector<BranchIdx>, ComplexVector>
 ComplexValueVector<sym> compute_link_solver(LinkSolver link_solver,
                                             SuperNodeSolverInput<sym> const& super_node_solver_input) {
+    if (std::ranges::empty(super_node_solver_input.links)) {
+        return {};
+    }
+
     if constexpr (is_symmetric_v<sym>) {
         return link_solver(super_node_solver_input.links | std::ranges::to<std::vector>(),
                            super_node_solver_input.get_total_injection_per_node());
@@ -199,18 +203,15 @@ template <symmetry_tag sym, typename BranchSolverOutputType>
               std::same_as<BranchSolverOutputType,
                            BranchShortCircuitSolverOutput<decode_symmetry_v<BranchSolverOutputType>>>)
 std::vector<BranchSolverOutputType> get_link_output(ComplexValueVector<sym> const& link_solver_result) {
-    std::vector<BranchSolverOutputType> link_output;
-    link_output.reserve(link_solver_result.size());
-
-    for (auto const& result : link_solver_result) {
-        if constexpr (std::same_as<BranchSolverOutputType,
-                                   BranchSolverOutput<decode_symmetry_v<BranchSolverOutputType>>>) {
-            link_output.emplace_back(BranchSolverOutputType{.s_f = result, .s_t = -result});
-        } else {
-            link_output.emplace_back(BranchSolverOutputType{.i_f = result, .i_t = -result});
-        }
-    }
-    return link_output;
+    return link_solver_result | std::views::transform([](auto const& result) -> BranchSolverOutputType {
+               if constexpr (std::same_as<BranchSolverOutputType,
+                                          BranchSolverOutput<decode_symmetry_v<BranchSolverOutputType>>>) {
+                   return BranchSolverOutputType{.s_f = result, .s_t = -result};
+               } else {
+                   return BranchSolverOutputType{.i_f = result, .i_t = -result};
+               }
+           }) |
+           std::ranges::to<std::vector>();
 }
 
 template <typename LinkSolver, main_model_state_c State, solver_output_type SolverOutput>
--- a/tests/cpp_unit_tests/main_core/test_topological_node_output.cpp
+++ b/tests/cpp_unit_tests/main_core/test_topological_node_output.cpp
@@ -524,12 +524,10 @@ TEST_CASE("Test topological node output") {
 
             auto const result = detail::solve_topological_nodes(std::ref(mock), state, math_output);
 
-            REQUIRE(mock.call_count == 2);
+            REQUIRE(mock.call_count == 1);
             CHECK(mock.recorded_edges[0] == links);
-            CHECK(mock.recorded_edges[1].empty());
             CHECK(mock.recorded_loads[0] ==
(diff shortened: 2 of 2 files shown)


mgovers:
@marhofmann would you be able to run the benchmark to see if this fixes the issue for you as well?

nitbharambe (approved):

power_grid_model_c/power_grid_model/include/power_grid_model/main_core/topological_node_output.hpp
```diff
@@ -162,6 +162,10 @@ template <symmetry_tag sym, typename LinkSolver>
     requires std::invocable<LinkSolver, std::vector<BranchIdx>, ComplexVector>
 ComplexValueVector<sym> compute_link_solver(LinkSolver link_solver,
                                             SuperNodeSolverInput<sym> const& super_node_solver_input) {
+    if (std::ranges::empty(super_node_solver_input.links)) {
```
mgovers, line 165: out of scope of this PR but maybe this function should actually be passed as `LinkSolver` to the parent of this. handing symmetric and asymmetric differently should be an implementation detail but now it's part of the interface.

marhofmann:
> @marhofmann would you be able to run the benchmark to see if this fixes the issue for you as well?

 @mgovers Thanks and I am glad it was useful! I'm benchmarking your version now.

I had the same fix in mind, but it drops call_count from 2 to 1 and means removing the recorded_edges[1] / recorded_loads[1] assertions. Those looked deliberate and I couldn't tell whether they were guarding something I was overlooking. Good to see you've made the same call.

mgovers:
> > @marhofmann would you be able to run the benchmark to see if this fixes the issue for you as well?
> 
> @mgovers Thanks and I am glad it was useful! I'm benchmarking your version now.
> 
> I had the same fix in mind, but it drops call_count from 2 to 1 and means removing the recorded_edges[1] / recorded_loads[1] assertions. Those looked deliberate and I couldn't tell whether they were guarding something I was overlooking. Good to see you've made the same call.

Hi @marhofmann, thanks! I already sent it to merge queue to get it out quickly, but feel free to reply once you have the numbers!

Indeed the calls were deliberate and that also amounts to my comment in https://github.com/PowerGridModel/power-grid-model/pull/1556#discussion_r3869776954

marhofmann:
> @marhofmann would you be able to run the benchmark to see if this fixes the issue for you as well?

@mgovers Benchmarked it: 
Produce output on the 2605-node benchmark grid, 1000-scenario batch, 8 interleaved rounds:

- pre-#1536: 0.219 s
- current main: 2.138 s
- your PR: 0.409 s — 5.2×, recovering ~90% of the regression

Whole batch run 4.24 s → 2.30 s. The remaining ~1.9× above pre-#1536 looks like the extra per-node vectors, so that should line up with what you expect from #1555.

PR #1564, opened 2026-09-01
Disable msvc reproducibility check temporarily 

nitbharambe:
Our nightly build intermittently fails because of reproducible build with msvc does not succeed. 
Since clang-cl succeeds, there is atleast some way possible for windows users to get reproducible builds. 
Hence lets disable this check, fix reproduciblity and then re-enable it again

### DIFF
--- a/.github/workflows/check-build-reproducibility.yml
+++ b/.github/workflows/check-build-reproducibility.yml
@@ -80,7 +80,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


figueroa1395 (commented):
@nitbharambe Can you run nightly on this just to check we didn't miss anything? After that I'll approve.

nitbharambe:
https://github.com/PowerGridModel/power-grid-model/actions/runs/33511602493
Ran successfully

TonyXiang8787 (approved):

PR #1568, opened 2026-09-03
Catch duplicate ids within a single update scenario

### LINKED ISSUE #1552: [FEATURE] Data validator: Multiple ids in a scenario should be caught
## Problem

Currently we don't support checking multiple ids update within same scenario in the data validator.  (See example below)

- The PGM core's behaviour is known and it is undefined behaviour: What ends up happening is that the last update gets applied.
- Checking it is also known and unsupported at data validator: What happens in this case is that data validator applies updates as per search of _get_indexer in the validation util.  Hence the duplicate id data gets ignored.

Adding this in the current implementation of validator is possible but not very simple. One motivation of this issue is to make users aware of this behaviour as it is easy to miss. Hence It would be beneficial to support this in the validator sometime.
Another is that we are thinking of migrating data validator to C++ core soon. And maybe we don't want to have large changes in python data validator right now. But lets remember to support this after migration.

## Example

```py
from power_grid_model import AttributeType, CalculationType, ComponentType, initialize_array, DatasetType, LoadGenType, PowerGridModel
from power_grid_model.validation import assert_valid_batch_data, assert_valid_input_data
    
# node
node = initialize_array(DatasetType.input, ComponentType.node, 1)
node[AttributeType.id] = [1]
node[AttributeType.u_rated] = [10.5e3]
# source
source = initialize_array(DatasetType.input, ComponentType.source, 1)
source[AttributeType.id] = [10]
source[AttributeType.node] = [1]
source[AttributeType.status] = [1]
source[AttributeType.u_ref] = [1.0]
# load
sym_load = initialize_array(DatasetType.input, ComponentType.sym_load, 2)
sym_load[AttributeType.id] = [4, 7]
sym_load[AttributeType.node] = [1, 1]
sym_load[AttributeType.status] = [1, 1]
sym_load[AttributeType.type] = [LoadGenType.const_power, LoadGenType.const_power]
sym_load[AttributeType.p_specified] = [0e6, 0e6]
sym_load[AttributeType.q_specified] = [5e6, 2e6]

input_data = {
    ComponentType.node: node,
    ComponentType.source: source,
    ComponentType.sym_load: sym_load,
}

# load
sym_load = initialize_array(DatasetType.update, ComponentType.sym_load, (2, 2))
sym_load[AttributeType.id] = [[4, 4], [7, 7]]
sym_load[AttributeType.p_specified] = [[10e6, 20e6], [30e6, 40e6]]

update_data = {
    ComponentType.sym_load: sym_load,
}

assert_valid_input_data(input_data=input_data, calculation_type=CalculationType.power_flow)
assert_valid_batch_data(input_data=input_data, update_data=update_data, calculation_type=CalculationType.power_flow)

model = PowerGridModel(input_data=input_data)
output_data = model.calculate_power_flow(update_data=update_data)

print(f"load output: {output_data[ComponentType.sym_load]['p']}")
```
Gives 
```
load output: [[20000000.        0.]
 [       0. 40000000.]]
```

### Community Meeting

- [ ] I wish to discuss this feature in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

appleweiping:
Fixes #1552.

`validate_ids()` checks that every update id exists in the input dataset, but never that an id occurs at most once within a scenario. Because `_update_component_array_data()` applies updates through a NumPy fancy-index scatter (`input_data[field][idx] = update_data[field][mask]`), duplicate indices silently keep only the last value — and validation runs on the already-merged data, so it can no longer see the duplicate. A batch with two rows sharing an id passes `validate_batch_data()` with no error while quietly discarding all but the last row.

This adds `ids_unique_in_update_data_set()` next to `ids_valid_in_update_data_set()`, returning the existing `NotUniqueError`, and calls it per scenario from `validate_batch_data` as a blocking id error so the bad merge is never validated. Unset ids are excluded — they mark update-by-position, and partially-unset ids are already reported by the existing rule. Reusing an id in a *different* scenario stays valid.

**One thing worth flagging for the C++ migration question raised in the issue:** the compiled core has the same behaviour. Calling `calculate_power_flow` directly with a duplicate-id batch, bypassing Python validation entirely, also succeeds silently and returns a result reflecting only the last row. So the current state is silent data loss rather than just a missing validator nicety, which may argue for landing a check now rather than waiting. A core-side fix is out of scope here.

While adding tests I also found that `update_sym_load_r` in `tests/unit/test_optional_ids.py` builds a `(2, 2)` array and assigns `[[4], [7]]`, which NumPy broadcasts to `[[4, 4], [7, 7]]` — updating the same id twice per scenario. Its columnar counterpart uses `(2, 1)`. I corrected the shape; that was the source of the new check's failures there.

## Verification

- `tests/unit/validation` → 358 passed, 1 skipped (324 passed before the added tests)
- `mypy .` clean, `ruff format` clean, `reuse lint` compliant at 1122/1122
- Full `tests/unit` shows 11 pre-existing failures in `test_0Z_model_validation.py`, identical with and without this change — numerical tolerances from a stale local C++ build on my machine, unrelated to this work. I did not rebuild the core, so I cannot confirm those pass on a fresh build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/user_manual/data-validator.md
+++ b/docs/user_manual/data-validator.md
@@ -62,6 +62,13 @@ In such cases, the latter is leading when only running batch calculations.
 Running single calculations on an incomplete input data set is, of course, unsupported.
 ```
 
+```{note}
+Within a single scenario, every component may be updated at most once: an update is applied by id lookup, so a
+duplicate id would silently result in only the last of the duplicated records being applied. `validate_batch_data`
+reports such duplicates as a `NotUniqueError` for the scenario in question. Updating the same component again in
+another scenario of the same batch is of course perfectly valid.
+```
+
 Validating a Cartesian product of datasets used in PGM's `update_data` via providing it with `list[BatchDataset]`is
 done by validating each individual `BatchDataset` that conforms the Cartesian product.
 
--- a/src/power_grid_model/validation/_rules.py
+++ b/src/power_grid_model/validation/_rules.py
@@ -746,6 +746,31 @@ def ids_valid_in_update_data_set(
     return []
 
 
+def ids_unique_in_update_data_set(update_data: SingleDataset, component: ComponentType) -> list[NotUniqueError]:
+    """
+    Check that for all records of a particular type of component, the ids are unique within a single scenario.
+
+    An id may be updated at most once per scenario: the update of a component is applied by an id lookup, so a
+    duplicate id would silently result in only the last of the duplicated records being applied. Note that the same
+    id may of course be updated again in any of the other scenarios of a batch.
+
+    Args:
+        update_data: A single scenario of the update data set for all components
+        component: The component of interest
+
+    Returns:
+        A list containing zero or one NotUniqueError, listing all ids that occur more than once within the scenario.
+    """
+    ids = update_data[component][AttributeType.id]
+    # Unset ids are not actual ids: they either mark an update-by-position scenario, or they are already reported
+    # by ids_valid_in_update_data_set. Either way they are not duplicates and are excluded from this check.
+    ids = ids[~is_nan_or_default(ids)]
+    _, inverse, counts = np.unique(ids, return_inverse=True, return_counts=True)
+    if np.any(counts != 1):
+        return [NotUniqueError(component, AttributeType.id, ids[(counts != 1)[inverse]].flatten().tolist())]
+    return []
+
+
 def all_finite(
     data: SingleDataset, exceptions: dict[ComponentType, list[AttributeType]] | None = None
 ) -> list[InfinityError]:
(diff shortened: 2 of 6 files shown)


appleweiping:
Validated the latest head 56909bbde. Targeted validation and optional-ID tests: 263 passed with the repository coverage gate disabled for a subset run. The default gate reports 72% only because a subset was selected; no test assertion failed. Duplicate-ID cases cover single-scenario errors, multiple scenarios, cross-scenario reuse, unset IDs, and columnar data. The Python fix and tests are ready for review; the C++ core behavior remains intentionally out of scope.

appleweiping:
Thanks for the guidance on #1552. I confirmed that the core also accepts duplicate IDs silently, but I do not want this Python change to work against the planned validator migration. I am closing this for now and will revisit it after the C++ migration if it is still needed.

figueroa1395:
@appleweiping Thank you! We'll tag you once we have migrated the validator to C++. Good that you are interested in this!

PR #1566, opened 2026-09-03
Multiple infinite faults on a bus incorrect fault output

nitbharambe:
See failing tests in the CI of first PR. Compare line i_to with sum of fault currents i_f for the bug.

### Checks

- [ ] If changes are related to CI/CD, are they verified via a manual run on this branch?
- [ ] Do you wish to discuss this PR in the bi monthly community meeting?
Please join the [PGM community meeting](<https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745>)

PR #1570, opened 2026-09-04
Multiple infinite faults on a bus incorrect fault output: 2nd test post merge links with nodes

nitbharambe:
Fixes test caused by #1566. New PR created as this test is common with #1550 .


mgovers (approved):

mgovers:
failing CI is because this PR depends on #1566 while the base branch depends on #1550 . Both are being sent to merge queue. The right way is to merge this branch into the #1584 now, so that we are notified by failing CI but we can already review the diff easily, which is the intention of this PR.

PR #1452, opened 2026-06-25
PV node: Support voltage-regulator Q limits and PV→PQ switching 

scud-soptim:
Implements #1236

## Summary

This PR adds reactive-power limit handling for voltage-regulated nodes in Newton-Raphson power-flow calculations.

An active `voltage_regulator` now models the regulated node as a PV node by enforcing the configured voltage magnitude `u_ref`. If the reactive power required to maintain this voltage violates `q_min` or `q_max`, the regulated object is clamped to the violated limit and the effective node type is switched from PV to PQ.

The PR also exposes the effective node type in steady-state node output and documents the voltage-regulator behavior, including PV→PQ switching and reactive-power allocation when multiple regulators are connected to the same node.

## Main changes

* Add effective `bus_type` to steady-state node output.

  * `0`: PQ
  * `1`: PV
  * `2`: Source/Slack
* Extend voltage-regulator steady-state output with a limit indicator.

  * `-1`: lower reactive-power limit `q_min` reached
  * `0`: no limit violation
  * `1`: upper reactive-power limit `q_max` reached
* Implement PV-node handling in Newton-Raphson power flow for active voltage regulators.
* Implement PV→PQ switching when the required reactive power violates configured Q limits.
* Clamp the regulated object to the violated Q limit after switching to PQ.
* Propagate final bus type and Q-limit status to steady-state output.
* Add and update tests for voltage-regulator Q-limit behavior, PV→PQ switching, batch updates, and output metadata.
* Update user documentation for voltage-regulator behavior and power-flow algorithm details.

## Reactive-power allocation with multiple voltage regulators

When multiple active voltage regulators are connected to the same node, they jointly regulate the same node voltage. The required reactive power is initially distributed equally over the active regulated objects. If one regulator reaches its individual `q_min` or `q_max`, it is clamped at that limit and removed from the remaining allocation. The unallocated reactive power is redistributed over the remaining regulators until all required reactive power has been allocated or all available regulators have reached a limit.

For asymmetric calculations, limit checks use the total three-phase reactive power. The phase distribution follows the available phase proportions where possible; if no usable phase proportion is available, the value is distributed equally over the three phases.

## Documentation updates

The documentation now describes:

* voltage-regulator PV-node behavior;
* PV→PQ switching on reactive-power limit violation;
* the meaning of `node.bus_type`;
* the meaning of `voltage_regulator.limit_violated`;
* reactive-power allocation for multiple voltage regulators at the same node;
[...]

PR #1573, opened 2026-09-05
Fix: Restore Voltage Regulator warning regarding support for NR only

furqan463:
In PR #1452 a warning was unintentionally removed and the comment https://github.com/PowerGridModel/power-grid-model/pull/1452#discussion_r3504944692 remained unresolved.

In my opinion it should be a warning, but at-least it should be a note to make it more visible.


mgovers (approved):

mgovers:
macos-26-intel was flaky. not an issue with this PR

PR #1574, opened 2026-09-07
Logger: Composite / Multiple loggers logger

nitbharambe:
Implementation from POC in pgm/feature/logger-api-poc

### DIFF
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
@@ -10,8 +10,9 @@
 
 #include <algorithm>
 #include <concepts>
+#include <format>
 #include <map>
-#include <string_view>
+#include <string>
 #include <type_traits>
 
 namespace power_grid_model {
@@ -92,6 +93,15 @@ class CalculationInfo : public Logger {
     Report report() const { return data_; }
     void clear() { data_.clear(); }
 
+    std::string string_report() const {
+        std::string result;
+        for (auto const& [tag, value] : data_) {
+            // Each line has format: EVENT_CODE\tVALUE
+            result += std::format("{}\t{}\n", std::to_underlying(tag), value);
+        }
+        return result;
+    }
+
     template <std::derived_from<Logger> T> T& merge_into(T& destination) const {
         if (&destination == this) {
             return destination; // nothing to do
@@ -109,7 +119,11 @@ class MultiThreadedCalculationInfo : public MultiThreadedLoggerImpl<CalculationI
     using Report = CalculationInfo::Report;
 
     Report report() const { return get().report(); }
-    void clear() { get().clear(); }
+    std::string string_report() const { return get().string_report(); }
+
+  protected:
+    std::string snapshot_locked() const override { return get().string_report(); }
+    void clear_locked() override { get().clear(); }
 };
 } // namespace common::logging
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
@@ -1,0 +1,108 @@
+// SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+//
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+#include "logging.hpp"
+
+#include <algorithm>
+#include <memory>
+#include <ranges>
+#include <string_view>
+#include <vector>
+
+namespace power_grid_model::common::logging {
+
+// Owns a list of child loggers (created by MultiThreadedCompositeLogger::create_child) and fans all log calls out to
+// each of them. The children are owned by this logger; their lifetimes are tied to this object.
+class CompositeChildLogger : public Logger {
+  public:
+    explicit CompositeChildLogger(std::vector<std::unique_ptr<Logger>> children) : children_{std::move(children)} {}
+
+    void log(LogEvent tag) override                       { log_all(tag); }
+    void log(LogEvent tag, std::string_view message) override { log_all(tag, message); }
+    void log(LogEvent tag, double value) override             { log_all(tag, value); }
+    void log(LogEvent tag, Idx value) override                { log_all(tag, value); }
+
+    using Logger::log;
+
+  private:
+    std::vector<std::unique_ptr<Logger>> children_;
+
+    template <typename... Args> void log_all(Args&&... args) {
+        for (auto& child : children_) {
+            child->log(std::forward<Args>(args)...);
+        }
+    }
+};
+
+// Owning fan-out MultiThreadedLogger. Holds shared ownership of MultiThreadedLogger instances and forwards
+// all log calls to each. create_child() creates a CompositeChildLogger that owns one child per registered logger.
+//
+// Lifetime contract: each registered logger is kept alive by this composite for as long as it remains
+// registered (shared ownership), regardless of whether any other owner (e.g. a C API wrapper) has released
+// its own reference. This is what makes destroying the wrapper while still registered safe.
+// Dedupe: registering the same logger twice is a no-op (idempotent, consistent with logging conventions).
+// UB: modifying the logger list while a calculation is in progress.
+class MultiThreadedCompositeLogger : public MultiThreadedLogger {
+  public:
+    MultiThreadedCompositeLogger() = default;
+    explicit MultiThreadedCompositeLogger(std::vector<std::shared_ptr<MultiThreadedLogger>> loggers)
+        : loggers_{std::move(loggers)} {}
+
+    // Add/remove a logger. The object address is unchanged so any existing reference_wrapper
+    // pointing to this composite remains valid. Do not call while a calculation is in progress.
+    void add(std::shared_ptr<MultiThreadedLogger> logger) {
+        if (logger == nullptr) {
+            return; // defensively ignore null registrations
+        }
+        if (std::ranges::any_of(loggers_, [&](auto const& existing) { return existing.get() == logger.get(); })) {
+            return; // already registered — dedupe silently, consistent with logging API conventions
+        }
+        loggers_.push_back(std::move(logger));
+    }
+    void remove(MultiThreadedLogger const* logger) {
+        if (auto it = std::ranges::find_if(loggers_, [&](auto const& existing) { return existing.get() == logger; });
+            it != loggers_.end()) {
+            loggers_.erase(it);
+        }
+    }
+    void reset() { loggers_.clear(); }
+
+    std::unique_ptr<Logger> create_child() override {
+        std::vector<std::unique_ptr<Logger>> child_loggers;
+        child_loggers.reserve(loggers_.size());
+        for (auto const& logger : loggers_) {
+            child_loggers.push_back(logger->create_child());
+        }
+        return std::make_unique<CompositeChildLogger>(std::move(child_loggers));
+    }
+
+    void log(LogEvent tag) override                           { log_all(tag); }
+    void log(LogEvent tag, std::string_view message) override { log_all(tag, message); }
+    void log(LogEvent tag, double value) override             { log_all(tag, value); }
+    void log(LogEvent tag, Idx value) override                { log_all(tag, value); }
+
+    using MultiThreadedLogger::log;
+
+    // Fan out clear() to every registered logger.
+    void clear() override {
+        for (auto const& logger : loggers_) {
+            logger->clear();
+        }
+    }
+
+    [[nodiscard]] bool empty() const { return loggers_.empty(); }
+
+  private:
+    std::vector<std::shared_ptr<MultiThreadedLogger>> loggers_; // owning
+
+    template <typename... Args> void log_all(Args&&... args) {
+        for (auto const& logger : loggers_) {
+            logger->log(std::forward<Args>(args)...);
+        }
+    }
+};
+
+} // namespace power_grid_model::common::logging
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
@@ -7,6 +7,7 @@
 #include "common.hpp"
 
 #include <cstdint>
+#include <functional>
 #include <memory>
 #include <string_view>
 
@@ -71,6 +72,13 @@ class Logger {
 
 struct MultiThreadedLogger : public Logger {
     virtual std::unique_ptr<Logger> create_child() = 0;
+
+    // The function is called exactly once with a string_view valid only for the duration of the call.
+    // Default: no op / delivers an empty view
+    virtual void get_output(std::function<void(std::string_view)> const& callback) const { callback({}); }
+
+    // Clear accumulated output. Default: no-op.
+    virtual void clear() {}
 };
 
 } // namespace common::logging
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
@@ -75,6 +75,27 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
 
     using MultiThreadedLogger::log;
 
+    // Lock-safe overrides. Marked final so subclasses cannot bypass the lock; override
+    // snapshot_locked / clear_locked instead to add type-specific behaviour.
+    void get_output(std::function<void(std::string_view)> const& fn) const final {
+        // Snapshot under the lock, then call fn without the lock so user callbacks
+        // cannot re-enter logger APIs and deadlock on the non-recursive mutex.
+        std::string snapshot;
+        {
+            std::lock_guard const lock{mutex_};
+            snapshot = snapshot_locked();
+        }
+        fn(snapshot);
+    }
+    void clear() final {
+        std::lock_guard const lock{mutex_};
+        clear_locked();
+    }
+
+  protected:
+    virtual std::string snapshot_locked() const { return {}; }
+    virtual void clear_locked() {}
+
   private:
     friend class ThreadLogger;
 
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
@@ -100,7 +100,7 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
     friend class ThreadLogger;
 
     LoggerType log_;
-    std::mutex mutex_;
+    mutable std::mutex mutex_;
 
     void sync(ThreadLogger const& logger) {
         assert(&logger != &log_);
--- a/power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
+++ b/power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
@@ -85,6 +85,7 @@ class TextLogger : public Logger {
         data_.clear(); // reset error flags
     }
     std::string report() const { return data_.str(); }
+    std::string_view report_view() const { return data_.view(); }
     void flush() {
         if (flush_handler_) {
             // exception swallowing: if the handler throws, we leave the logger in valid state and the caller handles it
@@ -113,8 +114,12 @@ class MultiThreadedTextLogger : public MultiThreadedLoggerImpl<TextLogger> {
     using MultiThreadedLoggerImpl<TextLogger>::MultiThreadedLoggerImpl;
 
     std::string report() const { return get().report(); }
-    void clear() { get().clear(); }
+    std::string_view report_view() const { return get().report_view(); }
     void flush() { get().flush(); }
+
+  protected:
+    std::string snapshot_locked() const override { return get().report(); }
+    void clear_locked() override { get().clear(); }
 };
 } // namespace common::logging
 
--- a/tests/cpp_unit_tests/logging/test_composite_logging.cpp
+++ b/tests/cpp_unit_tests/logging/test_composite_logging.cpp
@@ -1,0 +1,140 @@
+// SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+//
+// SPDX-License-Identifier: MPL-2.0
+
+#include <power_grid_model/common/composite_logging.hpp>
+
+#include <power_grid_model/common/logging.hpp>
+#include <power_grid_model/common/text_logger.hpp>
+
+#include <doctest/doctest.h>
+
+#include <memory>
+
+namespace power_grid_model::common::logging {
+namespace {
+using LoggerPtr = std::shared_ptr<MultiThreadedTextLogger>;
+
+LoggerPtr make_text_logger() { return std::make_shared<MultiThreadedTextLogger>(); }
+} // namespace
+
+TEST_CASE("Test MultiThreadedCompositeLogger") {
+    MultiThreadedCompositeLogger composite;
+
+    SUBCASE("Empty composite has no output and is empty") { CHECK(composite.empty()); }
+
+    SUBCASE("Adding a null logger is a no-op") {
+        composite.add(nullptr);
+        CHECK(composite.empty());
+    }
+
+    SUBCASE("Logging fans out to a single registered logger") {
+        auto logger = make_text_logger();
+        composite.add(logger);
+        CHECK_FALSE(composite.empty());
+
+        composite.log(LogEvent::total, Idx{1});
+
+        CHECK(logger->report().find("Tag:0") != std::string::npos);
+    }
+
+    SUBCASE("Logging fans out to multiple registered loggers") {
+        auto logger_a = make_text_logger();
+        auto logger_b = make_text_logger();
+        composite.add(logger_a);
+        composite.add(logger_b);
+
+        composite.log(LogEvent::total, Idx{1});
+
+        CHECK_FALSE(logger_a->report().empty());
+        CHECK_FALSE(logger_b->report().empty());
+    }
+
+    SUBCASE("Registering the same logger twice is idempotent") {
+        auto logger = make_text_logger();
+        composite.add(logger);
+        composite.add(logger); // second add — silent no-op
+
+        composite.log(LogEvent::total, Idx{1});
+
+        // Only one entry should be logged, i.e. exactly one occurrence of the tag.
+        auto const report = logger->report();
+        auto const first = report.find("Tag:0");
+        CHECK(first != std::string::npos);
+        CHECK(report.find("Tag:0", first + 1) == std::string::npos);
+    }
+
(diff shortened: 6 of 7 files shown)


power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -92,6 +93,15 @@ class CalculationInfo : public Logger {
     Report report() const { return data_; }
     void clear() { data_.clear(); }
 
+    std::string string_report() const {
+        std::string result;
+        for (auto const& [tag, value] : data_) {
+            // Each line has format: EVENT_CODE\tVALUE
+            result += std::format("{}\t{}\n", std::to_underlying(tag), value);
+        }
+        return result;
```
mgovers, line 102: please use `std::stringstream` or similar. `std::string` is not built for this kind of repeated appending in a loop

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
```diff
@@ -0,0 +1,108 @@
...
+    void log(LogEvent tag, Idx value) override                { log_all(tag, value); }
+
+    using Logger::log;
+
+  private:
+    std::vector<std::unique_ptr<Logger>> children_;
+
+    template <typename... Args> void log_all(Args&&... args) {
+        for (auto& child : children_) {
+            child->log(std::forward<Args>(args)...);
```
mgovers, line 35: you can't forward the same object multiple times. please add a test case that this is not accidentally done. i'd have expected sonar to warn about this

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -75,11 +75,32 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
...
+  protected:
+    virtual std::string snapshot_locked() const { return {}; }
+    virtual void clear_locked() {}
+
   private:
     friend class ThreadLogger;
 
     LoggerType log_;
-    std::mutex mutex_;
+    mutable std::mutex mutex_;
```
figueroa1395, line 103: This was initially strange for me, but it makes sense. See [this](https://herbsutter.com/2013/05/24/gotw-6a-const-correctness-part-1-3/) Herb Sutter article for a nice explanation.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
```diff
@@ -71,6 +72,13 @@ class Logger {
 
 struct MultiThreadedLogger : public Logger {
     virtual std::unique_ptr<Logger> create_child() = 0;
+
+    // The function is called exactly once with a string_view valid only for the duration of the call.
+    // Default: no op / delivers an empty view
+    virtual void get_output(std::function<void(std::string_view)> const& callback) const { callback({}); }
+
+    // Clear accumulated output. Default: no-op.
+    virtual void clear() {}
```
figueroa1395, line 81: Why is the default no-op? Shouldn't the default just be to clear the underlying logged data?

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
```diff
@@ -71,6 +72,13 @@ class Logger {
 
 struct MultiThreadedLogger : public Logger {
     virtual std::unique_ptr<Logger> create_child() = 0;
+
+    // The function is called exactly once with a string_view valid only for the duration of the call.
+    // Default: no op / delivers an empty view
+    virtual void get_output(std::function<void(std::string_view)> const& callback) const { callback({}); }
```
figueroa1395, line 78: Is this similar to `flush()` in the `TextLogger`? Or is their purpose different now?

I see `get_output` takes the callback as an argument, whereas `flush` takes the callback via the `TextLogger` constructor. It feels to me that both are attempting very similar things and only one should remain.

That said, taking it as an argument is a lot more flexible and perhaps aligns best with the C-API. So maybe `flush` can be removed?

Thoughts?

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
figueroa1395: See how it's done in the `TextLogger` for reference.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -109,7 +119,11 @@ class MultiThreadedCalculationInfo : public MultiThreadedLoggerImpl<CalculationI
     using Report = CalculationInfo::Report;
 
     Report report() const { return get().report(); }
-    void clear() { get().clear(); }
+    std::string string_report() const { return get().string_report(); }
+
+  protected:
+    std::string snapshot_locked() const override { return get().string_report(); }
+    void clear_locked() override { get().clear(); }
```
figueroa1395, line 126: A couple of questions:
- Why is `clear_locked` `protected`? It should be accessible by "everyone" now, right? Edit: I see now, CRTP, right?
- Why not just name it `clear` directly? The user would directly get this overload unless they explicitly cast the type to get the underlying `clear`. Also, this avoid potential naming confusion. Edit: Due to CRTP the way to access it is then via `clear`, as expected. This is just like `clear_impl`, right?
- Same questions from above but for `TextLogger`.

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
```diff
@@ -113,8 +114,12 @@ class MultiThreadedTextLogger : public MultiThreadedLoggerImpl<TextLogger> {
     using MultiThreadedLoggerImpl<TextLogger>::MultiThreadedLoggerImpl;
 
     std::string report() const { return get().report(); }
-    void clear() { get().clear(); }
+    std::string_view report_view() const { return get().report_view(); }
     void flush() { get().flush(); }
+
+  protected:
+    std::string snapshot_locked() const override { return get().report(); }
```
figueroa1395, line 121: Can this be made more efficient if you just get the "raw" data and the turn into a "string" or whatever you may need at the multi threaded logger side? Same for calculation info. 

I mention this because I believe this may copy the data twice, which can get expensive easily.

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -75,11 +75,32 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
...
 
+    // Lock-safe overrides. Marked final so subclasses cannot bypass the lock; override
+    // snapshot_locked / clear_locked instead to add type-specific behaviour.
+    void get_output(std::function<void(std::string_view)> const& fn) const final {
+        // Snapshot under the lock, then call fn without the lock so user callbacks
+        // cannot re-enter logger APIs and deadlock on the non-recursive mutex.
+        std::string snapshot;
+        {
+            std::lock_guard const lock{mutex_};
+            snapshot = snapshot_locked();
```
figueroa1395, line 86: I think this is an extra copy made. Maybe just passing around `string_view`s is fine and converting it once to `string` at the caller `fn` point below is sufficient?

Also, since this involves a callback which may `throw`, it may be a good idea to do Lippincot pattern or similar like in `flush` for the `TextLogger` such that we handle exceptions or at least we propagate to one that points towards hey, something is wrong with your callback, can't do anything.

figueroa1395 (commented):
Partial review. I'll continue later

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
```diff
@@ -0,0 +1,108 @@
...
+class CompositeChildLogger : public Logger {
+  public:
+    explicit CompositeChildLogger(std::vector<std::unique_ptr<Logger>> children) : children_{std::move(children)} {}
+
+    void log(LogEvent tag) override                       { log_all(tag); }
+    void log(LogEvent tag, std::string_view message) override { log_all(tag, message); }
+    void log(LogEvent tag, double value) override             { log_all(tag, value); }
+    void log(LogEvent tag, Idx value) override                { log_all(tag, value); }
+
+    using Logger::log;
```
figueroa1395, line 28: Can this be placed in `log_all`? I believe it's only relevant there and it may lead to confusion later if we add another member function with log in the "wrong" place and unexpected behaviour triggers.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
```diff
@@ -0,0 +1,108 @@
...
+};
+
+// Owning fan-out MultiThreadedLogger. Holds shared ownership of MultiThreadedLogger instances and forwards
+// all log calls to each. create_child() creates a CompositeChildLogger that owns one child per registered logger.
+//
+// Lifetime contract: each registered logger is kept alive by this composite for as long as it remains
+// registered (shared ownership), regardless of whether any other owner (e.g. a C API wrapper) has released
+// its own reference. This is what makes destroying the wrapper while still registered safe.
+// Dedupe: registering the same logger twice is a no-op (idempotent, consistent with logging conventions).
+// UB: modifying the logger list while a calculation is in progress.
```
figueroa1395, line 47: Since this logger is what will be shared, let's make sure to have these two things explicit in the documentation.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
```diff
@@ -0,0 +1,108 @@
...
+    void add(std::shared_ptr<MultiThreadedLogger> logger) {
+        if (logger == nullptr) {
+            return; // defensively ignore null registrations
+        }
+        if (std::ranges::any_of(loggers_, [&](auto const& existing) { return existing.get() == logger.get(); })) {
+            return; // already registered — dedupe silently, consistent with logging API conventions
+        }
+        loggers_.push_back(std::move(logger));
+    }
+    void remove(MultiThreadedLogger const* logger) {
```
figueroa1395, line 65: Should this also take in a `share_ptr` instead to keep it consistent? Probably not, but making sure.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
```diff
@@ -0,0 +1,108 @@
...
+        if (auto it = std::ranges::find_if(loggers_, [&](auto const& existing) { return existing.get() == logger; });
+            it != loggers_.end()) {
+            loggers_.erase(it);
+        }
+    }
+    void reset() { loggers_.clear(); }
+
+    std::unique_ptr<Logger> create_child() override {
```
figueroa1395, line 73: We probably want this one and below marked as `final` to avoid user overriding things and messing them up. Or should we leave that up to them?

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+    void log(LogEvent tag) override                           { log_all(tag); }
+    void log(LogEvent tag, std::string_view message) override { log_all(tag, message); }
+    void log(LogEvent tag, double value) override             { log_all(tag, value); }
+    void log(LogEvent tag, Idx value) override                { log_all(tag, value); }
+
+    using MultiThreadedLogger::log;
+
+    // Fan out clear() to every registered logger.
+    void clear() override {
```
figueroa1395, line 90: What's the difference in behaviour between `reset` and `clear`? Do we need both?

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
figueroa1395: An additional side note: Since we have some strong conventions about perfect forwarding, let's add a comment here for reference in the future. This cases do lay in one of the valid use cases: we don't care what `Args...` are nor about the qualification, we just pass them around. Same below.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
figueroa1395, line 1: It can become a bit obscure how the chain of logger, multithreadedlogger, compositelogger, multithreadedcompositelogger works, specially considering that after come the actual implementations. Can you add a brief description somewhere here explaining the flow a bit, otherwise in the tests.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
figueroa1395, line 73: Do we want to leave the user have this control? I make for C-API users it makes sense, but Python users and Cpp users (?) shouldn't need to, right?

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+// Owning fan-out MultiThreadedLogger. Holds shared ownership of MultiThreadedLogger instances and forwards
+// all log calls to each. create_child() creates a CompositeChildLogger that owns one child per registered logger.
+//
+// Lifetime contract: each registered logger is kept alive by this composite for as long as it remains
+// registered (shared ownership), regardless of whether any other owner (e.g. a C API wrapper) has released
+// its own reference. This is what makes destroying the wrapper while still registered safe.
+// Dedupe: registering the same logger twice is a no-op (idempotent, consistent with logging conventions).
+// UB: modifying the logger list while a calculation is in progress.
+class MultiThreadedCompositeLogger : public MultiThreadedLogger {
```
figueroa1395, line 48: I'm missing some reporting functionality at this stage, since the loggers will be under an abstraction, would reporting work directly via multithreaded? don't you need some overload where users can select from which or all loggers to report?

tests/cpp_unit_tests/logging/test_composite_logging.cpp
figueroa1395, line 1: I'm missing tests in which a "custom" logger inherits from `MultiThreadedCompositeLogger`. Also the `get_output` functionality with a custom callback must be tested (you can get inspiration from the `TextLogger` tests.

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -75,11 +75,32 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
...
+        }
+        fn(snapshot);
+    }
+    void clear() final {
+        std::lock_guard const lock{mutex_};
+        clear_locked();
+    }
+
+  protected:
+    virtual std::string snapshot_locked() const { return {}; }
```
mgovers, line 96:
```suggestion
    // Snapshot implementation. Thread-safety must be handled by the caller
    virtual std::string snapshot_thread_unsafe_impl() const { return {}; }
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -75,11 +75,32 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
...
+    // snapshot_locked / clear_locked instead to add type-specific behaviour.
+    void get_output(std::function<void(std::string_view)> const& fn) const final {
+        // Snapshot under the lock, then call fn without the lock so user callbacks
+        // cannot re-enter logger APIs and deadlock on the non-recursive mutex.
+        std::string snapshot;
+        {
+            std::lock_guard const lock{mutex_};
+            snapshot = snapshot_locked();
+        }
+        fn(snapshot);
```
mgovers, line 88: declarative

```suggestion
        // Snapshot under the lock, then call fn without the lock so user callbacks
        // cannot re-enter logger APIs and deadlock on the non-recursive mutex.
        std::string const snapshot = [this] {
            std::lock_guard const lock{mutex_};
            return snapshot_locked();
        }();
        fn(snapshot);
```

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers, line 103: why mutable?

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
nitbharambe: Default on any logger can ideally be a no-op.  
Currently this path gets used in NoMultiThreadedLogger and MultiThreadedCompositeLogger. Hence defaulted instead of pure virtual.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: Reset which is a MultiThreadedCompositeLogger method, and it empties/clears the loggers registered in composite logger. Its `loggers_.clear()` where `loggers_ ` is a vector here.
vs
clear is a generic Logger method applicable for all logger types. Its for means emptying / clearing output.
For MultiThreadedCompositeLogger, its clears all the child loggers inside the composite logger.

We need 2 distinct methods in the end.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: Slipped my mind. The only thing which is stopping from forwarding multiple times is that none of the downstream logs have a rvalue overload.
I dont see a use case for when we would like to forward instead of const&. Hence restricting this.

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
nitbharambe: (Renamed locked to thread_unsafe_impl as suggested by martijn)

Yes, for CRTP. we can make class friend / mark  specific places protected. I chose later. 
We do need separate handling in multithreaded logger to implement thread safe operations hence they were routed this way.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: We need it because we dont define all overloads of log function

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: This MultiThreadedCompositeLogger is not shared anywhere. Its owned fully by the handle. I guess this would be visible in C API PR.
Things inside it would have shared ownership

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: composite logger would not be exposed to the users. And indeed its get output would not be called. Its supposed to be "write only". The purpose of it is internal logging management. 

It would be a bit complicated to have a "report all" functionality and we dont see a use case for it.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: Yes, I think not. I suppose that would be redundant. After being passed to remove, the composite logger and C++ core should not need it anymore.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: As per above comment, composite logger is never available to user. And we dont have a special need to make it explicitly final yet.

nitbharambe: Answered above. No need to expose it to user.

power_grid_model_c/power_grid_model/include/power_grid_model/common/logging.hpp
nitbharambe: Well, flush gets the data-> passes to flush handler -> clears the data.
And get_output only gets the data and passes to callback without clearing.

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
```diff
@@ -75,11 +75,39 @@ class MultiThreadedLoggerImpl : public MultiThreadedLogger {
...
+    void get_output(std::function<void(std::string_view)> const& fn) const final {
+        // Snapshot under the lock, then call fn without the lock so user callbacks
+        // cannot re-enter logger APIs and deadlock on the non-recursive mutex.
+        std::string const snapshot = [&] {
+            std::lock_guard const lock{mutex_};
+            return snapshot_thread_unsafe_impl();
+        }();
+        fn(snapshot);
+    }
+    void clear() final {
```
nitbharambe, line 89: Thinking about this situation makes me think if we need special handling for multi threaded text logger and flushing. Will explore later

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
nitbharambe: To make get_output const. 
I see this seems like an acceptable way as described in the article suggested by https://github.com/PowerGridModel/power-grid-model/pull/1574/changes#r3976369778

power_grid_model_c/power_grid_model/include/power_grid_model/common/calculation_info.hpp
```diff
@@ -92,6 +93,15 @@ class CalculationInfo : public Logger {
     Report report() const { return data_; }
     void clear() { data_.clear(); }
 
+    std::string string_report() const {
+        std::ostringstream result;
+        for (auto const& [tag, value] : data_) {
+            // Each line has format: EVENT_CODE\tVALUE
+            result << std::to_underlying(tag) << '\t' << value << '\n';
+        }
+        return result.str();
```
mgovers: significant performance optimization
```suggestion
        return std::move(result).str();
```

mgovers: see also the 3rd instantiation from https://en.cppreference.com/cpp/io/basic_stringstream/str

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
mgovers: resolved

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
mgovers: this default-enables the base version. there's nothing wrong with it here, but maybe order-wise, you can add it to the top (right before line 23), so that all custom overrides come after it.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
mgovers: i think this can be resolved

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
mgovers: i don't think that's needed, but we should definitely add a `nullptr` check

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
mgovers: indeed

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
mgovers: you can maybe explicitly call it `flush()` or `clear_buffers()` instead

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: minor reminder^

mgovers (commented):
reviewed. LGTM except for some nitpicks but i still see some open comments from @figueroa1395

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
nitbharambe: Checked. Should be safe. Regarding user created threads, things are to be handled in API implementation to leave base log untouched.

tests/cpp_unit_tests/logging/test_composite_logging.cpp
nitbharambe: As before, we dont allow customization on MultiThreadedCompositeLogger because CompositeLogger is not available for users.

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: I think clear is generic enough for non text things. If flush or clear_buffers is more adequate, we can rename as such. Lmk if you think so

mgovers: hmmm let's keep it as is given the base class and we can change it later

power_grid_model_c/power_grid_model/include/power_grid_model/common/composite_logging.hpp
nitbharambe: Added check to skip nullptr

power_grid_model_c/power_grid_model/include/power_grid_model/common/text_logger.hpp
mgovers: i don't think there's a way out of this. since you use `std::move(sstr).str()`, you will only copy once up til here (or more if the sstream itself needed multiple reallocations, but there's not really a way around that), and whatever the user does is up to them. there's no such thing as a "string_view_stream" or something (since the stream must be owning the data)

mgovers: that said, this comment was made before `std::move(sstr).str()` so maybe it's just outdated

power_grid_model_c/power_grid_model/include/power_grid_model/common/multi_threaded_logging.hpp
mgovers: it originally was only a `move` but with the new implementation, it's also RVO

mgovers: it's changed to `scoped_lock` and scoped locks are freed on context release, even if that was done after an exception was raised. the exception propagates up and is caught by the C API implementation.

mgovers (approved):
LGTM

PR #1583, opened 2026-09-09
Merge links into nodes: temporarily remove all new code path registrations from #1550

mgovers:
This enables separation of concerns for #1550:

1. only refactor - no use of the new link registration.
2. merge nodes on links iff there are no node injection sensors.

In the original form of #1550 (before this PR), #1550 contains both steps. This PR isolates only step 1. A follow-up PR will then revert only this PR so that it only contains the change to step 2.


mgovers:
Benchmark run vs `v1.13.163`; looks OK

tests/data/state_estimation/dummy-test-sym/sym_output.json
figueroa1395, line 1: Let's keep this one. This should be fine in main as is.

nitbharambe: Same for above test of handling-of-links/merge-nodes-by-links-1-line-with-line-to-self

nitbharambe: But sicne the handling-of-links/merge-nodes-by-links-1-line-with-line-to-self is AssertionError anyway either is okay

mgovers: will do in a separate PR

figueroa1395 (approved):

nitbharambe (approved):

PR #1586, opened 2026-09-13
ThreadSanitizer: add nightly check + fix minor issue

mgovers:
* add TSan workflow to Nightly
* found small issue regarding status changed callback registration + fix

* [x] Test workflow in https://github.com/PowerGridModel/power-grid-model/actions/runs/35570786032 (do-not-merge label)

PR #1590, opened 2026-09-16
Logger: C/C++ api implementation

nitbharambe:
Implementation from POC in pgm/feature/logger-api-poc

Discuss over making the logger experimental.

Renaming of API comments will be addressed after all other reviews are done.

* [ ] TODO before merge: run ThreadSanitizer (see #1586 )

## AI Summary of API changes below

**C API**  
Defined in [logger.h](power_grid_model_c/power_grid_model_c/include/power_grid_model_c/logger.h).

```c
PGM_Logger* PGM_create_logger(PGM_Handle* handle, PGM_Idx logger_type);
void PGM_destroy_logger(PGM_Logger* logger);

void PGM_register_logger(PGM_Handle* handle, PGM_Logger* logger);
void PGM_unregister_logger(PGM_Handle* handle, PGM_Logger* logger);
void PGM_unregister_all_loggers(PGM_Handle* handle);

typedef void (*PGM_LogOutputCallback)(
    char const* data,
    PGM_Idx size,
    void* user_data
);

void PGM_logger_get_output(
    PGM_Handle* handle,
    PGM_Logger* logger,
    PGM_LogOutputCallback callback,
    void* user_data
);

void PGM_logger_clear(PGM_Handle* handle, PGM_Logger* logger);
```

Logger types are defined in [basics.h](power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h):

```c
enum PGM_LoggerType {
    PGM_text_logger = 0,
    PGM_benchmark_logger = 1
};
```

**C++ API**  
Defined in [logger.hpp](power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/logger.hpp).

```cpp
namespace power_grid_model_cpp {

class Logger {
public:
    explicit Logger(PGM_LoggerType logger_type);

    Logger(Logger const&) = delete;
    Logger& operator=(Logger const&) = delete;

    Logger(Logger&&) noexcept;
    Logger& operator=(Logger&&) noexcept;

    ~Logger();

    std::string get_output();
    void clear();
};

}
```

Logger attachment methods are available through [Model](power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/model.hpp):

```cpp
void Model::add_logger(Logger& logger) const;
void Model::remove_logger(Logger& logger) const;
void Model::remove_all_loggers() const;
```

The lower-level [Handle](power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/handle.hpp) exposes equivalent registration methods:

```cpp
void Handle::register_logger(Logger& logger) const;
void Handle::unregister_logger(Logger& logger) const;
void Handle::unregister_all_loggers() const;
```

The C API reports errors through `PGM_Handle`; the C++ wrapper converts them to exceptions. C callbacks receive non-null-terminated data valid only during the callback, while `Logger::get_output()` returns an owning `std::string`.


power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
```diff
@@ -141,6 +141,15 @@ typedef struct PGM_WritableDataset PGM_WritableDataset;
  */
 typedef struct PGM_DatasetInfo PGM_DatasetInfo;
 
+/**
+ * @brief Opaque struct for a logger object.
+ *
+ * A logger captures diagnostic output (text or benchmark timings) produced during calculations.
+ * Loggers are created with PGM_create_logger(), registered to a handle with PGM_register_logger(),
+ * and must be destroyed with PGM_destroy_logger() after unregistering.
```
Jerry-Jinfeng-Guo: Is this a two-step clean up? If that is the case: 
```suggestion
 * and must be destroyed with PGM_destroy_logger() after unregistering with PGM_unregister_logger().
```

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
```diff
@@ -248,6 +257,19 @@ enum PGM_ExperimentalFeatures {
...
 
+/**
+ * @brief Enumeration of logger types.
+ *
+ * Selects which kind of diagnostic output a logger captures.
+ * Calling output/clear operations on a do-nothing logger is a no-op.
+ *
+ */
+enum PGM_LoggerType {
+    PGM_do_nothing_logger = 0, /**< logger that discards all output (no-op) */
```
Jerry-Jinfeng-Guo, line 268: Would it make sense to just call it `PGM_no_op_logger` or even `PGM_no_logger`?

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/logger.h
```diff
@@ -0,0 +1,172 @@
...
+ * They are designed for advanced users: output is non-conclusive and intended as debugging hints.
+ *
+ * Lifecycle:
+ *   1. Create a logger:   PGM_create_logger()
+ *   2. Register it:       PGM_register_logger()
+ *   3. Run calculations.
+ *   4. Read output:       PGM_logger_get_output()
+ *   5. Optionally clear:  PGM_logger_clear()
+ *   6. Unregister it:     PGM_unregister_logger()
+ *   7. Destroy it:        PGM_destroy_logger()
```
Jerry-Jinfeng-Guo, line 18: So indeed a two step clean up

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/logger.h
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * Lifetime and safety:
+ *   - The underlying logging implementation is shared between the PGM_Logger returned by
+ *     PGM_create_logger() and every handle it is registered to. Calling PGM_destroy_logger()
+ *     while the logger is still registered to one or more handles is safe: the implementation
+ *     stays alive and keeps collecting output for those handles.
+ *   - After PGM_destroy_logger() is called, the caller no longer has a PGM_Logger* to target
+ *     that specific registration individually. PGM_unregister_all_loggers() (which detaches
+ *     everything on a handle) or destroying the handle itself is the only way to release it.
```
Jerry-Jinfeng-Guo: Is `PGM_unregister_all_loggers` the only actual interface function that can do this? Perhaps  `PGM_unregister*_loggers`? This last sentence in this paragraph is a bit confusing, not only grammatically.

power_grid_model_c/power_grid_model_c/src/logger.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <power_grid_model/common/exception.hpp>
+#include <power_grid_model/common/logging.hpp>
+#include <power_grid_model/common/multi_threaded_logging.hpp>
+#include <power_grid_model/common/text_logger.hpp>
+
+#include <memory>
+
+// The PGM_Logger struct is the C API wrapper for a polymorphic multi-threaded logger.
+// It is heap-allocated by PGM_create_logger and freed by PGM_destroy_logger.
+// The underlying logger implementation is shared with any handle it is registered to (see
```
Jerry-Jinfeng-Guo, line 27: "underlying logger" or "underlying logger instance"?

power_grid_model_c/power_grid_model_c/src/logger.hpp
```diff
@@ -0,0 +1,58 @@
...
+// MultiThreadedCompositeLogger), so destroying this wrapper while still registered does not
+// invalidate the implementation: it stays alive as long as any registration references it.
```
Jerry-Jinfeng-Guo, line 29: 1, "invalidate the logger"?
2, "it stays alive until all of its active registrations are unregistered."

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/model.hpp
```diff
@@ -54,6 +55,18 @@ class Model {
         handle_.call_with(PGM_calculate, get(), opt.get(), output_dataset.get(), nullptr);
     }
 
+    // Attach a logger so it receives output from calculations performed on this model.
+    // Attaching the same logger twice is a no-op. See logger.hpp for lifetime notes: the
+    // logger may safely be destroyed while still registered, but it can then no longer be
+    // targeted individually via remove_logger() (use remove_all_loggers() instead).
+    void add_logger(Logger& logger) { handle_.register_logger(logger); }
```
Jerry-Jinfeng-Guo: I think `attach_logger` is a better naming

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/model.hpp
```diff
@@ -54,6 +55,18 @@ class Model {
...
+
+    // Detach a specific logger from this model. A no-op if it is not registered.
+    void remove_logger(Logger& logger) { handle_.unregister_logger(logger); }
```
Jerry-Jinfeng-Guo: I think `detach_logger` is a better naming

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
nitbharambe: On a proper use case, its create register,  unregister from handle then destroy it.
But simply logger can be created and destroyed without registering.
Its not mandatory to unregister before destroying but logger would be active unnecessarily oterwise.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
nitbharambe: Should be fine. Do nothing is more user understandable i guess on the business side. But either should be okay.
NoLogger, DoNothingLogger or NoOpLogger

power_grid_model_c/power_grid_model_c/src/logger.hpp
nitbharambe: Underlying logger instance

power_grid_model_c/power_grid_model_cpp/include/power_grid_model_cpp/model.hpp
nitbharambe: I dont mind this either.

power_grid_model_c/power_grid_model_c/src/logger.hpp
nitbharambe: Its 2. Both written for same intention

tests/native_api_tests/test_api_logger.cpp
```diff
@@ -0,0 +1,644 @@
...
+    run_calculate(model_a);
+    auto const after_a = count_lines(logger.get_output());
+    CHECK(after_a > 0);
+
+    run_calculate(model_b);
+    auto const after_b = count_lines(logger.get_output());
+    CHECK(after_b > after_a); // combined output from both models
+}
+
+TEST_CASE("CPP Logger - move construction preserves registration and output access") {
```
nitbharambe, line 550: Behaviour of copy/move of model/loggers. Let me know what you think. IMO, this is kind of complicated but cant seem to determine a way which is correct as well as simple.

tests/native_api_tests/test_api_logger.cpp
Jerry-Jinfeng-Guo, line 1: The test is quite comprehensive. However, could we also consider adding tests in the following fronts?
- Test on multi-threading safety 
- test on large payload/stress tests
- tests on edge cases with special characters, null bytes or empty string
- test ensuring a logger can be registered, unregistered, and successfully registered again without leaking or breaking
- test verifying that calling `clear()` twice in a row acts as a no-op the second time and doesn't crash

Additionally, test `Logger - unregister non-registered logger is no-op` and `CPP Logger - remove_all_loggers detaches everything` is testing similar things (but fine since they are targeting different interfaces).

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
nitbharambe: Additionally, lets also think of renaming text logger.

power_grid_model_c/power_grid_model_c/src/handle.hpp
```diff
@@ -12,24 +12,36 @@
...
-    power_grid_model::Idx err_code;
+    power_grid_model::Idx err_code{};
     std::string err_msg;
     power_grid_model::IdxVector failed_scenarios;
     std::vector<std::string> batch_errs;
     mutable std::vector<char const*> batch_errs_c_str;
     [[no_unique_address]] power_grid_model::BatchParameter batch_parameter;
+    // Loggers registered on this handle. Owned by the caller; the composite forwards to them.
+    // Survives clear_error. Do not modify while a calculation is in progress.
+    power_grid_model::common::logging::MultiThreadedCompositeLogger composite_logger;
```
mgovers: this makes handle (and therefore also everything that includes handle, which is everything) incredibly expensive to compile. It's better to forward declare and make this a smart pointer.

power_grid_model_c/power_grid_model_c/src/handle.hpp
```diff
@@ -12,24 +12,36 @@
...
 
+#include <algorithm>
 #include <exception>
 #include <string_view>
+#include <vector>
 
 // context handle
 struct PGM_Handle {
-    power_grid_model::Idx err_code;
+    power_grid_model::Idx err_code{};
```
mgovers: optional nit-pick for clarity:
```suggestion
    power_grid_model::Idx err_code{PGM_no_error};
```

tests/native_api_tests/test_api_logger.cpp
mgovers, line 1: i guess that the removal of all the multi-threaded logger headers from main_model + unit tests to fix the compilation speed is out of scope of this PR?

power_grid_model_c/power_grid_model_c/src/logger.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+    switch (type) {
+    case PGM_do_nothing_logger:
+        return new PGM_Logger{std::make_shared<NoMultiThreadedLogger>()}; // NOSONAR(S5025)
+    case PGM_text_logger:
+        return new PGM_Logger{std::make_shared<MultiThreadedTextLogger>()}; // NOSONAR(S5025)
+    case PGM_benchmark_logger:
+        return new PGM_Logger{std::make_shared<MultiThreadedCalculationInfo>()}; // NOSONAR(S5025)
+    default:
+        throw power_grid_model::MissingCaseForEnumError{"make_logger", type};
```
mgovers: please use the `safe_memory_handling.hpp` tools for this, and probably also the `cast_to_c`/`cast_to_cpp` tools in `forward_declarations.hpp`

power_grid_model_c/power_grid_model_c/src/logger.cpp
```diff
@@ -0,0 +1,53 @@
...
+using power_grid_model_c::safe_ptr;
+using power_grid_model_c::safe_ptr_get;
+} // namespace
+
+PGM_Logger* PGM_create_logger(PGM_Handle* handle, PGM_Idx logger_type) {
+    return call_with_catch(handle, [logger_type] { return make_logger(logger_type); });
+}
+
+void PGM_destroy_logger(PGM_Logger* logger) {
+    delete logger; // NOSONAR(S5025)
```
mgovers: please use the `safe_memory_handling.hpp` tools for this

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
mgovers: why would we expose `PGM_do_nothing_logger` across the interface in the first place? the default logger should be no logging, right? The main reason for a `do_nothing_logger` is for internal mocking purposes, which can be achieved by creating a custom logger callback that is no-op in the C++ API integration tests

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
```diff
@@ -248,6 +257,19 @@ enum PGM_ExperimentalFeatures {
...
+    PGM_text_logger = 1,       /**< logger that captures timestamped text messages */
+    PGM_benchmark_logger = 2,  /**< logger that captures timing information per calculation phase */
+};
```
mgovers, line 271: 1. where's the custom callback logger?
2. no-op logger should probably be internal only
3. IMO the order should be this

```suggestion
enum PGM_LoggerType {
    PGM_text_logger = 0,       /**< logger that captures timestamped text messages */
    PGM_custom_logger = 1,     /**< logger that uses a custom callback */
    PGM_benchmark_logger = 2,  /**< logger that captures timing information per calculation phase */
};
```

nitbharambe: Original idea was that all logging vsisible to users use a custom callback via the composite logger. So that includes both text logger and benchmark logger. What you suggest reflects a different approach, lets sync up on this.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
nitbharambe: Yeah, makes sense. No need to expose it via enum.

mgovers: we can even remove the symbols (moving it to the tests entirely)

power_grid_model_c/power_grid_model_c/src/logger.cpp
nitbharambe: Done. Also did same for https://github.com/PowerGridModel/power-grid-model/pull/1590/commits/2dc2815186eb0b2d3d8cb4016df8491815e91cee and https://github.com/PowerGridModel/power-grid-model/pull/1590/commits/e1c3960c5a66adee12ac891f10fa67063c36409e.

power_grid_model_c/power_grid_model_c/src/handle.hpp
nitbharambe: Done in https://github.com/PowerGridModel/power-grid-model/pull/1590/commits/f0ea62c1ffd62f5e46b1e2cc93693fbfe871cfd2. I guess we cant copy because of it, but should be fine.

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
nitbharambe: Removed in https://github.com/PowerGridModel/power-grid-model/pull/1590/commits/709fd632a84113e038797aeda8f2851dbe62ff07

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/logger.h
```diff
@@ -0,0 +1,170 @@
...
+ * @code
+ *   #define MY_CTX_MAGIC 0x50474D4C  // "PGML"
+ *   struct MyCtx { uint32_t magic; std::string* out; };
+ *   void cb(char const* data, PGM_Idx size, void* user_data) {
+ *       assert(((MyCtx*)user_data)->magic == MY_CTX_MAGIC);
+ *       ((MyCtx*)user_data)->out->assign(data, (size_t)size);
+ *   }
+ * @endcode
+ */
+typedef void (*PGM_LogOutputCallback)(char const* data, PGM_Idx size, void* user_data); // NOLINT(modernize-use-using)
```
mgovers, line 137: explicit mention needs to happen here regarding compiler optimizations

power_grid_model_c/power_grid_model_c/src/handle.hpp
mgovers: it's fine indeed

tests/native_api_tests/test_api_logger.cpp
nitbharambe: Lets try to add this in scope. Atleast removal from main_model? Ill give an attempt and revert if change is bigger

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/logger.h
```diff
@@ -0,0 +1,181 @@
...
+ * @param handle The handle used to report errors.
+ * @param logger The logger to clear.
+ */
+PGM_API void PGM_logger_clear(PGM_Handle* handle, PGM_Logger* logger);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
```
mgovers, line 189: ?

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/logger.h
nitbharambe: Resolving as per offline update from you: No special addressal needs to happen regarding this.

tests/native_api_tests/test_api_logger.cpp
nitbharambe: Moved and added here

nitbharambe: 1. These are needed but we need to refine what needs to be done. 
2. Same as 1
3. Out of scope and should be part of respective logger tests. But it felt small enough to add one for text logger in this PR. Lmk if i should revert or remove altogether. 
Dont know if any other special characters are to be tested.

5. Added
6. Added

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
```diff
@@ -248,6 +259,17 @@ enum PGM_ExperimentalFeatures {
...
 };
 
+/**
+ * @brief Enumeration of logger types.
+ *
+ * Selects which kind of diagnostic output a logger captures.
+ *
+ */
+enum PGM_LoggerType {
+    PGM_text_logger = 0,      /**< logger that captures timestamped text messages */
```
mgovers, line 269: as discussed: `PGM_logger_type_info` or something similar + reserve `error`, `debug`, ...

power_grid_model_c/power_grid_model_c/include/power_grid_model_c/basics.h
```diff
@@ -248,6 +259,17 @@ enum PGM_ExperimentalFeatures {
...
+    PGM_benchmark_logger = 1, /**< logger that captures timing information per calculation phase */
```
mgovers, line 270: maybe for now we should not explicitly expose the benchmark logger via the C API yet. As you can see in #1598 , there's still some work to be done, like migrating formatting

power_grid_model_c/power_grid_model_c/src/logger.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include "logger.hpp"
+#include "handle.hpp"
+#include "input_sanitization.hpp"
+#include "logger_fwd.hpp"
+#include "safe_memory_handling.hpp"
+
+#include "power_grid_model_c/basics.h"
+#include "power_grid_model_c/logger.h"
+
+#include <power_grid_model/common/calculation_info.hpp>
```
mgovers:
```suggestion
```

power_grid_model_c/power_grid_model_c/src/logger.cpp
```diff
@@ -0,0 +1,117 @@
...
+namespace {
+MultiThreadedCompositeLogger&
+extract_handle_logger(HandleLogger& handle_logger) { // NOSONAR(S4998) // utility function for exactly this purpose
+    return dynamic_cast<MultiThreadedCompositeLogger&>(power_grid_model_c::get_logger(handle_logger));
+}
+
+PGM_Logger* make_logger(PGM_Idx type) {
+    using namespace power_grid_model::common::logging;
+
+    switch (type) {
```
mgovers:
```suggestion
    switch (type) { // NOLINT(hicpp-multiway-paths-covered) // temporary while benchmark is scoped out
```

mgovers (commented):
clang-tidy

PR #1594, opened 2026-09-18
feat: support ideal source handling

### LINKED ISSUE #733: [FEATURE] Ideal source handling

VITA2aishu:
Fixes #733

### Changes proposed in this PR include

- Support `sk = inf` as the representation of an ideal source.
- Internally cap the short-circuit power at a sufficiently large finite value to avoid numerical issues.
- Apply the same handling when `sk` is updated.
- Allow `inf` for source `sk` during input validation while retaining the existing positive-value validation.
- Add C++ tests covering ideal source construction, values above the internal cap, and updates to `sk = inf`.
- Add a Python input-validation test for an ideal source.
- Document how to configure an ideal source using `sk = inf`.

### Could you please pay extra attention to the points below when reviewing the PR

- Please review the handling of `sk = inf` and the chosen finite internal cap.
- Please review the validation behavior for ideal sources.

### Checks

- [ ] If changes are related to CI/CD, are they verified via a manual run on this branch?
- [ ] Do you wish to discuss this PR in the bi-monthly community meeting?


power_grid_model_c/power_grid_model/include/power_grid_model/component/source.hpp
```diff
@@ -27,13 +28,14 @@ class Source : public Appliance {
     using InputType = SourceInput;
     using UpdateType = SourceUpdate;
     static constexpr char const* name = "source";
+    static constexpr double ideal_source_sk = 10e50;
```
Jerry-Jinfeng-Guo, line 31: Please add a comment here stating the choice of value. Different from the feature desctiption, please also consider putting this as a constant included from `power_grid_model_c/power_grid_model/include/power_grid_model/common/common.hpp`.

tests/cpp_unit_tests/component/test_source.cpp
```diff
@@ -315,5 +316,98 @@ TEST_CASE("Test source") {
...
+                                         .sk = infinity,
+                                         .rx_ratio = 0.1,
+                                         .z01_ratio = 3.0};
+
+    SourceInput const capped_source_input{.id = 1,
+                                          .node = 2,
+                                          .status = 1,
+                                          .u_ref = 1.0,
+                                          .u_ref_angle = 0.0,
+                                          .sk = Source::ideal_source_sk,
```
Jerry-Jinfeng-Guo, line 337: We could share this across subcases with a using statement in the testcase.

tests/unit/validation/test_input_validation.py
```diff
@@ -771,6 +771,32 @@ def test_validate_input_data_sym_calculation(input_data):
...
+
+    data = {
+        CT.node: node,
+        CT.source: source,
+    }
+
+    validation_errors = validate_input_data(
+        data,
+        calculation_type=CalculationType.power_flow,
+    )
```
Jerry-Jinfeng-Guo, line 796: Please also add validate input for SE and SC.

VITA2aishu: Addressed. Added input validation coverage for both state estimation and short-circuit calculations.

TonyXiang8787:
@VITA2aishu,

Thanks for your contribution. Please also add a validation test case for power flow and one for short circuit with `inf` as the source `sk`. In the folder:

https://github.com/PowerGridModel/power-grid-model/tree/main/tests/data

tests/data/short_circuit/source-ideal-sk/params.json
```diff
@@ -0,0 +1,10 @@
+{
+  "calculation_method": "iec60909",
+  "rtol": 1e-8,
+  "atol": 1e-8,
+  "short_circuit_voltage_scaling": "maximum",
+  "xfail": {
+    "raises": "Failed",
+    "reason": "Ideal source with infinite short circuit power is not supported for short-circuit calculations."
```
TonyXiang8787, line 8: you add a test to test ideal source should fail in short circuit calculations. But in your adjustment in the core, you do not do such a test to throw the exception?

VITA2aishu: Thanks for pointing this out. Issue #733 specifies that sk == inf and sk > 10e50 should not be supported by input validation for short-circuit calculations, while the short-circuit calculation itself is described as undefined behavior. Currently, I reject these values in validate_input_data(), but I don't explicitly throw an exception in the C++ core. Would you prefer me to add an explicit exception in the short-circuit calculation path, or should the test only verify the input validation behavior?

TonyXiang8787: Indeed we decide not to support `inf` for short circuit calculation. If there is a fault at the source bus, the short circuit current will also be infinite. Leave them as undefined in the core is probably the preferred way.

That means we don't need a validation case for short circuit. As it is undefined and will not throw error from the core. Your `xfail` will not work since it only checks if the core throws error.

Maybe just remove this validation test case.

TonyXiang8787 (changes requested):
@VITA2aishu there are many build errors, please try to fix them.

tests/data/short_circuit/source-ideal-sk/params.json
mgovers: we can instead catch this in the data validator

